PR #218, opened 2014-12-23
Simulation extensions

plaa:
I'm providing this pull request for review and testing.  I implemented most of it in April-May, but it was left a bit halfway.  Now that I got back to reviewing it, I couldn't really find a lot that was missing.  Please test, comment and try to find things that are wrong or broken.

This implements configurable simulation extensions, i.e. user code that is easily accessible and configurable through the UI (in comparison with the current simulation listeners).  See the Simulation options tab.  There are two simulation extensions provided by default, Air start (launch altitude + velocity) and Java code (which allows using existing simulation listeners).

Simulation extensions are loaded using the plugin system, so you could just drop a .jar file into your plugin directory and then start using it through the UI.  This should make plugins much more accessible.

I'll have to write up a bit more details about simulation extensions, their internals and implementing them.  I'll also work on an external example simulation extension for reference purposes.


plaa:
I'm closing this in favor of another branch with additional functionality.

PR #226, opened 2015-01-09
Feature/scripting

plaa:
Functionality for simulation extensions + an extension that allows directly writing scripts that interact with simulations.

Scripts can be written in any language supported by the JVM's scripting API.  The scripts have full access, thus for security reasons all scripts that have not been explicitly indicated as trusted by the user are disabled when loading files.

(Sorry for the messy commits...)


kruland2607:
The tavis ci build failed because it couldn't find junit.  Any idea what's going on?

plaa:
One of my unit tests was accidentally in src/ instead of test/.  I also added a few more unit tests for the file saving/loading.

PR #229, opened 2015-01-27
Remove accidental debugging code

plaa:
Some debugging code was accidentally left over in my scripting branch.

### DIFF
--- a/core/src/net/sf/openrocket/simulation/extension/impl/JavaCode.java
+++ b/core/src/net/sf/openrocket/simulation/extension/impl/JavaCode.java
@@ -1,14 +1,9 @@
 package net.sf.openrocket.simulation.extension.impl;
 
-import java.math.BigInteger;
-import java.util.Arrays;
-import java.util.List;
-
 import net.sf.openrocket.simulation.SimulationConditions;
 import net.sf.openrocket.simulation.exception.SimulationException;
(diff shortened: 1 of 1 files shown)


jimmiedave:
Really sorry to bug you, but have tried to get in touch and get no answer: OpenRocket.info wiki has site configuration problems, and the site will also have to be updated (soon we hope) for a pending release.

If you're not able to work on the site config, and unwilling to accept my offer of assistance, is there a way we can get the site into the developers' hands for maintenance, or the domain pointed to a net-new site you don't have to maintain?

It doesn't have to be in my hands; just someone who can make it work.

JoePfeiffer:
I'm surprised by the lack of response -- it seems like in the past he has responded as needed.  I hope he's OK....

I don't know anything about the various wiki systems in use (on openrocket or in general), so this may be an obviously impractical suggestion -- but could the wiki be hosted as a subtopic of the wiki on github until a more permanent solution could be found?

jimmiedave:
I too, hope he's OK.

We might be able to use the GitHub wiki, though if no one can maintain the openrocket front page (looks nicely hand-coded to me) it'll still say 15.03 after the next release; and openrocket.info has all the SEO juice: it's the first Google hit. Number two: SourceForge (that should get fixed too, but one brushfire at a time).

PR #252, opened 2015-12-13
Parallel staging, Pods, Rail Buttons

teyrana:
Bug reports welcome! 
Questions and feedback will be addressed on the email list.  -DMW


kruland2607:
Even though the build is now broken, this is significant enough to merge first, fix later.

PR #258, opened 2016-03-23
Add concurrent mass info for both with and without motor configuration

wcravens:


wcravens:
Messing around with IntelliJ and Broke the build.  Cancelling the pull request.

PR #261, opened 2016-03-25
Merges with openrocket/master

teyrana:
1) Merges with Wes' mass with/without motor display code
2) Fixed a subtle bug emergent from rocket-flightconfiguration behavior changes in unstable branch


teyrana:
adjusting commit message.  Will resubmit.

PR #269, opened 2016-04-28
Moved 'no motors' text to translation system.

teyrana:
As requested via email.


teyrana:
Update: Added further translation string fixes
Update: fixed the DocumentConfigTest unit test failures

PR #272, opened 2016-09-13
Adds V2-style fins

teyrana:
Ability to add Freeform Fins onto Transitions.  

- Major rewrite of freeform finset
- Major rewrite of FinSetTest, FreeformFinSetTest
- Minor rewrites of many supporting classes
- only freeform fins are allowed because the trapezoid and elliptical fins already have many baked-in assumptions which make them prohibitively complicated to adapt.  Besides, each can simply be converted to freeform fins, and accomplish the same thing.
- other miscellaneous UI and event handling fixes


SpamWall:
By whom and when will these pull requests get merged with "unstable"?

teyrana:
@SpamWall pull requests are merged in by anyone with write access to the destination repository.

teyrana:
2nd note:  the above failure is due to test failures in destination repo.  I'll follow up with those separately, and remerge this branch into a clean destination.

SpamWall:
Sorry, this is my first open source work with Github.  As I understand the process, a developer clones the "unstable" repo and makes their changes.  Then they do a "Pull Request" and the person with write access reviews their changes and merges it with "unstable" if those changes are accepted.  My question is who is that person for this project?

Also, I assume the "unstable" branch is where the active work is going and I should keep my clone updated to "unstable".  Is that correct?

wcravens:
Fantastic.  You chose a pretty interesting project for your first.  I've always enjoyed talking about it with people, they are usually pretty fascinated and had no idea folks with our interests existed.

You are right that someone will review and merge pull requests.  Just like @teyrana said.  There are several people with the access rights to do it on this project and they all know when pull requests have been submitted.

Thanks again for choosing our project to help out with.  Welcome aboard, it's much appreciated.

teyrana:
@SpamWall ah, ok.  There doesn't seem to be a publicly viewable list of members with write access, and OpenRocket doesn't have specific assigned roles for this sort of thing. Happily though, you don't really need to know, because github notifies people automatically, as @wcravens pointed out :) Further discussion then takes place on these comment threads (like we're doing now), on the mailing list, or just via direct emails.

PR #273, opened 2016-09-14
Motor mass calculation fixes

teyrana:
Improved the mass calculation accuracy for the ThrustCurveMotor class: 

Total mass, propellant mass, and Center-of-Mass are now more accurate.

### DIFF
--- a/core/src/net/sf/openrocket/database/motor/ThrustCurveMotorSet.java
+++ b/core/src/net/sf/openrocket/database/motor/ThrustCurveMotorSet.java
@@ -271,8 +271,8 @@ public class ThrustCurveMotorSet implements Comparable<ThrustCurveMotorSet> {
 			}
 			
 			// 2. Number of data points (more is better)
-			if (o1.getTimePoints().length != o2.getTimePoints().length) {
-				return o2.getTimePoints().length - o1.getTimePoints().length;
+			if (o1.getSampleSize() != o2.getSampleSize()) {
+				return o2.getSampleSize() - o1.getSampleSize();
 			}
 			
 			// 3. Comment length (longer is better)
--- a/core/src/net/sf/openrocket/document/OpenRocketDocument.java
+++ b/core/src/net/sf/openrocket/document/OpenRocketDocument.java
@@ -415,7 +415,7 @@ public class OpenRocketDocument implements ComponentChangeListener {
 	 * Clear the undo history.
 	 */
 	public void clearUndo() {
-		log.info("Clearing undo history of " + this);
+		//log.info("Clearing undo history of " + this);
 		undoHistory.clear();
 		undoDescription.clear();
 		
--- a/core/src/net/sf/openrocket/file/rocksim/export/RocksimSaver.java
+++ b/core/src/net/sf/openrocket/file/rocksim/export/RocksimSaver.java
@@ -94,8 +94,8 @@ public class RocksimSaver extends RocketSaver {
 		
 		MassCalculator massCalc = new MassCalculator();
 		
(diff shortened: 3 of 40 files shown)


bkuker:
So I realize this is fun, not work, and this is more of a work comment, but
these PRs seem to have a LOT of changes unrelated to their description....

teyrana:
( This branch is currently failing because of the swing integration test;  we'll update that, based on whether or not we have consensus about these simulation numbers.  )

teyrana:
@bkuker Are you asking for more detailed commit messages, or smaller commits?

wcravens:
I think that this is at the branch level.  Commits should still be
contained in scope but that's not what pull requests are about.

In my thinking, work or fun, branches should add/remove/change one
thing/area.  It increases the amount of PR / merge tasks but it sure
does make for changes that are easier to reason about.

I personally even try to encapsulate whitespace/rename changes into
separate branches or at least separate commits.  So if I'm in a codebase
working on a feature and find styling or whitespace issue I try to deal
with it separately.  These are often labelled with a 'WhiteSpace Only:'
prefix in the commit message. It can be really confusing when there is a
200 change commit but only three lines actually changed the program
logic and the rest are whitespace or variable renaming.

With all that being said I don't even follow these rules all the time
and I haven't ever rejected work because it wasn't followed.  There have
been times that I've taken a PR and reworked it to account for this
stuff but I've never put that work back on the original developer.

Wes

teyrana:
Yea, I agree with that as a policy.  It is really nice to have separable commits that way.  

The problem I find (at the branch level) is that non-functional changes often are interleaved with the functional changes -- like what if you're changing whitespace on a function that   didn't exist in the original branch?   You can make that as a separate branch, but now you have dependencies among branches :P 

What have you works for those?

wcravens:
If I'm fixing the formatting of code that was created in the branch I'm working on then I will often rewrite history to correct it rather than adding a new commit.  But of course this is only valid if I've never pushed the branch anywhere.

Almost as equally often, I just plop down a whitespace correction commit but avoid making changes outside of the code that relates to the branch that I'm currently working on.  And these commits are 'Whitespace Only:' commits. It can be really hard to avoid the temptation to 'fix it all', especially with regard to refactoring.

When the non-functional changes span across code outside of the feature I'm working on that's when I save it for a separate branch all together.

Branches and commits are as much about communication with your fellow developers as anything else.  Most people would not like to read 10 books at once with the chapters and pages all interleaved.

Wes

bkuker:
You said it better than I was going to :)

kruland2607:
My biggest gripe about this change set is the use of:
if (debug) {
System.err()
}

Why not just use:

LOGGER.debug()

instead?

teyrana:
1) if you see any of the if( debug ) statements, let me know where and I'll
excise them.

2) the honest answer is that log.debug() has a lot of output.  I mean A
LOT.  It was much easier to use system.err, and just look at what I was
debugging.  I grant you this still isn't the "right" answer, but as a
temporary solution, it worked quite well :)   ... except when I forget to
clean up, and leave those statements there :P

bkuker:
I see that it is SLF...  is either logback or log4j backing it up? Both have options for you to reduce the log output to only the classes you are interested in

kruland2607:
Another option is to declare a special logger category and use that instead of the class name. Then just these messages can be enabled, and the rest disabled. 

----- Original Message -----

teyrana:
btw: a bit of more context:  The logging framework doesn't seem to load during unit tests, which means that everything gets dump to stdout, anyway.  At least in eclipse.

PR #277, opened 2016-09-27
"Fixes" jdk8 rounding bugs and forces point decimal

Vicilu:
This pull make the test assert and not fail for the rounding bugs of JDK 8
Forces point as decimal separator for locales where comma is used
Some minor refactoring done

### DIFF
--- a/core/src/net/sf/openrocket/preset/TypedKey.java
+++ b/core/src/net/sf/openrocket/preset/TypedKey.java
@@ -3,7 +3,7 @@ package net.sf.openrocket.preset;
 import net.sf.openrocket.unit.UnitGroup;
 
 public class TypedKey<T> {
-
+	
 	private final String name;
 	private final Class<T> type;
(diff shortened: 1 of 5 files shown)


teyrana:
"Forces point as decimal separator for locales where comma is used"

Pardon? isn't that the *purpose* of locales?  Perhaps you could expand on what you mean by that part? 

As a USian, I definitely would rather read 3,000.00 m,  but if someone is used to reading 3.000,00 meters, why are we forcing them otherwise?

Vicilu:
Yeah, changing the source code is indeed a bad idea, so I've made another pull request that changes the tests instead, sorry, my first pull request here (here I mean GitHub) 
The point I've tackled is that tests were failing just because my jdk is different, uses comma and there's this jdk 8 rounding bug, and that annoys because that always appears tests failed when using tdd  or refactoring, so for other to not have this problem, I've change the tests that were failing

PR #278, opened 2016-09-28
changes tests failing with false negative

Vicilu:
Allows locale with comma to run the toString tests without failling, changing the source code was a bad idea indeed
Also changes test to take into account the rounding bug of JDK 8 which doesn't round to even all the time

### DIFF
--- a/core/test/net/sf/openrocket/unit/FractionalUnitTest.java
+++ b/core/test/net/sf/openrocket/unit/FractionalUnitTest.java
@@ -1,206 +1,266 @@
 package net.sf.openrocket.unit;
 
-import static org.junit.Assert.*;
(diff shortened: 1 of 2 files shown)


kruland2607:
You failed to revert the change in Unit.java.

Also, digging through the java bugs, this is apparently a bug in jdk 1.7 
which was fixed in 1.8.

Vicilu:
oh yeah, I did fail at that, will redo the stuff here tomorrow, for now I'll close this, thanks
And I still wonder why, I'm with jdk 1.8 and those tests still

Vicilu:
I'm using jdk 1.8.101 and still the tests only passes when I change, so I'm gonna revert that change and keep the comma ones, and start ignoring the fails as this is a base item

PR #280, opened 2016-09-28
[Refactor] Replaced redundant method: 'PositionValue(...)' with 'AxialOffset(...)'

teyrana:
- These two methods did the same thing, so this patch removes the former from RocketComponent and descendant classes
- Wherever used, "PositionValue(...)" was replace with "AxialOffset(...)"
- Removed some extraneous redirect definitions of PositionValue(...)
- Fixed some UI references to the old method

(apologies for where eclipse auto-organized the imports)

### DIFF
--- a/core/src/net/sf/openrocket/file/rocksim/export/BasePartDTO.java
+++ b/core/src/net/sf/openrocket/file/rocksim/export/BasePartDTO.java
@@ -1,5 +1,10 @@
 package net.sf.openrocket.file.rocksim.export;
 
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
 import net.sf.openrocket.file.rocksim.RocksimCommonConstants;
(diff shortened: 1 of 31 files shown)


teyrana:
Since the other PR got merged pretty quickly, and this one did not-- did someone have questions I can clear up?  ( Or rework requests? ) 

For those wondering why PositionValue got changed to AxialOffset:  

While adding strap on stages, "positionValue' return a single double, while 'Position' refers to a 3-dimensional scalar.  (x,y,z).  In previous releases, 'positionValue' refers to the x / axial direction.  

Because 'double positionValue()' was so vague in that context, I'd added "axialOffset" and "radialOffset" and "angularOffset"; Each of which mapped directly to an element of the stages location in 3-dimensional space. 

We could also have PositionValue return that 3-vector as well.  The counter argument to that was that 90% of the use case in our code base use only the axial value, for obvious reasons.

Whichever way we go with this, I humbly suggest we decide on one scheme for getting those numbers, and then patch so that it's consistent across the codebase.

kruland2607:
Hi Dan, 

I accepted all but that one yesterday. It was slightly longer and I wanted to be able to look at it more closely. My biggest question was exactly what you answered here :) 

Kevin 

----- Original Message -----

teyrana:
Ah!  Good to hear :)

Issue #244, opened 2015-09-30
Multi-Select Motor & Configurations table for Removal/Deletion


wcravens:
Be able to multi-select rows from the 'Motors & Configuration' table to allow multiple simultaneous removal of configurations

SpamWall:
I have completed this and will be sending the Pull Request by the weekend.  I also modified the tables on the Motor Configuration page to add some height to the cells since they were too small for the text.  I also widened the Motor Mount table a bit so the component names didn't get cut off.

dkingsley:
It does not look the the PR from SpamWall ever happened?

PR #283, opened 2016-10-06
[#244] Multi-select delete motor configurations

SpamWall:
Motor configuration table now allows multi-row selections.  Also increased the size of the cells so text isn't cut off.  Refactored the changed files to modularize a little better (more work needs done there).


kruland2607:
There's some conflicts with this merge.  If you don't mind, please resolve and resubmit.

PR #286, opened 2016-10-10
adds doxyfile for doxygen documentation

Vicilu:
adds doxyfile for genearting docs,
maybe needs to change input, output paths to fit server needs
maybe also change the logo path to the original, which I didn't find


Vicilu:
btw, this thread may clear some bits about using doxygen and not javadoc:
http://stackoverflow.com/questions/225447/doxygen-vs-javadoc

as summary, doxygen do what javadoc does and more that makes easier

teyrana:
Ahhh, thanks for the clarification!

(Now all we have to do is document the code :D )

PR #289, opened 2016-10-13
adds documentation to aerodynmics package 

Vicilu:
and some refactoring of the files


Vicilu:
btw, I've notice the AbstractChangeSource class source keeps converting EventListener and checking every time in the fire event method if it's instance of StateChangeListener while the implemented methods only allows StateChangeListener to be added in the array, why is that? even if you extend, you can't add this other types

Vicilu:
if you want a more general, you can use a more general interface with accepting EventListeners with an array of List<? exenteds EventListener> while an extended class will change it?

kruland2607:
Can you describe the bug fixed in ecb467?

Vicilu:
it was my own bug, my bad at c9dadc1, and got fixed in it

PR #293, opened 2016-10-13
[Refactor] Rocket now subclasses ComponentAssembly 

teyrana:
. . . instead of RocketComponent

Discussion:
Arguably this is a non-trivial change, but this simplifies overrides, specifically when debugging mass-calculation code, and locating components.  My case is:
1. conceptually, a rocket _IS A_ ComponentAssembly
2. rocket still inherits all the RocketComponent methods.
3. This change reduces code duplication:  
    \* 7 redundant functions
    \* lines: +8 / -47
4. Also:  changed Coordinate.NUL -> Coordinate.ZERO to better reflect our intent.

[edited for format.]

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/ComponentAssembly.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/ComponentAssembly.java
@@ -32,6 +32,11 @@ public abstract class ComponentAssembly extends RocketComponent {
 	}
 	
 	@Override
+	public boolean allowsChildren(){
+		return true;
+	}
+	
+	@Override
 	public double getAxialOffset() {
 		return super.asPositionValue(this.relativePosition);
 	}
@@ -49,7 +54,7 @@ public abstract class ComponentAssembly extends RocketComponent {
 	 */
 	@Override
 	public Coordinate getComponentCG() {
-		return Coordinate.NUL;
+		return Coordinate.ZERO;
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


SpamWall:
I not familiar with the code, but be careful of thinking "Is A" means inheretance, it doesn't.   "Is A" means "is constructed as".  Proper inheritance is a "behaves as" relationship. 

Craig

⁣

Sent from TypeApp

​

Vicilu:
that actually makes sense, cleans up a lot of stub methods

Vicilu:
now just gotta figure our why that value is now half :/

teyrana:
> ... but be careful of thinking "Is A" means inheretance, (sic) it doesn't.
>  "Is A" means "is constructed as".

I use "is a" primarily because that's the most common way of phrasing
inheritance.

> Proper inheritance is a "behaves as" relationship.

What do you mean?  Obviously, I can define two different classes, with the
same code, rename one, and they "behave as" the same.... without being the
compiler thinking either is inherited. (a la python "duck typing"). Is that
what we're talking about?

teyrana:
@Vicilu WHAT value?   If you've found a bug in that code, please be more specific!

All the core/unit tests pass, excepting IntegrationTest, which we already knew about.

SpamWall:
For instance, a square "Is A" rectangle, but a square does not "Behave As" a rectangle. A rectangle can change length and width independently.  A square cannot.  Having a Square class inherit from a Rectangle class means some of the method are not polymorphic, some methods just don't make any sense in the scope of the derived class, and "instance of" calls start creeping in.

⁣

Sent from TypeApp

​

teyrana:
Yea, that makes sense.  I was making the is-a statment in contrast to
has-a, instead of the more nuanced distinction between is-a and behaves-as.

Obviously, we agree that just because it COULD be defined by inheritance,
doesnt mean it SHOULD...

but to bring this back on topic, what's the best choice here?  (PR#293)

Vicilu:
oh yes, that integration test was what I was talking about

Vicilu:
and I think the best choice is to apply this change, it makes more sense and has less "dead code"

Vicilu:
As this part is the source of almost all problems with unstable, I've been documenting it and found it a lot of mess, parent depending of child and other things, need a lot of refactoring, who know? my guess is this is the first step to solve the problems

PR #301, opened 2016-10-24
adds more documentation 

Vicilu:
More documentation and refactoring to the files


kruland2607:
The change to Appearance seems to loose the clamp call.

PR #324, opened 2017-02-08
Follow up to #321 - Update RealisticRenderer.java

ChrisMickelson:
Small fix so that 3D Unfinished renderer draws the back of body tubes to maintain "cutaway" look instead of "x-ray" look


ChrisMickelson:
Going to redo this change and resubmit

PR #343, opened 2017-06-06
Upgrade to guice-4.1 and guice-multibindings-4.1

tibbe:
This also requires adding guava, which is no longer shipped with guice.

This allows us to use Java8 and eventually Java9. Java9 is required for HiDPI support on new monitors.

Tested on Ubuntu 14.04.


tibbe:
If we still need JDK 1.6 we can use guava-20.0 instead. Preferences?

tibbe:
With this change I've managed to run openrocket on JDK-9 (from inside Eclipse) and have HiDPI work. All the buttons, text, etc. is the right size!

ChrisMickelson:
Could you create a pull request that's based on the unstable branch? That's what we're working with for the next release. I would love to test it however I'm stuck with 1080p screens everywhere. Perhaps running Ubuntu in a 4K VM might work?

tibbe:
Closed in favor of #345 against unstable.

PR #345, opened 2017-06-06
Upgrade to guice-4.1 and guice-multibindings-4.1

tibbe:
This also requires adding guava, which is no longer shipped with guice.

PR #348, opened 2017-06-07
WIP: Build a Fat-Jar style jar instead

tibbe:
I've put together this work-in-progress pull request to create some discussion.

Using a custom class loader can be hairy. The current `JarInJarStarter` class loading doesn't work in Java 9. One fix is to not use the jar-in-jar style jar anymore. It's not needed for JOGL, as shown by this pull request. However, I'm not sure how to do that and still support plugins. Any ideas?

With this pull request and #345 I've managed to use the Ant build.xml to build an OpenRocket.jar that works on Java 9 (on Windows 10) that seems to work.

### DIFF
--- a/swing/build.xml
+++ b/swing/build.xml
@@ -27,7 +27,7 @@
 	<property name="dist.src"    value="${jar.dir}/${pkgname}-src.zip"/>
 	
 	<!-- The main class of the application -->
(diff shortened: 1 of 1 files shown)


teyrana:
Confirmed: 
Works fine on:
- Ubuntu 17.04; with OpenJDK 1.8.0_131
- OSX 10.12.5; Oracle JDK 1.8.0_31

PR #352, opened 2017-06-18
Use a parent ClassLoader

tibbe:
Before we explicitly set the parent to null, which caused classes to fail to load on Java 9.

I'm not entirely sure why this is and Sampo doesn't remember why a null parent was used to begin with.

### DIFF
--- a/swing/src/net/sf/openrocket/startup/jij/JarInJarStarter.java
+++ b/swing/src/net/sf/openrocket/startup/jij/JarInJarStarter.java
@@ -30,7 +30,7 @@ public class JarInJarStarter {
 		}
 		
 		URL[] urlArray = urls.toArray(new URL[0]);
-		ClassLoader loader = new URLClassLoader(urlArray, null);
+		ClassLoader loader = new URLClassLoader(urlArray);
 		try {
 			Thread.currentThread().setContextClassLoader(loader);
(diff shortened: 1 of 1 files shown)


tibbe:
I've now updated the commits to apply on master.

tibbe:
I can confirm that this still fixes Java 9 not working. I can successfully start an use OpenRocket with this change. The benefit is that HiDPI screens will now work. :)

teyrana:
Minor Nitpick: (a) why are merging this against master branch, vs unstable branch, where the rest of development is happening?

Anyway, This branch works on my boxes.  Would anyone like to raise any objection to merging this? 

@wcravens @kruland2607 @maintainers

wcravens:
That’s not really a minor nitpick.  We shouldn’t be building anything on master.

Wes

dkingsley:
I agree leave master alone

teyrana:
So.... What do y'all think of me cherry-pick-ing this fix into the next release candidate?

wcravens:
If it works, I don't have any problem with it.

Wes

dkingsley:
If it makes newer Java installations work I am all for it.

Dennis

teyrana:
Fix has been pulled into release-candidate-3; and will be merged into unstable with the next batch of bug fixes.

Issue #351, opened 2017-06-13
Motor thrust not multiplied with number booster pods


ChrisMickelson:
I have noticed something off with the sims for rockets with motors in booster pods. For example, if a rocket has 4 booster pods, each with a motor, then the thrust should be about 4x a single motor, however it's only being simmed as a single motor. Exported simulation data verifies thrust levels equivalent to a single motor even when there are 2 or more booster pods.

teyrana:
Fix submitted in PR: #353

wcravens:
Dan confirmed that this is fixed in Unstable

PR #350, opened 2017-06-11
Configuration Fix for Travis-CI automated Tests.  Enables tests to run to completion.

teyrana:
attempt to figure a work-around for travis-ci tripping over itself.

PR #353, opened 2017-06-19
Motorfix - fixes motor simulation count for #351

teyrana:
Oneliner which includes booster and pod instancing when simulation motor clusters in pods and boosters.

Fixes issue: #351 

(bundled with a typo fix)

### DIFF
--- a/core/src/net/sf/openrocket/simulation/AbstractSimulationStepper.java
+++ b/core/src/net/sf/openrocket/simulation/AbstractSimulationStepper.java
@@ -174,7 +174,7 @@ public abstract class AbstractSimulationStepper implements SimulationStepper {
 	 * @param stepMotors				whether to step the motors forward or work on a clone object
 	 * @return							the average thrust during the time step.
 	 */
-	protected double calculateAvrageThrust(SimulationStatus status, double timestep,
+	protected double calculateAverageThrust(SimulationStatus status, double timestep,
 			double acceleration, AtmosphericConditions atmosphericConditions,
 			boolean stepMotors) throws SimulationException {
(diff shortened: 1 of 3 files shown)


teyrana:
test failures are due to the upstream change in travis configuration. ( Addressed in PR #350 )

ChrisMickelson:
I've been testing this one and somethings still off. After looking into it,
this.config.getMount().getInstanceCount(); and this.motorCount are still both being set to 1 when there are multiple booster pods. For example, if there are 4 boosters with motors, they should return 4 as is the case with a normal 4x cluster mount.

teyrana:
Hmmm, can you tell me more specifically where you're seeing the error?

And can you send me the .ork file?

ChrisMickelson:
[files.zip](https://github.com/openrocket/openrocket/files/1092394/files.zip)

Here is a zip of two files:
parallel_test_A.ork - simple rocket with normal cluster of 4 F26s
parallel_test_B.ork - same rocket but instead with 4 boosters, each with an F26.

These should perform pretty close but one only goes to 600ft, and the other gets up to 3300ft.

Also, if you export the simulation into a .csv file, you can see the thrust is consistent with a single motor versus 4x as with the normal cluster.

As another test, monitored the values for this.config.getMount().getInstanceCount(); and this.motorCount by just printing their values with System.out.println(). Their value was 4 with the cluster, but 1 with the booster pods.

teyrana:
@ChrisMickelson 
Ah! Found my mistake.  Corrected, and resubmitted in PR #354.  (to eliminate erroneous commit in PR ;)

PR #363, opened 2017-09-15
Implementing Coefficient of Drag Override

robocoder99:
When doing rocketry in the past I made a few changes to the OpenRocket platform in order to reduce the number of launches we had to do to calibrate height. It allows an override for the coefficient of drag value for individual parts which can adjust for small discrepancies between a physical rocket and a simulated version. I just updated my version to match with the master branch of the main repo. (I can change to the unstable branch if necessary)

This has been tested and used in a physical environment and was used to place high in a height-based competition.


teyrana:
@robocoder99 
Hey, thanks for making the patch -- it's good to see active development :)

First, active development the next release is (currently) happening  over at: https://github.com/teyrana/openrocket/tree/rc0.

Second: I tried to merge your changes, but this PR overrwrites the binary file: 
"thrustcurves.ser"  Could you please verify that this is a necessary change for this CD Override feature?   (Once we've figured that out, let's merge it.)

Overall the patch looks good!  So happy to hear that you guys are getting a lot of use out of the program :)

robocoder99:
It is not, let me fix that! Thanks!

robocoder99:
#363 Is now the official version of this. I fixed some of the issues and cleaned it up.

Issue #121, opened 2013-09-02
Illegal multiplier value, mul=1001.33


kruland2607:
Reported automatically:

net.sf.openrocket.gui.main.SwingExceptionHandler$InternalException: Illegal multiplier value, mul=1001.33
    at net.sf.openrocket.gui.main.SwingExceptionHandler.handleErrorCondition(SwingExceptionHandler.java:101)
    at net.sf.openrocket.gui.dialogs.ScaleDialog.doScale(ScaleDialog.java:431)
    at net.sf.openrocket.gui.dialogs.ScaleDialog.access$600(ScaleDialog.java:65)
    at net.sf.openrocket.gui.dialogs.ScaleDialog$4.actionPerformed(ScaleDialog.java:406)
    at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2028)
    at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2351)
    at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:387)
    at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:242)
    at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:236)
    at java.awt.Component.processMouseEvent(Component.java:6414)
    at javax.swing.JComponent.processMouseEvent(JComponent.java:3275)
    at java.awt.Component.processEvent(Component.java:6179)
    at java.awt.Container.processEvent(Container.java:2083)
    at java.awt.Component.dispatchEventImpl(Component.java:4776)

dkingsley:
If you scale a rocket by a large value like 1000.33 you can get the same exception to happen and the exception is not handled so OR crashes.

SpamWall:
Is this from choosing "Scale" from the "Tools" menu?

dkingsley:
Yes, then type a large scale factor into the textbox for the scale.

hcraigmiller:
This issue can no longer be reproduced in the current unstable.

[Windows 11 Pro; Version 21H2; OS Build 22000.739; Windows Feature Experience Pack 1000.22000.739.0]
[Java "11.0.15" 2022-04-19 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.15+8-LTS-149)]

SiboVG:
> This issue can no longer be reproduced in the current unstable.

Confirmed.

PR #408, opened 2018-05-29
Illegal multiplier value, mul=1001.33 #121

SpamWall:
Fixed this defect by putting up an error dialog instead of throwing an exception.


SpamWall:
Looking at the CI output, it looks like the JVM is still pointing to a 1.7 Java version.  If we are going to target Java 1.8, can someone update the JAVA_HOME on the CI machine to a 1.8 JVM?

wcravens:
I'll merge this when we decide what to do about the build failing.  I think that Dan set up travis so I'll let him figure out what to do.

SpamWall:
From the error message on the build server, it is still using the Java 7 JVM.  I changed the Ant file to use Java 8 since that is what the group said was the current target JVM.  I think pointing JAVA_HOME at a Java 8 JVM should fix things.

Craig

⁣Sent from TypeApp ​

teyrana:
@wcravens @SpamWall   
The travis issue was fixed months ago.  If the build fails, now it's because of a recent change.  
(probably in build.xml where the fix was ;)

teyrana:
Re: building with 1.7 vs 1.8.
One can build with a higher JVM, against the target bytecode @ 1.7 -- the setting allows backwards compatibility.  

If we no longer wish to support 1.7, I'm neutral on that.  But it's not a Travis limitation.

teyrana:
Also, let's split all these separate issues into separate Pull-Requests.  

Updating the Scale Dialog is not relating to changing the build.xml file....

SpamWall:
I agree that it is not a limitation of Travis – I didn’t mean to imply that, but Travis is configured (JAVA_HOME) to use a 1.7 JVM.  I suggest installing a 1.8 JVM on Travis and resetting the JAVA_HOME to point to it.  If we are supporting Java 1.8 we should set the language level to 1.8 also IMO.

 

Craig

SpamWall:
Ok.  I'll split them.

⁣Sent from TypeApp ​

SpamWall:
The Travis output says it is using  1.7 JVM.

⁣Sent from TypeApp ​

SpamWall:
I set the language and compiler levels back to 1.7 and removed the 1.8 language features I had used.  #121 now passes on Travis.

 

Should I submit another PR for the 1.8 upgrade of the build.xml files now or wait until the changes are made to Travis?

 

Craig

wcravens:
This PR has a merge conflict.

teyrana:
Closed due to inactivity.
And the submitter deleted the originating repository.
And the PR doesn't reference a known issue.

PR #410, opened 2018-05-30
Chris mickelson patch 1

ChrisMickelson:
Changing this back to ",null);" seems to fix the crash on startup when running as a jar file


teyrana:
So, we've gone back and forth with that 'null' parameter.  

Let's at least write down when the null works, and when it doesn't.  

e.g. When does this restoring the parameter solve it?

ChrisMickelson:
I have Java 8u171 and Java 10.0.1 installed on my Win10 machine. (Oracle seems to have already dropped 9 and went straight to 10). Java 8 needs the null parameter.
Java 8 with 'null': works
Java 8 without 'null': crash on startup
Java 10 with 'null': crash on startup
Java 10 without 'null': crash on startup

teyrana:
@ChrisMickelson 
Thanks for spelling that out :D

Issue #434, opened 2018-08-03
resource leak in SerializeMotors.java


JoePfeiffer:
SerializeMotors opens a new input stream for every motor to be loaded, but doesn't close it.  
This resource leak eventually causes a build of the Master branch to fail with a "Too many files open" error.  I'll be submitting a pull request in a moment.

JoePfeiffer:
This is in Master, so no real need to apply the fix

PR #435, opened 2018-08-03
Close input streams after use to avoid build crashing due to resource leak.

JoePfeiffer:
Also closes stream oos, to match ofs which was already being closed.

Closes issue #434


JoePfeiffer:
Right, this is against master, not unstable.  Not really important as we're mainly working on unstable, but I'm rebuilding master frequently since I'm comparing simulations results from it against unstable (and adding lots of logging along the way), and fixing it makes those builds more reliable.

JoePfeiffer:
This was just maintenance on the master branch.  Sounds like that branch is dead

Issue #441, opened 2018-08-21
Average Thrust Calculation in Unstable has Error


JoePfeiffer:
The calculation of getAverageThrust throws out the first section of the time interval if the time from that to the next point in the thrust curve is less than .001, but doesn't adjust the length of the interval to compensate when calculating the weighted average of thrusts.  This results in an underestimated average thrust when a time interval starts near a thrust curve point.  For example (in particular, the anomaly I first noticed!) with an A8 motor, average thrust at 0.46656 shows a dip:

# Time (s) | Thrust (N)
0.46123    | 2.3774
0.46656    | 2.2037
0.47285   | 2.3626

The correct values are
0.46123   | 2.37741
0.46656   | 2.36835
0.47285   | 2.36260

Pull request to follow.

PR #442, opened 2018-08-21
Fix getAverageThrust (fixes issue #441)

JoePfeiffer:
Remove test for short time interval before first data point in thrust
curve.  Comment said it was for numerical stability; multiplying by a
small number and then adding doesn't introduce any instabilities I'm
aware of in this code.

Add parentheses to clarify that values are being multiplied by time
intervals, not divided.


teyrana:
@JoePfeiffer 

First: thank you for providing such a clean, concise pull request :)

Unfortunately, I recommend you rework this fix.  Full discussion follows:

Numerical computations are complicated, finicky, and fail in a lot of non-obvious, non-physical, and non-intuitive ways.  And Openrocket is complicated enough that we have many of the same issues.  

In this case, the reason for line 325 in the original is likely because if really small time steps can distort the calculations. Check out this example: 
https://docs.google.com/spreadsheets/d/1MYhiYVmxwajlDKhapi6LGc_xTs-vYFZyFXnpmCgIp-A/edit?usp=sharing

Look at row two, which shows we have a perfectly common situation: The engine is just starting, so thrust is low, almost negligible.  The Δt for integrations also typically starts out very small, so dividing a small number by another small number is very unpredictable, and may not be what you expect.  It is also the kind of issue that is not always present.  It would be relatively rare, and we probably don't have unit tests for it :(  *

However, obviously, #441 is a real issue, so we'd like to fix it.  How hard would it be to simply adjust the time interval for averaging instead of removing this small-time check? 

* here's where I plug for people to add unit tests both for the instability situation and your bug of interest ;)

JoePfeiffer:
Thanks for the kind words!  I'm a big fan of little tiny patches whenever possible.

I had considered your alternative, and believe my solution to be preferable.  Mine is actually working to avoid division by small denominators (something I've been bitten by in the past!).

Consider what would happen if we decided to change the time interval for averaging.  First, of course, we're dividing by a smaller denominator.  Worse, if the test is needed on the first subinterval (ie the test at line 325), then it is also needed before the calculation of the last subinterval (the code starting at line 334 if the PR is accepted).  This further reduces the interval, and in fact a very small interval right around a data point could result in an adjusted interval of 0 -- about as unstable as you can get!

The calculation of the impulse over the first subinterval (and all the others, for that matter) only involves a division by 2.0.  Since this only adjusts the exponent of the floating point number and not the significand, no instability is introduced until the numbers we're dealing with are denormals.  It's easy to read the code as dividing by (time[timeIndex+1] - startTime), and I suspect the test was added because it was read that way -- in fact, the first time I read it I made that mistake myself!  However, it isn't divided by that number, it's multiplied by it.  That's why I added the parentheses -- to try to make clearer that it's a multiplication.

The only division by a time interval in this section of code is by the total time interval, in the return statement.  Keeping this unchanged should avoid increasing instability.

Your example of unexpected instability is a good one; unfortunately, the calculation of the impulse is incorrect.  Impulse is the integral of force, not the derivative.  Column C should be A * B, not A / B.

teyrana:
@JoePfeiffer, @kruland2607 

Ah, of course; you're right.  My example is incorrect and irrelevant, to boot. :P  Thank you for the correction.

So let's back up -- What DOES that code do, anyhow?

After a bit of investigation, this code was added by @kruland2607 in commit: https://github.com/openrocket/openrocket/commit/a5b083ade75b462da0f00924bb71d2b2030ed2e0#diff-7d883789f817e5d5a6e0bc869ff92149
I've asked him to chime in.

### Let's clarify what the actual issue is:
How does the time issue lead to the error? And which time intervals, specfically, are you talking about? 

What is our test case we're evaluating with? What's output do we expect? 
Seriously, can this be an automated unit test?  At this point, it sounds like we need a test case just to clarify what the assumptions and expectations are.

JoePfeiffer:
I've massaged the code a little to try to clarify the order of operations.

The code is computing the average thrust over the interval between startTime and endTime by using a Riemann sum to compute the impulse over the interval, and then dividing by the interval to get the average thrust.

Looking some more, I've actually found a time interval in the simulation I'm running that shows the issue more clearly than the interval in my original issue report.  Consider the time interval from 0.305 seconds to 0.3075 seconds.  The Estes A8 motor thrustcurve in the database includes the following datapoints:
Time   Thrust
 0.277  3.717
 0.306  3.205
 0.351  2.884

The code will divide the calculation into two subintervals:  one from .305 to .306, and a second from .306 to .3075.  The thrust at the three points of interest is

Time Thrust
.305   3.2227
.306   3.205
.3075 3.1943

The average thrust and the impulse over those two subintervals is
Interval        Thrust    Impulse
.305-.306     3.2138  .0032138
.306-.3075   3.1997  .0047995

The correct average thrust (so the expected value) is 3.2053.
The value in the code as it stands now is 1.9198

I'm attaching the .ork file for the rocket being simulated.  github doesn't support the .ork file type, so I've renamed it 1-stage.zip so I can attach it.
[1-stage.zip](https://github.com/openrocket/openrocket/files/2321955/1-stage.zip)
(I realize it's not a practical design; it's stripped down to the minimum I can simulate so if I have to reproduce calculations by hand I can).

As for a unit test, I know nothing about writing them...

PR #446, opened 2018-09-21
Fix clone configuration

JoePfeiffer:
Closes issue 445.

Copies original active stage flags into clone, instead of having all stages marked as active.


JoePfeiffer:
Needs more work, sorry.

Issue #439, opened 2018-08-15
Alpha 8: cannot delete points in freeform fin shape editor


neilweinstock:
I can add points, but I can't seem to delete them with control-click.

teyrana:
Fix implemented.  Btw, was this even in 15.04?  Beacuse it didnt look like it...

neilweinstock:
Yup, worked fine in previous release.

teyrana:
Ah, ok.  I found the original code.  The old code is now fixed, instead of adding new stuff.

neilweinstock:
This seems to be officially fixed.  at this point.  Closing it.

Issue #436, opened 2018-08-06
Alpha 8: side view is not centered vertically


neilweinstock:
I see this new behavior consistently in Alpha 8:
![image](https://user-images.githubusercontent.com/243872/43737619-e44356d2-998f-11e8-9857-1181aad78406.png)

neilweinstock:
I haven't seen this in a while, and definitely not in Alpha 11.  Closing.

Issue #431, opened 2018-07-09
Can't save fins on pods with fin count=1


neilweinstock:
This is on Daniel's build, still labeled alpha 5.
Using this design:
[alc_new_pods.zip](https://github.com/openrocket/openrocket/files/2177152/alc_new_pods.zip)
On the pods, there are two fins: "fin tip" and "pylon".  When opening the file, each of those will open with a fin count of "3", despite the fact that I save the file with fin count of "1".

neilweinstock:
Works correctly on the file that was previously failing in Alpha 8.

teyrana:
Sorry, can you rephrase that? alpha8 should have fixed it :/

neilweinstock:
Sorry, I knew the wording sounded funny when I wrote it.


It seems to work now.  I tried it on the file that was previously breaking, and it reads back correctly with one fin in each set.

________________________________

teyrana:
Understood.  Thanks for getting back to us :)

neilweinstock:
This problem has re-appeared for me in alpha 8.  See attached ORK file.  The trapezoidal fins on the pods should have 1 per set, but they always come back as 3 when I reload the file.
[faddle.zip](https://github.com/openrocket/openrocket/files/2298305/faddle.zip)

neilweinstock:
I still see this problem in Alpha 9.

neilweinstock:
I'm re-opening this one.  In the alpha I'm using now, pods with instance count = 1 are again not being saved.  Or at least, when re-opening the file they all show up with instance count = 2.

teyrana:
@neilweinstock 

This should be fixed in unstable, now.   Are you stil seeing this?

neilweinstock:
Nope, seems to be fixed.

neilweinstock:
Time to close it down again, hopefully for good this time. :)

Issue #426, opened 2018-06-22
FinSetConfig Improvements (4/5)


teyrana:
When a row is selected in the [FinPointFigure] coordinate editor, the corresponding vertex in the editor should be highlighted.

neilweinstock:
This is fantastic.

Might I suggest a slightly different way of highlighting the selected point?  Right now you have them growing slightly, but not centered on the same point.  It's subtle and a little weird.

How about filling the selected square with translucent gray or something like that?  E.g:
![image](https://user-images.githubusercontent.com/243872/43737161-5927d0d8-998e-11e8-80bd-c6b1e2aef3b5.png)

teyrana:
Haha, yea, I did that because it was easy.  I must admit I dint spend a loy
of time thinking about it.

neilweinstock:
It’s not a big deal; having the feature at all is more important than anything, but I figured a slightly more visible highlight would be easy enough to implement.  I also thought of changing the outline of the selected point to green (that would probably look nice but not good for the colorblind) or adding a second green outline outside the first.

Attached are mockups of a few additional different approaches.  Pick whichever you like, or a new one. 😊

[cid:image001.png@01D42E2D.5420D1B0]


  *   Neil

teyrana:
Well, its not a big deal either way.  Y'all discuss, And decide what makes
most sense :)

PR #447, opened 2018-09-23
Alpha8 - Bugfix Collection

teyrana:
Contains fixes for #424, #425, #419, #431, #426, #439, #436, #425, #419, #426, and other code cleanup.


JoePfeiffer:
Looks like it fixes 424 and 425; fixes 436, but should the figure also be centered horizontally?  I don't use the freeform fin editor so I can't really comment on the others.  The comment on 
4cb8a03 says it resolves #423, but describes 423 as the Barrowman bug.  I don't think that ever got formally added as an issue, did it?  Too much to try to review all the code changes.

teyrana:
Hmmm, fair point.  I can split them up, if you prefer.

JoePfeiffer:
At the moment I think it's more productive to just observe whether the changes address the bugs, and see if new ones crop up.

teyrana:
Yup. Good eyes. I've removed #423 from the to-review list.  That commit probably should not have a 'fixes' tag, at all.

PR #451, opened 2018-09-29
Fix ignition type

### LINKED ISSUE #450: Second stage ignition on "burnout" ignites on "ejection charge" instead
The attached design (name changed from 2-stage.ork to 2-stage.zip) has second stage ignition set for "burnout"; however, it doesn't ignite until the first stage's ejection charge fires.

My next task...
[2-stage.txt](https://github.com/openrocket/openrocket/files/2429493/2-stage.txt)

JoePfeiffer:
Fixes #450.  Ignition type was being ignored, and all upper stages were being based on Ejection Charge with no extra delay


JoePfeiffer:
This has to be applied after pull request 444, so I'm deleting this and adding the same commit there.

PR #456, opened 2018-10-16
Multiple fixes and Refactoring of fin code

teyrana:
- Refactors AxialPositionMethod to apply to FinTabs.
- FlightConfiguration caches its bounds
- FreeformFinsets may be placed on Transitions and NoseCones.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -473,7 +473,7 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 				FinSet f = (FinSet) c;
 				double mac = ((FinSetCalc) calcMap.get(c)).getMACLength();
 				double cd = componentCf * (1 + 2 * f.getThickness() / mac) *
-						2 * f.getFinCount() * f.getFinArea();
+						2 * f.getFinCount() * f.getPlanformArea();
 				finFriction += cd;
 				
 				if (map != null) {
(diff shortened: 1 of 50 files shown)


neilweinstock:
Ability to attach freeform fins to transitions and nose cones is going to make a lot of people very very happy!

wcravens:
I've not looked back at history, but I have a hazy memory of this being
discussed before and that the end result was that there would be issues
with simulation modeling.  Do we know that the simulation and cp
calculations can handle this?

Wes

teyrana:
It has not really been tested, no.   I would put moderately favorable odds
on it working without too much work...

But at this point, it's just not certain.

JoePfeiffer:
A quick look shows that I get the same results for a transition whose fore and aft diameters are the same as for a section of body tube, and CP movement as I change aft diameter seems reasonable.

On simulation modelling...  after a similarly brief look, I couldn't find where the existing code computes body-fin interference?

wcravens:
Have you looked at 'Fin–body interference' on Pg. 34 of the OpenRocket
technical documentation
<file:///Users/wcravens/Downloads/OpenRocket_technical_documentation-v13.05.pdf>
?

Wes

JoePfeiffer:
Yes, but I'm not finding it in the code.  I'd expect it to be somewhere in calculateAssemblyNonAxialForces() in BarrowmanCalculator.java, or in merge() in AerodynamicForces.java.  I've got this sense I must just be blind....

JoePfeiffer:
Ah, OK, got it.  Yes, the fin-fin interference effects (eq 3.54) are there; it's the others that I see in the documentation but not the code (and it's nice to know that if I'm blind I'm not the only one!).  Barrowman's dissertation doesn't derive them, but does cite sources that I haven't tracked down.

From context, it seems to be assumed that the interference effects are being calculated on a cylindrical section of the body tube.

teyrana:
Ah, so the r_t is just the (aft) radius of the body tube?

JoePfeiffer:
Yes, and the span s is also constant.  I'll hazard a guess that what we really want is the mean of K_T(B) from beginning to end of the fin.  That shouldn't be too hard to figure out, but for now I'd suggest adding it to the issue list as something to add to the simulation for the next release.

teyrana:
Closed why the unit-test issues are resolved.

PR #473, opened 2018-11-04
[version] bump version string to '19-xx-alpha10'

teyrana:
- last release candidate was alpha-9
- next round of bug fixes will be alpha-10

### DIFF
--- a/core/resources/build.properties
+++ b/core/resources/build.properties
@@ -1,7 +1,7 @@
 
 # The OpenRocket build version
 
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
Rather than 18-xx-alpha10, should it be 19-xx-alpha10?  I think there's hope of a beta before the end of the year, but not a production release.

teyrana:
:/ 
Yea, you're right. I was just hoping...

(updated)

Issue #478, opened 2018-11-05
Numerical instability in simulation


JoePfeiffer:
The "Booster stage" in the attached model (as usual, filename changed so github will upload it) suddenly starts showing wild swings shortly after its apogee roughly nine out of ten times I simulate it, ending up with the simulation stopping with a simulation exception.  Working to understand what's happening....
[main.txt](https://github.com/openrocket/openrocket/files/2546635/main.txt)

JoePfeiffer:
The problem is caused by the calculation of the pitch and yaw damping moments in calculateDampingMoments().  They are quadratic in pitch (resp. yaw) rate over velocity, so if the rates become large in relation to velocity the damping force can dominate the pitch and yaw moments.  If this happens the rates increase until they overflow, and the simulation gets an exception.

There appears to be an earlier attempt to limit this by setting a minimum velocity for the calculation of 1.  No doubt that helped, but isn't sufficient for the case we're encountering here.

PR to follow.

PR #491, opened 2018-12-02
Clamp pitch and yaw damping moments to avoid numerical instability

JoePfeiffer:
Closes issue #478 

Clarifies variable names

Removes previous attempt to limit damping moments by using minimum
velocity of 1.0

Clamps damping moments so they cannot exceed actual moments, avoiding
numerical instability


teyrana:
+1  for clarifying variable names :D

PR #514, opened 2019-01-27
Update .travis.yml to omit the extra jar

teyrana:
travis attempted to upload `swing/build/jar/OpenRocket-Core.jar`, which (a) doesn't exist at that path, and (b) isn't needed to run OpenRocket.jar


JoePfeiffer:
This looks like a straightforward bit of housekeeping, let's go ahead and commit it.

PR #517, opened 2019-02-04
Fix 481

JoePfeiffer:
Changes BarrowmanCalculator.calculateNonAxialForces() to iterate through an InstanceMap, and adds a unit test for it.


teyrana:
Overall, it looks good. 

I notice the fins are still special-cased.  Did you take a look at what it'd take to fix that, so the FinSetCalc class works the same way as the others?

And of course, testing is going to be a big deal... but I don't see a way around that.

JoePfeiffer:
Much to my pleasant surprise, modifying FinSetCalc to only provide a result for one fin and BarrowmanCalculator to iterate through the fins was trivial.  The results passed Barrowman unit tests on the first try, and the FinSetCalc unit test only required modification to iterate through the fins -- no change to the actual expected results was needed.

PR #519, opened 2019-02-10
fix issue 516

JoePfeiffer:
Modified thrustcurve serialization to be more robust in the case of
some common (but generally not important) errors in thrustcurves, and
not enter thrustcurves that will cause simulation failures into the
database.

(1) Permits thrustcurves to start at time 0 with a nonzero thrust.
Previous code inserted a datapoint with time 0 and thrust 0; this
caused thrust interpolation code to fail while running a simulation.
If the first datapoint is at a time greater than 0 it does enter a
(0, 0) datapoint, as this is the correct documented behavior for RASP
files.

(2) Permits thrustcurves to end with a nonzero thrust.  Similarly to
the previous case, the old code entered a new datapoint at the same
time as the previous end of thrust with thrust 0, which would break
thrust interpolation.

(3) Thrustcurves that have a zero-thrust datapoint and a non-zero
thrust datapoint at time 0 have the first datapoint removed.

(4) Thrustcurves with more serious errors -- mainly multiple thrusts
at a single time -- aren't entered into the database, as they will
break thrust interpolation and there is no obvious way to guess which
of the datapoints is correct, or to see how to average them.

(5) When a thrustcurve isn't entered into the database, reports the simfile
number and dumps the file under the name simfile-NNN (where NNN is the
simfile number at thrustcurve.org) for inspection.


JoePfeiffer:
Oops, forgot to mention:  also reports motor simfile ID and dumps a copy of the simfile for inspection in the cases of thrustcurves that aren't entered into the database.

PR #520, opened 2019-02-17
[Fixes #482][graphics] adjusts line color and width in FinPointFigure

### LINKED ISSUE #482: Adjust Fin shape editor grid colors
In the midst of all the great improvements to the fin shape editor, I just noticed that it lost its grid.

15.03:
image

Alpha 9: 
image

Was this an intentional removal?  The grid was handy.

Note also that the root was shown with a thick grey line, vs. the 1-pixel rendering of the parent object in the new release.  Although the parent object is sometimes useful, the thick grey line contributed more towards readability.

teyrana:
See:
https://github.com/openrocket/openrocket/issues/482

Improved contrast between the fin, fin-body, and the background grid.

Cosmetic only.


JoePfeiffer:
I feel like I'm being really nit-picky here, but I like the change in line width in this PR but prefer the blue fin outline from the prior version.

But I don't have a strong opinion.  Either is really fine (a possible future feature would be to let the user configure this stuff).

PR #521, opened 2019-02-22
Fix 361

JoePfeiffer:
Run simulation to the bitter end, to better inform users of problems with models.  Previously, each simulation branch would end when the corresponding part of the rocket landed, even if there were still flight events to be processed.  It now processes all flight events, and posts a Warning for every event occurring after landing.


teyrana:
Sorry, what's the motivating situation here? 

What could cause a flight to end after landing? Like a powered landing before the engine has burned out?

JoePfeiffer:
(oops, I replied to openrocket-devel by email assuming my response would show up here.  It didnt)

Issue 361.  The rocket was horribly underpowered (among other things) and the second and third stages landed after bottom stage separation but before second stage ignition.  This appeared to the user as if only the first and third stages were being simulated due to the naming of the simulation branches.

The goal is to more clearly tell the rocket designer what went wrong.

teyrana:
@JoePfeiffer 
Do you happen to still have the inducing .ork file? 

The code looks good, but I'd love to see it in action :)

teyrana:
Reviewed. 👍 

Granted the final graphs *do* look a little bit odd, but that's probably a good thing, in this case.

Issue #522, opened 2019-02-24
Rocket shows very different CG in 15.03 and current unstable


JoePfeiffer:
Here's a simple DD 3FNC rocket of mine:
[airframe.ork.zip](https://github.com/openrocket/openrocket/files/2898357/airframe.ork.zip)
CG location under 15.03 is at 29.655"
![or15 03](https://user-images.githubusercontent.com/3065196/53304856-89564580-3837-11e9-89fc-b46ead3f06e3.png)
CG under current unstable branch is at 16.81"
![unstable](https://user-images.githubusercontent.com/3065196/53304878-e0f4b100-3837-11e9-8441-e1d703603df2.png)

Starting to look into it now...

teyrana:
closed as resolved.

teyrana:
Closing as resolved.

PR #523, opened 2019-03-08
Fix 522

JoePfeiffer:
The main thing this does is separate calculation of motor mass data from calculation of structure mass data, addressing issue #522.  It also fixes some comments to match code, and fixes an error in calculating moments of inertia when a component's mass and CG or overridden while overriding its children's mass and CG (an extra component was being included in the MOI calculation).  The testFalcon9HeavyBoosterStageMassOverride expected MOI is modified to reflect this.


teyrana:
Reviewed.  +1

PR #529, opened 2019-03-13
Fix issue 527

JoePfeiffer:
This resolves the issue, but I'm not really happy with adding three methods which are called in exactly one place each, and which couldn't really be refactored from the validating versions of the same methods (so it ends up with some duplicate code).  The problem does need to be resolved somehow, so if someone sees a better way to do it I'm all ears!


JoePfeiffer:
No, let's think about this some more.

PR #530, opened 2019-03-13
Fix #527

### LINKED ISSUE #527: fin tabs loaded incorrectly.
When saving and then reading a design with TTW fins, my fin tabs are only 1.969 inches long.  Looking at the .ork file I find <tablength>0.20319999999999994</tablength> (which is the 8 inches I set them to), so it looks like the problem is in loading the design.

JoePfeiffer:
This is the cleanest way to fix it that I'm seeing at present...  essentially, make the three methods setting tab parameters private and give them a second parameter saying whether or not to validate the fin tab.

Create two public wrappers for each, one with and one without validation.

Hook up file loaders to the NoValidate wrapper for each.


teyrana:
There are a couple ways to approach this, but I'm not sure any of the rest are better...

1. The way you've done: multiple functions, overloaded, with optional validating
2. Validating doesn't happen in the setTabHeight etc functions -- the caller calls setTabHeight(..), then the validate function.
3. guarantee the order of save-files (and I think *that* horse is out of the barn)

JoePfeiffer:
Yeah, I went down several garden paths before I decided on the way I did it in the PR.

The problem with your (2) is the way calls to the setters are constructed in the configuration popups.  Those calls don't explicitly appear in the code, they're constructed by taking the name of the element being configured (for instance, "Tab length"), massaging it ("TabLength") and putting "set" in front of it ("setTabLength").  I've got no idea how I'd put a separate validation call in there.

Your (3) would require the least code modification, but requiring XML elements to appear in a particular order seems like a trap for a later maintainer.  Also, of course, to the user this would present as "I loaded my file and the fin tabs were wrong, but then when I saved it and loaded it again it worked" (assuming they loaded a file that had been saved out of order the first time); now I can't duplicate it".

There's also

4.  Write a custom handler like AtmosphereHandler, and have it manipulate the tab directly.  I really don't like modifying a parameter in multiple places

5.  Validate when a fin tab parameter is read, rather than written.  So you'd just save the requested value no matter what, and then restrict it as necessary as part of the getter.  This was actually my second favorite approach, but it seemed like having the fin tab track the fin length up to some value (the original length requested by the user), and not past that, would be confusing.

teyrana:
I think (2) is fixable, as long as "validate()" is defined for all of those classes.

I'm actually starting to wonder whether we need the validate on the fun properties, at all, actually....

I mean, whats the downside if the fin tab is longer than the fin?

JoePfeiffer:
Hmmm, you raise a good point.  Would it make sense to drop the validation, but add two flags for automatic length and height -- if "true", the tab length tracks the fin length, and the height tracks the distance from inner to outer tube... that would match my typical use case better than validation and the "calculate automatically" button.  Shall I pursue that?

teyrana:
¿Automatic full length tabs?

If you think so, I don't really object. I've never had a build with them,
myself.

Is that something currently in the codebase?  Won't it require UI fixes?

JoePfeiffer:
Interesting -- that's the way I always do it, since just coming straight down from the root of the fin is easier than making any sort of cut to have a shorter tab.

No, it isn't in the current codebase, but would be easy to implement.  Maybe go ahead and commit the current PR, and think about it for the future?

teyrana:
>  Would it make sense to drop the validation, but add two flags for automatic length and height 

Yea, I think those are the two PRs we should be talking about: 

1. (this one) "drop the validation" 

2. "add two flags for automatic length and height"

This particular PR is well implemented, but (IMHO) it solves a problem we don't need to solve.  Let's just not worry about fin-tab-validation. :)

JoePfeiffer:
No, it solves a problem we definitely need to solve -- at the moment, when we save and then load a design, the fin tab ends up wrong, which needs to be fixed.  I could easily rework it to drop fin tab validation all together if that's preferable.  Replacing the "calculate" button with a couple of flags would also be straightforward, but should wait until after we get a release out.

teyrana:
On Tue, Mar 19, 2019, 12:48 PM Joe Pfeiffer <notifications@github.com>
wrote:

> No, it solves a problem we definitely need to solve -- at the moment, when
> we save and then load a design, the fin tab ends up wrong, which needs to
> be fixed. I could easily rework it to drop fin tab validation all together
> if that's preferable.
>
Yea, that's what I was trying to say.  I thino it would be simpler, and
more clear to a user, to simply drop the unnecessary validation, and load
exactly what the file says.

Replacing the "calculate" button with a couple of flags would also be

JoePfeiffer:
That's what I'm doing here -- no validation on file loads, but maintains it on GUI interactions.  The other alternative I mentioned would be to drop it in the GUI too.

JoePfeiffer:
Latest version removes validation when modifying fin tab in GUI.

teyrana:
Awesome!  :+1:

PR #533, opened 2019-03-23
Fixes #525; Main Figure display rocket diameter again.

### LINKED ISSUE #525: Rocket max diameter is 0
The rocket max diameter, as reported in the second line of the RocketInfo pane, is always 0.  This doesn't seem like another manifestation of #329, since the other items described in the pane (length, mass with motors, mass without motors) do get updated correctly, and it looks (from a cursory look) like they're all updated in the same method.

teyrana:
The displayed diameter is also restricted to centerline components, now.
Also: corrects rocket length to be the overall max x span of the rocket.


JoePfeiffer:
Reworked version looks good.

PR #534, opened 2019-03-29
Modified Hypertek_I205 Length

JKPOKE:
Modified Hypertek_I205 Length
521mm → 546mm


JoePfeiffer:
The motor database is downloaded from thrustcurve.org when OpenRocket is compiled.  Checking...  yes, the length of that motor has been corrected there, so it will be correct in the next release (someday...) of OpenRocket.

teyrana:
Close as resolved.  
This fix will be in the next release.

PR #543, opened 2019-07-17
Fix 525

JoePfeiffer:
This is a much, much bigger PR than I like to make, but I didn't see a good way to break it up in smaller increments that made sense.  In essence, the deprecated getComponentBounds function was being used to compute lengths, and it was a mess:  I'm not sure it was guaranteed to produce a convex hull in some places; it was inconsistent as to whether it was computing a given component's bounds relative to the rocket or relative to the component, and more.

So I've ripped it out, and finished replacing it with getBoundingBox(), iterating through the rocket components using Dan's InstanceMap.


teyrana:
I'm pleased to see that the BoundingBox class is useful. It's a lovely bit of validation that you actually picked it over the previous design choice :)

On a more serious note, 90% of this is just refactoring APIs, and I think those are migrating to a clearer, more succinct form.   So, fwiw: 👍 +1

PR #546, opened 2019-09-10
Maintain component position when changing axial method. Fixes #541

### LINKED ISSUE #541: Changing "Position Relative To" setting creates unexpected results
After discussion on the dev mailing list, I'm reporting this as an issue.

In the previous release version, if you changed the "Position Relative To" option, the offset value would be automatically adjusted so that the component stays in the same place on the rocket.

In the development versions of the new release, this has changed.  Now, when you change the "Position Relative To" option, the offset value is not adjusted, and so the component moves.  For example, fins previously positioned at the bottom of the rocket move to the top.

It is unclear if this was an intentional change or not.  Either way, it seems unanimous among those who gave their opinion that the old behavior was preferred.

bweissinger:
This addresses #541. It keeps the component in the same location and modifies the offset value when the axial method is changed. It also maintains the current behavior for events such as parent length changes.

I don't like the use of `axialMethodChange`, but as far as I can tell, within `setAxialOffset()` there is no other way to tell that the axial method was changed. My first thought was to check if `this.axialMethod == requestedMethod`. However, `axialMethod` has already been changed before that function is called, so they are always the same.

Still getting acquainted with the code base. Thoughts?

Edit: `this.axialMethod = requestedMethod;` was left out in `setAxialOffset()`, cause of the majority of failed tests.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/RocketComponent.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/RocketComponent.java
@@ -89,6 +89,9 @@ public abstract class RocketComponent implements ChangeSource, Cloneable, Iterab
 	 * Defaults to (0,0,0)
 	 */
 	protected Coordinate position = new Coordinate();
+
+	// A flag to signal the axial method has changed
+	private boolean axialMethodChange = false;
 	
 	// Color of the component, null means to use the default color
 	private Color color = null;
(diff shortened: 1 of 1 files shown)


teyrana:
Overcome by events.

This issue this PR fixes has since been resolved.

PR #559, opened 2019-11-22
[Feature] Added velocity initial condition to simulation options

SilentSys:


teyrana:
Hey, sorry it took so long to get to this.  

1. I'm not really clear on the purpose of the PR... when you simulate the second stage, do you define the booster AND the sustainer in the same .ork file?  Or are do you just simulate the sustainer in a standalone file? 

2. Most of the changes contained here seem to support setting an inertial vertical velocity.  Is that all you guys needed? 

3. Do you have a test case? or can you send over an example rocket that you would use with this code?

SilentSys:
Sorry it took me so long as well. I haven't been working with this anymore.

1. Yes the use case was to simulate the sustainer alone, without the booster in the file.
2. Yes, initial velocity was all we needed.
3. Sorry, as I am no longer working on the project I don't have any test cases or files I can send. But I think as you can imagine there are other use cases where this may be useful as well: e.g. spring-loaded pre-ignition assist.

JoePfeiffer:
I don't understand the advantage of putting an initial vertical velocity in simulation conditions, rather than simply simulating the whole flight including both stages.  Should this be closed?

PR #564, opened 2020-02-15
Revert "Fix 525" -- PR causes Stack Overflow when simulating other rockets.

teyrana:
@JoePfeiffer 

Hey, I was taking a look at another bug, and discovered openrocket was exhausting the stack, merely on opening a rocket :/  (Probably due to the auto-simulation done in the background.)

It's pretty much a breaking change, so If we can't find a solution, we'll have to revert the whole thing.  Do you have time to take a look at your PR #543 with the .ork below, and see what happened? 

[Inducing .ork file]( https://github.com/openrocket/openrocket/files/3095876/alcubierre_cranberry.zip )


JoePfeiffer:
Ouch!  It'll be a couple of days before I can spend any real time looking at it, so please revert.

Issue #565, opened 2020-02-15
Canted fins do not display correctly


teyrana:
To reproduce, open the 'canted fin rocket' example from the 'file' menu.

Fixed in: #566

PR #566, opened 2020-02-15
[review-requested][fix] Adds explicit Angle Unit Conversion to/from FinSet cant and base angles.

teyrana:
## Ready for Review - comments welcome
- Discovered while debugging #538 
- Also addresses: #565


teyrana:
(rebased on top of `unstable` branch.)

PR #569, opened 2020-03-02
Fix large rail button default size.

### LINKED ISSUE #554: Default rail button size is ginormous
Default outer diameter and total height are 1 meter, which is big enough for a Falcon 9 I think. :)

image

(also there is no parts library for buttons at this point).

derived from issue #550.

wolsen:
The default values assigned to RailButton sizes were configured with
a 1.0 meter outer diameter and 0.8 meter internal diameter (for 1010
rail buttons) which resulted in overly large rail buttons when added
using the default values.

Presuming that 1010 rail buttons are the common and an acceptable
default, change the default button sizes to align with the 1010 size.

Additionally, this change fixes some of the layout issues present in
add railbutton dialog.

Closes #554


teyrana:
@wolsen 
FYI: if you push a new commit to the same branch as an open pull request, it automatically updates the pull request.  If that's what you're looking for.

Closing as duplicate of #570.

PR #570, opened 2020-03-02
Add railbutton presets

### LINKED ISSUE #554: Default rail button size is ginormous

wolsen:
Allow rail buttons to define presets. This change doesn't actually add
any default presets to the library, but enables the ability to add some
preset options for rail buttons.

Partially Closes #554


teyrana:
@wolsen @neilweinstock

Will this do?: 
[OpenRocket-wolsen:fix-railbuttons.jar](https://drive.google.com/open?id=1V5Q6QQphOiGfArVK4idaYgP_qHsPfRpR)

swing/src/net/sf/openrocket/gui/configdialog/RailButtonConfig.java
```diff
@@ -90,20 +97,30 @@ private JPanel buttonTab( final RailButton rbc ){
...
+			panel.add(new BasicSlider(offsetModel.getSliderModel(
+					new DoubleModel(component.getParent(), "Length", -1.0, UnitGroup.UNITS_NONE),
+					new DoubleModel(component.getParent(), "Length"))),
+					"w 100lp, wrap para");
 			
 		}
 		
 		//// Material
-		panel.add( instanceablePanel(rbc), "cell 4 0, spany 3, wrap para");
+		/* TODO (wolsen) confirm this removal
```
teyrana: @wolsen 
Hey, rather than making a unilateral decision about this, I'd like to poll the list.

I understand your opinion. And I have the opposite opinion.
In addition, users don't *need* to use this panel.  Adding two separate rail buttons still works as expected.

teyrana:
Reviewed. Most of the changes look good! 

The removal of the instancing panel is not something I'd like to do unilaterally, though.  (1) It's significant enough that it ought to be mentioned in the commit messages, btw. (2) because we have differing opinions on this, I'll ask on the mailing list as a tie-breaker.

neilweinstock:
Personally I like the instancing panel and would keep it.  The reason is simple: Rockets are going to have exactly two buttons roughly 99% of the time.  Why shouldn’t the program make it easier to enter that way?

What is the reasoning for removing it?

swing/src/net/sf/openrocket/gui/configdialog/RailButtonConfig.java
wolsen: @teyrana 

Fair enough. I actually didn't intend to remove it as part of this PR. In my previous git commit, you can clearly see a comment with a TODO from me trying to find out if we should remove it. You've sent an email to the mailing list, so I'll respond in kind there.

wolsen:
Added back the removed instance panel option, but did do some slight tweaking to get the layout from 3 columns back to 2. There's a few UI spacing issues that could be improved upon in a future round.\\
Screenshot from 2020-03-03 09-40-15

teyrana:
@wolsen 
Hey, I didn't intend to call you out or anything.  I'm sorry if it came off that way.  

We honestly haven't had a lot of discussion about the instancing feature yet, and I wanted to get a sense of the community before pushing any point-of-view.   (feel free to follow up offline, if you want: equipoise (at) gmail )

Anyway, the whole PR looks good!  Thanks for contributing!

p.s. The instance options do also appear in Pods, Boosters, and FinSets.  Mind you, it looks different, of course.  Check it out, if you like:

booster-instanceing
(see the 'number of copies' field.)

wolsen:
@teyrana no worries :-) It was good to call it out, it honestly was unintentional to remove it. Thanks for working with me to merge this PR. I'll be coming back at ya with some other PRs after I look through the bug list a bit and figure out which one I want to tackle next. Still getting my feet wet with how this particular community works, but I'm not new to Open Source and realize that everyone has differences of opinion and mine is not always the right one. We work better as a community to come up with something that serves the largest percentage of users we can.

PR #577, opened 2020-03-06
Fix ground hit velocity calculation

### LINKED ISSUE #576: Ground hit velocity reported as 0
On the tip of the unstable branch, the ground hit velocity is reported as 0.

This appears to be a regression introduced by commit f11a3e4. That commit introduced a new groundstepper to the simulation which adds additional flight data with 0.0 as a TYPE_VELOCITY_TOTAL. This causes in turn causes the FlightData.calculateIntrestingValues() to return 0, since it simply grabs the last total velocity of the flight data.

wolsen:
The ground hit velocity is calculated by taking the last recorded
total velocity for the flight data and using that as the ground
hit velocity if the last altitude measurement is less than 10 m.

This method worked until commit f11a3e4 introduced the ability to
track and report all events that happened after a GROUND_HIT event.
That commit introduced a GroundStepper that will step between the
GROUND_HIT event and the SIMULATION_END event and will always
add a measurement to the flight data to record the rocket being on
the ground.

A better way of measuring the ground hit velocity is to interpolate
the velocity at the time the GROUND_HIT event, as that would actually
be the appropriate ground hit velocity.

Closes #576

### DIFF
--- a/core/src/net/sf/openrocket/simulation/FlightData.java
+++ b/core/src/net/sf/openrocket/simulation/FlightData.java
@@ -195,13 +195,7 @@ public class FlightData {
 		maxMachNumber = branch.getMaximum(FlightDataType.TYPE_MACH_NUMBER);
 		
 		flightTime = branch.getLast(FlightDataType.TYPE_TIME);
-		if (branch.getLast(FlightDataType.TYPE_ALTITUDE) < 10) {
-			groundHitVelocity = branch.getLast(FlightDataType.TYPE_VELOCITY_TOTAL);
-		} else {
-			groundHitVelocity = Double.NaN;
-		}
 		
-
 		// Time to apogee
 		List<Double> time = branch.get(FlightDataType.TYPE_TIME);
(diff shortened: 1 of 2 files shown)


JoePfeiffer:
This fixes my error (thanks), and also makes it consistent with the velocity calculation of other events.

teyrana:
Hey, I like the idea involved here, but I couldn't find "ground hit velocity" in the UI.  

Where would a user see this?
What test data / test case shows the difference in behavior? (either .ork or unittest)

wolsen:
@teyrana I’ll look at adding a unit test. That is a very fair ask and should have been included.

The ground hit velocity is the far right column of the simulation results table. You can recreate this by loading one of the example rockets and running the a simulation. The ground hit velocity is always 0 after the aforementioned commit, where it would normally report the velocity of the rocket as it touched down (i.e. 19.7 FPS etc)

teyrana:
confirmed -- passes under java 8.

PR #578, opened 2020-03-07
Fixing deprecated macOS classes and adjusting JarInJarStarter classpath

itshanney:
@teyrana can you please take a look at this and incorporate as I now have a running version of your Java11/Java13 updates on macOS.


itshanney:
@teyrana  broke the build so will try again with a new PR after verifying it doesn't break locally.

JarInJarLoader.java class loader fix works though if you want to make that change.

PR #587, opened 2020-03-11
Upgrade/ci/xenial

teyrana:


teyrana:
Useful, but the bionic image runs our tests, too, and that's more up-to-date.

PR #589, opened 2020-03-12
Fix macOS after java11 upgrade

itshanney:
Switches to the java.awt.desktop package as required by JEP 272 when upgrading beyond Java 9


itshanney:
@teyrana Fix to get OSXSetup.java to compile correctly on macOS after the update to Java 11. Note this will be the official dropping of Java 8 due to JEP 272.

teyrana:
hey, I don't have a osx machine handy  to test these, so I'll settle for making sure they don't break the build on other OSes.

Can you briefly describe what issues you're addressing here? Is it literally just the builds?

itshanney:
@teyrana  take a look at the diff in OSXSetup.java. You see that the old version was using classes from com.apple.eawt package. That package is officially deprecated and replaced with java.awt.desktop package as specific in JEP 272 (http://openjdk.java.net/jeps/272).

This change needs to be done so you can compile on macOS as those packages no longer exist. From the user perspective on a non-macOS environment, it should be equivalent to a no-op and they won't see anything. For macOS developers, they will be able to compile.

*Note:* This will be the official dropping of Java 8 because you can't compile with a compiler older than Java 9. Since we're all syncing on Java 11, it shouldn't be a big deal, but we will have to make note of it to the developer list.

teyrana:
So this just fixes build-time errors, yes? 

And if it compiles elsewhere, then that's a reasonable validation test?

itshanney:
Yes, it fixes build errors. Any developer not on macOS won't notice it, but any who are will notice it's broken because that older API is unavailable.

Yes, so long as it compiles elsewhere, that should be a sufficient validation test.

teyrana:
> Yes, it fixes build errors. ...
> so long as it compiles elsewhere, that should be a sufficient validation test.

Okay, now I understand.  Thanks for bringing me up to speed :)

I've verified the build and run with windows + ant + java11.  Looks good!  merging.

PR #591, opened 2020-03-15
[Fixes #531] Fix printing/export as pdf

### LINKED ISSUE #531: Uncaught exception printing design
When I try to "Print/Export PDF" for the attached design, I'm getting an uncaught exception "IllegalStateException:  Attempted to add motor data with an invalid fcid".  The pdf file is created, but is empty.
[airframe.ork.zip](https://github.com/openrocket/openrocket/files/2971886/airframe.ork.zip)

wolsen:
This pull requests addresses a few issues with the print/export as PDF function. As identified in #531, an attempt to print or export as pdf will result in an IlllegalStateException being raised.

1. Correct the logic for detecting if a motor has an invalid FlightConfigurationId and throw an IllegalStateException only in the case that it has an error.
2. Fix the lookup for the stageMass. The wrong value was passed to the MassCalculator.getCGAnalysis(..) which resulted in a NullPointerException.
3. Fix issues with drawing rocket diagram in print/pdf
4. Clean up some of the code in DesignReport, including some refactoring TODOs in the addMotorData method.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
@@ -315,12 +315,6 @@ public class RocketFigure extends AbstractScaleFigure {
 		
 	}
 	
-	@Override
-	public void updateFigure() {
-		updateShapes(this.figureShapes);
-		super.updateFigure();
-	}
-	
 	public RocketComponent[] getComponentsByPoint(double x, double y) {
 		// Calculate point in shapes' coordinates
 		Point2D.Double p = new Point2D.Double(x, y);
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
@@ -315,6 +315,12 @@ public class RocketFigure extends AbstractScaleFigure {
 		
 	}
 	
+	@Override
+	public void updateFigure() {
+		updateShapes(this.figureShapes);
+		super.updateFigure();
+	}
+	
 	public RocketComponent[] getComponentsByPoint(double x, double y) {
 		// Calculate point in shapes' coordinates
 		Point2D.Double p = new Point2D.Double(x, y);
--- a/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
@@ -372,6 +372,11 @@ public class BodyTube extends SymmetricComponent implements MotorMount, Coaxial 
 	}
 	
 	@Override
+	public MotorConfigurationSet getMotorConfigurationSet() {
+		return this.motors;
+	}
+	
+	@Override
 	public MotorConfiguration getMotorConfig( final FlightConfigurationId fcid){
 		return this.motors.get(fcid);
 	}
@@ -432,7 +437,7 @@ public class BodyTube extends SymmetricComponent implements MotorMount, Coaxial 
 		
 	@Override
 	public int getMotorCount() {
-		return this.motors.size();
+		return this.getClusterConfiguration().getClusterCount();
 	}
 		
 	@Override
--- a/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
@@ -105,7 +105,7 @@ public class FlightConfiguration implements FlightConfigurableParameter<FlightCo
 		this._setAllStages(true);
 		this.updateMotors();
 	}
-		
+
 	private void _setAllStages(final boolean _active) {
 		for (StageFlags cur : stages.values()) {
 			cur.active = _active;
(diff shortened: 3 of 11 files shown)


wolsen:
Marked as work in progress. There's a few additional changes needed and required and will take some time for me to study them and write the unit tests

wolsen:
@teyrana and others - would love some feedback on this pull request before I spend more time on it. I'm going to look writing some tests around the printing stuff. The patch I'm most iffy about in this patch set is the last patch, in which the scaleFactor was added back in for the PrintableNoseCone scenario. 

Other transition shapes and fin shapes appear to work because the shapes are built within the code that's printing the PDF. The PrintableNoseCone delegates the shape behavior to the TransitionShape.getShapeSides(...) method - which I think is the *right* thing to do in order to not duplicate the code, however the scaleFactor is not taken into consideration in the TransitionShapes.

swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
```diff
@@ -315,6 +315,12 @@ public void paintComponent(Graphics g) {
 		
 	}
 	
+	@Override
+	public void updateFigure() {
+		updateShapes(this.figureShapes);
```
teyrana: if you're updating the shapes on every change, does that make the `updadeShapes` call at [194 ](https://github.com/openrocket/openrocket/blob/ba37faa16eda9344190177d60c51b56c74474a75/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java#L194)  redundant?

teyrana:
Ohhh, there's a lot of good fixes in here! -- Some great changes, and some that I'm a little bit less clear on.

Okay, so how are you using scaleFactor, here?  We see it a bunch of places.   Is it the scaling of an individual component? Or the zoom? or a dpi scaling for 1m => 100% zoom on the print-out? Or something else? 

nit: Ideally, all the changes to `core/src/` would be more distinct from all the UI changes, but I'm not sure it's worth stressing about, atm.

swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
wolsen: Good question, I'll have to look at it a bit more and get back to you. I originally added this in as the figure was lacking any of the figureShapes until the paintComponent. I left the call on line 194 b/c the updateShapes method will also handle any transform changes which are applied as part of the paintComponent method. However, I'll look at this code path a bit more and come back with more information.

wolsen:
> Ohhh, there's a lot of good fixes in here! -- Some great changes, and some that I'm a little bit less clear on.
> 
> Okay, so how are you using scaleFactor, here? We see it a bunch of places. Is it the scaling of an individual component? Or the zoom? or a dpi scaling for 1m => 100% zoom on the print-out? Or something else?

In the PrintableNoseCone [draw](https://github.com/openrocket/openrocket/blob/ba37faa16eda9344190177d60c51b56c74474a75/swing/src/net/sf/openrocket/gui/print/PrintableNoseCone.java#L61) logic, the shapes are collected using a scaling factor (which is ultimately the number of display points per meter). Since all of the values for length are in meters, when rendering the objects were _tiny_ and appeared invisible.

Ideally, this would switch over to use the scaling that's built into the RocketFigure itself (like the DesignReport does). All of the printing logic could be converted over to using the RocketFigure's paint logic, which would ensure consistency between what is displayed and what is printed (no weird bugs due to differing implementations).

I actually have a pretty good idea on how to do this, however, in the effort to reduce the amount of code change/design change at this point I chose to simply add back in the scaleFactor that was mostly removed in the limited shapes that need it for the PrintableNoseCone. This was the lesser of the two changes in terms of size and risk IMO.

I'm happy to work to combine this logic after a release is out the door. Hopefully this answers your questions.

> nit: Ideally, all the changes to `core/src/` would be more distinct from all the UI changes, but I'm not sure it's worth stressing about, atm.

In general I agree here. I originally had many of these changes as one big blob and split them up into logical chunks and kept these together to allow the code to fully compile at this commit. This aids in some debug activities like a git bisect, which I frequently use to determine when a breaking change was introduced and to better understand what the change was that caused the breakage.

swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
wolsen: @teyrana after further review and testing, overriding the updateFigure method is not necessary here. I'll push an additional commit to address this.

teyrana:
@wolsen 
Hey, this PR is having trouble building under intellij -- but it was quickly solved by rebasing to the tip of 'unstable'.   And from _*there*_ it looked good :)

Would you mind doing that, and re-pushing to the source branch, above? If so, I'm cool with merging.

wolsen:
> @wolsen
> Hey, this PR is having trouble building under intellij -- but it was quickly solved by rebasing to the tip of 'unstable'. And from __there__ it looked good :)
> 
> Would you mind doing that, and re-pushing to the source branch, above? If so, I'm cool with merging.

done

PR #594, opened 2020-03-19
Minor issues in InstanceContext

JoePfeiffer:
Added missing instanceNumber parameter to String.format() call in toString()

Added explicit call to component.toString() so intellij can print component name


teyrana:
o.O
Oh, my.

PR #597, opened 2020-03-19
Fixes #540  - Fins are created with a default count of 3

### LINKED ISSUE #540: Default number of instances for fins should be 3
At some point in the new release development, the default number of fins changed from 3 to 1.  I don't think 1 makes a lot of sense as default, should go back to 3 IMHO.

teyrana:


wolsen:
LGTM +1

PR #599, opened 2020-03-21
Remove OrangeExtensions-1.2.jar from eclipse classpath

wolsen:
Commit a7c0cb2 removed the OrangeExtensions-1.2.jar file from the
swing project, but failed to remove it from the eclipse classpath.
This fixes that.


teyrana:
Ah. Good catch.  +1

PR #602, opened 2020-03-23
[Fixes 452] Reflect length of selected stages

wolsen:
Fix the length of the rocket displayed based on the selected stages. This change switches from calculating the bounds of the rocket to using the length already calculated in the flight configuration, which properly reflects the length of active stages and components.

Additionally, while debugging and testing this - I noticed that the display toggle buttons are always set as selected. The second commit changes the button to reflect the state of the active stage.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/components/StageSelector.java
+++ b/swing/src/net/sf/openrocket/gui/components/StageSelector.java
@@ -10,21 +10,17 @@ import javax.swing.JPanel;
 import javax.swing.JToggleButton;
 
 import net.miginfocom.swing.MigLayout;
-import net.sf.openrocket.l10n.Translator;
 import net.sf.openrocket.rocketcomponent.AxialStage;
 import net.sf.openrocket.rocketcomponent.ComponentChangeEvent;
 import net.sf.openrocket.rocketcomponent.FlightConfiguration;
 import net.sf.openrocket.rocketcomponent.Rocket;
 import net.sf.openrocket.rocketcomponent.RocketComponent;
-import net.sf.openrocket.startup.Application;
 import net.sf.openrocket.util.StateChangeListener;
 
 
 @SuppressWarnings("serial")
 public class StageSelector extends JPanel implements StateChangeListener {
 
-	private static final Translator trans = Application.getTranslator();
-	
 	private final Rocket rocket;
 	
 	private List<JToggleButton> buttons = new ArrayList<JToggleButton>();
@@ -41,7 +37,7 @@ public class StageSelector extends JPanel implements StateChangeListener {
 		this.removeAll();
 		for(AxialStage stage : configuration.getRocket().getStageList()){
 			JToggleButton button = new JToggleButton(new StageAction(stage));
-			button.setSelected(true);
+			button.setSelected(configuration.isStageActive(stage.getStageNumber()));
 			this.add(button);
 			buttons.add(button);
 		}
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
@@ -606,20 +606,8 @@ public class RocketPanel extends JPanel implements TreeSelectionListener, Change
 		figure3d.setCG(cg);
 		figure3d.setCP(cp);
 
-		// Length bound is assumed to be tight
-		double length = 0;
-		Collection<Coordinate> bounds = curConfig.getBounds();
-		if (!bounds.isEmpty()) {
-			double minX = Double.POSITIVE_INFINITY, maxX = Double.NEGATIVE_INFINITY;
-			for (Coordinate c : bounds) {
-				if (c.x < minX)
-					minX = c.x;
-				if (c.x > maxX)
-					maxX = c.x;
-			}
-			length = maxX - minX;
-		}
-
+		double length = curConfig.getLength();
+		
 		double diameter = Double.NaN;
 		for (RocketComponent c : curConfig.getCoreComponents()) {
 			if (c instanceof SymmetricComponent) {
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -714,11 +714,32 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 		final double finLength = singleFinBounds.max.x;
 		final double finHeight = singleFinBounds.max.y;
 		
-		BoundingBox compBox = new BoundingBox().update(getComponentLocations());
+		Coordinate[] locations = getInstanceLocations();
+		double[] angles = getInstanceAngles();
+		BoundingBox finSetBox = new BoundingBox();
 		
-		BoundingBox finSetBox = new BoundingBox( compBox.min.sub( 0, finHeight, finHeight ), 
-												compBox.max.add( finLength, finHeight, finHeight ));
-		return finSetBox; 
+		/*
+		 * The fins themselves will be offset by the location itself to match
+		 * the outside radius of a body tube. The bounds encapsulate the outer
+		 * portion of all the tips of the fins.
+		 * 
+		 * The height of each fin along the Y axis can be determined by:
+		 * yHeight = cos(angle) * finHeight
+		 * 
+		 * The height (depth?) of each fin along the Z axis can be determined by:
+		 * zHeight = sin(angle) * finHeight
+		 * 
+		 * The boundingBox should be that box which is the smallest box where
+		 * a convex hull will contain all Coordinates.
+		 */
+		for (int i = 0; i < locations.length; i++) {
+			double y = Math.cos(angles[i]) * finHeight;
+			double z = Math.sin(angles[i]) * finHeight;
+			finSetBox.update(locations[i].add(0, y, z));
+			finSetBox.update(locations[i].add(finLength, y, z));
+		}
+		
+		return finSetBox;
 	}
 	
 	/**
--- a/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
@@ -1,6 +1,7 @@
 package net.sf.openrocket.rocketcomponent;
 
 import java.util.ArrayDeque;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -551,16 +552,71 @@ public class FlightConfiguration implements FlightConfigurableParameter<FlightCo
 		return cachedBounds;
 	}
 
+	/**
+	 * Calculates the bounds for all the active component instances
+	 * in the current configuration.
+	 */
 	private void calculateBounds(){
 		BoundingBox bounds = new BoundingBox();
-			
-		for (RocketComponent component : this.getActiveComponents()) {
-			BoundingBox componentBounds = new BoundingBox().update(component.getComponentBounds());
-			bounds.update( componentBounds );
-		}
 
+		InstanceMap map = getActiveInstances();
+		for (Map.Entry<RocketComponent, java.util.ArrayList<InstanceContext>>  entry : map.entrySet()) {
+			RocketComponent component = entry.getKey();
+			List<InstanceContext> contexts = entry.getValue();
+						
+			Collection<Coordinate> coordinates = new ArrayList<Coordinate>();
+			/* FinSets are a bit different in how they store their bounds,
+			 * so we'll use the fin points for calculations. 
+			 */
+			if (component instanceof FinSet) {
+				coordinates.addAll(Arrays.asList(((FinSet) component).getFinPoints()));
+			} else {
+				coordinates.addAll(component.getComponentBounds());
+			}
+			BoundingBox componentBox = new BoundingBox();
+			List<Coordinate> transformedCoords = new ArrayList<Coordinate>();
+			for (InstanceContext ctxt : contexts) {
+				/*
+				 * If the instance is not active in the current context, then
+				 * skip the bound calculations. This is mildly confusing since
+				 * getActiveInstances() implies that it will only return the
+				 * instances that are active, but it returns all instances and
+				 * the context indicates if it is active or not.
+				 */
+				if (!ctxt.active) {
+					continue;
+				}
+				for (Coordinate c : coordinates) {
+					Coordinate tc = null;
+					/* These components do not need the transform performed in
+					 * order to provide the proper coordinates for length calculation.
+					 * The transformation will cause the values to be calculated
+					 * incorrectly. This should be fixed in the appropriate places
+					 * not handled as one-offs in here.
+					 */
+					if ((component instanceof AxialStage) || (component instanceof BodyTube) ||
+						(component instanceof PodSet)) {
+						tc = c;
+					} else {
+						tc = ctxt.transform.transform(c);
+					}
+					componentBox.update(tc);
+					transformedCoords.add(tc);
+				}
+			}
+			
+			bounds.update(componentBox);
+		}
+		
 		boundsModID = rocket.getModID();
 		cachedLength = bounds.span().x;
+		/* Special case for the scenario that all of the stages are removed and are
+		 * inactive. Its possible that this shouldn't be allowed, but it is currently
+		 * so we'll just adjust the length here.  
+		 */
+		if (cachedLength == Double.POSITIVE_INFINITY || cachedLength == Double.NEGATIVE_INFINITY) {
+			cachedLength = 0;
+		}
 		cachedBounds.update( bounds );
 	}
 	
--- a/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
@@ -565,11 +565,11 @@ public class FlightConfiguration implements FlightConfigurableParameter<FlightCo
 			List<InstanceContext> contexts = entry.getValue();
 						
 			Collection<Coordinate> coordinates = new ArrayList<Coordinate>();
-			/* FinSets are a bit different in how they store their bounds,
-			 * so we'll use the fin points for calculations. 
+			/* FinSets already provide a bounding box, so let's use that.
 			 */
 			if (component instanceof FinSet) {
-				coordinates.addAll(Arrays.asList(((FinSet) component).getFinPoints()));
+				bounds.update(((FinSet) component).getBoundingBox());
+				continue;
 			} else {
 				coordinates.addAll(component.getComponentBounds());
 			}
@@ -614,7 +614,7 @@ public class FlightConfiguration implements FlightConfigurableParameter<FlightCo
 		 * inactive. Its possible that this shouldn't be allowed, but it is currently
 		 * so we'll just adjust the length here.  
 		 */
-		if (cachedLength == Double.POSITIVE_INFINITY || cachedLength == Double.NEGATIVE_INFINITY) {
+		if (getActiveStages().isEmpty()) {
 			cachedLength = 0;
 		}
 		cachedBounds.update( bounds );
--- a/core/test/net/sf/openrocket/rocketcomponent/FlightConfigurationTest.java
+++ b/core/test/net/sf/openrocket/rocketcomponent/FlightConfigurationTest.java
@@ -44,7 +44,7 @@ public class FlightConfigurationTest extends BaseTestCase {
 		// preconditions
 		assertThat("active stage count doesn't match", config.getActiveStageCount(), equalTo(2));
 
-		final double expectedLength = 0.33;
+		final double expectedLength = 0.335;
 		final double calculatedLength = config.getLength();
 		assertEquals("source config length doesn't match: ", expectedLength, calculatedLength, EPSILON);
 
@@ -70,7 +70,7 @@ public class FlightConfigurationTest extends BaseTestCase {
 		int expectedMotorCount = 2;
 		int actualMotorCount = config1.getActiveMotors().size();
 		assertThat("active motor count doesn't match", actualMotorCount, equalTo(expectedMotorCount));
-		double expectedLength = 0.33;
+		double expectedLength = 0.335;
 		assertEquals("source config length doesn't match: ", expectedLength, config1.getLength(), EPSILON);
 		double expectedReferenceLength = 0.024;
 		assertEquals("source config reference length doesn't match: ", expectedReferenceLength, config1.getReferenceLength(), EPSILON);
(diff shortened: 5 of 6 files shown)


wolsen:
I noted that the stage selection has a different behavior than the one in the 15.03 release. In the 15.03 (stable) release, selecting a stage will activate all stages up to and including the stage that was selected. The current (unstable) branch allows different stages to be toggled on or off - allowing for gaps in the rocket that is displayed. I can't quite tell if that change was intentional or not based on the commit history. If it should be restored to the 15.03 behavior, that's a simple change that I can propose in a subsequent pull request.

JoePfeiffer:
My hazy recollection is we wanted to be able to look at the design on a stage-by-stage basis; the intent was you could either pick the rocket or pick a stage.  Sounds like the way it was implemented was too general.

wolsen:
Thanks for the clarification - it sounds like its working as intended in unstable currently which IMHO is a better behavior than the original.

teyrana:
Yes, the intent was to toggle the stages individually, and independently--
I.e. on a per stage basis.   (And yes, it's currently buggy :/)

wolsen:
Great thanks @teyrana - good news is that this fixes a couple of bugs :-)

teyrana:
1. +1 for fixing the stage-toggle buttons
2. I'm seeing the bounds being calculated on the *entire* rocket, not on just the active stages.  Is this the desired behavior?

wolsen:
Hmm, I tested this on the Three-stage rocket example and I see the length update appropriately. I see the length of the rocket drop appropriately as I drop stages from right to left.

Here's a few screenshots of mine (just collected so as to not lose my sanity):

Full Rocket
full-rocket

Upper 2 Stages
upper-stages

Top-most Stage
top-stage

Missing Middle Stage
Note, the length of the rocket is the length from tip to tail.
missing-middle-stage

wolsen:
Interestingly, if I hide the top stage only then the length doesn't change - which means it isn't changing for the rocket either.

Can you confirm which form your seeing @teyrana

teyrana:
oh! i was looking at the zoom extents....

Okay, looking at the length field -- It still doesn't seem to behave as expected: 

1. Open example 3-stage rocket
2. deactivate bottom (first booster) -> length changes
3. leave middle stage active
4. disable top (sustainer) stage -> length remains at 50 cm.

So, yea, I corroborate your observation ...
> which means it isn't changing for the rocket either."
What does this mean?

JoePfeiffer:
I think you may be headed down the rabbit hole that led to my refactor of the size calculation code. #543, which had to be reverted as it was somehow causing a stack exhaustion...

wolsen:
> > What does this mean?

What I mean is that the BoundingBox used in the FlightConfiguration.getLength() doesn't accurately reflect the length in this situation (removing a top stage). This may be going down the rabbit hole that @JoePfeiffer is referring to. I'll look at his changes and investigate this some more.

wolsen:
Yeah, fairly positive I'm going down the same rabbit hole that @JoePfeiffer went down. When debugging this particular issue, I can see that TubeCoupler on the first booster stage (in the 3-stage example rocket), returns a bounding box starting at x=0 rather than the offset to its global position. This causes the getLength() calculation to actually reset its range to [0, last point) and the length of the overall BoundingBox essentially calculates for the full length of top stages to bottom-most active stage.

wolsen:
Here's a question for the group. I'll look at @JoePfeiffer's changes and attempt to determine what was causing the stack overflow error. However, if that is a large change I'm curious if it would better to revert to the previous release's behavior of activating all stages through the selected stage. We're working towards stability, so that means thinking about reducing the size of changes we introduce. Just thinking about alternatives here, should we need it.

JoePfeiffer:
My hazy recollection is the bounding box code has enough problems that it really needs to be fixed.  If the stack overflow problem is small enough to be reasonable to correct that then that would be best; if not and it turns out simply reverting to the previous behavior masks the problems so we can get a release out and then go back and fix bounding boxes for real I'm all for it.
Go ahead and take a look.

wolsen:
Working on unit tests still, however thought I'd push this up for some feedback on the approach (@teyrana, @JoePfeiffer). Essentially, this change corrects the x offsets of the various RocketComponent.getComponentBounds() methods to always return absolute rocket coordinate positions (rather than relative instance-level coordinates). The code base is somewhat mixed in whether it provides absolute/relative and this change aims to correct that.

I fell short of going through the full migration to the RocketComponent.getBoundingBox() implementation as it was getting bigger than a bread basket. I did, however, include some groundwork for moving this in.

Also, ignore the Fix 2D TubeFin rendering (will clean up the commit later) - as I was simply fixing the rendering for my own tests. I know @teyrana raised a bug with little detail, so I assume a patch is coming.

JoePfeiffer:
Looking good to me.  Restricting this fix to just the bug, planning to migrate to getBoundingBox() later, seems like the right aproach.

teyrana:
> Essentially, this change corrects the x offsets of the various RocketComponent.getComponentBounds() methods to always return absolute rocket coordinate positions (rather than relative instance-level coordinates). 

The reason that we'd been moving to relative coordinates is that basically, using global coordinates at individual components breaks instancing trees.  

For instance, say you have two identical pods; on on each side of the main rocket.  This is represented in the program by the following rocket tree:
```
[-] Main stage
   [-] Center body
      [-] Pods
         [-] Pod Body
```

So that Pod Body -- which instance does it represent?   If you want to each instance of the pod body (2) to be calculated differently, then at minimum, you need to pass an instance_number for each instanced parent into `getComponentBounds`  -- which would change the function signature.

Also, you're duplicating the math about how we calculate the offsets for each of those instanced components in the hierarchy -- which is already calculated by those other components -- i.e. the Pods. 

At the top-level (i.e. get-active-stage-bounds) I recommend migrating to `getActiveInstances` based code, which was added for this purpose.
https://github.com/openrocket/openrocket/blob/faf770450c1082e8a35fec665f7d23e5ed2e53df/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java#L287

`getActiveComponents`  is deprecated for this very purpose -- it basically ignores instanced component assemblies.

> The code base is somewhat mixed in whether it provides absolute/relative and this change aims to correct that.

Yea, that's true.  It is a bit broken.  And mixed -- sometimes by mistake, sometimes by oversight, and sometimes on purpose (and it could be better documented).
Where it was left mixed, it was generally left that way, to keep change-sets to a smaller, more reasonable size (like you're doing as well ;)

JoePfeiffer:
Is instancing an issue for this?  Won't all of the instances be at the same x offset?

My impression when I was looking at the code was that the relative/absolute lengths weren't deliberate, and the bug just didn't trigger any visible behavior until it became possible to enable/disable individual stages.

You're right that it should probably be going to relative rather than absolute, since that's what we'll end up with when we go to getBoundingBox.  That will also be the time to switch over to active instances rather than active components.

teyrana:
> Is instancing an issue for this? 

In a function that calculates the bounding box of the rocket?   Yes, absolutely!  

.... also, it's been instancing bugs like, 5 times already :/ 

> Won't all of the instances be at the same x offset?

In the specific instance, above, the _ _x_ _ offsets are the same.  But the y,x offsets are *absolutely* not the same.  And it would be a mess to calculate x in one way and y and z in another.  (besides just being simply more work)

> My impression when I was looking at the code was that the relative/absolute lengths weren't deliberate, and the bug just didn't trigger any visible behavior until it became possible to enable/disable individual stages.

That sounds right, considering how old some of those code blocks are.... it depends on which LoC we're talking about...

JoePfeiffer:
This is turning into the same line of reasoning that led to my #543.  I wonder if I abadoned that too soon.

wolsen:
> > Essentially, this change corrects the x offsets of the various RocketComponent.getComponentBounds() methods to always return absolute rocket coordinate positions (rather than relative instance-level coordinates).
> 
> The reason that we'd been moving to relative coordinates is that basically, using global coordinates at individual components breaks instancing trees.

The inconsistency of using rocket coordinates vs relative coordinates is unfortunately the problem here though. I see that while I selected absolute rocket coordinates, the preference is to use relative coordinates. OK.

> 
> For instance, say you have two identical pods; on on each side of the main rocket. This is represented in the program by the following rocket tree:
> 
> ```
> [-] Main stage
>    [-] Center body
>       [-] Pods
>          [-] Pod Body
> ```
> 
> So that Pod Body -- which instance does it represent? If you want to each instance of the pod body (2) to be calculated differently, then at minimum, you need to pass an instance_number for each instanced parent into `getComponentBounds` -- which would change the function signature.

I'm not afraid to change the function signature, but I don't think this is the right function signature to go after. I think its more clear if we get rid of the getComponentBounds altogether and make something like ```public Collection<ComponentInstance> getInstances()```. Each of these ```ComponentInstance``` objects are responsible for providing the instance details like the specific bounds.

> 
> Also, you're duplicating the math about how we calculate the offsets for each of those instanced components in the hierarchy -- which is already calculated by those other components -- i.e. the Pods.

Yes, I realize that. I was hoping that by making it part of the RocketComponent itself we could move away from having this logic performed outside of the rocket tree. I believe I understand why it was done this way, but to me, it doesn't feel right to have the calculations done outside of the tree.

> 
> At the top-level (i.e. get-active-stage-bounds) I recommend migrating to `getActiveInstances` based code, which was added for this purpose.
> https://github.com/openrocket/openrocket/blob/faf770450c1082e8a35fec665f7d23e5ed2e53df/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java#L287
> 
> `getActiveComponents` is deprecated for this very purpose -- it basically ignores instanced component assemblies.

Fair enough. I knew that it was deprecated and why it was deprecated, but was aiming to keep the scope low.

> 
> > The code base is somewhat mixed in whether it provides absolute/relative and this change aims to correct that.
> 
> Yea, that's true. It is a bit broken. And mixed -- sometimes by mistake, sometimes by oversight, and sometimes on purpose (and it could be better documented).
> Where it was left mixed, it was generally left that way, to keep change-sets to a smaller, more reasonable size (like you're doing as well ;)

It feels like its time to just rip the band-aid off and make the bigger change. Its *very* confusing to have the mix all over the place.

teyrana:
> The inconsistency of using rocket coordinates vs relative coordinates is unfortunately the problem here though. 

Well, we're doing a lot of hand waving, here... let's talk more specific -- Which classes are you saying should deal with absolutes, and which with relative coordinates? 

> I'm not afraid to change the function signature,...

Not saying we can't, I was just pointing out that changing the signature means a bigger changeset.  That's all.

>  I think its more clear if we get rid of the getComponentBounds altogether and make something like ```public Collection<ComponentInstance> getInstances()```. Each of these ```ComponentInstance``` objects are responsible for providing the instance details like the specific bounds.

Uhhh, with this level of detail, 'ComponentInstance' sounds very similar to `InstanceContext`.  And the `Collection<>` sounds very similar to `InstanceMap` (with a slightly different interface).

> Yes, I realize that. I was hoping that by making it part of the `RocketComponent` itself we could move away from having this logic performed outside of the rocket tree. I believe I understand why it was done this way, but to me, it doesn't feel right to have the calculations done outside of the tree.

Yea, I see what you mean.  That _would_ be nice, if you can find a way to do.   I couldn't see one, but that doesn't prove anything.  :)   Just keep that any solution needs to perform correctly, and not break anything else.

> Fair enough. I knew that it was deprecated and why it was deprecated, but was aiming to keep the scope low.

Yea. Generally, I appreciate that.  But this fix may have to be bigger.

> It feels like its time to just rip the band-aid off and make the bigger change. Its *very* confusing to have the mix all over the place.

Well first, let's agree on what the "bigger change" _is_ ;)  Working at cross purposes doesn't solve anything.

I don't know that `InstanceMap` is the perfect solution, but any replacement for it has to solve the same problems, as well.

wolsen:
> > The inconsistency of using rocket coordinates vs relative coordinates is unfortunately the problem here though.
> 
> Well, we're doing a lot of hand waving, here... let's talk more specific -- Which classes are you saying should deal with absolutes, and which with relative coordinates?

As you pointed out, all the instanced components use relative coordinates while most (if not all) of the non-instanced components use absolute coordinates. Regardless of the coordinates, all of the *realized* instances of all components essentially use absolute coordinates (if you consider the transforms as part of the coordinates).

> 
> > I'm not afraid to change the function signature,...
> 
> Not saying we can't, I was just pointing out that changing the signature means a bigger changeset. That's all.
> 
> > I think its more clear if we get rid of the getComponentBounds altogether and make something like `public Collection<ComponentInstance> getInstances()`. Each of these `ComponentInstance` objects are responsible for providing the instance details like the specific bounds.
> 
> Uhhh, with this level of detail, 'ComponentInstance' sounds very similar to `InstanceContext`. And the `Collection<>` sounds very similar to `InstanceMap` (with a slightly different interface).

It actually is very close to be honest. I couldn't remember some of the class names at the time I was writing my comment and the commentary is more around the idea than concrete classes.

> 
> > Yes, I realize that. I was hoping that by making it part of the `RocketComponent` itself we could move away from having this logic performed outside of the rocket tree. I believe I understand why it was done this way, but to me, it doesn't feel right to have the calculations done outside of the tree.
> 
> Yea, I see what you mean. That _would_ be nice, if you can find a way to do. I couldn't see one, but that doesn't prove anything. :) Just keep that any solution needs to perform correctly, and not break anything else.

That's always a trick isn't it? :-)

> 
> > Fair enough. I knew that it was deprecated and why it was deprecated, but was aiming to keep the scope low.
> 
> Yea. Generally, I appreciate that. But this fix may have to be bigger.

Yeah, I think that's a fair statement.

> 
> > It feels like its time to just rip the band-aid off and make the bigger change. Its _very_ confusing to have the mix all over the place.
> 
> Well first, let's agree on what the "bigger change" _is_ ;) Working at cross purposes doesn't solve anything.
> 
> I don't know that `InstanceMap` is the perfect solution, but any replacement for it has to solve the same problems, as well.

Well, to be fair - this was a bit of a grandiose statement on my part founded in my frustrations with  inconsistencies between the use of relative and absolute coordinates. Inconsistencies drive me absolutely bonkers b/c they always lead to bugs and/or misunderstandings.

From what I can tell, the whole getActiveInstances() method exists entirely to ensure that a) instances are honored and b) to ensure that enough information is available to determine the instance's absolute coordinates.

I think the practical change that should be done for the near-term, is to use the getActiveInstances as you've pointed out b/c this ultimately does most of the absolute location (though it appears the transforms have to be applied as well). Anything bigger will than this will introduce risk that I think we should naturally be inclined to reject if we're moving towards stability.

However, I think the bigger set of changes I would like to see/make in the long run is:

1. Move the instance map details into the rocket tree itself (assuming we can figure out how to make this work).
2. Explicitly declare when coordinates/bounds are absolute or relative (either via objects or function names)
3. Get rid of functions that provide nonsensical answers for instanced components

Anyways, I really appreciate the patience, feedback and discourse. Some of the nuances and code details are being made more clear through these discussions.

wolsen:
Okay, before I tread any further down this (new) path, I thought I'd offer the change up so far to get some feedback.

This code primarily handles the calculation in the FlightConfiguration.calculateBounds() method using the FlightConfigureation.getActiveInstances() as previously discussed. There's a few oddball component types which, when applying the transform in the InstanceContext cause the bounds to be skewed. Part of the problem may be my use of the getComponentBounds() for several components and then applying the transform, but it generally seems to work with a few notable exceptions (FinSets, BodyTubes, AxialStages, and PodSets).

Thoughts on fixing the length calculation, with all the special caveats in calculateBounds and addressing each caveat as it appears later? These need additional investigation to see if something in the transform is incorrect or if its the bounds that are calculated differently (at first glance, it seems PodSets fall into column A, the others into column B). FinSets are a whole different ballgame - for good reason IMO.

Anyways, feedback welcomed :-)

teyrana:
> 2. Explicitly declare when coordinates/bounds are absolute or relative (either via objects or function names)

Yea, I like this idea -- do you think updating the Javadocs would be enough? or something like adjusitng the function signatures, or adding type information to "Coordinates" to differentiate RelativeCoordinates from AbsoluteCoordinates? 

> Get rid of functions that provide nonsensical answers for instanced components.  Anyways, I really appreciate the patience, feedback and discourse. 

For sure.  yes.   I've started on that.  Obviously, I started on the instanced components because those were the most broken, but if you find others that need to be fixed / changed, let's do those, too! 

> Thoughts on fixing the length calculation, with all the special caveats in calculateBounds and ... 

Disclaimer: my work with InstanceMap so far has been focused around things like CG and CP.   I'm not surprised that the 'calculateBounds' and `getComponentBounds` functions are broken. They might still expect to return absolute coordinates across the board, for all I know.  

The machine behind `FlightConfiguration.getActiveInstances` I actually trust more, because we've validated it with things like Cg and Cp calculations.   They're not finished, I don't think.  But they're in a relatively stable state. :)

wolsen:
> > 1. Explicitly declare when coordinates/bounds are absolute or relative (either via objects or function names)
> 
> Yea, I like this idea -- do you think updating the Javadocs would be enough? or something like adjusitng the function signatures, or adding type information to "Coordinates" to differentiate RelativeCoordinates from AbsoluteCoordinates?

I lean towards both. Updating the Javadocs would be helpful and should be done, but experience tells me this will be wrong at some point in the future (I trust code more than docs). I'm currently of the mindset that it's better to be explicit than implicit. However, I do think this is a longer term goal to be perfectly honest.

> > Get rid of functions that provide nonsensical answers for instanced components.  Anyways, I really appreciate the patience, feedback and discourse.
> 
> For sure. yes. I've started on that. Obviously, I started on the instanced components because those were the most broken, but if you find others that need to be fixed / changed, let's do those, too!

Yeah, I've noticed that you've started on this and I noticed it perhaps too late. Spending even more time on it recently (and trying to make the bold changes I thought) lead me to some of the same trade-offs it is now clear that were made in the code. There's a lot to grok in here :-)

> > Thoughts on fixing the length calculation, with all the special caveats in calculateBounds and ...
> 
> Disclaimer: my work with InstanceMap so far has been focused around things like CG and CP. I'm not surprised that the 'calculateBounds' and `getComponentBounds` functions are broken. They might still expect to return absolute coordinates across the board, for all I know.
> 
> The machine behind `FlightConfiguration.getActiveInstances` I actually trust more, because we've validated it with things like Cg and Cp calculations. They're not finished, I don't think. But they're in a relatively stable state. :)

Yeah it seems relatively stable and so far. Its a bit counter-intuitive due to the nomenclature used that all instances are returned with an active flag, but that's a nit more than anything and simple to address/discuss with a separate PR.

I'm actually happier with this approach than I was with my initial take as this is a more localized, small change so far. While it has its hackiness with the special casing of a few components (so far), it actually reduces the risk of additional bugs being introduced since its small - which works towards stability.

I guess what I'm saying is, thanks for pushing me around to your way of thinking :-)

swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
```diff
@@ -606,20 +606,8 @@ private void updateExtras() {
...
-			for (Coordinate c : bounds) {
-				if (c.x < minX)
-					minX = c.x;
-				if (c.x > maxX)
-					maxX = c.x;
-			}
-			length = maxX - minX;
-		}
-
+		double length = curConfig.getLength();
```
teyrana, line 609: I think we want to actually preserve the previous behavior, here.  -- see the comment about increasing fin-height.  (for example)

swing/src/net/sf/openrocket/gui/components/StageSelector.java
```diff
@@ -41,7 +37,7 @@ private void updateButtons( final FlightConfiguration configuration ) {
 		this.removeAll();
 		for(AxialStage stage : configuration.getRocket().getStageList()){
 			JToggleButton button = new JToggleButton(new StageAction(stage));
-			button.setSelected(true);
+			button.setSelected(configuration.isStageActive(stage.getStageNumber()));
```
teyrana, line 40: 😁

teyrana:
Heh, glad to hear at least _some_ of what I said made sense :) 

1. The length (as displayed) seems to be correct, for the test cases I tried.
2. Increasing the fin *height* --  not length -- seem to increase the zoom... but not enough to contain the fins.  (fyi -- this works as desired in 'unstable').  The easiest way to reproduce this is to open the first example rocket -- "Simple Example Rocket"
3. Do we have example rockets, that we're testing against? 

(I took a look at the code -- it looks much cleaner :)

swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
wolsen: That's not actually causing the fin-height scaling issue, but the fin height-issue is a good catch. Using the old method in here essentially just recalculates the value in essentially the same way that has already been calculated so it feels like just extra processing. As such, I'd prefer to keep this change here and fix the fin-height scaling issue.

It took me a few minutes to figure out what was going on with the fin-height issue, but its pretty clear and straightforward. The fin-height issue is that I'm not properly calculating the max height the fins will span in all the way around the rocket. Using the approach I used for the FinSets, you get the tightest bounds for the fins in the side profile.

Fortunately, the FinSet already has a getBoundingBox() method which _does_ take this into account. The FinSet.getBoundingBox() method sets the bounds for its bounding box to be the length of the fins x 2 * (finHeight - 0).

wolsen:
> Heh, glad to hear at least _some_ of what I said made sense :)

Heh, I'm glad that I'm able to catch up on you train of thoughts on many of these. Believe it or not, the more time I spend looking at this stuff the more understanding I am.

> 1. The length (as displayed) seems to be correct, for the test cases I tried.
Yay \o/. I spent the most time validating this aspect and also compared it to the various values displayed in the 15.03 version. It turns out, the Boosted Dart example rocket is off by 2 cm in its length in 15.03.

> 2. Increasing the fin _height_ --  not length -- seem to increase the zoom... but not enough to contain the fins.  (fyi -- this works as desired in 'unstable').  The easiest way to reproduce this is to open the first example rocket -- "Simple Example Rocket"

This is a good catch, the zoom fit actually doesn't change in my tests but its pretty easy to replicate the issue you identified and simple to resolve. See my reply to your inline comment for an explanation of what is causing it.

> 3. Do we have example rockets, that we're testing against?

I'm actually testing against all of the example rockets. I went through and manually calculated the length of every stage in all the rockets, which is tedious and really calls for unit tests rather than manual tests. Those are coming.

> 
> (I took a look at the code -- it looks much cleaner :)

teyrana:
Okay, yea.  This looks like it's doing what we want it to. 

We were also talking about tests -- do we think the tests you've already updated are sufficient for this fix?

wolsen:
Okay, I spent a good amount of time today revisiting this and manually testing a variety of options. My previous change to use the FinSet.getBoundingBox() did improve the increasing fin height scenario in some ways, but failed in others. You can see it fails with the simple example rocket by simply rotating the fins 60 degrees.

Checking the rear view, it doesn't have this problem... which, of course it doesn't because it is calculating the height to be the diameter of the circle which touches all the corners of the bounding box in the Y, Z plane. This ensures that the points will be contained within the circle, and thus the height.

As such, I've revisited my last change and still use the FinSet.getBoundingBox(), but updated this to be a tight bounding box around the FinSet itself. The previous implementation of the FinSet.getBoundingBox() didn't quite do as I stated before (the doubling of the fin height), but it almost did this (a quirk in the location math prevented it from doing this). Instead, I converted this to update the bounds of the box based on the actual fin's outer-most point along the Y, Z plane which creates a tightly bound BoundingBox.

At this point, I feel that the tests previously changed are sufficient to exercise this code path. I will write some new unit tests around the BoundingBoxes on the test rockets, but can do that in a subsequent pull request. I think this is ready to get some more usage on it.

PR #603, opened 2020-03-25
[Fixes 568] Consider all instances of components when calculating drag

### LINKED ISSUE #568: Inconsistent CP and flight results when using pods or not
This image shows two outwardly identical rockets.  The one on the left, is a simple 3FNC.  The one on the right is too, except the fins are mounted to phantom body tubes on pods rather than directly on the main airframe.  The CP location and flight sim results are quite different between them.  Ideally should be the same.

This running on an Alpha build from April 2019, because later builds have been generally unusable for me.

image

[3fnc.zip](https://github.com/openrocket/openrocket/files/4216072/3fnc.zip)

JoePfeiffer:
Fixes #568 by iterating through all instances of rocket components when calculating drag.
Adds unit test for fins on pods as in bug report.

Jar file is at https://www.dropbox.com/s/rnzs2i0xp9e5l0z/OpenRocket-PR603.jar?dl=0 if anyone would like to take it for a spin.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -571,11 +571,9 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 	 * @return
 	 */
 	private double calculateBaseDrag(FlightConfiguration configuration, FlightConditions conditions,
-			Map<RocketComponent, AerodynamicForces> map, WarningSet warnings) {
+									 Map<RocketComponent, AerodynamicForces> map, WarningSet warnings) {
 		
 		double base, total;
-		double radius = 0;
-		RocketComponent prevComponent = null;
 		
 		if (calcMap == null)
 			buildCalcMap(configuration);
@@ -583,36 +581,51 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 		base = calculateBaseCD(conditions.getMach());
 		total = 0;
 		
-		for (RocketComponent c : configuration.getActiveComponents()) {
+		final InstanceMap imap = configuration.getActiveInstances();
+	    for(Map.Entry<RocketComponent, ArrayList<InstanceContext>> entry: imap.entrySet() ) {
+			final RocketComponent c = entry.getKey();
+			
 			if (!(c instanceof SymmetricComponent))
 				continue;
 			
 			SymmetricComponent s = (SymmetricComponent) c;
-
-			if(c.isCDOverridden()) {
-				total += c.getOverrideCD();
-				continue;
-			}
 			
-			if (radius > s.getForeRadius()) {
-				double area = Math.PI * (pow2(radius) - pow2(s.getForeRadius()));
-				double cd = base * area / conditions.getRefArea();
-				total += cd;
-				if (map != null) {
-					map.get(prevComponent).setBaseCD(cd);
+			// iterate across component instances
+			final ArrayList<InstanceContext> contextList = entry.getValue();
+			for(InstanceContext context: contextList ) {
+				if(c.isCDOverridden()) {
+					total += c.getOverrideCD();
+					continue;
 				}
-			}
-			
-			radius = s.getAftRadius();
-			prevComponent = c;
-		}
-		
-		if (radius > 0) {
-			double area = Math.PI * pow2(radius);
-			double cd = base * area / conditions.getRefArea();
-			total += cd;
-			if (map != null) {
-				map.get(prevComponent).setBaseCD(cd);
+				
+				// if aft radius of previous component is greater than my forward radius, set
+				// its aft CD
+				double radius = 0;
+				final SymmetricComponent prevComponent = s.getPreviousSymmetricComponent();
+				if (prevComponent != null) {
+					radius = prevComponent.getAftRadius();
+				}
+				
+				if (radius > s.getForeRadius()) {
+					double area = Math.PI * (pow2(radius) - pow2(s.getForeRadius()));
+					double cd = base * area / conditions.getRefArea();
+					total += cd;
+					if ((map != null) && (prevComponent != null)) {
+						map.get(prevComponent).setBaseCD(cd);
+					}
+				}
+				
+				// if I'm the last componenet, set my base CD
+				// note I can't depend on the iterator serving up components in order,
+				// so I can't just do this after the last iteration.
+				if (s.getNextSymmetricComponent() == null) {
+					double area = Math.PI * pow2(s.getAftRadius());
+					double cd = base * area / conditions.getRefArea();
+					total += cd;
+					if (map != null) {
+						map.get(s).setBaseCD(cd);
+					}
+				}
 			}
 		}
 		
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -392,84 +392,90 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 		
 		double[] roughnessLimited = new double[Finish.values().length];
 		Arrays.fill(roughnessLimited, Double.NaN);
-		
-		for (RocketComponent c : configuration.getActiveComponents()) {
+
+		final InstanceMap imap = configuration.getActiveInstances();
+	    for(Map.Entry<RocketComponent, ArrayList<InstanceContext>> entry: imap.entrySet() ) {
+			final RocketComponent c = entry.getKey();
 			
 			// Consider only SymmetricComponents and FinSets:
 			if (!(c instanceof SymmetricComponent) &&
 					!(c instanceof FinSet))
 				continue;
-			
-			// Calculate the roughness-limited friction coefficient
-			Finish finish = ((ExternalComponent) c).getFinish();
-			if (Double.isNaN(roughnessLimited[finish.ordinal()])) {
-				roughnessLimited[finish.ordinal()] =
-						0.032 * Math.pow(finish.getRoughnessSize() / configuration.getLength(), 0.2) *
-								roughnessCorrection;
-			}
-			
-			/*
-			 * Actual Cf is maximum of Cf and the roughness-limited value.
-			 * For perfect finish require additionally that Re > 1e6
-			 */
-			double componentCf;
-			if (configuration.getRocket().isPerfectFinish()) {
-				
-				// For perfect finish require Re > 1e6
-				if ((Re > 1.0e6) && (roughnessLimited[finish.ordinal()] > Cf)) {
-					componentCf = roughnessLimited[finish.ordinal()];
-				} else {
-					componentCf = Cf;
-				}
-				
-			} else {
-				
-				// For fully turbulent use simple max
-				componentCf = Math.max(Cf, roughnessLimited[finish.ordinal()]);
-				
-			}
-			
-			//Handle Overriden CD for Whole Rocket
-			if(c.isCDOverridden()) {
-				continue;
-			}
-			
-			
-			// Calculate the friction drag:
-			if (c instanceof SymmetricComponent) {
-				
-				SymmetricComponent s = (SymmetricComponent) c;
-				
-				bodyFriction += componentCf * s.getComponentWetArea();
-				
-				if (map != null) {
-					// Corrected later
-					map.get(c).setFrictionCD(componentCf * s.getComponentWetArea()
-							/ conditions.getRefArea());
-				}
-				
-				double r = Math.max(s.getForeRadius(), s.getAftRadius());
-				if (r > maxR)
-					maxR = r;
-				len += c.getLength();
-				
-			} else if (c instanceof FinSet) {
-				
-				FinSet f = (FinSet) c;
-				double mac = ((FinSetCalc) calcMap.get(c)).getMACLength();
-				double cd = componentCf * (1 + 2 * f.getThickness() / mac) *
-						2 * f.getFinCount() * f.getPlanformArea();
-				finFriction += cd;
-				
-				if (map != null) {
-					map.get(c).setFrictionCD(cd / conditions.getRefArea());
-				}
-				
-			}
 
-		
+				// iterate across component instances
+			final ArrayList<InstanceContext> contextList = entry.getValue();
+			for(InstanceContext context: contextList ) {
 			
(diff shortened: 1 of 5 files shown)


core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
```diff
@@ -515,39 +515,46 @@ private double calculatePressureDrag(FlightConfiguration configuration, FlightCo
...
+				double cd = calcMap.get(c).calculatePressureDragForce(conditions, stagnation, base,
+																	  warnings);
+				total += cd;
+				
+				if (map != null) {
+					map.get(c).setPressureCD(cd);
 				}
 				
-				radius = s.getAftRadius();
+				if(c.isCDOverridden()) continue;
```
wolsen: nit: while I'm not a style nut, I find it better to keep styles consistent so I think the continue should be on the next line. I realize this is carry over from indentation change, but since the code is being touched - maybe worthwhile to make it consistent.

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
```diff
@@ -515,39 +515,46 @@ private double calculatePressureDrag(FlightConfiguration configuration, FlightCo
...
-					if (map != null) {
-						map.get(c).setPressureCD(map.get(c).getPressureCD() + cd);
-					}
+				// Pressure fore drag
+				double cd = calcMap.get(c).calculatePressureDragForce(conditions, stagnation, base,
+																	  warnings);
+				total += cd;
+				
+				if (map != null) {
+					map.get(c).setPressureCD(cd);
```
wolsen: nit: I realize this is carry-over, however there's a couple of places in this path which lookup the AerodynamicForces for the component and use it. Each time I had to re-lookup what the map.get(c) was returning. I wonder if it would be more clear if the AerodynamicForces was set at the beginning of the loop and use it throughout. i.e.

```java
AerodynamicForces forces = (map != null) ? map.get(c) : null;
...
if (forces != null) {
    forces.setPressureCD(cd);
}
```

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
```diff
@@ -392,84 +392,90 @@ private double calculateFrictionDrag(FlightConfiguration configuration, FlightCo
...
-				componentCf = Math.max(Cf, roughnessLimited[finish.ordinal()]);
-				
-			}
 			
-			//Handle Overriden CD for Whole Rocket
-			if(c.isCDOverridden()) {
-				continue;
-			}
+				//Handle Overriden CD for Whole Rocket
+				if(c.isCDOverridden()) {
```
wolsen: I think this condition could be moved prior to processing the active instances. None of the iterations above this check within the loop contribute to the final calculations returned at the end. Moving this check up will save some unnecessary loop iterations.

core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java
```diff
@@ -645,7 +645,7 @@ public double calculatePressureDragForce(FlightConditions conditions,
 		// Airfoil assumed to have zero base drag
 		
 		// Scale to correct reference area
-		drag *= finCount * span * thickness / conditions.getRefArea();
+		drag *= span * thickness / conditions.getRefArea();
```
wolsen: good catch here - it took me a few minutes to realize why the finCount was removed here (due to using the activeInstances rather than activeComponents)

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
```diff
@@ -477,7 +481,7 @@ private double calculateFrictionDrag(FlightConfiguration configuration, FlightCo
 		}
 		
 		// fB may be POSITIVE_INFINITY, but that's ok for us
-		double fB = (len + 0.0001) / maxR;
+		double fB = (maxX - minX + 0.0001) / maxR;
```
wolsen: Does this affect when the overall CD is overridden for the rocket? The changes will cause this to go from 0.0001 / maxR to Double.MAX / maxR. The correction applied to each of the components will be significantly different in this scenario, and I'm just not positive how its used.

core/src/net/sf/openrocket/util/TestRockets.java
```diff
@@ -1,13 +1,16 @@
 package net.sf.openrocket.util;
 
+import java.util.ArrayList;
+import java.io.FileOutputStream;
```
wolsen: nit: this will add a warning for an unused import since the code below is commented out for reference.

core/test/net/sf/openrocket/aerodynamics/BarrowmanCalculatorTest.java
```diff
@@ -231,4 +231,50 @@ public void testRadialDiscontinuityWithStrapOns() {
...
+		// This doesn't turn out to cause a problem in an actual
+		// simulation, so we are just not testing for it.  Test with
+		// correction if we want some time is here but commented out
+		
+		final Coordinate cpNoPods = calcNoPods.getCP(configNoPods, conditionsNoPods, warningsNoPods);
+		final Coordinate cpPods = calcPods.getCP(configPods, conditionsPods, warningsPods);
+		
+		assertEquals(" 3FNC With Pods rocket cp x value is incorrect:", cpPods.x, cpNoPods.x - 0.00672, EPSILON);
+		// assertEquals(" 3FNC With Pods rocket cp y value is incorrect:", cpPods.y, cpNoPods.y - 0.00548, EPSILON);
+		// assertEquals(" 3FNC With Pods rocket cp z value is incorrect:", cpPods.z, cpNoPods.z, EPSILON);
```
wolsen: are these commented out for a reason?

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
JoePfeiffer: Yeah, coding styles are all over the map with this project -- I've been trying to keep my changes to just functional stuff, and not change things that are just style.  All the same, I didn't realize what a complete outlier that one was; fixed it.  I think that was the only one in the files I touched?

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
JoePfeiffer: I noticed a few possible optimizations like that while I was working on it -- I was thinking for this PR I wanted to stick with just fixing things that are truly broken, and then maybe revisit later.  That would still be my preference, but if others would like to see things like this improved while I'm at it I certainly can.

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
JoePfeiffer: Another one I'd prefer to fix later

core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java
JoePfeiffer: :)

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
JoePfeiffer: The problem I was trying to address was in the calculation of the length of the rocket -- the old code simply summed up the lengths of all the axially symmetric components, which had problems when there were pods:  if there was a central body tube and a pod off to the side, the lengths would be added together so len would be too long.  This code simply replaces that sum with the max and min x encountered while processing the rocket.
Unless I'm missing something, this shouldn't make any difference when the overall CD is overridden.

core/src/net/sf/openrocket/util/TestRockets.java
JoePfeiffer: On the FileOutputStream -- nice catch.  Fixed
On the java.util.ArrayList -- the code using that isn't commented out.

core/test/net/sf/openrocket/aerodynamics/BarrowmanCalculatorTest.java
JoePfeiffer: Yes.  When the center of pressure is calculated for a with an odd number of fins, there is a Y component to the calculated CP because one of the fins is taken as being at angle of 0 and doesn't contribute to the CP; the error doesn't end up affecting flight simulations because the rocket is assumed to only vary in pitch (which strikes me as something to fix to make the simulation more accurate, but that's for another day), so indeed that fin doesn't contribute.
When you've got  a finset with three or more fins, the Y and Z components of the CP are forced to 0; that's not practical when you're putting together the CP from fins from different instances.
My initial thought was to explicitly run the test with a correction for the offset, but upon reflection it seemed like writing an error into the test case was a mistake.  So, I commented out the Y and Z tests, and added the comment just before the tests.

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
wolsen: Yeah, that was the only one - and agree to sticking with the functional changes in general. I commented probably a bit much, but I figure a review is a review :-) Anyways, thanks for fixing that!

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
wolsen: Works for me. Additional changes are perfectly fine outside of this PR.

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
wolsen: +1

teyrana:
@JoePfeiffer @wolsen 
Are you guys happy with the style discussion, for now?

core/src/net/sf/openrocket/util/TestRockets.java
```diff
@@ -1,7 +1,7 @@
 package net.sf.openrocket.util;
 
 import java.util.ArrayList;
-import java.io.FileOutputStream;
+// import java.io.FileOutputStream;
```
teyrana: why not just remove?

core/test/net/sf/openrocket/aerodynamics/BarrowmanCalculatorTest.java
teyrana: @JoePfeiffer 
> When the center of pressure is calculated for a with an odd number of fins, there is a Y component to the calculated CP because one of the fins is taken as being at angle of 0 and doesn't contribute to the CP; the error doesn't end up affecting flight simulations because the rocket is assumed to only vary in pitch (which strikes me as something to fix to make the simulation more accurate, . . .

Yea, I ran into this in relation to https://github.com/openrocket/openrocket/issues/538
And may be fixing that.  Let me know in a separate thread if you have opinions on the right way to do so.

core/src/net/sf/openrocket/util/TestRockets.java
```diff
@@ -1629,5 +1634,84 @@ public static OpenRocketDocument makeTestRocket_for_estimateFileSize() {
 		rocket.enableEvents();
 		return rocketDoc;
 	}
+
+	// the following two models are used in testing
+	// otherwise-identical rockets, one created in the obvious way
+	// using a single finset and the other creating three pods, each
+	// with a single fin.
+	public static final Rocket make3FNCNoPods()  {
```
teyrana: this sounds almost identitcal to the `makeAlphaIII` function.  Is it possible to combine these? 

(where the *withPods function would correspond to a pod-ized version of the AlphaIII?)

core/src/net/sf/openrocket/util/TestRockets.java
JoePfeiffer: I had to figure out how to dump a rocket so I could look at it all over again, and I left that code in the file but commented out because I couldn't really think of a good place to put it as documentation.  If that code is uncommented by somebody adding a new test case, the import will need to be as well (ignorance of github here -- how do I reference some lines of code in my comments, as Billy did above)?
Better solution would be to document it outside the file.

teyrana:
Overall looks good.  Just waiting for discussions to resolve.

core/src/net/sf/openrocket/util/TestRockets.java
JoePfeiffer: I'll take a look.

core/src/net/sf/openrocket/util/TestRockets.java
teyrana: Yea, we really should have a common way to debug that.... I've run into that use case a few times, as well. :P 

Well, github's documentation explains it better than I can: 
https://help.github.com/en/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet

Also, I generally prefer to simply remove commented loc, but if you want to leave it, I'll defer in this case.

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
wolsen: Actually, I was wrong. Regardless of whether the overall CD being overridden or not, this value doesn't actually change/affect the computations. The key is that maxR is set to 0, so fb will always result in Double.INFINITY when the overall CD is overridden. The correction that's used will then always be 1 + 1/(2 * INFINITY) which will always be 1.0. So regardless of the change, it doesn't affect the correction value applied.

Apologies for the noise!

wolsen:
Overall +1 from me

teyrana:
(waiting for joe to replay to the testrockets.java discussion)

JoePfeiffer:
OK, I'm an idiot -- of course the right thing to do with the "dumpRockets" code was to make it a function and put it in the file so the next person building a unit test can find it.
Also, it turned out that basing the "3fnc with pods rocket" on the Alpha III was trivial, so I made that change.  I almost forgot to change the offset in the y value to match the new rocket; in order to avoid leaving a trap for some future maintainer I uncommented those lines.

teyrana:
So I presume the 
[make3FNCNoPods function ](https://github.com/openrocket/openrocket/blob/ecc8a32891d314d2e7239a2d9edc62069b3e0406/core/src/net/sf/openrocket/util/TestRockets.java#L1638-L1666)
is dead code, now (i.e. no longer called / no longer needed?)

JoePfeiffer:
Correct.  Failed to commit that last change until now.

teyrana:
Excellent! LGTM.

PR #610, opened 2020-03-30
[fixes #608] save all instance counts, even if equal to 1 

### LINKED ISSUE #608: Fins Counts default  to 3 -- even if specified from a load-file.
Investigation:
- the finset causing the error is solved without an explicit instance count or fin-count attribute.  Therefore it defaults to 3.  (it's actual default)

teyrana:
(because different components have different default instance counts

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
+++ b/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
@@ -90,8 +90,8 @@ public class RocketComponentSaver {
 			
 			if( c instanceof Clusterable ){
 				; // no-op.  Instance counts are set via named cluster configurations
-			}else if( 1 < instanceCount ) {
-				emitInteger( elements, "instancecount", c.getInstanceCount() );
+			}else {
+				emitInteger(elements, "instancecount", c.getInstanceCount());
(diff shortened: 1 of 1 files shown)


core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
```diff
@@ -90,9 +90,8 @@ protected void addParams(net.sf.openrocket.rocketcomponent.RocketComponent c, Li
 			
 			if( c instanceof Clusterable ){
```
JoePfeiffer: Is this still needed?

teyrana: yup.  There's a comment the line after that which explains why.  

Also, that version was buggy. Try the new version :)

JoePfeiffer: Ah, got it.  It was the bugginess that threw me -- emitInteger wasn't in the else clause with that version.

teyrana: Yea.  That was definitely a bug.

PR #621, opened 2020-04-13
Fix TubeFinSet 2D rendering

### LINKED ISSUE #605: tube fins don't display -- 2d / side-views

wolsen:
The TubeFinSetShapes were applying the supplied transform too many
times, resulting in the the shapes being offset off the viewing area.
The offset transform is already applied to the component instance
that is provided to the TubeFinSetShapes methods, so applying the
transform resulted in the offset being moved to the right.

This change removes the transform being applied to the component
locations when retrieved and only applies a linear transformation
to the component locations when applying the base rotational
transform. Using the linear transform avoids applying the offset
in addition to the rest of the transform.

The problem existed in both the side view and the back view, however
the back view was not as obvious due to the nature of the view. The
common code between both functions applying the transform is
refactored into a common location to reduce the amount of code.
Additionally, documentation was added explaining the transform process
that is being applied to the component locations for the final shapes.

Fixes #605

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
+++ b/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
@@ -294,6 +294,10 @@ class DocumentConfig {
 				Reflection.findMethod(TubeFinSet.class, "setOuterRadius", double.class),
 				"auto",
 				Reflection.findMethod(TubeFinSet.class, "setOuterRadiusAutomatic", boolean.class)));
+		setters.put("TubeFinSet:instancecount", new IntSetter(
+				Reflection.findMethod(TubeFinSet.class,  "setInstanceCount", int.class)));
+		setters.put("TubeFinSet:angleoffset", new AnglePositionSetter() );
+		setters.put("TubeFinSet:radiusoffset", new RadiusPositionSetter() );
 		
 		// InternalComponent - nothing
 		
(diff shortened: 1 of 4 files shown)


teyrana:
A) Javadoc comments.  yay! 
B) Unequivocal improvement +1
C) This helps, but does not resolve the issue -- tube fins + pods still looks ... odd. 
     I encourage you to take a look at what happens in that case, because it's a relevant example of what I was talking about in the other issue...


## To Reproduce:
1. Open new (blank) rocket:
2. add a (centerline) body tube to the default stage
3. add a pod to body tube from (2)
4. add a body tube to the pod-set from (3) 
5. add some tube fins to the body tube from (4)

The default "fin" count there is 3.  And already it looks odd.  Check out the rear-view as that's more instructive....  It shows 6 tube fins all arranged around the centerline body, where it should show three fins around the *pod* bodies, instead.

wolsen:
Haha, um ... yeah - I think odd is the right word. I'll look at this a bit more

teyrana:
Most Recent Travis Build (Passing):
https://travis-ci.org/github/openrocket/openrocket/builds/674251757

teyrana:
These look fine on center bodies, but aren't quite right on pods -- (1) they rotate half as slow as they should (input of 54 deg ~= where 90deg should be), and (2) they seem centered about.... something?   Something not the center of their parent mount.

teyrana:
@wolsen 
I took a quick look, and created a patch for this PR  (it's a PR onto your branch.... if you like it, then we can merge it, and then this one)
https://github.com/wolsen/openrocket/pull/1

wolsen:
Thanks for the help @teyrana - I over-thought the problem clearly. Also had to fix the 3D rendering as the extra rotation was applied there as well. Should all be good to merge now!

wolsen:
Rebased and pushed

teyrana:
yup. lgtm.

PR #622, opened 2020-04-16
[Fixes 620] avoid prevent null strings in new motors

JoePfeiffer:
Ensure motors have sane defaults (notably, no null strings)
Create unit test for motor row filter in motor selection dialog
Fix spelling error getTotalImpuse => getTotalImpulse


wolsen:
Looks fairly straightforward, but you said there was a test created?

JoePfeiffer:
Oops.  Considering how much more effort the test took than the actual fix, that was a big omission!

teyrana:
LGTM.

PR #623, opened 2020-04-16
Add GPL v3 license file

itshanney:
@teyrana Adding LICENSE file with license details for OpenRocket for reference sake.


JoePfeiffer:
Isn't this already in core/LICENSE.TXT?  That file also mentions Sampo and others as copyright holders, and grants permission to package.

itshanney:
@JoePfeiffer ah, didn't notice that. Any objection to moving those files to the root folder and out of the /core folder? Will follow-up with an update to this PR to see what that looks like.

wolsen:
I believe the license is pulled into the jar file by the build.xml script as well fwiw.

itshanney:
@wolsen Fixed the jar build. I also noticed then when I run OpenRocket.java through IntelliJ it wasn't correctly picking it up since it was a level higher. I ended up copying it to swing/resources as /resources was already on my classpath and is the norm.

wolsen:
I think the move of the LICENSE file to the top directory and dropping the .txt extension is good, however the git history is lost on a couple of the moved files the way the files are moved. I especially think that the git history is relevant around any changes to the LICENSE file (even a relocation).

I'm not sure we want to go down the path of copying the license file to swing/resources, but it does solve the issue of the license file not showing up from a non-built jar file. Multiple copies will surely end up out of sync and while the official ant script will include the LICENSE file from the top level jar, this will probably lead to at least confusion in the future. I'm thinking maybe you should break that change into a separate PR and we can have some discussion around that as there may or may not be alternative ways to solve this.

PR #624, opened 2020-04-17
Text files cleanup

itshanney:
@teyrana @wolsen Files cleanup of various text files over time.

For broader context on if you're wondering why I'm doing this, it's because I'm going through to clean up documentation and try to ensure things are current.

FWIW, I also noticed there's a core/TODO file that's *really* old. I would argue this creates more harm than good because it can't really be trusted and people are skeptical of it. Any objection to deleting it?


JoePfeiffer:
Briefly looking at the TODO list, you're right -- it will do a lot more harm than good if people look at it.  Getting rid of it seems like a really good idea.

wolsen:
+1 to removing TODO list, that kind of list should be kept on the wiki or bug tracker going forward.

teyrana:
For reference, I assume you're talking about this one: 
http://wiki.openrocket.info/Feature_wishlist

README.md
```diff
@@ -19,14 +19,39 @@ The main features include:
...
+To start the software from the JAR file run:
+```
+$ java -jar OpenRocket-<VERSION>.jar
+```
+
 License
 -------
 
-OpenRocket is an Open Source project licensed under the [GNU GPL](http://openrocket.sourceforge.net/license.html). This means that the software is free to use for whatever purposes, and the source code is also available for studying and extending.
+Copyright (C) 2007-2020  Sampo Niskanen and others
```
teyrana, line 32: Honestly, this just seem redundant with the "Contributors" Section Below.

As the GPL doesn't require attributation, it's not particularly relevant.

README.md
```diff
@@ -19,14 +19,39 @@ The main features include:
 
 Read more about it on the [OpenRocket Wiki](http://wiki.openrocket.info).
 
+Running OpenRocket
```
teyrana, line 22: This is a great  section to add.   

However, the first thing mentioned should be the installers: (which we don't yet have a download page on the website... but we could still mention them, and add the url later?)

I wouldn't even recommend running the jars *directly*, like the below.   If they're going to github, wouldn't they want to use an IDE?   Like eclipse, Intellij or VSCode? 
Should we get people to contribute instructions for each of those?

teyrana:
+1 for removing the todo list.  That' doesn't need to live in source.

-1 for moving the Changelog from core -> root.  Actually this is more of a question:  when Android was a full fledged client, it was much more obivous, because both the desktop UI and the android UI used the same core "library".  So they really were all separate modules.  (And still are, as far as the IDEs are concerned.    

So it's more a question -- do we want to continue to iterate with the core library separate from the UI?   Or just treat it as one big integrated project? 

I vote for the first option.

itshanney:
Will remove the TODO file in a follow-up commit. Seems we have critical mass on agreement.

Will migrate the Feature Wishlist from wiki to Github over time. I agree with Billy that we should have everything centralized in Github Issues, whether it's feature requests or bugs. In fact, I'm going to volunteer to move the entire wiki to Github wiki so we have a single source of truth - the Github project. It's on my TODO list with documentation cleanups.

@teyrana I vote to keep it separate too. This allows some specialization to occur from people who may have skills in certain areas of the project. For example, if a UI expert joins, they can work on the UI without worrying about underlying code. On the flip side, if we have experts that are great at the math calculations and other lower-level aspects of the code, they can work too. Keeping the modules separate presents more benefits over the long-term than combining them back. Doing the split back when was a good choice.

Can you expand a bit more on the ChangeLog move? Is this file particularly referring to the Core, or is it referring to the entire project? If it's just Core, sure, we can move it back. But if it's the entire project, then root seems more appropriate as a reference.

teyrana:
> But if it's the entire project, . . .  

Hmmm.  That's also a good point.  Maybe we want to have separate change longs with different scope?  Also, how much of the change log is contained in the history of git commits, and how much do we want summarized in a separate (more) human readable changelog.

README.md
JoePfeiffer: Please don't get rid of instructions for building and running directly from the command line (in the wiki instead of the README is fine, just don't get rid of them completely). This may reflect the fact that I'm older than dirt, but every time I've tried to use an IDE I've felt like I was trying to thread a needle while wearing boxing gloves. I'm doing all my editing in emacs and using good ol' print statements for debugging.

itshanney:
@teyrana FWIW, I never really look at a changelog. I always look at a the source control history. The changelog in its current form is basically equivalent to release notes, which we also already have and are quite good.

This may sound blasphemous, but I would propose we just eliminate ChangeLog file entirely. Anybody that's truly interested in what the code looked like at any point in time can go through git history. And git history is way more valuable because it provides the full diff of what changed.

teyrana:
@itshanney 
Oh, sure.  **We** never use it, we're developers :)  But it's not for us, it's for the users.  It saves us from answering a bunch of users going "why doesn't this work the way it used to!?"  Not everyone wants to dig into the code.  And not everyone even _wants_ to....

In this context, it makes more sense to just have one per release -- corresponding to the end-user program.  i.e. Openrocket Swing UI.

README.md
itshanney: @JoePfeiffer wiki updated: https://github.com/openrocket/openrocket/wiki/Instructions-to-Build-from-Terminal

@teyrana Removed Running OpenRocket section to be deferred to the wiki above. Added Installers section.

README.md
itshanney: Removed copyright declaration.

itshanney:
@teyrana What if we remove the file and use the wiki instead? https://github.com/openrocket/openrocket/wiki/Change-Log

README.md
JoePfeiffer: I just made a slight change to the instructions -- corrected where to build from, and said where the JAR ends up.

teyrana: Okay, that's fine, I have no object to leaving the native builds in there. . We should probably mention all the methods, here.  

The more I think about it, this README is probably an entry point for new developers -- so it should mention enough to get people started, for their preferred method. Not all the details-- it sounds like y'all want to move full details to the wiki.   But enough 

I do prefer repository readme  files to stand on their own.... but they don't need to answer all the questions.

teyrana:
As much as I like wikis for most cases, I don't think changelogs are a good match -- we want those to be read-only, and only updated on releases, right?   I think a single authoritative file in the repo should be the source-of-truth. (and optionally, we copy that to the website each release)

itshanney:
@teyrana That's fine to not put it in the wiki and keep it in the repo. My question now is what's the difference between ChangeLog and core/ReleaseNotes.md? Can they be "merged" in some manner? I ask because I would to propose we simplify to a single file. The presence of the two file presents some confusion in that the benefits of each are unclear. 

If we say ChangeLog is developer-focused, do we still need it given that developers will look at git commit history? If it's user-focused, it needs to be out of the code since most users aren't usually looking at code and the wiki and/or website is the best.

An alternate proposal here is that Release Notes move to the website, which is under git control, limiting access to who can update the file. In summary, the changes would look like:
1. Move core/ReleaseNotes.md to openrocket.github.io repo.
2. Delete core/ReleaseNotes.md from openrocket git repo.
3. Delete ChangeLog from openrocket git repo.

itshanney:
@teyrana Additional comment to my last comment. Looks like Release Notes are already on the main openrocket.info website: http://openrocket.info/release_notes.html

Would definitely propose we use that as the single source of truth and remove duplication of them from the main repo. Will update README.md and Github wiki to point to them on openrocket.info.

teyrana:
Hmm, yea, ok. You make a good case.  

Have a single version of the "release notes" on the website, which is written for end users.  And developers will refers to github commit history / issues / PRs.  

Nit:  the "release notes" in this scheme are not a source of truth, but merely an imperfect summary of the git-history.

wolsen:
Yeah, I think the release notes are what matters most as that's where we will outline the relevant user oriented changes.

I think a ChangeLog is also valuable for developers who are consumers of the OpenRocket artifacts (i.e. anyone who writes code which depends on the jars). However, I think this is a release artifact and can be as simple as a `git log --pretty=oneline 15.03..20.?`. This *can* live in the git repo, but doesn't have to.

Relying on the git history is useful for the *contributing developers*, but not necessarily for consuming developers. While the consuming developers may not use git specifically, they may check the changelog to see if it contains a specific fix. While there's probably not a ton of developers consuming the libraries, it would be nice to include this for them as a release artifact.

README.md
wolsen: > @teyrana What if we remove the file and use the wiki instead? https://github.com/openrocket/openrocket/wiki/Change-Log

Actually, now that I think about this more. The Changelog should not be a wiki entry. It can be a page on openrocket.github.io but it should *not* be a wiki entry which is editable.

itshanney:
@teyrana @wolsen @JoePfeiffer here's the summary of this PR and latest changes:

1. README.md and Github wiki updates to point to ReleaseNotes on openrocket.info website
2. openrocket/openrocket.github.io#10 out for review by @wcravens to ensure Release Notes are accurate. Will delete from this repo once live on openrocket.info.
3. ChangeLog remains, but lives in root folder.

If this all looks good, would propose we call this PR quits and merge upon approval.

teyrana:
Most Recent Travis Build (Broken):
https://travis-ci.org/github/openrocket/openrocket/builds/677385859

itshanney:
@teyrana that's from @wolsen's unit test failure as this PR has no code changes - just text files. 

I can't figure out in Travis how to make it rebuild the PR without submitting a change to the PR, which I don't want to submit a "blank" change just to make it run the build. 

Let me dig on Travis to see if I can trigger a build manually.

teyrana:
Are you kidding me?  So `unstable` itself fails builds, now? 🤢

teyrana:
added a fix for travis-ci commenting on builds:
https://github.com/openrocket/openrocket/pull/634

I imagine we'll want to get that straightened out, to clean up the rest of these, no?

wolsen:
> Are you kidding me? So `unstable` itself fails builds, now? 
> 
> Okay, if that's true, then we need to submit a fix pr, anyway. And that can fix both the tests and the travis updating... if it's even a code issue.

Yes, I broke the build when I merged commit 4326c71c. I submitted a PR fix as quick as I could get to it (@itshanney reached out to me on slack). It is now resolved in unstable. This is my fault, my apologies - this was also when I noticed the travis-ci stuff was missing.

Once I saw the travis run succeed, I merged that into unstable to resolve the unstable builds. This was my mistake and its also when I realized that travis wasn't commenting/etc.

Unstable should be fixed now and if @itshanney rebases on unstable and pushes the rebase it should clean this up.

itshanney:
@teyrana build passes and Travis seems to work correctly.

@wolsen can you confirm I did the rebase correctly?

itshanney:
@wolsen @teyrana definitely looks like I messed up the rebase. Will abandon this and replace with #635.

wcravens:
I realize that I'm late to the party.... but I have $0.02 that I need to spend...

Repositories are about recording change.  And arguably the Release Notes should change at the same time that a feature branch changes something in the application code.  So the changes should be recorded together.... and hence part of the same repo.  So there ya go... my $0.02 worth. :-)

PR #625, opened 2020-04-18
[address 618] improves display of freeform fin tabs on transition bodies

teyrana:
Minor change to display how fin tabs are display on transition bodies. 

Previously it was possible to display non-physical tabs which looked like the tab cut into the fin body, (outside of the mounted body) where the fin tab should be entirely contained within the mount-body.

Now, zero-height is interpreted as a "no fin tab" and a fin tab height of 0.00001 is interpreted as "give me a fin tab of minimum height"   On a transition, this results in an almost triangular fin tab, which 0 height at one end, and a non-zero height on the other end of the fin tab.  (see image below)

## Pre-existing Behavior 
renderfix

## Resultant Behavior
renderfix

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -203,7 +203,7 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	
 	/**
 	 * 
-	 * @param cant -- new cant angle, in radians
+	 * @param newCantRadians -- new cant angle, in radians
 	 */
 	public void setCantAngle(final double newCantRadians) {
 		final double clampedCant = MathUtil.clamp(newCantRadians, -MAX_CANT_RADIANS, MAX_CANT_RADIANS);
@@ -264,16 +264,16 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	 * Note this function also does bounds checking, and will not set
 	 * a tab height that passes through it's parent's midpoint.
 	 *  
-	 * @param newHeightRequest how deep the fin tab should project
+	 * @param newTabHeight how deep the fin tab should project
 	 * from the fin root, at the reference point
 	 * 
 	 */
-	public void setTabHeight(final double heightRequest) {
-		if (MathUtil.equals(this.tabHeight, MathUtil.max(heightRequest, 0))){
+	public void setTabHeight(final double newTabHeight) {
+		if (MathUtil.equals(this.tabHeight, MathUtil.max(newTabHeight, 0))){
 			return;
 		}
 		
-		tabHeight = heightRequest;
+		tabHeight = newTabHeight;
 
 		fireComponentChangeEvent(ComponentChangeEvent.MASS_CHANGE);
 	}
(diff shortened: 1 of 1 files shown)


core/src/net/sf/openrocket/rocketcomponent/FinSet.java
```diff
@@ -885,24 +884,25 @@ public boolean isRootStraight( ){
 		
 		final double xTabFront = getTabFrontEdge();
 		final double xTabTrail = getTabTrailingEdge();
+
+		// // limit the new heights to be no greater than the current body radius.
+		// this.tabHeight = Math.min( this.tabHeight,  bodyRadius );
```
wolsen, line 889: Can remove commented out code added here, unless you think its intended to add something?

core/src/net/sf/openrocket/rocketcomponent/FinSet.java
```diff
@@ -264,16 +264,16 @@ public double getTabHeight() {
...
 	 * a tab height that passes through it's parent's midpoint.
 	 *  
-	 * @param newHeightRequest how deep the fin tab should project
+	 * @param newTabHeightRequest how deep the fin tab should project
 	 * from the fin root, at the reference point
 	 * 
 	 */
-	public void setTabHeight(final double heightRequest) {
-		if (MathUtil.equals(this.tabHeight, MathUtil.max(heightRequest, 0))){
+	public void setTabHeight(final double newTabHeightRequest) {
```
wolsen, line 271: nit: Since you're changing the variable name anyways, does the Request part of the name really make any sense? I think its just as clear to set it as newTabHeight.

wolsen:
Looks like a fairly straightforward patch. A couple of nits inline.

core/src/net/sf/openrocket/rocketcomponent/FinSet.java
teyrana: Yea, good point.  remove the 'request' suffix in this method.

core/src/net/sf/openrocket/rocketcomponent/FinSet.java
teyrana: yea, removed.

teyrana:
nits fixed.

wolsen:
LGTM, I think this is clearly the right behavior

neilweinstock:
I didn't get to test it but from the rendering it certainly looks reasonable.

wolsen:
Hey Neil, you can still test it if you want. The code is logical in that its choosing the best place to draw the fin tab from (in the Y direction) rather than how it was doing it previously and I think overall its a change we want, but it may not address the issue fully - which is why I left the issue open.

The code change makes sense as it now considers both the transition height at the front of the fin tab as well as the back of the fin tab when determining how to draw the tab. Previously it was only considering the front which is why it has that weird line which cuts through the transition.

neilweinstock:
Behavior seems fine as described.  I only incurred this by accident by having a tab height of .001” or something nonsensical like that.  I was more concerned about the invisible fin problem that occurred at the same time as this one (that’s in a separate issue).

If you *want* me to test it, I’ll be happy to, given a jar file.


  *   Neil

PR #628, opened 2020-04-20
Move MotorRowFilterTest into correct package

wolsen:
MotorRowFilterTest lives in the wrong package and causes Eclipse
to fail building. Oddly, it still builds using the ant commandline,
but since it declares it should be in the package
net.sf.openrocket.gui.dialogs.motor.thrustcurve this change moves
it to the location it declares.


teyrana:
huh. This wasn't causing issues under ant or Intellij :/

PR #633, opened 2020-04-22
Fix 627

JoePfeiffer:
Janitorial -- removing some unused code


teyrana:
@JoePfeiffer 
Hey, can you comment briefly about how we know this code is, in fact, unused?

JoePfeiffer:
A recursive grep through the source tree doesn't find the code anywhere except in these definitions.  Also, OR compiles, runs and passes unit tests without it.

teyrana:
Okay, sounds good to me :)

Also, imho, compiling and passing unit tests is necessary, but not sufficient.

PR #634, opened 2020-04-24
[fix] restores travis-ci commenting on PRs

teyrana:


wolsen:
Aaaand for whatever reason travis-ci is back to commenting as well :-) whatever :-) LGTM!

teyrana:
> Aaaand for whatever reason travis-ci is back to commenting ....

It's not "for whatever reason"  I made a change specificially for this!  This PR fixes it :P

wolsen:
@teyrana Sorry I should have elaborated a bit more. Yes, this PR actually fixes it and is why I went ahead and merged it :-). My comment was clearly confusing, maybe a bit due to the lockdown craziness.

This PR adds a notification script to integrate with travisbuddy to add comments on the PRs as they are done. That's clear to me and travisbuddy absolutely comments, so yes - that fixes the general issue of not getting the notifications and feedback loop.

What was unclear to me, is why this particular change _also_ caused the travis-ci checks to start working again:

Screenshot from 2020-04-25 08-00-39

I suspect its something to do with the yaml itself, but these checks should have been working without the travisbuddy change itself. So the problem for me is that I still don't understand what specifically caused that piece to start working again, other than it has something to do with the yaml file itself - but it might be irrelevant.

This is where my confusion was born ^. But its clearly a fix to the problem, I just wished I understood better why it fixed the problem. I understand the travisbuddy commentary and notifications, I don't understand the travis-checks bit.

teyrana:
ahhh, okay. Gotcha.  

Yea, didn't really dig into the WHY why.   I'm happy enough that it works :)

PR #636, opened 2020-04-26
Update icons for Tube Fins and Rail Buttons

wolsen:
Rail Button icons have been updated to better match the theme
with the rest of the icons. Tube Fin icons have been updated
with a larger tube for the fin (rather than the launch lug icon)
in order to disambiguate the two options.

Here's a screenshot with the new icons - both for the buttons and in the component tree view

Screenshot from 2020-04-25 20-57-15


teyrana:
Those look nice!  Did you make them yourself?

wolsen:
> Those look nice! Did you make them yourself?

Yes and no. I made the rail button and a tube fin outlines using Inkscape then replaced the layers of relevant xcf.gz files within gimp (the new rail button one started from the launchlug base file). I haven't used either inkscape or gimp for anything useful before, so ... it took awhile

teyrana:
Hmmm, the background transparency seems to be spotty, when I pull your branch down.  Have you seen this before?
buttons-buggy-transparency

neilweinstock:
Those look great.

wolsen:
Ah right, didn’t see that. Probably botched the transparency on export. Will fix it a bit later today

wolsen:
Updated the transparency and tweaked the small icons a bit. The tube fine small icon is larger to match the size of the other finsets. The small rail button icon was switched to a side-profile view as that makes it a little sharper. 

Also added the source files for the small icons as they are slightly different than the large button icons.

Obligatory screenshot

Screenshot from 2020-05-08 20-50-15

teyrana:
There we go!  Looks good :)

PR #647, opened 2020-05-02
[Fixes #645] fix auto radius calculation on Symmetric Components

### LINKED ISSUE #645: Change  calculation method of "getPreviousSymmetricComponent"
Triggered by undesired behavior of the "auto Radius" check box on a trailing Transition component.
The component should refer to the previous component _in-line_, and ignore those in attached pods.

teyrana:
Fixes the "auto" calculation on Symmetric Components (to include transitions and bodies) when pods are present (and previously would interfere)

Addresses: #645 
See that issue for more details, and example file.

## Todo   
- [x] 1. Auto-sizing within a stage
- [x] 2. Auto-sizing across multiple "AxialStage" instances, within the same rocket
    - [x] Add unit test
    - [x] Adjust behavior
- [x] 3. Unit Tests pass
- [x] 4. 'getNextSymmetricComponent' Usages:
  - [x] a. BarrowmanCalculator.java:635 -- desired behavior
  - [x] `swing/src/net/sf/openrocket/gui/dialogs/preset/ComponentPresetChooserDialog.java`
- [x] 5. 'getPreviousSymmetricComponent' Usages:
  - [x] a. BarrowmanCalculator.java:557 -- desired behavior
  - [x] b. BarrowmanCalculator.java:618 -- desired behavior
  - [x] `swing/src/net/sf/openrocket/gui/dialogs/preset/ComponentPresetChooserDialog.java`

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -629,9 +629,9 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 					}
 				}
 				
-				// if I'm the last componenet, set my base CD
-				// note I can't depend on the iterator serving up components in order,
-				// so I can't just do this after the last iteration.
+				// if I'm the last component, set my base CD
+				// note:  the iterator *should* serve up the next component.... buuuut ....
+				//        this code has is tested, and there's no compelling reason to change. 
 				if (s.getNextSymmetricComponent() == null) {
 					double area = Math.PI * pow2(s.getAftRadius());
 					double cd = base * area / conditions.getRefArea();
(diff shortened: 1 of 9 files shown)


teyrana:
## Done:
1. The explicit behavior this addresses is fixed. 
2. Unit Tests pass.

## TODO:

3. 'findNextSymmetricComponent' is referenced from BarrowmanCalculator.java:635.  Verify calculations.
4. 'findPreviousSymmetricComponent' is reference from BarrowmanCalculator.  Verify calculations.

JoePfeiffer:
This doesn't auto-size lower stages of a rocket.  The attached rocket has a 1" diameter booster stage in 15.03 and current unstable (which I think is the intended and expected behavior), but has the default 1.969 inch diameter with this PR.

Do we need to search the component tree until we find a symmetric component with the same axis as the component we're auto-sizing?

[two-stage-auto-diameter.ork.zip](https://github.com/openrocket/openrocket/files/4570877/two-stage-auto-diameter.ork.zip)

teyrana:
> This doesn't auto-size lower stages of a rocket. ... 

Oh, good point. I hadn't thought of that case.... I'll have to rework this bit.
(Well, I guess that's why we have reviews, right?

teyrana:
note: the component preset chooser dialog is still crashing.... so we need to fix #650 before we can verify behavior -.-

Issue #641, opened 2020-05-01
CG calculation error


neilweinstock:
CG is sometimes shown off-axis in 3D Finished view.  See attached design.

Side view:
![image](https://user-images.githubusercontent.com/243872/80773963-89cc2980-8b29-11ea-9318-8037c9a6b1ad.png)

3D Finished view:
![image](https://user-images.githubusercontent.com/243872/80773980-994b7280-8b29-11ea-9125-ec32e62001d5.png)

Tested on latest unstable build.

[cg_render_bug.ork.zip](https://github.com/openrocket/openrocket/files/4562127/cg_render_bug.ork.zip)

This issue was previously glommed onto #387, I've separated it out here.

teyrana:
Notes: 
1. iirc, The 2d view ignores the y,z components of the CG and CP: 
[RocketPanel.java#L634](https://github.com/openrocket/openrocket/blob/8288de651f19a0ad7fee0f82770a0c87b344df4e/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java#L634-L635)

neilweinstock:
Interesting.  Is it a bug that the 2d view ignores Y and Z?  I'm not sure why it should.

JoePfeiffer:
I'm guessing it should not.  CP is forced to be on-axis because its calculation with individual fins in a FinSet (instead of calculating the FinSet as a whole) erroneously calculates an off-axis CP (I've spent some time this morning trying to find where we concluded that, and haven't been able to locate the discussion nor the PR that does it.  I know I'm just looking in the wrong place).  I'll hazard a guess that before we forced the CP calculation, the side view rendering did it instead and what Dan is finding there is an artifact.

CG off axis seems like it has to be an error in the calculation, which was being masked in the side view rendering.

(all this is speculation and memory, of course)

teyrana:
> Is it a bug that the 2d view ignores Y and Z? I'm not sure why it should.

I interpret it as a bug, certainly.   It's an optimization which makes sense when everything is on one axis, but it should be updated. 

Why?:  it's simply ignored at display:  check it out:
https://github.com/openrocket/openrocket/blob/8288de651f19a0ad7fee0f82770a0c87b344df4e/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java#L634-L635

I can add that particular change to the PR fixing this issue.

PR #652, opened 2020-05-05
Avoid considering cp with 0 weight when calculating worst-case CP for display in RocketInfo

### LINKED ISSUE #648: Error in calculating CP with phantom body tubes:
CG is sometimes shown off-axis in 3D Finished view.  See attached design.

Related to #641, but this issue is for CP calculations.

Side view:
image

3D Finished view:
image

Tested on latest unstable build.

[cg_render_bug.ork.zip](https://github.com/openrocket/openrocket/files/4562127/cg_render_bug.ork.zip)

If the pod-fins are offset *past* the end of their parent bodies (Offset via BOTTOM: x > 0) ... CP is NaN.

If the pod-fins are offset *past* the end of their parent bodies (Offset via BOTTOM: x > 0) ... CP  is reasonably.

JoePfeiffer:
This fixes #648 (errrors in CP calculation).  The fins in the first example design in the issue are extremely small, and at some angles of attack can result in a CP weight so small as to be meaningless.  This PR only considers CP's with a non-zero weight in finding the worst case for display in the RocketInfo panel.

Also, in RocketPanel, a comparison was being made between CP weight at worst case and rocket minimum mass.  These have nothing obvious to do with each other, so this PR changes the test to be against MathUtil.EPSILON to avoid confusion in the future (this doesn't change behavior, as MIN_MASS is set to EPSILON in the MassCalculator class).

This doesn't address the off-center rendering of CG in some cases (#641), which is referenced in #648 but appears to be unrelated.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/AbstractAerodynamicCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/AbstractAerodynamicCalculator.java
@@ -4,6 +4,7 @@ import java.util.Map;
 
 import net.sf.openrocket.rocketcomponent.FlightConfiguration;
 import net.sf.openrocket.rocketcomponent.RocketComponent;
+import net.sf.openrocket.util.MathUtil;
 import net.sf.openrocket.util.Coordinate;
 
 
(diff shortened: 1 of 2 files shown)


teyrana:
LGTM.

Issue #654, opened 2020-05-06
CP is still off-center if pods.


JoePfeiffer:
In Neil's 3fnc test rockets, CP is still shown off-center in 3fnc-take2-pods.ork.  On-center in 3fnc-take2.ork, however.

JoePfeiffer:
Just clarified the title -- the display isn't the problem (in fact, was masking the problem).

JoePfeiffer:
Ah.  The CP code for FinSets recognizes when there are three or more fins in the set, and (because they're evenly distributed around the axis) pins CP Y and Z to 0.  This doesn't work when there are pods, since there may be only a single fin on the pod, even though the pods are evenly distributed.

Need to look more closely at the moment calculations...

teyrana:
We've merged #654, so I suppose this is due for re-validation.

teyrana:
@JoePfeiffer I'm assumed you're referencing #568?

teyrana:
I just ran the 3fnc files referenced above through our `unstable` branch, and the CP is on-axis.  
It looks like this was resolved by PR #657.

Closing as resolved.

JoePfeiffer:
I wasn't specifically referencing any other issue in particular, though the examples in #568 are a good test case.  Yes, issue is resolved.

PR #655, opened 2020-05-06
[fix] 2D Rocket Figure will now display off-axis CoM and CoP

teyrana:
This does _NOT_ affect any calculations -- it just more accurately displays those same values on the 2D view.

This also means that 2d views and 3d views display these values more similarly.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
@@ -24,6 +24,8 @@ import javax.swing.JPanel;
 import javax.swing.JSlider;
 import javax.swing.JViewport;
 import javax.swing.SwingUtilities;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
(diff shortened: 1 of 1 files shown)


teyrana:
( Spin off from #641, #654 )

JoePfeiffer:
I think you're right, we need to do this -- we may decide to pin the CP to Y=Z=0 in some (most?) cases, but the figures need to show whatever we end up calculating.

But, in 2D view, as I rotate the rocket the CP stays stationary instead of rotating with the rocket.

JoePfeiffer:
This isn't rotating the CP with the rocket -- with the attached design (which is the off-axis CP example I created before with the nose cone removed so the CP is *really* off axis) the CP just sits at the top of the body tube as the rocket is rotated.
[two-fins.ork.zip](https://github.com/openrocket/openrocket/files/4595955/two-fins.ork.zip)

teyrana:
(display now respects the rotation slider)

JoePfeiffer:
That's got it.  My vote is we merge this so we actually see what we're calculating (whatever that turns out to be).

PR #657, opened 2020-05-07
Pin CP to axis (Fix 654)

JoePfeiffer:
My test for three or more fins in BarrowmanCalculator.calculatedNonAxialForces() turns out to be the only place in the code where it's possible to set the CP off-axis.  Everywhere else, CP is pinned to the axis; the code would be clearer if CP were just a Double instead of a Coordinate (but that's a bigger change than I want to make right now).

So anyway, this pins CP to the axis in this one case as well by removing the "if" statement.  In essence, the code now assumes radial symmetry; even if we're only looking at a single fin we assume there is a matching fin somewhere.

We may want to investigate whether the CP should be allowed to be off-center (say, modeling a glider) in the future, but not right now.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/AerodynamicForces.java
+++ b/core/src/net/sf/openrocket/aerodynamics/AerodynamicForces.java
@@ -407,24 +407,4 @@ public class AerodynamicForces implements Cloneable, Monitorable {
 		return this;
 	}
 	
-	public AerodynamicForces multiplex(final int instanceCount) {
-		
-		this.cp = cp.setWeight(cp.weight*instanceCount);
-		if(1 < instanceCount) {
-			this.cp = cp.setY(0f).setZ(0f);
-		}
-		this.CNa = CNa*instanceCount;
-		this.CN = CN*instanceCount;
-		this.Cm = Cm*instanceCount;
-		this.Cside = Cside*instanceCount;
-		this.Cyaw = Cyaw*instanceCount;
-		this.Croll = Croll*instanceCount;
-		this.CrollDamp = CrollDamp*instanceCount;
-		this.CrollForce = CrollForce*instanceCount;
-	
-		modID++;
-		
-		return this;
-	}
-	
 }
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -204,8 +204,7 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 					
 					Coordinate cp_inst = instanceForces.getCP();
 					Coordinate cp_abs = context.transform.transform(cp_inst);
-					if ((comp instanceof FinSet) && (((FinSet)comp).getFinCount() > 2))
-						cp_abs = cp_abs.setY(0.0).setZ(0.0);
+					cp_abs = cp_abs.setY(0.0).setZ(0.0);
 					
 //					if( 1e-6 < cp_inst.weight) {
 //						System.err.println("_________ cp:inst: (rel): " + cp_inst.toString());
--- a/core/test/net/sf/openrocket/aerodynamics/BarrowmanCalculatorTest.java
+++ b/core/test/net/sf/openrocket/aerodynamics/BarrowmanCalculatorTest.java
@@ -338,20 +338,13 @@ public class BarrowmanCalculatorTest {
 		// results with and without it
 		// cpNoPods (0.34125,0.00000,0.00000,w=16.20502) -- interference disabled
 		// cpNoPods (0.34797,0.00000,0.00000,w=19.34773) -- interference enabled
-
-		// another note:  the fact that this is seen as three one-fin
-		// FinSets instead of a single three-fin FinSet means the CP
-		// will be off-axis (one of the fins is taken as having an
-		// angle of 0 to the airstream, so it has no contribution).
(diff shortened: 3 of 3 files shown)


teyrana:
Okay, this is the opposite of #655.   It's silly to do both.  We should either pin both the calculation AND the display, or neither.

And fwiw, I'd rather calculate it in 3d.  If a rocket has an off center Cp, that's REALLY important.

Even for gliders, I would argue.  Wouldn't hurt to get some feedback from gliders designers, ofc

JoePfeiffer:
I believe this is behavior that goes back a long, long time -- where can I find the source for 15.03?  Was that the initial checkin to the old Master branch?  If 15.03 calculates it off-center this is a regression we need to look at now (and fixing will be painful); if it has been the behavior all along investigating it can wait until after next release.  I sort of like the idea of having the renderings display it wherever we calculate it and (assuming this isn't a regression) pinning it in the calculation.  That way if we do end up calculating off-center when appropriate in the future we won't get bitten by the rendering not showing it.

FWIW, I created the attached rocket in 15.03 with two fins at a 120 degree angle; the component analysis only shows the CP x location (if it's calculating it off-center, the only place I can see to get that information is in the renderings, which may well be pinning it in the display).
[two-fins.ork.zip](https://github.com/openrocket/openrocket/files/4593893/two-fins.ork.zip)

teyrana:
> I believe this is behavior that goes back a long, long time -- where can I find the source for 15.03? Was that the initial checkin to the old Master branch? 

It's just a [ tag ]( https://github.com/openrocket/openrocket/tree/release-15.03 )

> If 15.03 calculates it off-center this is a regression . . . 

I don't quite follow here... do you mean if 15.03 calculates a non-zero CP on the example rocket in ~#654~ #648 ?

I'm pretty sure (and we'll check this) that 15.03 is just too simple in the way it calculates Cp -- iirc, it only calculates in the x-y plane, and simply *ignores* the x-z plane)

Personally, I consider this behavior to be somewhere between a barely-tolerable-simplification, and an outright bug (_in 15.03_); but this question has been lurking behind a _bunch_ of the Cp issues & PRs for the past year.    We'll can't really sweep it under the rug any longer, and this is as good a reason to surface it, as any.

> I sort of like the idea of having the renderings display it wherever we calculate it and (assuming this isn't a regression) pinning it in the calculation. That way if we do end up calculating off-center when appropriate in the future we won't get bitten by the rendering not showing it.

/agreed

> FWIW, I created the attached rocket in 15.03 with two fins at a 120 degree angle; the component analysis only shows the CP x location (if it's calculating it off-center, the only place I can see to get that information is in the renderings, which may well be pinning it in the display).
two-fins.ork.zip

Yea, so that's the other thing -- if we're pinning x = y = 0 in the 2d render, we should pin them in the 3d render, too.   Or the converse.

> Just marked this PR as a draft until we figure out whether it's a regression.

Good point.  I've converted #655, as well.

JoePfeiffer:
> > If 15.03 calculates it off-center this is a regression . . .
> 
> I don't quite follow here... do you mean if 15.03 calculates a non-zero CP on the example rocket in ~#654~ #648 ?
I meant if the code in 15.03 could calculate an off-center CP, whether we have an example at present that shows it or not.
> 
> I'm pretty sure (and we'll check this) that 15.03 is just too simple in the way it calculates Cp -- iirc, it only calculates in the x-y plane, and simply _ignores_ the x-z plane)
> 
> Personally, I consider this behavior to be somewhere between a barely-tolerable-simplification, and an outright bug (_in 15.03_); but this question has been lurking behind a _bunch_ of the Cp issues & PRs for the past year. We'll can't really sweep it under the rug any longer, and this is as good a reason to surface it, as any.

After a look at the FinSet CP calculations, I'm confident 15.03 can't generate an off-axis CP.  My view is that we've been at this release long enough that anything that isn't a a regression or new bug is acceptable; more accurate calculation of the CP (in particular, whether it should be off-axis in some cases) should come after the next release.

The issue that's been behind the CP and rendering issues is that I introduced a single circumstance in which an off-center CP can be calculated, without making that possibility be the general case (in which case all the off-center CPs ought to cancel out in the various fins-on-pods examples we've been fighting with).

teyrana:
fyi, the code that calculates CP in 15.03 is [here for fins](https://github.com/openrocket/openrocket/blob/f3440bac6846fd60da5dcac76e772f0539541e69/core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java#L116-L232)   ... (for example).

And you can see from [Line 125]( https://github.com/openrocket/openrocket/blob/f3440bac6846fd60da5dcac76e772f0539541e69/core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java#L125 )  That it only calculates the CP in the x-y plane.

And you can see from [line 232](https://github.com/openrocket/openrocket/blob/f3440bac6846fd60da5dcac76e772f0539541e69/core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java#L232) and especially [260 ](https://github.com/openrocket/openrocket/blob/f3440bac6846fd60da5dcac76e772f0539541e69/core/src/net/sf/openrocket/aerodynamics/barrowman/FinSetCalc.java#L260) that it's hard coded to only calculate the x-component.

teyrana:
So you're saying that if we just merge this (657) and 655, we'll be done with the off-axis CP values?

JoePfeiffer:
> So you're saying that if we just merge this (657) and 655, we'll be done with the off-axis CP values?

I believe so, yes.

PR #658, opened 2020-05-08
Add snapcraft information for building snaps

wolsen:
Snaps are a universal packaging format for Linux platforms and is available for most major Linux distributions.

This commit adds support for building OpenRocket as a snap. The snap convention is to place snap relevant information in the top level snap directory. The snapcraft.yaml file provides information regarding what should be in the snap and how to build it.


teyrana:
Oh, interesting.  

So how would we go about installing this?   (I have a 18.04 laptop I can test this with.)

Also,  OR developers ourselves may not have experience with Snaps / snapcraft -- _even_ among those of us who use linux.  Do you have a quick-start guide from Canonical?  (Probably that we can link from our documentation / wiki?)

wolsen:
Great questions, and apologies I didn't provide them with the PR initially. I'm on a funky work schedule this week and it was well past my bedtime :-) I saw the commentary on the Ubuntu package installation and figured it was worth pushing this up right before I went to bed. I wanted to take some time to provide a detailed write-up as many folks will certainly be new to snaps and building snaps.

**Snapcraft Information**
The snapcraft.io [docs](https://snapcraft.io/docs) site has a lot of good information on (including quick start, how to use, and lots of other details). While the docs are fairly good, there are some edges which aren't necessarily well documented. For help, there's a [discourse forum](https://forum.snapcraft.io/categories) setup and the devs are fairly responsive there. I'm also available to help anyone as necessary - hit me up in slack in the openrocket-dev group.

Some of the things that are easy to get tripped up on are the permissions. For example, snaps do not have access to all hidden folders in user home directories. The logic is that these may contain sensitive information and they'd prefer not to expose it. For example, the ~/.java folder is where userPrefs are stored and the snap won't have access to it.

There's some really good first snap [tutorials](https://snapcraft.io/first-snap) available on the snapcraft.io site and they are worth walking through if you're brand new to snapping.

**Building a snap on this Branch**
Building a snap on this branch is fairly straight forward. It simply requires you to install snapd, install snapcraft and then install the built snap.

1. **Install snapd**
   If you're already running on Ubuntu 18.04, snapd is pre-installed and you should be ready to go. If you happen to be running another distro, it may or may not have snapd installed. It's generally as easy as running `sudo apt install snapd` (debian and variants), `sudo dnf install snapd` (fedora), etc.   The docs will help you [install it on your platform of choice](https://snapcraft.io/docs/installing-snapd).

2. **Install snapcraft**
  Snapcraft is the tool used to build snaps. Snapcraft itself is delivered as a _classic*_ snap, which means that it has full access to the system. To install snapcraft, use snap:
```
  $ snap install snapcraft --classic
```

3. **Snap OpenRocket**
  At this point, you're ready to create the OpenRocket snap. The first time you run this, it will likely take a long time as it has to download an image to run in a virtual machine (using [multipass](https://multipass.run/)), then install some gnome libraries into the virtual machine (due to the gnome-3-28 extension defined in the yaml file), and then finally build OpenRocket. This is all done when you run the `snapcraft` command (within your OpenRocket working directory):

```
  $ snapcraft
  ...
  Priming launcher
  Priming openrocket
  Could not find '/snap/core/current/usr/bin/snap', validation of the license string will only take place 
  once pushed to the store.
  Snapping 'openrocket' |
  Snapped openrocket_19-xx-alpha-12_amd64.snap
```
  This will create the file openrocket_\<version\>_amd64.snap in the working directory.

4. **Trying out the Snap**
  You can try out the built snap in a couple of different ways, but the easiest way is just running it in devmode:

```
  sudo snap install ./openrocket_19-xx-alpha-12_amd64.snap --devmode
```

  You can launch the program from the command line by simply running `openrocket` or finding the OpenRocket icon in the applications menu.

5. **Rebuilding OpenRocket**
  Subsequent builds run much faster as you generally don't need to re-run the gnome-3-28 extension stuff. Some packages may get updated from time to time (for security fixes, etc), but its pretty easy to simply rebuild it. I recommend doing a clean of the openrocket _part_ and rebuild rather than cleaning everything and rebuild (to avoid the gnome-3-28 extension penalty).

```
  $ snapcraft clean openrocket
  $ snapcraft
```

**Trying without building it yourself (i.e. a test snap I'm building for you)**

I have built and registered my own space in the [snap store](https://snapcraft.io/store), however it is not publicly visible (but anyone can install it if they know about it).

It's fairly straightforward:

```
$ sudo snap install --edge wolsen-openrocket
```

It will now be installed and ready to use... however because it is registered and published with my namespace `wolsen-openrocket`, the name `openrocket` is not setup as the executable name. We can fix this by simply aliasing the `openrocket` command to the one in wolsen-openrocket (which is <snap-name>.<app-name>):

```
$ sudo snap alias wolsen-openrocket openrocket
```

You can now run `openrocket` from the command line.

** Limitations / Known Issues **
- ~/.java/.userPrefs cannot be read by the snap since it is confined. Instead the java userPrefs are located in the ~/snap/openrocket/common/.java/.userPrefs folder. So initial switch from a debian package/java jar file to a snap will lose user preferences (last files opened, etc). This can be worked around by _copying_ (not symlinking) the user preferences into the user's snap dir:
```
$ mkdir -p ~/snap/openrocket/common
$ cp -r ~/.java ~/snap/openrocket/common/
```
- .openrocket folder has the same limitation as above, though I need to look at this a bit more to make sure its working as intended. I'll mark the PR as a draft for the time being.
- Print preview does not work. I haven't figured this one out just yet, but I _think_ it's because the java.awt.Desktop.open() command is trying to open an external process, for which the snap is prevented from doing since it doesn't have permissions. Its a minor nit for now, as you can export a PDF and print from that (which is technically all the print preview does on linux anyways, but just using a temporary file).

_* Sidebar: classic snaps differ from standard snaps in that they have unsandboxed access to your system, kind of like a tradional linux package would (e.g. deb, rpm, etc). They have the benefits of including all their dependencies, but do not have all the security benefits. Some non-Ubuntu distributions will fall back to degraded confinement, primarily due to some kernel level details. Its outside the scope of this discussion, but I think its interesting to understand._

JoePfeiffer:
I've started trying this against the little menagerie of Linux distributions I accumulated while building my 15.03 appimage -- thoughts so far (particularly in comparison against appimage)

1) I really like how easy it appears to be to set it up and build it.  How exactly does it figure out what dependencies to include?  The yaml file doesn't seem to require enough...  for the appimage, I basically had to construct a directory tree with everything needed, which took several iterations.

2) I'm not so thrilled with the need to install snapd.  I don't like having multiple package managers; both apt and pip is bad enough...  appImage just needs FUSE, which pretty much every distribution has.  Of course, anyone who doesn't want snapd can just run the jar file (I'd like to say hopefully we'll have people packaging the new release, but I'm not sure how hopeful I am).

3) How concerned are we about antique systems?  I notice snapd isn't available for centos 6 through its standard repositories (centos 6 stopped receiving full updates in 2017, and is scheduled to stop receiving maintenance updates this December).

On balance, I don't see my #2 and #3 above as deal-breakers.  Unless there's a *much* easier way of producing appimages than the way I did it (actually creating the image is easy; figuring out what goes in it takes effor), we're better off with snap.

It would be interesting to see how flatpak compares.  Also, of course, there's no particular reason we couldn't have multiple packages, if other people want to go to the effort.

To wrap up this long-winded comment -- I had no problem building the snap, and it runs fine on my Debian 10 workstation.  I'll keep trying against the other specimens in my menagerie, but at the moment it looks good to me.

wolsen:
> I've started trying this against the little menagerie of Linux distributions I accumulated while building my 15.03 appimage -- thoughts so far (particularly in comparison against appimage)
> 
> 1. I really like how easy it appears to be to set it up and build it.  How exactly does it figure out what dependencies to include?  The yaml file doesn't seem to require enough...  for the appimage, I basically had to construct a directory tree with everything needed, which took several iterations.

There's a few ways the dependencies are declared in this snapcraft.yaml file. I'll be detailed as it is a learning exercise for all the developers involved (and will make good notes for what should be included in the Wiki). Some of the dependencies are explicit, while others are provided by other abstractions/facilities.

First, there's the key `base`, which is set to the value `core18`. This tells snapcraft to base the builds on the Ubuntu Core 18 image (which is 18.04 Bionic). Subsequently, packages will generally be sourced from the Ubuntu archives.

Next is there's the plugin for the openrocket part itself:

```yaml
parts:
  openrocket:
    plugin: ant
```
The ant plugin will indicate that this part depends on the ant build system and jdk libraries. The [ant plugin](https://snapcraft.io/docs/ant-plugin) has additional options which can be specified which allow for finer grained control. The defaults work well for OpenRocket so I'm not tweaking things like the specific jdk version for building (default jdk in 18.04 is Java 11).

Next, specific build dependencies are declared as build-packages and the packages which should be included in the final image are declared as stage-packages.

```yaml
    build-packages:
      - wget
    stage-packages:
      - openjdk-11-jre
      - ca-certificates
      - ca-certificates-java
```
You can see that I've declared wget as a build-package (which is wrong). Build-packages are those that are explicit dependencies. Since the OpenRocket project includes the java dependencies already, nothing really needs to be specified here.

The `stage-packages` are those packages which should be staged for inclusion in the final snap. The [stage cycle](https://snapcraft.io/docs/parts-lifecycle) is where all of the built parts come together, so we can add additional packages at this point which are required for runtime-dependencies. Here, we depend on the openjdk-11-jre and the certificates files.

The one (big) thing that I've skipped over so far is that this is a GUI application, so there are naturally graphical library dependencies. In order to make this easier for application developers, the snapcraft community has provided what they call [extensions](https://snapcraft.io/docs/snapcraft-extensions) which allows the developer to leverage someone else's experience and knowledge to incorporate it into their packaging system.

We leverage the `gnome-3-28` extension in the OpenRocket snapcraft.yaml file to include all of the graphics dependencies. When the extension is declared, the snapcraft.yaml file is modified just before the build process to incorporate all of the information contained in the extensions. You can view the actual snapcraft.yaml file that will be built by running:

```bash
$ snapcraft expand-extensions
```

The output looks like [this](https://gist.github.com/wolsen/e9b2b35a5198cc6ad368f34d27990de1), which shows some of the complexity that is hidden.

Hopefully this clears up some of how the dependencies are declared. Speaking from my own experience with snaps and snapcraft, it took me awhile to wrap my head around some of the dependency stuff, but I like that it is fairly simple.

> 2. I'm not so thrilled with the need to install snapd.  I don't like having multiple package managers; both apt and pip is bad enough...  appImage just needs FUSE, which pretty much every distribution has.  Of course, anyone who doesn't want snapd can just run the jar file (I'd like to say hopefully we'll have people packaging the new release, but I'm not sure how hopeful I am).

I hear you on the multiple package manager front on my local system. I'm not entirely thrilled with it either.

> 3. How concerned are we about antique systems?  I notice snapd isn't available for centos 6 through its standard repositories (centos 6 stopped receiving full updates in 2017, and is scheduled to stop receiving maintenance updates this December).
> 
> On balance, I don't see my #2 and #3 above as deal-breakers. Unless there's a _much_ easier way of producing appimages than the way I did it (actually creating the image is easy; figuring out what goes in it takes effor), we're better off with snap.
> 
> It would be interesting to see how flatpak compares. Also, of course, there's no particular reason we couldn't have multiple packages, if other people want to go to the effort.
> 

Agreed. I think it would be desirable to make OpenRocket as easy to consume as possible. The Linux ecosystem tends to have more opinions and varieties than other platforms do. I'm personally happy to see OpenRocket delivered in as many formats as possible; as long as the means to build the packages are included in the source trees somewhere (it doesn't even have to be this main source tree, it could be a separate repository within the OpenRocket namespace on github).

> To wrap up this long-winded comment -- I had no problem building the snap, and it runs fine on my Debian 10 workstation. I'll keep trying against the other specimens in my menagerie, but at the moment it looks good to me.

Thanks for the feedback! I'm glad to see it works for you. Hopefully my long response here helps clear up some of your questions.

teyrana:
@wolsen 
Hey, this PR is still a draft.  Are you still working on it, or should I take a look?

wolsen:
@teyrana 
I'm still looking at a couple of aspects for this regarding the print stuff and the local directory stuff (e.g. the .openrocket and .java preferences files). I want to make sure the migration path would be clear for anyone wanting to try this. I just haven't had time recently to get back to it (but will do so in the not too distant future)

That being said, it should be fairly functional as it is. If you wanted to give it a go and provide any feedback, it'd be more than appreciated.

wolsen:
Last two commits in this branch help handle the printing scenario and the migration path from package/jar installed versions to the snap installed version.

The first commit enables a user to connect their local Java preferences folder and their default openrocket folder for access by the application.

Upon installing the snap, the user must _manually_ connect the defined plugs to allow for the OpenRocket application to access the hidden folders, since they are restricted by default. It is possible to petition that these be auto-connected upon installation once the snap is available in the store. I'm happy to work with the Snap/Snapcraft folks to petition for this, but also include the relevant information in the docs as necessary.

To try this branch out, you can build the snap.

If you've already built the snap, first clean the openrocket part:
```bash
$ snapcraft clean openrocket
```

Then rebuild the snap:
```bash
$ snapcraft
```

Install the local snap artifact and connect the plugs:
```bash
$ sudo snap install --devmode ./openrocket_19-xx-alpha-12_amd64.snap
$ sudo snap connect openrocket:preferences :personal-files
$ sudo snap connect openrocket:openrocket-db :personal-files
```

This commit also changes the default tmpdir to use for any temporary files created by OpenRocket. It turns out that the snapd daemon which sets up the snap, sets up an overlayfs for the snap to have its own `/tmp` folder, but the actual path to the `/tmp` folder is something different than `/tmp`. To work around this, the java tmpdir is simply set to the $SNAP_USER_COMMON dir, which is a fully accessible path. Various bits in OpenRocket use temporary files (printing, decal editing, etc).

The second commit switches the printing/print preview behavior to use the xdg-open application on Linux when confined within a snap. If running outside of a snap confined environment, the standard java.awt.Desktop.open() method is used.

This is necessary because the java.awt.Desktop.open() method on Linux platforms uses the `gnome_url_show` method from the gtk libraries. A set of gtk libraries are bundled within the snap in order for the graphics to work. An alternative approach would be to bundle additional libraries within the snap (like evince for PDFs). There are pros and cons to that approach and I favored user preferences by using xdg-open here.

wolsen:
Okay, my apologies for the novels when writing these comments - but I figure its best to share the knowledge that I have learned when going through this so far.

After digging some more, I think there's a couple of things that still need to be addressed for this to really be ready and would love some input from @JoePfeiffer, @teyrana, and any others that are interested.

**Question 1**
There are several entry points (`public static void main`) methods throughout, which should probably have convenient entry points to startup. After looking through them all, most of them appear to be utilities for development purposes and not necessarily used by the end-user. I believe the following utilities may have value for the end-user:

```java
net.sf.openrocket.utils.ComponentPresetEditor
net.sf.openrocket.thrustcurve.SerializeThrustcurveMotors
```

The full list of entry points is available [here](https://paste.ubuntu.com/p/YfX43cjM4n/)

Adding a new entry point to the snap is relatively straightforward, simply adding an extra bash script or two for the launchers and adding a new 'app' in the list of available apps in snapcraft.yaml file but before I endeavor on that, I'd like to get some input on the relevant set of utilities which should be included. What are the relevant utilities that should be exposed to the end-user?

**Question 2**
Editing decals (something I never do) is an option available to users of OpenRocket, however the snap confinement makes things a bit tricky. I do believe that this is a power user option as it requires knowledge of graphics programs, etc but I want to make sure that we consider this use case.

On Linux today, the java.awt.Desktop.edit() option is not available so users must manually specify the application that they want to run. However, the snap confinement does not allow for the application to launch any other application that exists because a misbehaving application could launch a nefarious program without the user's knowledge or consent.

There's a few options that we could go down to resolve this:

1. Require that the user installs and uses the jar file for this scenario. This partially defeats the purpose of having the snap in the first place, but they are in a power user scenario (IMO). Its an option that we could take, but I'm not sure its the right option.
2. Include a photo editing app (i.e. gimp) within the snap. This will increase the bundled application size, and packaging some of these applications can be a bit tricky. Additionally, since photo and graphics editing generally requires some enhanced training the user will likely have a preference for an application they are used to and we risk selecting the wrong one (though its our right to have an opinion here).
3. Use the xdg-open utility. This will likely open an image viewer that handles png files by default for the system. The user could choose to change the default application or they could get the path and edit the file using their favorite editor. The UI would need to gracefully handle this change of behavior as the user shouldn't be allowed to specify their own command to run since that will always fail.

At this point, my preference is leaning towards option 3 as it still provides the user a means to edit the graphics file and can be documented as to how this works. However, what do you guys think about the right behavior here?

teyrana:
hmmmm, it seems to run well enough.
1. Although I had to manually install Multipass.  Does the snap installer not recurse to dependencies? 
2. I saw some minor windowing glitches on my machine.   I know you're running 20.04, but have you seen this?:

snap_windowing_glitches

(note the lack of lines around the tabs -- particularly the lack of highlighting the currently-selected tab.)

I also don't develop java on my ubuntu machine very often, so I'm not sure how widespread this is.

wolsen:
> hmmmm, it seems to run well enough.
> 
> 1. Although I had to manually install Multipass.  Does the snap installer not recurse to dependencies?

Hmm, it recurses some dependencies (i.e. core snaps and such). I already had Multipass installed so I missed this. You can also run using lxd containers instead of vms by Multipass, I think the flag is snapcraft --lxd, but you'd probably need lxd installed via snap as well.

> 2. I saw some minor windowing glitches on my machine.   I know you're running 20.04, but have you seen this?:

I actually haven't seen this before. I'll fire up an 18.04 machine and see if I can chase down the root cause of that.

wolsen:
> I think the one you've currently packaged is the main one. I've never used the others, and I haven't heard much about other people using them, either -- so they don't need to be packaged.
> 
> I think they're for command-line usage of the openrocket.

Yeah, we can always add them in later if there's a request for them. There's little (if any) documentation around them and while they may be useful for some, there's some work-arounds available.

> 
> ## Question 2
> > Editing decals (something I never do) is an option available to users of OpenRocket,  . . .
> 
> Do you mean placing existing images on a rocket, or literally editing the image itself?

If you take a component and look at the Appearance tab, there's a Texture option. The edit option next to it opens the graphics editor. I just want to make sure the UI degrades appropriately to prevent the user from entering a condition that would result in errors.

This reminds me that you can import a file for the graphics, and maybe that's the right way to go on Snap confined installations. Just remove/disable the edit option and Graphics Editor section on the Graphics tab in preferences.

N.B. Neil actually has a fairly good [tutorial](https://www.rocketryforum.com/threads/decals-in-openrocket-a-not-so-quick-and-fairly-complete-tutorial.129192/) on the rocketryforum about decals.

> 
> I wasn't aware Openrocket had this ability...

I think its used by a small subset of the installation base, and if I had to guess - most of those users are not on Linux.

wolsen:
> I actually haven't seen this before. I'll fire up an 18.04 machine and see if I can chase down the root cause of that.

I can confirm this is seen just using the jar file as well. It appears to be something to do with GTK3, but not yet sure what it is. I don't see it on 20.04, but there's this JDK bug report - https://bugs.openjdk.java.net/browse/JDK-8232865 that suggests its a more general problem.

teyrana:
>  The edit option next to it opens the graphics editor. I just want to make sure the UI degrades appropriately to prevent the user from entering a condition that would result in errors.

Yea, that seems like an unnecessary, and brittle feature. -- there are many image-editing programs, how would we know which one to open, even if OR did have access?   This seems like a lot of complexity for not much benefit.

teyrana:
>  I don't see it on 20.04

If that's the case, I'm inclined to just not worry about it. The number of users who use linux AND ubuntu 18.04 AND snaps is just not that many people.

If someone actually complains, we can file an issue then.

wolsen:
Last push was to rebase on latest unstable

wolsen:
Okay, with the last couple of commits here I think this is ready. The last couple of commits ensure the functionality for the edit decals work within the snap confinement space, using the xdg-open bits.

>The number of users who use linux AND ubuntu 18.04 AND snaps is just not that many people.

Yep, fair enough - I think there are some and we can spend more time investigating it when users complain. I did note that the same issue exists not using the snap (aka with normal java) on my son's 18.04 laptop. I'll dig into that more in my "infinite free time" as I think its an Ubuntu issue rather than an OpenRocket issue.

teyrana:
@JoePfeiffer 
I'm okay with merging this. Do you have any objections?  Or just want to take it for a spin?

JoePfeiffer:
> @JoePfeiffer
> I'm okay with merging this. Do you have any objections? Or just want to take it for a spin?

I haven't had a chance to look at it (or much of anything, lately!).  If it's OK with you I'm fine -- If you want, I should be able to give it a look tomorrow afternoon.

PR #659, opened 2020-05-08
[ui] Moves new assembly buttons to their own row in the New Component panel.

### LINKED ISSUE #643: Two button issues
image

1) Why do the "New Stage", "New Boosters", and "New Pods" buttons have the word "new" in them?  *All* the buttons create a new component, that is their function.  Propose removing the word "new" from these three button names.

2) I'm pretty sure the "Mass Objects" section should start on a new row.

Tested on latest unstable.

teyrana:
This PR addresses, but does not completely fix #643.

result


neilweinstock:
New layout looks good, I think having assembly components up top like that will be a win with users first experiencing the new version.

wolsen:
Agree, the new layout looks nice. Everything looks fine code wise. While I generally think the consensus is this is fine, will let @JoePfeiffer weight in before merging.

JoePfeiffer:
I like the new layout.  The categories still need to be consistent in plural/singular and capitalization FWIW, my preference would be 
Assembly Components (as it is now)
Body Components and Fin Sets
Inner Components
Mass Objects

PR #662, opened 2020-05-09
[fixes #638] Scale Rocket now scales the radial distance of pods & parallel stages

### LINKED ISSUE #638: Scaling rocket doesn't scale radial distance of pods
Scaling seems to forget to scale the radial distance of pods.  Here's a design, before scaling:
image

Here it is after scaling 400%:
image

Tested on the latest unstable, based on commit 1c7171e.

teyrana:
There's no picture to show, because the rocket looks the same as before (in fact I had to undo-redo to see the size change :)


wolsen:
LGTM, can confirm scaling now works correctly.

PR #663, opened 2020-05-09
[Fixes #618] "Transparent" Fin now renders correctly

### LINKED ISSUE #618: Problem with display of fins attached to transition or nose cone
I am working with the build that Justin recently distributed in package form.

I experienced the following weirdness.  The attached design is stripped down to the bare essentials to show the bug.

I have a set of freeform fins that were originally attached to the body tube.  All was well:
image

I then moved the fins to the tail cone.  Invisible fins!
image

It also had problems display the root of the fin in side view:
image

Moving the fins to the nose cone behaves the same as on the tail cone.

Note that if I create *new* fins, they work just fine.  Just this pre-existing set of fins is having trouble.  It could be related to the fact that the *original* version of these fins had a "cutout in them" which may have created a backward loop at one point.  But I've deleted all those extra points and now they're simple four-point fins, and they persist in displaying this way.  Hopefully you can see what's wrong in the ORK file.

[clear-fins.ork.zip](https://github.com/openrocket/openrocket/files/4441590/clear-fins.ork.zip)

teyrana:
- It looked like this was due to the buggy fin-tab (i.e. crosses the body boundary) 
- upon setting fin-tab to zero-size, the fin renders correctly
- Therefore, the fin tab is simply disabled here.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -941,12 +941,7 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	 * but the minor performance hit is not worth the code complexity of dealing with.
 	 */
 	public Coordinate[] getFinPointsWithTab() {
-		final Coordinate[] finPoints = getFinPoints();
-		final Coordinate[] tabPoints = getTabPoints();
-		
-		Coordinate[] combinedPoints = Arrays.copyOf(finPoints, finPoints.length + tabPoints.length);
-		System.arraycopy(tabPoints, 0, combinedPoints, finPoints.length, tabPoints.length);
-		return combinedPoints;
+		return combineCurves(getFinPoints(), getTabPoints());
 	}
 	
 	@Override
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -299,7 +299,7 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	/** 
 	 * internally, set the internal offset and optionally validate tab
 	 * 
-	 * @param offsetRequest new requested shift of tab -- from
+	 * @param offsetRequest new requested tob offset
 	 */
 	public void setTabOffset( final double offsetRequest) {
 		tabOffset = offsetRequest;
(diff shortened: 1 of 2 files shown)


wolsen:
Hmm, aren't the fin tabs rendered in the 3d non-finished models today?

teyrana:
> Hmm, aren't the fin tabs rendered in the 3d non-finished models today?

Meaning?  

Do you want this change to apply to other modes?  Or you think fin-tabs should still be rendered in 3d mode?

wolsen:
Well, I personally use the 3D renderings to witness details like the fintabs needing adjustment (though I also use the 2D view. Initially, my thoughts were "oh yeah, no need to display fintabs in 3d renderings" but then I remembered that the "3D Figure" and "3D Unfinished" renderings today will draw the internal components (parachutes, centering rings, etc). Here's a before and after this change

**Before**
Screenshot from 2020-05-09 13-15-09

**After**
Screenshot from 2020-05-09 13-16-10

It _feels_ incomplete to me not seeing the fintabs. Maybe because I'm used to them.

teyrana:
Okay, adjusted the fix to include tabs.   

It looks like an inconsistent point-ordering was confused the OGL library.

Seems to look better, now, but please kick the tires, here.

core/src/net/sf/openrocket/rocketcomponent/FinSet.java
```diff
@@ -943,9 +943,15 @@ public Coordinate getFinFront() {
...
-		
+
 		Coordinate[] combinedPoints = Arrays.copyOf(finPoints, finPoints.length + tabPoints.length);
-		System.arraycopy(tabPoints, 0, combinedPoints, finPoints.length, tabPoints.length);
+
+		//System.arraycopy(tabPoints, 0, combinedPoints, finPoints.length, tabPoints.length);
+		combinedPoints[finPoints.length + 3] = tabPoints[0];
+		combinedPoints[finPoints.length + 2] = tabPoints[1];
+		combinedPoints[finPoints.length + 1] = tabPoints[2];
+		combinedPoints[finPoints.length + 0] = tabPoints[3];
```
wolsen, line 953: It turns out there's already a method which takes care of this within FinSet.java itself. This whole method can simply be replaced with:

```java
	public Coordinate[] getFinPointsWithTab() {
		return combineCurves(getFinPoints(), getTabPoints());
	}
```

teyrana: hahahaha, I wrote that function!  Guess I had a good reason for it :)

teyrana:
However, I've also found a further bug:
Negative tab dimensions still cause this -- I'll have to add more validations into FinSet.

teyrana:
@wolsen 
Okay, I added another commit, which tightens up the consistency checking on FinTabs.
Take a look, when you get a chance. :)

core/src/net/sf/openrocket/rocketcomponent/FinSet.java
```diff
@@ -299,7 +299,7 @@ public void setTabLength(final double lengthRequest) {
 	/** 
 	 * internally, set the internal offset and optionally validate tab
 	 * 
-	 * @param offsetRequest new requested shift of tab -- from
+	 * @param offsetRequest new requested tob offset
```
wolsen, line 302: nit: s/tob/tab/

PR #669, opened 2020-05-16
Resolves #666 -- fix the "Exception loading stream" error while loading certain rockets.

### LINKED ISSUE #666: Error opening file with newest unstable
Opening attached ORK file gives the following error:
image

This cropped up recently; it works fine in the build from late April tagged 1c7171ef.

[blackfish_golden_sunset.ork.zip](https://github.com/openrocket/openrocket/files/4610684/blackfish_golden_sunset.ork.zip)

teyrana:
The culprit was treating PodSets like stages in the `SymmetricComponent.getPreviousSymmetricComponent` and ``SymmetricComponent.getNextSymmetricComponent` methods.

[Testable Jar file: Openrocket-fix666.jar](https://drive.google.com/open?id=1YDi1j6dKpxm5zCRofQRmLnqPFpyntrKL)

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -299,7 +299,7 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	/** 
 	 * internally, set the internal offset and optionally validate tab
 	 * 
-	 * @param offsetRequest new requested tob offset
+	 * @param offsetRequest new requested tab offset
 	 */
 	public void setTabOffset( final double offsetRequest) {
 		tabOffset = offsetRequest;
--- a/core/src/net/sf/openrocket/rocketcomponent/SymmetricComponent.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/SymmetricComponent.java
@@ -608,21 +608,21 @@ public abstract class SymmetricComponent extends BodyComponent implements Radial
 			return null;
 		}
 
+		final ComponentAssembly assembly = this.getAssembly();
 		// might be: (a) Rocket -- for centerline stages
 		//           (b) BodyTube -- for Parallel Stages
-		final AxialStage stage = this.getStage();
-		final RocketComponent stageParent = stage.getParent();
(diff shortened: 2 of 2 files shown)


teyrana:
@neilweinstock
This works on the ornate rocket you posted.  Let me know if you have time to poke at it further.

neilweinstock:
Would be happy to, just send me a jar.  To be fair, this was the only rocket that I found which failed to load before, but I didn't go through much of my library.  This time I'll try the whole pile.

teyrana:
It was linked in the first comment :)

And it only seems to affect the pods & boosters rockets, if I understand the issue, correctly.

neilweinstock:
Sorry, dunno how I missed that. Will test tonight.

teyrana:
> Sorry, dunno how I missed that. Will test tonight.

Awesome! thanks.

neilweinstock:
All designs open successfully.  However, all my fin decals are now even more screwed up than before.  Likely not due to this PR, though; more likely (if I had to guess) a previous one where you were mucking about with fin tab rendering.

teyrana:
> all my fin decals are now even more screwed up than before. 

Sigh.  yea, that would be because we changed the point order, so consequently we need to change the texture coordinate order :(

neilweinstock:
Are fin tabs included in the object that gets the decal texture map?  It would be great if they weren’t.  That is, if the tab were a separate object (only from a 3D-render perspective) from the fin, and only the actual fin part got the texture.

I know I’m probably asking too much here but what the heck. 😊



teyrana:
@Neilweinstock

Yea, that sounds accurate.  Maybe we can do that.
I'd have to take a look.

neilweinstock:
I'm having second thoughts on my request, although I really think it's how it should work.  Reason: it would be a change from 15.03, which means existing models would "break".  Although it's quite easy to fix (just crop the decal images), it still feels like a no-no to change the way old models are rendered, even though the new way would be much "better".  What do you think?

teyrana:
@neilweinstock 

If you think this should be fixed in this release, please submit a new issue :)

with screenshot + ork file, preferably ;)

neilweinstock:
Agreed.

PR #670, opened 2020-05-17
[Resolve #571] Add message fins-on-phantom-bodies

### LINKED ISSUE #571: Add more descriptive messages for Fins-on-phantom-bodies
Currently the warning is something like "thick fins -- accuracy may suffer".  Change this to something like "fins on phantom bodies are allowed, but discouraged -- simulation accuracy may suffer"

Comments on wording are welcome!

@neilweinstock @JoePfeiffer

teyrana:
Adds a new error message to warn users about mounting fins on phantom bodies:

implementation

Runnable jar:
https://drive.google.com/open?id=1wY6ydj5zBnv4VT6QQlKepTxEuNWOLmQC

This is a purely cosmetic change; as such it's subjective.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1695,6 +1695,8 @@ Warning.RECOVERY_LAUNCH_ROD = Recovery device device deployed while on the launc
 Warning.RECOVERY_HIGH_SPEED = Recovery device deployment at high speed
 Warning.TUMBLE_UNDER_THRUST = Stage began to tumble under thrust.
 Warning.EVENT_AFTER_LANDING = Flight Event occurred after landing:  
+Warning.ZERO_LENGTH_BODY = Zero length bodies may not result in accurate simulations.
+Warning.ZERO_RADIUS_BODY = Zero length bodies may not result in accurate simulations.
 
 ! Scale dialog
 ScaleDialog.lbl.scaleRocket = Entire rocket
--- a/core/src/net/sf/openrocket/aerodynamics/Warning.java
+++ b/core/src/net/sf/openrocket/aerodynamics/Warning.java
@@ -380,7 +380,7 @@ public abstract class Warning {
 	public static final Warning FILE_INVALID_PARAMETER = new Other(trans.get("Warning.FILE_INVALID_PARAMETER"));
 	
 	public static final Warning PARALLEL_FINS = new Other(trans.get("Warning.PARALLEL_FINS"));
-	
+
 	public static final Warning SUPERSONIC = new Other(trans.get("Warning.SUPERSONIC"));
 	
 	public static final Warning RECOVERY_LAUNCH_ROD = new Other(trans.get("Warning.RECOVERY_LAUNCH_ROD"));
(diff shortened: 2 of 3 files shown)


neilweinstock:
Might I suggest a very slight trim, so maybe it'll fit a little more easily?

> Phantom bodies (zero length or radius) may produce inaccurate simulations.

Also note capitalization; in all warnings I checked, only first word was capitalized.

teyrana:
> Also note capitalization; in all warnings I checked, only first word was capitalized.

Aye.  Sentence casing?  Sure, we can change that.    In any case, I don't have any strong opinion there..

My concern -- does the message convey the actual rocket situation? in your opinion?

neilweinstock:
> My concern -- does the message convey the actual rocket situation? in your opinion?

Yes, although thought my edit, suggested above, made it just a bit more clear and succinct.

Since then, though, I've had a couple more thoughts:
1. This warning was originally a substitute for a thick fin warning.  That concerned only phantom bodies with zero radius.  Therefore, unless this warning is generated in other places as well, it shouldn't mention zero length.
2. There's no reason to include a definition of phantom bodies in the warning, much like "Thick fin warning" doesn't have a full definition of what a thick fin is.

Therefore, my new proposed edit would be:

> Body tubes with zero radius may produce inaccurate simulations.

If this message is indeed used in other places, triggered by a zero *length* tube, then you can say

> Body tubes with zero length or radius may produce inaccurate simulations.

These say exactly what your original wording did, just more succinctly.  What do you think?

- Neil

JoePfeiffer:
Maybe "Fins on body tubes with zero length or radius may produce inaccurate results"?

neilweinstock:
> Maybe "Fins on body tubes with zero length or radius may produce inaccurate results"?

Are we actually checking for zero length tubes anywhere?  If not we should leave out the "length or".

If this message is exclusively for fins on zero radius tubes (alternative to the "thick fin warning") then I'm in favor of Joe's wording.  It's quite precise.

Anyway, if this warning is just for fins on phantom tubes then

JoePfeiffer:
To the best of my knowledge, we aren't.  Yes, we could have the warning only talk about 0 radius (I'm just developing a pessimism that even if we aren't checking it now, sooner or later we're going to discover we should be).

neilweinstock:
I don't disagree, but in that case couldn't we just change it then?

JoePfeiffer:
Or even make that a separate warning, I suppose.

neilweinstock:
Either way.


teyrana:
Okay, I've split this into two separate warnings -- one for zero-length bodies, and one for zero-radius bodies.   

Exact messages are: 
1. "Zero length bodies may not result in accurate simulations."
2. "Zero radius bodies may not result in accurate simulations."

neilweinstock:
I would still prefer "may produce inaccurate simulations" or "may produce inaccurate results", but I'll sign off on what you have there.

JoePfeiffer:
lgtm

neilweinstock:
Looks like the messages both say "zero length" instead of one of them saying "zero radius".

Issue #671, opened 2020-05-18
simulation results in rocket design pane don't match results in flight simulations tab


JoePfeiffer:
In the attached rocket, the Altitude in the rocket design pane is significantly different from the value shown in the Flight Simulations tab (Max velocity and Max acceleration are also different, though not by as much).

Rocket design pane:
![designpane](https://user-images.githubusercontent.com/3065196/82165279-b09c9680-9871-11ea-8d9d-1b67fda190cf.png)

Simulation tab:
![simulationtab](https://user-images.githubusercontent.com/3065196/82165333-e2adf880-9871-11ea-9c11-64a2c0e998e9.png)

[Rocket Double Stage Constant Diameter.ork.zip](https://github.com/openrocket/openrocket/files/4641514/Rocket.Double.Stage.Constant.Diameter.ork.zip)

neilweinstock:
Is this one fixed?  Should it be closed?

JoePfeiffer:
Not fully -- I got distracted, but am back on it.

PR #672, opened 2020-05-20
Use simulation options (addresses 

JoePfeiffer:
This PR addresses #671 (data in RocketPanel doesn't match simulation results) by searching for a Simulation associated with the current FlightConfiguration.  Two things I'd like input on before proceeding:

(1) Yes, I said "a" simulation, not "the" simulation.  It is possible to have multiple Simulations associated with a single FlightConfiguration; this PR simply chooses one of them so there is no guarantee it'll pick the one the user prefers.  A change that would get around this would be to directly choose a current Simulation in the current "Flight configuration:" dropdown; we could either just put the Simulation names in the dropdown, or we have plenty of room to use something like
Simulation 1 [J381-8;J150-0]
concatenating the Simulation name with the FlightConfiguration name

(2) When the simulation is run to generate the statistics in the RocketPanel, it only runs to apogee of each stage.  If the user proceeds to go up and plot the simulation without rerunning it, the graph is disconcerting to say the least!  I wonder if I need to mark the Simulation as needing to be rerun before plotting?

### DIFF
--- a/core/src/net/sf/openrocket/simulation/SimulationOptions.java
+++ b/core/src/net/sf/openrocket/simulation/SimulationOptions.java
@@ -574,5 +574,28 @@ public class SimulationOptions implements ChangeSource, Cloneable {
 		
 		return conditions;
 	}
+
+	public String toString() {
+		return "SimulationOptions [\n"
+			.concat("    AtmosphericModel: " + getAtmosphericModel().toString() + "\n")
+			.concat(String.format("    launchRodLength:  %f\n", launchRodLength))
+			.concat(String.format("    launchIntoWind: %b\n", launchIntoWind))
+			.concat(String.format("    launchRodAngle:  %f\n", launchRodAngle))
+			.concat(String.format("    windDirection:  %f\n", windDirection))
+			.concat(String.format("    launchRodDirection:  %f\n", launchRodDirection))
+			.concat(String.format("    windAverage:  %f\n", windAverage))
+			.concat(String.format("    windTurbulence:  %f\n", windTurbulence))
+			.concat(String.format("    launchAltitude:  %f\n", launchAltitude))
+			.concat(String.format("    launchLatitude:  %f\n", launchLatitude))
+			.concat(String.format("    launchLongitude:  %f\n", launchLongitude))
+			.concat("    geodeticComputation:  " + geodeticComputation.toString() + "\n")
+			.concat(String.format("    useISA:  %b\n", useISA))
+			.concat(String.format("    launchTemperature:  %f\n", launchTemperature))
+			.concat(String.format("    launchPressure:  %f\n", launchPressure))
+			.concat(String.format("    timeStep:  %f\n", timeStep))
+			.concat(String.format("    maximumAngle:  %f\n", maximumAngle))
+			.concat(String.format("    calculateExtras:  %b\n", calculateExtras))
+			.concat("]\n");
+	}
 	
 }
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
@@ -682,9 +682,27 @@ public class RocketPanel extends JPanel implements TreeSelectionListener, Change
 			extraText.setCalculatingData(true);
 
 			Rocket duplicate = (Rocket) document.getRocket().copy();
-			Simulation simulation = ((SwingPreferences) Application.getPreferences()).getBackgroundSimulation(duplicate);
-			simulation.setFlightConfigurationId( document.getSelectedConfiguration().getId());
 
+			// find a Simulation based on the current flight configuration
+			FlightConfigurationId curID = curConfig.getFlightConfigurationID();
+			Simulation simulation = null;
+			for (Simulation sim : document.getSimulations()) {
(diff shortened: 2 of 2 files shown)


teyrana:
> (1) Yes, I said "a" simulation, not "the" simulation. It is possible to have multiple Simulations
> associated with a single FlightConfiguration;

Wait... what? I thought that was the definition of a FlightConfiguration -- that it's a unique configuration of the rocket?   

What changes then, are the multiple simulations at different flight environments? -- Launch site, weather, etc?

As far as resolutions -- 
(a) I'd vote to have simulation results on the simulation page, where the correspondence is clear ... but I don't expect that to be too popular.

(b) Second easiest would be for the front-page simulations to always be at 0 launch angle, zero wind, sea level, etc for that Configuration... and then a user must specifically choose and run simulations at "non-standard" conditions.

(c) Maybe flight configurations should link 1-1 with simulations?   You might have to reconfigure an engine or a parachute delay, but it would make things simpler.

(d) Adding a simulation drop-down _in addition_ to flight configuration drop down on the front page... is doable, but I wouldn't vote for it.  That seems like it's just adding more complexity.

teyrana:
> (2) When the simulation is run to generate the statistics in the RocketPanel, it only runs to apogee of each stage.  . . . I wonder if I need to mark the Simulation as needing to be rerun before plotting?

Yea, I would vote for that.  

Or just remove the false optimization of ending the simulation early.

I think we agree the current behavior is not ideal :/

JoePfeiffer:
> > (1) Yes, I said "a" simulation, not "the" simulation. It is possible to have multiple Simulations
> > associated with a single FlightConfiguration;
> 
> Wait... what? I thought that was the definition of a FlightConfiguration -- that it's a unique configuration of the rocket?
> 
> What changes then, are the multiple simulations at different flight environments? -- Launch site, weather, etc?

So far as I've found at this point, yes -- the multiple simulations for a single flight configuration are just the different simulation options, like wind, launch rod angle, etc.  I wonder how many designs have ever actually had more than one simulation for a given configuration -- there's absolutely nothing on the "Flight simulations" to indicate what the differences between simulations is, though the user can make a note in the simulation name.  It seems like a pretty useless feature to me...
> 
> As far as resolutions --
> (a) I'd vote to have simulation results on the simulation page, where the correspondence is clear ... but I don't expect that to be too popular.

I think that would have been the best idea long ago, but I expect users have gotten used to having that flight data in the design pane.  I don't think we can get rid of it now.
 
> (b) Second easiest would be for the front-page simulations to always be at 0 launch angle, zero wind, sea level, etc for that Configuration... and then a user must specifically choose and run simulations at "non-standard" conditions.

That's pretty much what we have now -- but as the recent TRF thread on the topic demonstrates, and how long it took me to figure out what was going on, it isn't what users expect.  Once they've set the options for a simulation, they expect the background simulation in the design pane to use those options.

> (c) Maybe flight configurations should link 1-1 with simulations? You might have to reconfigure an engine or a parachute delay, but it would make things simpler.

That would be the simplest solution, both from the standpoint of implementing it (at first glance, all it would require is getting rid of the "New simulation" button in the "Flight simulations" tab) and from a UI standpoint.  If a consensus develops that it's acceptable, it would be my first choice.  Note, incidentally, that the examples you give of things that might need to be reconfigured already require a new configuration, not a new simulation.

> (d) Adding a simulation drop-down _in addition_ to flight configuration drop down on the front page... is doable, but I wouldn't vote for it. That seems like it's just adding more complexity.

As I've thought about it, I think I see a way clear to linking Simulations instead of Flight configurations cleanly.  For the 99.99% of designs that only have one simulation per flight configuration, just display the flight configuration.  For the others, have the flight configuration followed by the simulation name in parentheses -- so if a rocket had two configurations, and one of the configurations had two simulations, the dropdown would look like

    [J1000-7;J1000-0]
    [J381-8;J150-0] (Simulation 1)
    [J381-8;J150-0] (Simulation 3)

(deliberately using simulation names that aren't consecutive, since they may well not be)

JoePfeiffer:
> Or just remove the false optimization of ending the simulation early.

This is the way I'll go forward on this.

teyrana:
> Note, incidentally, that the examples you give of things that might need to be reconfigured already require a new configuration, not a new simulation.

I probably explained myself poorly, there -- I mean that if i want to launch simulate a F150 at two different launch sites, with the current code, that's one configuration, but 2 simulations. . . But with the change I suggested, it would be 2 configurations and 2 simulations.

Anyway, doubling up the configuration + simulation like that is slick -- and probably what people are expecting.  I expect that'll work okay.

JoePfeiffer:
> Anyway, doubling up the configuration + simulation like that is slick -- and probably what people are expecting. I expect that'll work okay.

Actually... this discussion has given me an idea (for a later release):  save named sets of simulation options, so users can have configurations for the various launch fields they use.  Then they can hook up a flight configuration to a set of options, giving them an easy way to simulate a given configuration  at a given launch field (or whatever).

More immediately, I'll proceed with the "doubling up" idea.

JoePfeiffer:
Oops -- something I should have recognized but didn't is that simulation options aren't stored between runs of OR (see #803), so being able to look up simulation/option pairs for the design window is of very limited usefulness until #673 (save named simulation options) is added which won't be on this release.  So let's just go with this PR as the fix for #672.

PR #682, opened 2020-06-12
[fixes #681] Fires an change event from InnerTube `setMotorMount(...)`

### LINKED ISSUE #681: Selecting Inner Tube as the only motor mount doesn't enable "new configuration"
Only reproducible during same run when adding other components:

1. Add BodyTube
2. Add InnerTube
3. Select InnerTube as a motor mount
4. go to MotorSelection Tab, and try to select "New Configuration"
5. It will be disabled.

teyrana:
basically, if an innertube is enabled or disabled, fire an event, so UI components can update.


wolsen:
LGTM, confirmed fixed.

PR #684, opened 2020-06-21
[fixes #502] Refactors Bounds Calculation

### LINKED ISSUE #502: Side view horizontal centering problems
No ORK file is attached to this one, just follow along to reproduce.  Screenshots show 15.03 on left, new alpha 10 on right.

Start with a new design.  Add a default body tube:
centering1

So far so good.  Now add to the body tube a default freeform fin set.  In the new build, set the number of fins to 3.  Things are starting to go awry:
centering2

Now edit the shape of the fins (do the same change to each).  Between the first and second points, add a new point at (-20, 1).  In the new build, having all that extra fin area way to the left somehow pulls the side view even more to the left,  Old build handles it properly.
centering3

As an aside, neither the old nor new build knows how to zoom this properly in the 3D view:
centering4

Back to side view.  For the final step, delete the fin set, leaving only the body tube.  15.03 returns to the original view, while alpha 10 holds the same position as before:
centering5

teyrana:
1. Simplifies code in FlightConfiguration
2. Adds Interface to indicate implementation of `getInstanceBoundingBox` function
3. Adjusts test to ensure that refactoring still behaves as desired
4. Adds a fix to the Intersection testing code in `FreeformFinSet.setPoint`

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/AxialStage.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/AxialStage.java
@@ -54,25 +54,7 @@ public class AxialStage extends ComponentAssembly implements FlightConfigurableC
 	public void reset( final FlightConfigurationId fcid){
 		separations.reset(fcid);
 	}
-	
-	/**
-	 * {@inheritDoc}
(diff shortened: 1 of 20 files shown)


teyrana:
merged with PR #686

Issue #472, opened 2018-11-04
Zoom problems in fin set shape editor


neilweinstock:
Showing the component the fin attached to is a nice addition (and necessary for transitions and nose cones), but it has brought a couple of related problems regarding zooming.  I'm reporting this as one issue, if you'd prefer to split it into two that's fine.

This issue is *related* in some way to issue #471.

Basically, the editor seems really fixated on the front of the parent component as a Very Important Thing, when in fact it is not particularly relevant to editing a fin shape.  The editor should focus on the extents of the fin outline itself when doing zoom and positioning calculations.

Here are two examples:

1) As far as the fin editor is concerned, the front of the parent component is the farthest point forward it is interested in.  If a point on a fin extends significantly in front of the front of the parent component, there's no way to ever zoom or scroll the editor to get it in view, e.g.:
![image](https://user-images.githubusercontent.com/243872/47959278-34f03700-dfb6-11e8-81a2-829f5676baf0.png)

2) When zooming in, the editor does not try to maintain the center of the editor area, or keep the fin visible.  Instead it seems to be locked on the front of the viewable area, which initially is the front of the parent.  For example, if I start here:
![image](https://user-images.githubusercontent.com/243872/47959298-9a442800-dfb6-11e8-80fd-d9f4b8285879.png)
and then zoom in once, I get this:
![image](https://user-images.githubusercontent.com/243872/47959306-b0ea7f00-dfb6-11e8-9c6d-fbcc58b9aceb.png)
The scrollbar is pinned to the left after the zoom, which is the least useful result.  Aggregating across all rocket designs, I think it would be fair to say that the majority of fins are located towards the rear of the rocket.  So this means that almost every time you try to zoom in, you need to zoom and then scroll to the right.  It's annoying.

3) The fin editor always opens up zoomed out to the parent component.  Often, the means the the fin itself starts out very small, particularly if it's mounted to a long BT.  For example:
![image](https://user-images.githubusercontent.com/243872/47959323-1d657e00-dfb7-11e8-9a51-1231e5fcc9da.png)
OK, so naturally I want to zoom in before editing it.  And then issue 2 crops up; every time I zoom I must also then scroll to the right.

I'm not sure that opening the editor zoomed to the parent component is the best choice.  I think I'd rather have it zoomed to fit the fin itself.  I can always keep the main window open to see the fin in the larger context of the rocket, or zoom out in the editor.

[...]

PR #686, opened 2020-06-21
[fixes #502][Fixes #476] Fixes auto-scrolling of FinPointFigure bounds.

### LINKED ISSUE #476: FreeformFinConfig: On initial open, auto-scroll should center to fin bounds.
> The fin editor always opens up zoomed out to the parent component.  
1. the zoom bounds (size)  don't make sense. 
    - [ ] scroll bars not present when they should be  (canvas size is too small?)
    - [ ] bounds are too small for the fin itself.  (should be mount + fins)
2. The zoom bounds (origin) seems to start at the beginning of the fin (which is probably the desired behavior)

### LINKED ISSUE #502: Side view horizontal centering problems

teyrana:
1. Simplifies code in FlightConfiguration
2. Adds Interface to indicate implementation of getInstanceBoundingBox function
3. Adjusts test to ensure that refactoring still behaves as desired
4. Adds a fix to the Intersection testing code in FreeformFinSet.setPoint

@neilweinstock
This fixes issue #476, as reported.

### UI Changes:

Here's the new zoom results:
results

### Jar for testing:
https://drive.google.com/file/d/1rrOCsJG-2JZ9C_G5fI8uY3GvHrVkDW7R/view?usp=sharing

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
@@ -8,7 +8,7 @@ import org.slf4j.LoggerFactory;
 import net.sf.openrocket.l10n.Translator;
 import net.sf.openrocket.rocketcomponent.position.AxialMethod;
 import net.sf.openrocket.startup.Application;
-import net.sf.openrocket.util.ArrayList;
+import net.sf.openrocket.util.BoundingBox;
 import net.sf.openrocket.util.Coordinate;
 import net.sf.openrocket.util.MathUtil;
 import net.sf.openrocket.util.StateChangeListener;
@@ -83,7 +83,16 @@ public class Rocket extends ComponentAssembly {
 		configSet = new FlightConfigurableParameterSet<>( defaultConfig );
 		this.selectedConfiguration = defaultConfig;
 	}
-	
+
+	/**
+	 * Return a bounding box enveloping the rocket.  By definition, the bounding box is a convex hull.
+	 *
+	 * Note: this function gets the bounding box for the entire rocket.
+	 *
+	 * @return    Return a bounding box enveloping the rocket
+	 */
+	public BoundingBox getBoundingBox (){ return selectedConfiguration.getBoundingBox(); }
+
 	public String getDesigner() {
 		checkState();
 		return designer;
--- a/.idea/runConfigurations/Openrocket_UI_Jar.xml
+++ b/.idea/runConfigurations/Openrocket_UI_Jar.xml
@@ -1,10 +1,12 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Openrocket UI Jar" type="JarApplication">
-    <option name="JAR_PATH" value="build/jar/OpenRocket.jar" />
+    <option name="JAR_PATH" value="$PROJECT_DIR$/build/jar/OpenRocket.jar" />
     <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="ALTERNATIVE_JRE_PATH" value="11" />
     <method v="2">
-      <option name="BuildArtifacts" enabled="true" />
+      <option name="BuildArtifacts" enabled="true">
+        <artifact name="openrocket:jar" />
+      </option>
     </method>
   </configuration>
 </component>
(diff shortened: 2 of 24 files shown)


neilweinstock:
Looks great int the screenshot, can't wait to try it out.

teyrana:
@neilweinstock 
You saw the jar link above, right?

neilweinstock:
No I did not!  Sorry about that, testing now.

neilweinstock:
OK, just tried it.

For me, this is only a partial fix.

The good: it fixes the problem where the entire fin was not on-screen at the start.

The bad: it still seems determined to show the complete parent object in the view.  This is not IMHO useful behavior.  Example:
image
This forces me to do a bunch of zooming and scrolling before I can reasonably start editing.  I want the initial zoom to be bounded by the fin, and I honestly don't care about the bounds of the parent object.  E.g.:
image

That would enable me to start editing immediately.  Note that 15.03 works this way, although it is a bit buggy (sometimes doesn't show the entire fin, but it's usually close).

I mentioned this in the original issue (#472) but it might have gotten a little lost among the other things discussed there.

core/src/net/sf/openrocket/rocketcomponent/Rocket.java
```diff
@@ -2,17 +2,14 @@
...
 import org.slf4j.LoggerFactory;
 
 import net.sf.openrocket.l10n.Translator;
 import net.sf.openrocket.rocketcomponent.position.AxialMethod;
 import net.sf.openrocket.startup.Application;
-import net.sf.openrocket.util.ArrayList;
-import net.sf.openrocket.util.Coordinate;
-import net.sf.openrocket.util.MathUtil;
-import net.sf.openrocket.util.StateChangeListener;
-import net.sf.openrocket.util.UniqueID;
```
hola-holo, line 13: our static analysis engine detects that your pull request includes a change from an explicit import to a wildcard import. According to the best practice (http://www.javadude.com/articles/importondemandisevil.html), import star (import on demand) is a bad practice, wildcard import can sometimes cause name conflicts and ambiguities.

teyrana:
@neilweinstock 
Yea, I see what you mean.  I'll circle back and adjust the auto-zoom bounds.

neilweinstock:
Thanks!

teyrana:
Okay, that was a lot more complicated than expected.  It turns out the figures didn't really understand auto-zooming to anything except their full bounds.  So this was more of a new feature than a fix, exactly.  But it should do what you want, now.

neilweinstock:
Daniel,
Sorry for my slowness getting back on this one.

It's closer but not quite there yet.  It seems as if the zoom level is being set correctly, but not the scroll position.

Consider the following rocket (ORK also attached):
image

Starting with the main (middle) fins: opening up the editor looks good:
image

Now one of the front fins:
image
It has set the scroll position to the back of the parent body tube, and so the fin is off the left side of the window.  If I scroll it into view, you can see the that zoom is perfect:
image

I'm not sure why the fin is jammed into the upper half of the edit window but that's minor.

Now let's try one of the rear fins that's mounted to the tail cone:
image

This time the vertical scroll is off.  If I adjust it, the whole fine does (barely) fit in the window:
image

Here's the ORK file.
[blackfish_golden_sunset.ork.zip](https://github.com/openrocket/openrocket/files/4878978/blackfish_golden_sunset.ork.zip)

- Neil

teyrana:
@neilweinstock 
[Okay, third time's the charm? ](https://drive.google.com/file/d/1edHNbPGd4wP_O7x_U29O8zG751md8xUv/view?usp=sharing)

I've made a few changes, which afaict fix the auto-zooming issues.

(code is not yet ready for review)

neilweinstock:
Looks good!

PR #688, opened 2020-06-29
fixes Fin MOI calculations 

### LINKED ISSUE #679: Fin Moment of Inertia calcs.
On line 712 of the finset file, the rotational MOI is calculated using a rectangular approximation for the fin, and the parallel axis theorem is included, as well as a multiplication by the number of fins. The mass is excluded, and then later included, in line 324 of the massCalculation file. The mass is referenced from the compCM.weight, I believe that when the compCM.weight is referenced, it uses the mass of all the fins in the set, put together. Since the MOI calculation has already been done with the 3 fins considered, I believe that value should only be multiplied by the mass for one fin. 

I took some steps to check this myself to be sure,

I rebuilt the identical rocket in solidworks, as well as calculated the MOI for the rocket by hand. when the fins are excluded, my hand calculations, solidworks, and the MOI that OpenRocket gives all agree. Then I added 1 fin, and the once again, still all agreed. When I have 3 fins, the rotational inertia that is contributed by the fins in openrocket is about 3 times what solidworks gives, and what my hand calculations give. My last step to check this, was I clicked 'split fins', creating individual instances of each fin as opposed to a set. When the split fin option is selected, the rotation moment of inertia agrees with solidworks, and is different than what I get when I have all the fins together as one component.

JoePfeiffer:
Fixes #679 by removing extra multiplication by number of fins, updates unit tests to match

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -642,14 +642,14 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
      *
 	 * The process is the following:
 	 * 
-	 * 1. Approximate a fin with a rectangular fin
+	 * 1. Approximate a fin with a rectangular thin plate
 	 * 
 	 * 2. The unitary moment of inertia of one fin is taken as the average
 	 *    of the unitary moments of inertia through its center perpendicular
 	 *    to the plane (Izz/M), and through its center parallel to the plane (Iyy/M)
 	 *    
 	 * 3. If there are multiple fins, the inertia is shifted to the center of the
-	 *    FinSet.
+	 *    FinSet using the Parallel Axis Theorem
 	 */
 	@Override
 	public double getLongitudinalUnitInertia() {
@@ -684,10 +684,11 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 		if (finCount == 1)
 			return inertia;
 
-		// Move axis to center of FinSet.  We need to apply parallel axis theorem
+		// Move axis to center of FinSet.  We need to apply the Parallel Axis Theorem
 		// to Izz, but not to Iyy (as the displacement as we move to the new axis
-		// is along Y).  Since our moment of inertial is the average of Iyy and Izz,
-		// this is accomplished by just weighting the transformation from the theorem by 1/2
+		// is along Y).  Since our moment of inertia is the average of Iyy and Izz,
+		// this is accomplished by just weighting the transformation from the theorem
+		// by 1/2
 		return inertia + MathUtil.pow2(MathUtil.safeSqrt(h2) / 2 + getBodyRadius()) / 2;
 	}
 	
@@ -696,9 +697,10 @@ public abstract class FinSet extends ExternalComponent implements RingInstanceab
 	 * Return an approximation of the rotational unitary inertia of the fin set.
 	 * The process is the following:
 	 * 
-	 * 1. Approximate the fin with a rectangular fin and calculate the
-	 *    unitary rotational inertia (Ixx/M) of the rectangular approximation,
-	 *    about the center of the approximated fin.
+	 * 1. Approximate the fin with a rectangular thin plate
+	 *
+	 * 2. calculate the unitary rotational inertia (Ixx/M) of the
+	 *    rectangular approximation, about the center of the approximated fin.
 	 *    
 	 * 2. If there are multiple fins, shift the inertia axis to the center
 	 *    of the Finset.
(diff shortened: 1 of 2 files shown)


core/src/net/sf/openrocket/rocketcomponent/FinSet.java
```diff
@@ -707,9 +707,14 @@ public double getRotationalUnitInertia() {
 			h = MathUtil.safeSqrt(h * singlePlanformArea/ w);
 		}
 
+		final double inertia = h * h / 12;
```
teyrana, line 710: nit:  in an entire function about calculating inertia,  naming a variable just "inertia" is a bit vague.

Recommend better naming or at least a comment about which inertia this is.
(afaict, it's an approximation for a "thin rod about it's center": 
https://d2vlcm61l7u1fs.cloudfront.net/media%2Fe27%2Fe2782b33-c9c3-4e4a-aac3-1e245d5167a1%2FphpRvs8d4.png )

teyrana: But, to be clear, pulling this out into a separate variable is probably a good thing, and adds to readability  (in addition to correct code ;)

JoePfeiffer:
Yeah, let me improve those comments.  I hadn't looked at the code in a week and I had to re-derive the equations!  Note in that png file the "thin rod about center" and "thin rod about end" cases are actually the same as the "plane or slab" cases -- in the "thin rod" cases they refer to the length of the rod as L, while in the "plane" cases they refer to the same dimension as "a" (so "a" is not area as one would expect) and "b" is irrelevant except as a component of M.

JoePfeiffer:
I will confess to being a little unclear on the physical meaning of a "longitudinal moment of inertia" that is the average (arithmetic mean) of Iyy and Izz. However, it seems clear that when moving the axis to the center of the FinSet, only Izz is adjusted (if instead of this "longitudinal MOI" we had functions returning Iyy and Izz, only Izz would be modified).  Therefore, we should only be adjusting the "longitudinal MOI" by the weight of Iyy, or 0.5

teyrana:
> ... thin rod about center" and "thin rod about end" cases are actually the same as the ... 

Yea, that's the part I didn't follow... I'd have expected that the fin was approximated as a flat plate, about its edge.  
if you're going to add comments, the useful part would be to explain what the approximation is, and why.

>I will confess to being a little unclear on the physical meaning of a "longitudinal moment of inertia" that is the average (arithmetic mean) of Iyy and Izz. 

Yea, I think it's a bit of a kludge.   It shouldn't be an _average_ though.  It should be the max value (i.e. calculate Iyy _and_ Izz independently, the use the greater)  Then "longitudinal stability should obey the stability equations as normal.

I think it's currently a grey area, and I expect the current implementation to have quite a few  faults.   It works well enough for 3FNC rockets, but one of the issues is if you have 2-count pods, or boosters, Iyy ends up being MUCH different than Izz, so we may need to expand it to a more flexible scheme, anyway.   (I_long_max and I_long_min, for sake of argument)

JoePfeiffer:
> > ... thin rod about center" and "thin rod about end" cases are actually the same as the ...
> 
> Yea, that's the part I didn't follow... I'd have expected that the fin was approximated as a flat plate, about its edge.
> if you're going to add comments, the useful part would be to explain what the approximation is, and why.

What's going on with those formulas is that the extent along the axis doesn't end up being a factor in the unitary MOI calculation once we've divided by the mass.  So the thin plate formula is the same as the thin rod formula (it's unfortunate that png file didn't maintain consistent variables).  Latest commit tries to make this a little more explicit.

> 
> > I will confess to being a little unclear on the physical meaning of a "longitudinal moment of inertia" that is the average (arithmetic mean) of Iyy and Izz.
> 
> Yea, I think it's a bit of a kludge. It shouldn't be an _average_ though. It should be the max value (i.e. calculate Iyy _and_ Izz independently, the use the greater) Then "longitudinal stability should obey the stability equations as normal.
> 
> I think it's currently a grey area, and I expect the current implementation to have quite a few faults. It works well enough for 3FNC rockets, but one of the issues is if you have 2-count pods, or boosters, Iyy ends up being MUCH different than Izz, so we may need to expand it to a more flexible scheme, anyway. (I_long_max and I_long_min, for sake of argument)

I'd have to go back and review some simulation code, but my recollection from when I last went through it is that it only considers pitch and rotation -- so consistency with that code would imply ignoring Izz completely.  Doing it "right" would be to have separate Iyy and Izz calculations, and actually track rotations in all three axes.  That's a can of worms I don't think we want to open right now.

For this release, I think our best options are to either leave the application of the parallel axis code as is (so don't weight the adjustment), or weight it by 1/2 (as I've done in my PR).

teyrana:
@JoePfeiffer 
Okay, given that we both seem ... I'm not sure "satisfied" is the word, but given that this patch is good enough for now, are we ready to merge? 

> That's a can of worms I don't think we want to open right now.
Yea, that seems to be the consensus.   -.-

JoePfeiffer:
I'm satisfied with the patch; my dissatisfaction is with what it's actually calculating, which is a level removed.

PR #703, opened 2020-07-19
[fixes #385] Restores ability to select components in the 3d views

### LINKED ISSUE #385: Cannot click-select components in 3d views
On this build I cannot click on components in the 3D view to select them.  This has never worked on the Mac (at least not since I've been using it), but it has always worked on Windows.  Now clicking on the 3D view in Windows has no effect.

teyrana:
note: @bkuker  is the original author of the rendering code I modified for this.


bkuker:
>
> * This function is a bit.... unusual. Instead of computing an inverse
> transform from the UI window into design-space,
> * this renders each component with a unique identifiable color ... to a
> dummy, throwaway canvas:
> *
> * Then, we read the pixel (RGB) color value at a point on the canvas, and
> use that color to identify the component
>

Heh, color picking is I think one of the most oldschool picking methods.
Even Duck Hunt on the NES used it! I may be showing the age of my OpenGL
knowledge!
I am not sure, but you might have forgotten to turn multisample back on
after picking. Sorry for my old disorganized code... Not sorry for using
immediate mode ;)

-Bill

On Sun, Jul 19, 2020 at 4:07 PM Daniel Williams <notifications@github.com>
wrote:
>
> ________________________________
>
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/openrocket/openrocket/pull/703
>
> Commit Summary
>
> [fix] enabled selection of 3d components ... in some cases
> [fix] enabled select of 3d components in the general case
> [debug] removed excess debug output in FinSet
> [debug] removed excess/redundant debug output in GUIUtil
> [refactor] RocketFigure3D converted to use 'util.BoundingBox'
> [rm] remove duplicate jogl jar
> [refactor] changed function names: 'reduce360' to 'reduce2PI' because the
numbers are in radians

teyrana:
> Heh, color picking is I think one of the most oldschool picking methods.
Even Duck Hunt on the NES used it! I may be showing the age of my OpenGL
knowledge!

Hey, no worries.  Once I understood it, it totally makes sense.   It's probably quicker and more elegant than the alternative.   I was just surprised, is all.


> I am not sure, but you might have forgotten to turn multisample back on
after picking. 

err, hmm.  That's a good point.  What _does_ multisample_ do, anyhow?   Should it be enabled for the picking method, or for normal rendering?

teyrana:
Ah, here.  The lines disabling those two are at: 
https://github.com/openrocket/openrocket/blob/0f2edc002499fd18413684fc38ee18ff75fe3410/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java#L297-L298

and the lines ENabling them are at:
https://github.com/openrocket/openrocket/blob/0f2edc002499fd18413684fc38ee18ff75fe3410/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java#L321-322

.... basically the disable/enable calls wrap the 'pick' call.    Is that the right ordering?

PR #709, opened 2020-07-26
[fixes #701] Re-enables Rocket Optimizer

### LINKED ISSUE #701: Add back the "Rocket Optimization" feature
"Rocket Optimization" was dropped out of the unstable branch at some time in the past.  Should be added back in before next release.

teyrana:
https://drive.google.com/file/d/1SnGZY9PUC86x4rIgK-Vx4S7oS9WtD9dv/view?usp=sharing

### DIFF
--- a/core/resources/build.properties
+++ b/core/resources/build.properties
@@ -1,6 +1,6 @@
 
 # The OpenRocket build version
-build.version=20-08-alpha-14
+build.version=20-08-alpha-15
 
 
 # The source of the package.  When building a package for a specific
--- a/core/src/net/sf/openrocket/document/Simulation.java
+++ b/core/src/net/sf/openrocket/document/Simulation.java
@@ -497,6 +497,7 @@ public class Simulation implements ChangeSource, Cloneable {
 			Simulation copy = new Simulation(newRocket);
 			
 			copy.name = this.name;
+			copy.configId = this.configId;
 			copy.options.copyFrom(this.options);
 			copy.simulatedConfigurationDescription = this.simulatedConfigurationDescription;
 			for (SimulationExtension c : this.simulationExtensions) {
--- a/core/src/net/sf/openrocket/document/Simulation.java
+++ b/core/src/net/sf/openrocket/document/Simulation.java
@@ -164,7 +164,12 @@ public class Simulation implements ChangeSource, Cloneable {
 		}
 		
 	}
-	
+
+	public FlightConfiguration getActiveConfiguration() {
+		mutex.verify();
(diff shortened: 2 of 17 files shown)


neilweinstock:
First attempt to use this crashed; see attached.
Steps to reproduce:
1) Open "Simple Model Rocket" example design
2) Open Rocket Optimization
3) Select body tube length as parameter to optimize
4) On "Optimize Simulation", Select "Basic Simulation 5 (C6-7)"
5) Click "Start Optimization".
[optimization_crash.txt](https://github.com/openrocket/openrocket/files/4978927/optimization_crash.txt)

Also I note that the side view window has a familiar centering problem.

This build:
image

(and in this build, Rocket Optimization always opens for me at full screen width, vs. 15.03 which opens at a more "natural" size)

15.03, after Optimization window has been stretched to full screen width:
image

neilweinstock:
OK, it appears that the crashing was Java 14-related.  No crashes on Java 11.  The default window size and horizontal centering issues are still there.  I am also still analyzing the results I get from the optimization, there may be an issue there but still working on it.

neilweinstock:
I don't think it's working correctly in some respects.  I haven't tested it exhaustively, but here are some results that should hopefully shed some light.

Start with the Simple Rocket example.

Optimize for body tube length, choose "Basic Simulation (C6-5)", leave the rest as defaults, and go.  Result suggests body tube length of 5.906".  That might be correct.  Then click "Plot Path" and you get this:
image
That "Optimization Result" plot can't be correct.

Close the plot and Click "Reset".

Now add Nose Cone Length as another parameter to optimize.  Click "Start optimization".  Now the optimized result has both lengths at their original values.  That can't be right; if the nose cone optimal length is unchanged then the optimal body length should be the same as the previous run.  Here's the plot:
image

teyrana:
@neilweinstock 
@hcraigmiller

do you guys use all the extra simulations this dialog invents? 

I'm thinking of reducing the available simulations to the currently defined simulations.   What you you think? 
(i.e. whatever simulations are available in the "Flight Simulations" tab.")

hcraigmiller:
I have never used the rocket optimization basic simulations, I only optimize the specific configurations that I define on the motors and configuration tab, if I understand your question correctly.

neilweinstock:
I've never really used it much, didn't even realize it was inventing
simulations like that.  I admit that's pretty weird.

Using only the currently defined simulations certainly seems reasonable to
*me*, dunno what folks would think about it who actually use it.

We could try to gather some opinions on TRF if you like.

PR #717, opened 2020-07-27
Center rocket in side-view 2d renderings

wolsen:
Center the rocket within the side-view 2d renderings. This is similar to
how the side-view is rendered in 15.03 version.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java
@@ -1,7 +1,14 @@
 package net.sf.openrocket.gui.scalefigure;
 
 
-import java.awt.*;
+import java.awt.BasicStroke;
+import java.awt.Color;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
+import java.awt.Point;
+import java.awt.Rectangle;
+import java.awt.RenderingHints;
+import java.awt.Shape;
 import java.awt.geom.AffineTransform;
 import java.awt.geom.Ellipse2D;
 import java.awt.geom.NoninvertibleTransformException;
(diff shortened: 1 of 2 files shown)


teyrana:
I've not seen this issue, and I was fixing bugs in this code last week.   Are screenshots of the bug and/or the fix behavior available anywhere?

wolsen:
Ah yeah, I'll get some screenshots tomorrow. I looked at this based off some comments from @neilweinstock in slack. If you look at 15.03 the rockets are centered in the side profile and in unstable they are left aligned.

I can also raise a proper bug to track it.

neilweinstock:
There was #502 but that was closed recently.

teyrana:
Ahh, so is that last screenshot the issue this PR is addressing?  Is that what were talking about?

neilweinstock:
@teyrana PR #686 claimed to fix this... should we not still be seeing it?

teyrana:
@neilweinstock 
Issue #502 had multiple issues inside it.   PR #686 fixed the first issue -- fins going out of bounds -- but (apparently) not address the later issue.

neilweinstock:
Actually #502 was only about the above issue.  Should we reopen it just so we can now close it again? :)

wolsen:
> Ahh, so is that last screenshot the issue this PR is addressing? Is that what were talking about?

Yes, that's what this issue is addressing. Here's some screen caps of before and after this change.

**Before Change**
or-nose-cone-unstable-left-justified

**After change**
or-nose-cone-unstable+fix

wolsen:
Opened #722 for this specific one as a spinoff of #502

teyrana:
@neilweinstock 
Yea, it's one of those complicated things where what looks like the same issue can be caused my multiple issues on the back end.  
It's imperfect, and basically something fell through the cracks.  I think we're handling it the right way, though :)

@wolsen
I did some spot checks -- looks good!

neilweinstock:
Does this also fix the centering of the rocket in the side view inside the Rocket Optimization dialog?

PR #730, opened 2020-08-01
[fixes #727][gui][cosmetic] Adjusted layout in RocketComponentConfig layout

### LINKED ISSUE #727: UI: Object configuration dialogs have gotten larger
I just noticed that the object configuration dialogs have gotten somewhat larger in the new builds due to expanded spacing around the header and footer:
image

This is not a major issue but to me it is better the old way.  The extra spacing does not (IMHO) enhance usability, it just makes the dialogs take up more space on the screen.

teyrana:
The new layout is should improve legibility and reduce dead space.
This is a purely cosmetic change.

## Before:
unstable v0

## After:
v6


neilweinstock:
That looks roughly back to the way it was, excellent.  Thanks!

PR #731, opened 2020-08-01
[Fixes #726] Fixes multiple issues when changing Fin-Offset-Methods

### LINKED ISSUE #726: Changing "Position Relative To" setting creates unexpected results
This is a re-issue of #541; which was fixed and closed but has reappeared.

When changing the "Position relative to" selector in a component, the expected behavior (from 15.03 and before) is that the value in the "plus" field will adjust to leave the component in the same actual position.  What happens now is that the value in the "plus" field stays the same, and the object moves.

In the following screenshots, 15.03 is on the left, recent unstable build on the right.

Starting with a new design, add a body tube (leave settings as default), then add a fin set to the body tube:
image

Now change "Position relative to" to "Top of the parent component":
image

Expected behavior is on the left.

teyrana:
Addresses #726

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/RocketComponent.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/RocketComponent.java
@@ -934,11 +934,14 @@ public abstract class RocketComponent implements ChangeSource, Cloneable, Iterab
 			// no change.
 			return;
 		}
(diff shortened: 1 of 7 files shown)


teyrana:
(updated to fix the unit tests)

PR #737, opened 2020-08-08
Fix/FreeformFinConfig - editing fin points

teyrana:
## Note: this addresses #472, partially, but not completely.

Contains the fixes: 
1. fixes the auto-zoom for fin points
2. if a user drags a fin point out-of-bounds, it will now scroll with the (new) point.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FreeformFinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FreeformFinSet.java
@@ -24,7 +24,10 @@ public class FreeformFinSet extends FinSet {
 	
 	private static final double SNAP_SMALLER_THAN = 5e-3;
 	private static final double IGNORE_SMALLER_THAN = 1e-12;
-	
+
+	// attempts to set a fin value any larger than this will be snapped to this max value
+	public static final double SNAP_LARGER_THAN = 2.5; // in meters
+
 	public FreeformFinSet() {
 		points.add(Coordinate.ZERO);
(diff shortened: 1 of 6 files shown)


neilweinstock:
@teryana is there a newer jar you want me to test?  Looks like you added additional changes since the last one.

teyrana:
@neilweinstock 
Nope. I was still working on it.

But _now_ it's ready.  
[Here ya go](https://drive.google.com/file/d/1GTpXg1Qfq_1S1Vwjvl7zhgmqfeSPLFP6/view?usp=sharing)

In addition to cleaning up the zooming logic, I also cleaned up the scrolling logic, so that it tries to keep the point you're dragging within frame.

neilweinstock:
OK thanks, I requested access to the file on Google drive.

teyrana:
Okay, fixed the permissions.  You should have acess, now.

neilweinstock:
Mostly good now.  One issue:  If I drag a point way off an edge and keep going, eventually it'll get a stack overflow, e.g.:
image.

teyrana:
Yea, it's basically a positive feedback loop that runs off to infinity.  what do you think is a safe max size?   1m?  10m?

neilweinstock:
Well, if I recall correctly, Sampo once constructed a model for the full-scale Saturn V in OpenRocket, so I say a safe max size is "pretty big". :)

1m definitely too small, 10m probably covers out to 99.99%, maybe 100%.  I think it's a reasonable number.

PR #769, opened 2020-09-01
[fixes #755] Correctly calculate cg.x override

### LINKED ISSUE #755: Broken CG calculation with 3-stage model
On the attached model, CG calculation breaks when third stage is added.
[Talos Terrier Recruit.ork.zip](https://github.com/openrocket/openrocket/files/5109921/Talos.Terrier.Recruit.ork.zip)

Here it is with the top stage ("Recruit") only.  The CG matches the override set for the stage (15"):
image

Here it is with the next stage (Terrier) included.  CG moves back to a reasonable place:
image

Now we add the bottom stage (Talos).  For some reason, the CG moves *forward*:
image

The problem seems to be caused by the CG override of the "Talos" stage.  If that is disabled, CG moves back to a more reasonable position:
image

teyrana:
Apply CG override distance to be the distance from the beginning of the component.

### DIFF
--- a/core/src/net/sf/openrocket/masscalc/MassCalculation.java
+++ b/core/src/net/sf/openrocket/masscalc/MassCalculation.java
@@ -341,6 +341,17 @@ public class MassCalculation {
 
 			// mass data for *this component only* in the rocket-frame
 			compCM = parentTransform.transform( compCM.add(component.getPosition()) );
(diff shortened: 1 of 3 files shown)


hcraigmiller:
See recent comment to Issue #755 concerning additional CG quirk.

PR #782, opened 2020-09-12
Only consider active stages when calculating drag

### LINKED ISSUE #781: Drag is calculated as if no stage separation occurs
Another one found in the course of tracking down #768.  My branch fix-768a contains a fix for this, but this exposes a new bug:  when only calculating upper stages of a multistage rocket, resulting drag is substantially less than drag under 15.03, and drag coefficient is nuts.

In the case of the rocket in issue #768, immediately following first stage separation total drag in version 15.03 goes from 75.225 to 27.033.  In current unstable, it goes from 75.51 to 73.492.

Following commit 20ab8ef8b15b5bea46dee56815b9b64a78a3ccef the force goes from 75.499 to 22.094.  Meanwhile, drag coefficient goes from 0.62039 to 0.18621.

JoePfeiffer:
This fixes #781, which was discovered while tracking down #477 

Three commits:
the first corrects FlightConfiguration:getActiveInstances() to indeed only return active instances by modifying FlightConfiguration:getContextListAt() to only return active contexts

the second fixes the FlightConfigurationTest unit test by making stage 1 active.

third is a bit of clarification, changing the name of FlightConfiguration:getContextListAt() to getActiveContextListAt, since that's what it's now doing.


JoePfeiffer:
Ah, didn't mean to merge that without comments from others -- I was just trying to merge the two sets of commits together.  If there's a problem, feel free to revert.

PR #789, opened 2020-10-01
Last fix on issue 768

### LINKED ISSUE #768: Strange flight sim behavior with 3-stage design
There's a lot of weirdness here that is hard to explain.  I will try to do it with pictures.  There may be multiple issues here, but at this point I have no idea how to separate them.

Here is the design in question:
[Talos Terrier Recruit.ork.zip](https://github.com/openrocket/openrocket/files/5147969/Talos.Terrier.Recruit.ork.zip)

Here is 15.03 on left, latest unstable on right:
image
**_First, note that the apogee is *way* higher on 15.03._**

Next, note that 15.03 has a sim error: "Simulation values exceeded limits.  Try selecting a shorter time step".  This seems to be a common problem in multi-stage designs in 15.03 that is fixed in unstable.  Still, as I'll show below, the error seems to occur at apogee, so the apogee estimate looks reasonable.

**_Next, note that the unstable has a different sim error: "Recovery device opened while motor is still burning".  I cannot explain this at all; the plots do not show it happening._**  

Let's look at the plots (again, 15.03 on left, unstable on right).  I included lateral velocity just to ensure that everything was pretty stable and flying vertically:
image
The 15.03 plot looks pretty clean, other than that the plot stops at the top stage apogee, when the sim crashes apparently.

The unstable plot, on the other hand is... not so much.  The apogee of each stage is much lower than before.  Also, the top stage hits apogee almost immediately after ignition.  Also the vertical velocity plots are weird.

**_I cannot offer any reasonable hypothesis about what's going on here, other than "it's not working right"._**

Finally: note that the top stage is mostly base-drag-stabilized; either way, it is showing good stability margin and OR is definitely simulating it as if it is stable.  Nonetheless, I don't know if that is somehow contributing to the problem here.  But 15.03 has no problem with it.

Hope someone can sort out this mess! :)

JoePfeiffer:
On multistage rockets, there can be a next symmetric component that is not active.  Check for that.

I think this is the last fix required to close #768.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -4,6 +4,9 @@ import static net.sf.openrocket.util.MathUtil.pow2;
 
 import java.util.*;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import net.sf.openrocket.aerodynamics.barrowman.FinSetCalc;
 import net.sf.openrocket.aerodynamics.barrowman.RocketComponentCalc;
 import net.sf.openrocket.rocketcomponent.ComponentAssembly;
@@ -29,6 +32,7 @@ import net.sf.openrocket.util.Reflection;
  * @author Sampo Niskanen <sampo.niskanen@iki.fi>
  */
 public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
+	private static final Logger log = LoggerFactory.getLogger(BarrowmanCalculator.class);
 	
 	private static final String BARROWMAN_PACKAGE = "net.sf.openrocket.aerodynamics.barrowman";
 	private static final String BARROWMAN_SUFFIX = "Calc";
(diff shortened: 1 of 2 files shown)


core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
```diff
@@ -667,8 +671,10 @@ private double calculateBaseDrag(FlightConfiguration configuration, FlightCondit
 				
 				// if I'm the last component, set my base CD
 				// note:  the iterator *should* serve up the next component.... buuuut ....
-				//        this code has is tested, and there's no compelling reason to change. 
-				if (s.getNextSymmetricComponent() == null) {
+				//        this code has is tested, and there's no compelling reason to change.
+				final SymmetricComponent n = s.getNextSymmetricComponent();
+				if ((n == null) ||
+					!configuration.isStageActive(n.getStageNumber())) {
```
teyrana, line 677: line 676 -> 677 is misleading indentation :/

teyrana (approved):

core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
JoePfeiffer: That's the problem with using tabs for indentation...  just merged a PR (793) to clean it up.

teyrana: Ah, it was a nitpick anyway.  But thanks for cleaning it up ^^

PR #795, opened 2020-10-11
Fixes #754 #755 -- Overriding Center-of-Mass x coordinate, mass works better

### LINKED ISSUE #754: Mac: cannot click-select components in 3D-view
This is essentially a re-issue of #385, but I now believe it is Mac-specific, and it applies to the latest unstable builds.

Clicking on the 3D finished view does nothing, it is simply not possible to select components this way.

teyrana:
fixes more corner cases when performing mass calculations.

### DIFF
--- a/core/src/net/sf/openrocket/masscalc/MassCalculation.java
+++ b/core/src/net/sf/openrocket/masscalc/MassCalculation.java
@@ -344,12 +344,13 @@ public class MassCalculation {
 			// mass data for *this component only* in the rocket-frame
 			compCM = parentTransform.transform( compCM.add(component.getPosition()) );
 			if (component.getOverrideSubcomponents()) {
-				if (component.isMassOverridden()) {
-					double newMass = MathUtil.max(component.getOverrideMass(), MIN_MASS);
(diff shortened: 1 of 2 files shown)


core/src/net/sf/openrocket/masscalc/MassCalculation.java
```diff
@@ -300,10 +300,10 @@ MassCalculation calculateStructure() {
...
 		final double[] allInstanceAngles = component.getInstanceAngles();
 
-		// // vvv DEBUG
-		// if( this.config.isComponentActive(component) ){
-		// 	System.err.println(String.format( "%s>>[%s]....", prefix, component.getName()));
-		// }
+		// vvv DEBUG
+		//if( this.config.isComponentActive(component) ){
+		//	System.err.println(String.format( "%s>>[%s]....", prefix, component.getName()));
+		//}
```
wolsen, line 306: nit, this seems like it could have just been removed

wolsen (approved):
LGTM, nit inline, but the code on the nit is essentially the same so nothing worth holding up

Issue #786, opened 2020-09-27
Rocket mass does not change when motor is installed


hcraigmiller:
If you create a new design, or open a design that does not have a motor installed when saved, when a motor is installed the upper left corner of the display window does not change to add "Mass with motors..." below "Mass with no motors..." But, if you save a design with the motor installed, it does. Also, in the latter, if you uninstall all motors, it keeps both lines of text, both showing the same mass.

teyrana:
- [ ] update window when a motor mount is added
- [ ] update window when a motor mount is removed
- [ ] update window -- motor mass -- on _any_ motor change

hcraigmiller:
Fixed.

PR #798, opened 2020-10-18
[fixes# 786] Updates config when drawing extras on 2D RocketFigure

teyrana:
Fixes Issue #786 
Caused by the FlightConfiguration not updating figure display code.


swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
```diff
@@ -573,6 +573,8 @@ private void updateExtras() {
 		FlightConditions conditions = new FlightConditions(curConfig);
 		warnings.clear();
 
+		extraText.setCurrentConfig(curConfig);
```
teyrana, line 576: This is the line that actually fixes the bug.

The rest is refactoring to make intent more clear.

PR #847, opened 2020-12-13
Fix rocket scaling in 3D views

### LINKED ISSUE #788: Display smaller in current than 15.03
I am editing this issue to reflect the 12/03/2020 merge as a number of issues that appear in the side view drawings have been corrected.

There is a significant difference in the 3D functionality between 15.03 and the current unstable; the current unstable is not recognizing the system display settings in the same way the 15.03 does. Using the sample Apocalypse rocket, you can see the problem.

Sample 01
Sample 02

See also #719. 

Even lowering the screen resolution, the current unstable 3D Finished view is smaller than in 15.03.

Sample 03

In the current unstable, the off-screen rendering also has an effect on size and the placement of the CG and CP markers.

Sample 04

Although not recognizing the system display settings for 3D functions is not otherwise affecting program functionality, when the Beta is release this bug will likely trigger a host of reports.

See also, #790

wolsen:
When setting up the view for 3D figures, the Y span is doubled
incorrectly. This appears to be legacy behavior and the BoundingBox
itself accounts for the full span. Removing the 2x scaling resolves the
rocket scaling appropriately.

Fixes #788

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
@@ -496,7 +496,7 @@ public class RocketFigure3d extends JPanel implements GLEventListener {
 		// Add 10% for space around it.
 		final double dX = (b.span().x * 1.2 / 2.0)
 				/ Math.tan(Math.toRadians(fovX / 2.0));
-		final double dY = (b.span().y * 2.0 * 1.2 / 2.0)
+		final double dY = (b.span().y * 1.2 / 2.0)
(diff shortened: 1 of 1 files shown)


wolsen:
Before and after screenshots:

*Before*
before-change

*After*
after-change

hcraigmiller:
That looks great...

neilweinstock:
Is that an original design?  I love it.  Can I get the file?

wolsen:
> Is that an original design? I love it. Can I get the file?

Its the Apocalypse with decals example rocket actually.

neilweinstock:
Ha, I guess I should look at those one day.

PR #866, opened 2020-12-24
Fix scaling on high dpi monitors

### LINKED ISSUE #854: 3D display not rendering properly with display scaling (Windows 10)
Although the 3D display after the recent merge is much better, it is still different than version 15.03.

Differences Compare

In viewing what was intended, I looked back at the "_after_" screen shot in Issue #847 where the display appeared the same as Version 15.03, but the recent merge is not under Windows 10.

wolsen:
GLJPanel handles display scaling better than the GLCanvas does natively.
Change to use the GLJPanel always when rendering in 3D mode. FBO
(offscreen rendering) will be set on the GLJPanel via the
GLCapabilities.

Also, ensure when creating a Graphics2D object from a GLOverlay to set
the current transform to the default transform for the component. The
default transform of the swing components will take into account the
scaling of the screen that it is rendered on. By setting the transform
of the overlay, we can be sure that it gets the scaling information as
well.

Fixes #854

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
@@ -4,6 +4,7 @@ import java.awt.BorderLayout;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Graphics2D;
+import java.awt.GraphicsConfiguration;
 import java.awt.Point;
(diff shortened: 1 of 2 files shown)


wolsen:
Received positive feedback on patch and tests from both @hcraigmiller and @neilweinstock as well as my own. This has been waiting for long enough I'm merging.

Issue #865, opened 2020-12-24
Fin tabs on tail cone not functioning properly


hcraigmiller:
These are 4 related issues pertaining to fin tabs on tail cones [transitions] that are identified as A through D; these can be split into 4 separate issues if materially unrelated. 

Using this design file: 
[ERROR.Transition Fin Tab.zip](https://github.com/openrocket/openrocket/files/5737616/ERROR.Transition.Fin.Tab.zip)

Open and set side view to 150% before viewing the transition and fin root. 

A.  Select the Lower fin set, Fin tab, and click automatic. Two errors are apparent, the fin tab does not go from centering ring to centering ring, and the tab height extends below the inner tube. See "A" below.

![Tail Cone Fin Tab  01 Automatic Reference](https://user-images.githubusercontent.com/68821492/103046421-10140080-453d-11eb-8caf-b3ca606b568f.png)

B. Set the Tab position relative to the bottom of the parent component, increase the tab length by 1". Instead of increasing in length toward the top of the component, the tab increases in length toward the bottom of the component and the Tab position is incremented to match. Again, the Tab position should remain unchanged and the tab should increase in length toward the top of the component. See "B" above.

C. When the tab is created manually, the Tab height appears to be the shortest distance between the transition and the inner tube, not the longest distance between the transition and the inner tube that should _always_ be used. See "C" below. 

![Tail Cone Fin Tab  02 Manual Reference](https://user-images.githubusercontent.com/68821492/103047045-173c0e00-453f-11eb-8ffd-95a64c13f096.png)

D. SAVE the modified design file. RE-OPEN the design file and view the transition and fin root at 150%. The Fin length and Tab position have changed, numerically and visually. See "D" above.

JoePfeiffer:
Ah.... when I looked at this before, the fin was only on a single component.  In this case, the fin's parent is the transition, but the fin extends both forward and aft of the transition.  Hmmm...

hcraigmiller:
See also #868

hcraigmiller:
Fixed by way of PR #1620.

PR #869, opened 2020-12-25
[fixes #868] Fin tabs correctly adjust their location when the fin tab _length_ is changed.

### LINKED ISSUE #868: Fin tab length changes are referenced only to the top of the parent component
Rather than referencing the change in length from the selected relative to position, the change is always referenced to the top of the parent component. The design file used is a Simple model rocket.

00 Automatic

Although only the trapezoidal fin has been used above, the same issue exists for all fin types, and for both fins on body tubes and fins on transitions.

teyrana:
This fix addresses Issue #868, and may help with #865.

Basically, the update function in fin set wasn't checking the fin-tab properties, so I inserted a call to `validateFinTab()`.   It appears to work, now, according to the description / images in #868.    Thanks to @hcraigmiller for the high-quality bug report 👍 

@JoePfeiffer I got you fam.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/EllipticalFinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/EllipticalFinSet.java
@@ -71,7 +71,7 @@ public class EllipticalFinSet extends FinSet {
 		if (MathUtil.equals(this.length, length))
 			return;
 		this.length = length;
-		validateFinTab();
+		validateFinTabLength();
 		fireComponentChangeEvent(ComponentChangeEvent.BOTH_CHANGE);
 	}
 	
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -274,7 +274,7 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 		}
 		
 		tabHeight = newTabHeight;
-
+		validateFinTabHeight();
 		fireComponentChangeEvent(ComponentChangeEvent.MASS_CHANGE);
 	}
 	
@@ -293,7 +293,13 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 		
 		tabLength = lengthRequest;
 		
+		setTabPosition();
+		
 		fireComponentChangeEvent(ComponentChangeEvent.MASS_CHANGE);
+	}
+	
+	protected void setTabPosition(){
+		this.tabPosition = this.tabOffsetMethod.getAsPosition(tabOffset, tabLength, length);
 	}
 	
 	/** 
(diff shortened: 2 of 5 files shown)


JoePfeiffer:
Can't look at it carefully right now -- could this bring #527 (fin tabs loaded incorrectly) back? Problem there was validating the tabs in the process of reading a .ork file.

teyrana:
@JoePfeiffer 
Ah. Thanks for the heads up -- yes, it restores the #527 issue behavior.

teyrana:
After looking at it in more detail, the problem is that basically we want to validate from the gui, but not while loading from a file....

What do we think of happing a validate=true|false parameter to the `setTab*()` functions?

JoePfeiffer:
I think we explored that option in the discussion of PR #530.  

It isn't immediately obvious to me that the solution here is validation -- if the fin tab is set relative to the parent's aft end, adjusting the length should just move the forward end.  It doesn't seem like revalidating will necessarily make that happen -- won't it let the aft end of the tab move until it fails validation, then move the forward end?

teyrana:
Okay, I think I've massaged it to resolve both issues. 

Re-reading the discussion on #530 I'm not clear on why we dropped all that validation on UI modification.    Do we want all that validation, at all? 

(for instance -- fin tabs that extend in front or in back of their fins)

I mean, if nobody's going to complain about it, I'm fine with that.... but I do remember complaints about some of it.

Maybe the question should be "which validations" do we want the UI to enforce?

Also, I've split the `validateFinTab` function into `validateFinTabPosition` and ...Length and ..Height functions.

teyrana:
My tests are:
1. Unit tests (of course)
2. Opening a simple rocket with 10-fins and 10-inch tabs.   (the default chord is 5in)

JoePfeiffer:
On dropping validation, it's the question of how much we want to ensure users only do things make sense vs. letting them do what they want.  I remember us as thinking that since there's an "automatic" option that tries to do them most reasonable thing, if they want to change from that they probably have a reason.

JoePfeiffer:
Finally back in town...  I created the attached ork file (as usual, filename changed to .txt to get around github), and with this PR I don't see any change in the display when I change tab length or height, or click "calculate automatically".  If I change tab position, that change and any prior changes are visible immediately. It does appear to be adjusting length relative to the specified part (front, back, middle) of fin now, as it should be.

I guess I'm happy either way on validation.

[fintest.txt](https://github.com/openrocket/openrocket/files/5758491/fintest.txt)

teyrana:
Ooooph, it's been a while.   iirc, this worked when submitted, but I haven't tested it in a while.   I'll take a look this weekend.

teyrana:
Okay, I thiiiiiink this works, relative to what I thought the issue was.  But someone else should probably test it.

hcraigmiller:
Can you provide a .jar file for testing?

teyrana:
@hcraigmiller 
https://drive.google.com/file/d/1o-zqxMqAFlhZFsTTxldMWtsdN54cR67S/view?usp=sharing

hcraigmiller:
OpenRocket-pull-869.jar fixes Issues #868 and #925. Anomalies remain with Issue #865, but the cross-over with Issues #868 and #925 is fixed; where Issue #865 is completely separate from this issue, I will post those anomalies after this PR is merged.

It appears that this PR is complete, no anomalies found.

PR #884, opened 2021-01-22
correct linear interpolation of motor CG

### LINKED ISSUE #883: Motor CG calculation wrong
See the attached .ork file (uploaded as a .zip) and simulation plot with CG in yellow. CG should move forward smoothly during motor burn.  The error stems from an error in calculating the motor CG.  PR in a moment.
bad-cg
[3fnc.zip](https://github.com/openrocket/openrocket/files/5857560/3fnc.zip)

JoePfeiffer:
fixes #883


wolsen:
It took me awhile to convince myself this is correct :-)

JoePfeiffer:
It took me a while to realize why the old one wasn't!

PR #893, opened 2021-02-20
Update Bug report dialog

### LINKED ISSUE #860: Confirm/Update "Bug report" transmission to location
When a user makes a "Bug report" within OpenRocket, the location to which the report is being transmitted should be confirmed/updated.

wolsen:
Updates the Bug report dialog to send users to either raise issues
via the github issues URL for OpenRocket or to send an email to the
forum.

This removes code to auto-submit bugs from the bug report dialog. The
service does not appear to be working and github issues cannot be
anonymous which complicates auto-submission. Instead, just direct users
to the right place.

Fix #860


neilbalch:
I don't see a problem with this. Should we merge it now?

PR #894, opened 2021-02-21
Fix transparency display issue

### LINKED ISSUE #878: Transparency does not display correctly
@ChrisMickelson (I think) added support for transparency a while back.  I am fairly certain it worked correctly in the past., which is to say that the alpha value was properly reflected in the object appearance in 3D views.  No longer, at least on my Mac running Mojave: now, any alpha value other than 255 displays exactly the same, e.g.:

Color = 0,0,0, 255
image

Color = 0,0,0, 254
image

Color = 0,0,0, 0
image

wolsen:
Use the provided alpha from the color when it is available when
rendering the model in the 3d finished view. The 3d unfinished view will
continue to override the alpha to show interior components.

Fixes #878

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/RealisticRenderer.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/RealisticRenderer.java
@@ -101,15 +101,23 @@ public class RealisticRenderer extends RocketRenderer {
 		render(gl, geom, Surface.EDGES, app, false, alpha);
 	}
 	
+	protected float[] convertColor(Appearance a, float alpha) {
+		float[] color = new float[4];
+		convertColor(a.getPaint(), color);
+		return color;
+	}
+
 	private void render(GL2 gl, Geometry g, Surface which, Appearance a, boolean decals, float alpha) {
 		final Decal t = a.getTexture();
 		final Texture tex = textures.getTexture(t);
 		
 		gl.glLightModeli(GL2.GL_LIGHT_MODEL_COLOR_CONTROL, GL2.GL_SEPARATE_SPECULAR_COLOR);
 		
+		float[] convertedColor = this.convertColor(a, alpha);
+		for (int i=0; i < convertedColor.length; i++) {
+			color[i] = convertedColor[i];
+		}
 		
-		convertColor(a.getPaint(), color);
-		color[3] = alpha;//re-set to "alpha" so that Unfinished renderer will show interior parts.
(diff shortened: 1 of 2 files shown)


wolsen:
Some screenshots

**unfinished 3d rendering**
Screenshot from 2021-02-20 20-44-23

**finished 3d rendering**
Screenshot from 2021-02-20 20-44-14

neilweinstock:
Can you show images with fins at 200 alpha and 50 alpha for comparison?  The problem reported in 878 is that all alpha values look the same, other than 255 which looks opaque.

wolsen:
Good call @neilweinstock. It was missing a slight bit. Here's a screen cap with 4 body tubes of different alpha settings - 255, 200, 125, and 50.

Screenshot from 2021-05-31 20-40-34

neilweinstock:
Excellent.  A minor bug, but a bug nonetheless. :)

JoePfeiffer:
Trying to review this... how do you set the transparency?

wolsen:
@JoePfeiffer its a mess on Linux b/c the default Linux widget doesn't provide alpha color options. You have to set the plaf to a Java one rather than a native one. I'll follow up with an example tomorrow or this weekend.

neilweinstock:
In Mac/Windows, you set the alpha channel in any of the various color picker tabs other than the Swatches.

This is something to fix in the future (can't remember if I filed an issue; if not I should).  This should work more like how Rocksim does it: put a separate transparency control on the appearance tab, under "Shine", and then (ideally) remove it from the color picker, if that's possible.

But for now... in the color picker.

JoePfeiffer:
Managed to set it to the Metal look and feel; I've now got the transparency control but I'm only seeing opaque vs. transparent, not the "levels" of transparency shown above.  Oddly, the color picker button appears to show different levels of transparency, though the 3D Finished view does not.
(note that setting a different plaf requires hacking GUIUtil::setBestLAF to keep it from overriding)

neilweinstock:
Sounds weird.  Can I get a JAR file to try out?

JoePfeiffer:
> Sounds weird. I'd like to try it out, although I won't be able to for about a week. If you can wait, then I'll come back and ask for a JAR file next weekend.

Since others are reporting it's working, I'm guessing it's a Linux thing.  I'll try to give it a look in the morning

JoePfeiffer:
I've held up the other platforms long enough.  Since it isn't possible to set the alpha under Linux anyway (see #956 -- specifying a plaf that allows it on the command line gets overriden) let's merge this.

PR #895, opened 2021-02-23
Update thrustcurves from latest Thrustcurve.org data

### LINKED ISSUE #888: Motor data: 29/40-120 casing shows up in modern reloads as 29/40
Note the case descriptions for G53, G76, G64 and G138. These are all for the 29/40-120 case (I buy and use all of these regularly). The case shows up as 29/40. The outdated reloads below (see prior bug on filtering, whence came this image) show up with the proper designation.

image

Using Unstable
MacOS Catalina 10.15.7, 2020 16in MacBook Pro
openjdk version "15.0.2" 2021-01-19
OpenJDK Runtime Environment (build 15.0.2+7-27)
OpenJDK 64-Bit Server VM (build 15.0.2+7-27, mixed mode, sharing)

wolsen:
Update the thrustcurves from the latest Thrustcurve.org data. This
corrects some of the case information.

Fixes #888


JoePfeiffer:
Which ones are corrected with this?

wolsen:
I double checked the same motors identified in issue #888, though I did notice a few others as well - but did not capture them.

I can do this if requested. It did take me some time to narrow it down to the database itself. Fetching the results from ThrustCurve itself did not reveal the error and when refreshing the database the problem disappeared.

I narrowed it down to the data inside the ThrustCurve database to be reporting the 29/40 for the case info, whereas the data pulled did not have this.

PR #919, opened 2021-05-11
Fix proper UTF-8 encoding for CSV export

### LINKED ISSUE #912: Wrong encoding for CSV export
When exporting to CSV, units have question marks instead of actual unit. Please, add a possibility to select a proper encoding (so user can select if he/she wants to have questionmarks or actual units, in unicode or utf-8 or something).
The problem: (should be m/s^2)
image

neilbalch:
This fixes #912 by having `PrintWriter` follow the UTF-8 character set for CSV export.

I found this solution method on Stack Overflow: https://stackoverflow.com/questions/29556253/keep-unicode-characters-in-java-string

### DIFF
--- a/core/src/net/sf/openrocket/file/CSVExport.java
+++ b/core/src/net/sf/openrocket/file/CSVExport.java
@@ -3,6 +3,7 @@ package net.sf.openrocket.file;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
Small point for future reference -- there are quite a few lines whose only change is whitespace.  Avoiding doing that if possible makes reviewing easier.

neilbalch:
Oh, okay. My editor does that automatically and I usually leave the setting on so help me keep good form.

But will do in the future.

PR #920, opened 2021-05-13
Fix #881, #907 issues

### LINKED ISSUE #881: RockSim coversion error: Distance of fin relative to base/bottom reversed
When RockSim measures a distance from the bottom of the component, distances toward the tip (to the left) have positive and distances toward the bottom (to the right) have negative values. In OpenRocket, the reverse is true. OpenRocket correctly inverses the distance for centering rings relative to the base/bottom of the component during conversion...

Ring Placement

But, during conversion, OpenRocket fails to inverse the distance for fins relative to the base/bottom of the component...

Fin Placement

This error results in the fin placement being reversed when relative to the base/bottom of the component.

### LINKED ISSUE #907: Position component relative to tip of nose cone needs to be brought back
I hadn't noticed it was gone until K'Tesh posted on the mailing list...

neilbalch:
I apologize for the (again) whitespace changes... I can't seem to figure out how to make VSCode not change them. I've tried several user setting overrides, but it keeps changing correcting the whitespace even when I think I'm telling it not to.

I believe this fixes #881. I thought about (_and tried_) changing the `RocksimCommonConstants.ROCKSIM_TO_OPENROCKET_LENGTH` constant with a negative sign, but I soon realized that only the relative position field was bungled, and changing this constant screwed up other parts of the import. I tested it on my own RockSim model, the [OEM-provided PublicMissiles Spitfire kit file.](https://publicmissiles.com/images/spitfire.zip) The issue wasn't present for me in the official 15.03 `.jar` app, but I did replicate the issue easily with the current unstable branch.

As Craig pointed out, the fins were positioned wrong on import:
image

But this patch corrects it for me:
image

**EDIT:**
This PR now also fixes #907 by re-adding the ABSOLUTE axial offset enum type to the AxialMethod[] axialOffsetMethods array that every component config references.

As far as I can tell, all of the functionality was still built into the code. The option appears to have been removed by just taking away the reference to the enum type option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,4 @@ openrocket.log
 
 *.snap
 prime/*
+swing/resources/datafiles/presets/system.ser
--- a/core/src/net/sf/openrocket/file/rocksim/importt/FinSetHandler.java
+++ b/core/src/net/sf/openrocket/file/rocksim/importt/FinSetHandler.java
@@ -39,7 +39,7 @@ class FinSetHandler extends AbstractElementHandler {
 	 * The parent component.
 	 */
 	private final RocketComponent component;
-	
+
 	/**
 	 * The name of the fin.
 	 */
@@ -68,13 +68,13 @@ class FinSetHandler extends AbstractElementHandler {
 	 * The length of the tip chord.
 	 */
 	private double tipChord = 0.0d;
-	
+
 	/**
 	 * The length of the mid-chord (aka height).
 	 */
 	@SuppressWarnings("unused")  // stored from file, but not used.
 	private double midChordLen = 0.0d;
-	
+
 	/**
 	 * The distance of the leading edge from root to top.
 	 */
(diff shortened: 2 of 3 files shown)


neilbalch:
By the way, this last commit adds the file `swing/resources/datafiles/presets/system.ser` to the `.gitignore`. `ant` changes this file on every build, which makes me think that it's a compiler temporary and should be ignored by Git.

neilbalch:
As far as I can tell/test, this last commit addresses #907 by re-adding the `ABSOLUTE` axial offset enum type to the `AxialMethod[] axialOffsetMethods` array that every component config references.

As far as I can tell, all of the functionality was still built into the code. The option appears to have been removed by just taking away the reference to the enum type option.

jonnypjohnston:
I have Rocksim. Bought it to compare sim data between the OR and some Jupyter notebooks. I can test the file...

JoePfeiffer:
Fixes #907.

Issue #822, opened 2020-11-12
Feature Request: Option to display stability as a percentage of rocket length


neutronium95:
Calibers of stability are a nice rule of thumb for most normal rocket designs, but they tend to break down when working with especially thin or fat rockets. Some people instead prefer to think of stability in terms of percentage of body length, aiming for around 10-15%. It's kind of annoying to have to calculate that from the CG CP distance all the time.

JoePfeiffer:
This seems like something we might want to make possible as we consider making the rocket information panel configurable (#802)

hcraigmiller:
This issue has been fixed by way of PR #921; this issue should be closed.

Issue #802, opened 2020-10-25
Make data in rocket design panel configurable


JoePfeiffer:
A user on Rocketry Forum suggests adding more data (such as ground hit velocity) to the rocket design panel.  While it seems to me like adding much more to that panel would quickly make it more cluttered, it seems like it could be reasonable to allow users to configure the data in that panel and make their own decisions as to what to display.  Definitely not for the current release though!

PR #921, opened 2021-05-16
Allow rocket info to show stability margin as percentage of length

raketenolli:
* create a new class PercentageOfLengthUnit
* add new unit to stability units

### DIFF
--- a/core/src/net/sf/openrocket/unit/PercentageOfLengthUnit.java
+++ b/core/src/net/sf/openrocket/unit/PercentageOfLengthUnit.java
@@ -1,0 +1,99 @@
+package net.sf.openrocket.unit;
+
+import net.sf.openrocket.rocketcomponent.FlightConfiguration;
+import net.sf.openrocket.rocketcomponent.Rocket;
+import net.sf.openrocket.util.BugException;
+
+public class PercentageOfLengthUnit extends GeneralUnit {
+
+    private final FlightConfiguration configuration;
+    private final Rocket rocket;
+
+    private int rocketModId = -1;
+    private int configurationModId = -1;
+
+    private double referenceLength = -1;
+
+    public PercentageOfLengthUnit(FlightConfiguration configuration) {
+        super(0.01, "%");
+        this.configuration = configuration;
+
+        if (configuration == null) {
+            this.rocket = null;
+        } else {
+            this.rocket = configuration.getRocket();
+        }
+    }
+
+    public PercentageOfLengthUnit(Rocket rocket) {
+        super(0.01, "%");
+        this.configuration = null;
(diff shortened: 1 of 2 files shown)


hcraigmiller:
See also, duplicate issue #822, and new feature request #802.

neilbalch (approved):

JoePfeiffer:
Rather than just labelling it "%" in the Preferences->Units dialog, would it be clearer to label it "% of length"?  If I hadn't already known what it meant, I'm not sure it would have been obvious to me.

wolsen:
> Rather than just labelling it "%" in the Preferences->Units dialog, would it be clearer to label it "% of length"?  If I hadn't already known what it meant, I'm not sure it would have been obvious to me.

I had the exact same thought you did here @JoePfeiffer. However, after having run it and play with it, I think % is actually fine. For brand new rocketeers, they might confused with the caliber unit as well. I think we should document this, but % is better screen real-estate wise; especially when you know what it is.

So in short, I guess personally I'm fine with % if we release note it and document it.

neilweinstock:
Is this handled as a global units preference, so I'll always see either calibers *or* percentage?

If so, then speaking only for myself, I'd much prefer one of the following approaches:
1) Option to show % *in addition* to calibers (this is best IMHO)
2) Preference selectable on a per-rocket basis

I have a few rockets that I would like to use % on, but not all of them.  If I could only see percentage if I select it as a global preference then I'd never use it (or rather, I'd continue to do it in my head like I do now).

If I'm misunderstanding how this is working, then disregard.

wolsen:
> Is this handled as a global units preference, so I'll always see either calibers _or_ percentage?

Global preference, the same way that stability units are elsewhere (like mm, cm, etc).

> 
> If so, then speaking only for myself, I'd much prefer one of the following approaches:
> 
> 1. Option to show % _in addition_ to calibers (this is best IMHO)

Agreed, however I think landing % is fine and a subsequent combined one to get something like `1.2 cal (12 %)` can be a follow up patch and actually fairly straight-forward.

> 2. Preference selectable on a per-rocket basis

This is a larger change and I think Option 1 is better in general. I would argue it can replace just calibers as the default.

> 
> I have a few rockets that I would like to use % on, but not all of them. If I could only see percentage if I select it as a global preference then I'd never use it (or rather, I'd continue to do it in my head like I do now).
> 
> If I'm misunderstanding how this is working, then disregard.

neilweinstock:
> Agreed, however I think landing % is fine and a subsequent combined one to get something like `1.2 cal (12 %)` can be a follow up patch and actually fairly straight-forward.

So... new issue or leave the existing one open?

PR #932, opened 2021-06-03
Fix FinSetTest.testTabLocation

### LINKED ISSUE #931: FinSetTest failing since commit 93633a2
Commit 93633a27 re-introduced the ABSOLUTE AxialMethod for offsets. However, the unit test was not updated to reflect the correct expected settings for the new offset method.

wolsen:
Commit 936333a2 re-introduced the AxialMethod.ABSOLUTE as an offset
method but did not update the corresponding unit test. Add in the
expected offset value for AxialMethod.ABSOLUTE.

Fixes #931


wolsen:
Trivial patch. Merging

Issue #763, opened 2020-08-27
Selection of deployment configuration for one motor on recovery tab opens the config menu for a different motor 


hcraigmiller:
Open the attached OR design file.
[MDRM.Fighter.Rev.02.zip](https://github.com/openrocket/openrocket/files/5133223/MDRM.Fighter.Rev.02.zip)

Go to the Motors & Configurations tab and click on the Recovery tab. On the Recovery tab, select the Main Parachute, Apogee setting for the H115 motor (either double-click on the word "Apogee" or single-click on the word "Apogee" and click the "Select deployment" button). The Select Deployment Configuration window shows that the recovery configuration for the H115 motor is being modified. Click OK or cancel to exit.

Now, select the Main Parachute, Apogee setting for the G138 motor. The Select Deployment Configuration window shows that _the recovery configuration for the **H115** motor is still being modified_.  Change deployment to first ejection charge of the stage then click OK, and it confirms that the recovery configuration for the H115 motor, not the G138 motor, was modified.

![Recovery Wrong Motor](https://user-images.githubusercontent.com/68821492/91369448-a66b2c00-e7c0-11ea-87f0-a01a20a266f8.png)

Even if you go straight from opening the design file to selecting the Main Parachute, Apogee setting for the G138 motor. The Select Deployment Configuration window will show that the recovery configuration for the **H115** motor is being modified.

FYI, may be related, new motor configurations just seem to be ordered without any consistency, sometimes place first in order, sometimes last, and sometimes in the middle of other existing configurations.

JoePfeiffer:
The underlying issue is the selected configuration isn't being kept in sync between the Motors tab, Recovery tab, Stages tab, and the main rocket display window. When selecting a motor it uses the configuration selected in the pane so it isn't as noticeable; when selecting recovery options it's really noticeable since the configuration selected in the main display window determines which configuration is changed, which is not expected.

15.03 had this right...

hcraigmiller:
May be related to #771 and/or #845.

JoePfeiffer:
I *think* they're separate issues.  But I'll check.

SiboVG:
From the first looks, it seems that the pop-up window uses the motor that is selected in the flight configuration. Selecting G138-5 in flight configuration makes the G138-5 to be edited in the deployment configuration, in both cases, both when selecting the G138-5 and H115-14 in the Recovery window. Will look into fixing it.

hcraigmiller:
[...]

PR #942, opened 2021-06-14
Restore 15.03 selection behavior for flight config

### LINKED ISSUE #916: When adding new motor configuration, should automatically select the newly-created config
After creating a new motor configuration, that new configuration should be automatically selected.  The current behavior is weird and a notable regression from 15.03.

Here's the good behavior in 15.03:

First, an existing config is selected:
image

Next, a new config is created:
image
The new configuration is correctly placed at the bottom of the list, and it is selected.

Here's the behavior in the current unstable:

First, an existing config is selected:
image

Next, a new config is created:
image
The new configuration is placed at the *beginning* of the list (this time, at least... position is random).  The selection moves up one position to the second-to-last item.  It appears as if the selection is fixed at the sixth row, even when the whole list shifts down.

I suspect this is related to #845, the new random ordering of the motor configurations in the new unstable.

wolsen:
Restore the behavior from the 15.03 version of the flight configuration
panel. Selected flight configuration is synchronized with the rocket
panel where the design is being done and vice versa.

Fixes #916

### DIFF
--- a/swing/src/net/sf/openrocket/gui/components/ConfigurationComboBox.java
+++ b/swing/src/net/sf/openrocket/gui/components/ConfigurationComboBox.java
@@ -1,14 +1,17 @@
 package net.sf.openrocket.gui.components;
 
+import java.util.EventObject;
+
 import javax.swing.JComboBox;
 import javax.swing.MutableComboBoxModel;
-import javax.swing.event.PopupMenuEvent;
 import javax.swing.event.ListDataListener;
+import javax.swing.event.PopupMenuEvent;
 import javax.swing.event.PopupMenuListener;
 
-import net.sf.openrocket.rocketcomponent.Rocket;
 import net.sf.openrocket.rocketcomponent.FlightConfiguration;
 import net.sf.openrocket.rocketcomponent.FlightConfigurationId;
+import net.sf.openrocket.rocketcomponent.Rocket;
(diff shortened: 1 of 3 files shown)


wolsen:
@neilweinstock mind confirming the behavior of this patch? Test jar file is at https://www.dropbox.com/s/lv2dgvznawvg7xi/OpenRocket-bug916.jar?dl=0

neilweinstock:
Behavior looks good to me.

hcraigmiller:
Window 10 Pro...
The Flight configuration window and the motor/recovery configurations are correctly syncing.

But, there is an abnormality when a new configuration is created. Using A simple model rocket, create a new motor configuration, then select the new motor (I used an Estes D12-7). You will see the D12-7 in the Flight configuration window, but the flight information (lower left corner) says NA in each of the three information fields. _Another motor must first be selected before coming back to the D12-7 for the flight information to be generated._

My hunch is that the command to generate the flight information is _somewhere other than at the end of the motor selection routine_. **But, this is more of an FYI...** I have found no other anomalies.

**This fix also appears to fix Issue #763, PR #943.**

JoePfeiffer:
Looks like it fixes the issue.  I'm going to go ahead and merge this (I note @SiboVG's comment on his #943 that this fixes it on a more fundamental level); I think Craig's observation should be raised as a new issue.

PR #943, opened 2021-06-14
Selection of deployment configuration for one motor on recovery tab opens the config menu for a different motor

SiboVG:
Fixed the issue of wrong configuration selection when trying to edit the Deployment Configuration of the configuration selected in the Recovery tab of the motors.

The issue was that the Deployment Configuration was based on the flight configuration that was selected for the rocket, not the one that was selected in the recovery tab. Fixed this by fetching the flight configuration ID in selectDeployment in RecoveryConfigurationPanel and passing this ID to the DeploymentSelectionDialog (also added an extra parameter in its constructor for this).

Fixes [#763](https://github.com/openrocket/openrocket/issues/763)

Here is the [jar file](https://www.dropbox.com/s/85ebgbdn72uopi2/OpenRocket.jar?dl=0) for testing.

### DIFF
--- a/.idea/ant.xml
+++ b/.idea/ant.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AntConfiguration">
+    <buildFile url="file://$PROJECT_DIR$/build.xml" />
+  </component>
+</project>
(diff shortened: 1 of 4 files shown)


hcraigmiller:
PR #943 fixes Issue #763... But...

It appears that PR #942 and PR #943 are overlapping, i.e. PR #942 fixes the underlying issue encompassed by PR #943, as well as synchronizing the selected flight configuration with the rocket panel where the design is being done, and vice versa. A comparison should be undertaken to avoid conflicts between the two.

SiboVG:
It seems like #942 tackles the problem on a deeper level, hence this branch might be obsolete.

PR #944, opened 2021-06-14
[fixes #879] Added real-time color change appearance

### LINKED ISSUE #879: Changes to the color picker should be reflected in real time
When making changes to the color of a component, the 3D view is only updated after you close the color picker.  This is highly frustrating if you're experimenting.  The 3D view should be updated as changes are made in the picker, or at least within a second or two.

SiboVG:
Edited the ability for the 3D Finished preview to change its color in real-time when changing a component's appearance. When 'OK' is clicked in the appearance window, the color is applied, if cancel is clicked, the initial color will be reverted.

Fixes #879 

Here is the [jar](https://www.dropbox.com/s/c7i2nye5cheitin/OpenRocket_issue-879.jar?dl=0) for testing.


hcraigmiller:
Works in Windows 10 Pro; much more convenient.

PR #946, opened 2021-06-15
[fixes #928] Fix launch lug reference

### LINKED ISSUE #928: Bug: Launch lug distance to a reference point does not change when the reference point is changed
Launch lug components **DO NOT** correctly calculate the component distance to a changed reference point (the distance to a reference point does not change when the reference point is changed).

_I will post specific screenshots and files to show the errors._

* Conversions from Trapezoidal and Elliptical to Freeform fins _may not be_ consistent with the original fin. More testing will be done and further specifics posted.

SiboVG:
Fixed the issue of the lug distance spinner not updating after another reference point was set. Was fixed by trigger a stateChange on the DoubleModel of the AxialOffset of the lug after an item was selected in the reference JComboBox.

Fixes #928 

Here is the [jar](https://www.dropbox.com/s/zdelttkrufxp3nb/OpenRocket-928.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/LaunchLugConfig.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/LaunchLugConfig.java
@@ -20,6 +20,9 @@ import net.sf.openrocket.rocketcomponent.position.AxialMethod;
 import net.sf.openrocket.startup.Application;
 import net.sf.openrocket.unit.UnitGroup;
 
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
 @SuppressWarnings("serial")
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Using Windows 10 with Java version 11.0.9 2020-10-20 LTS. No abnormalities found in operation of test .jar functions. Issue #928  fixed.

wolsen (changes requested):
Per my other review commentary, I'd prefer if we kept the changes focused on what is needed for the fix. I'm generally in agreement with the formatting changes, but those belong in another PR, not this one

wolsen (approved):
LGTM, thanks for the patch!

Issue #940, opened 2021-06-11
Change request: Set launch lug/rail button default rotation to 180 degrees


hcraigmiller:
Currently, the launch lug/rail button default rotation is set to 0 degrees (the top of the rocket). However, a default rotation setting of 180 degrees (the bottom of the rocket) seems more appropriate.

When creating a new design, the user will select either an odd or even number of fins. And, when an odd number of fins is selected, a fin is always placed on the "top" of the rocket, at 0 degrees. This also means that a fin is never placed on the "bottom" of the rocket at 180 degrees and that (if the fins are evenly spaced around the body tube) _a launch lug or rail button with a rotation of 180 degrees_ will always be equally distanced from the fins to the right and left.

If the user selects an even number of fins, those fins can easily (and should) be rotated _the number of degrees between fins divided by two_. Doing so, allows side or top decals to be properly oriented with regard to fin rotation so that (using 4 fins as an example) side decals appear at 90 and 270 degrees and the top decal is at 0 degrees; again, zero degrees is the top of the rocket. As with an odd number of fins, by so rotating an even number of fins, _a launch lug or rail button with a rotation of 180 degrees_ will always be equally distanced from the fins to the right and left.

Those users who wish to place a launch lug against a fin will not be adversely affected because those users must manually rotate the launch lug anyway.

hcraigmiller:
Fixed by way of PR #949

PR #949, opened 2021-06-15
Lug & rail default rotation 180°

SiboVG:
Changed the default launch lug & rail button rotation to 180°, as per suggestion on #940 .

[Java file](https://www.dropbox.com/s/anonqe9iu7m1ye9/OpenRocket-940.jar?dl=0) for testing.

(also changed the layout of the smokeShader, not needed for this issue, but it ended up in this commit, sorry)

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/RailButton.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/RailButton.java
@@ -56,7 +56,7 @@ public class RailButton extends ExternalComponent implements AnglePositionable, 
 
 	private double radialDistance_m=0;
 	protected static final AngleMethod angleMethod = AngleMethod.RELATIVE;
-	private double angle_rad = 0;
+	private double angle_rad = Math.PI;
 	private int instanceCount = 1;
 	private double instanceSeparation = 0; // front-front along the positive rocket axis. i.e. [1,0,0];
 	
--- a/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
@@ -21,7 +21,7 @@ public class LaunchLug extends ExternalComponent implements AnglePositionable, B
 	private double radius;
 	private double thickness;
 	
-	private double angleOffsetRadians = 0;
+	private double angleOffsetRadians = Math.PI;
 	private double radialOffset = 0;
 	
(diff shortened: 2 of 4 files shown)


hcraigmiller:
No abnormalities found; works great.
OS Windows 10 Pro service pack 120.2212.2020.0 with Java version 11.0.9 2020-10-20 LTS.

core/src/net/sf/openrocket/rocketcomponent/RailButton.java
```diff
@@ -66,7 +66,7 @@ public RailButton(){
 		this.totalHeight_m = 0.0097;
 		this.innerDiameter_m = 0.008;
 		this.flangeHeight_m = 0.002;
-		this.setStandoff( 0.002);
```
wolsen, line 69: This change doesn't seem relevant to the current commit. Don't get me wrong, I prefer the spacing changes you have provided, however the change is irrelevant for this issue and we should address some of the formatting issues in subsequent patches.

swing/resources/datafiles/flame/smokeShader.glsl
```diff
@@ -2,35 +2,32 @@ uniform sampler2D uNormal;
...
 
-void main(void)
-{
-	vec3 normal = 2.0 * texture2D (uNormal, gl_TexCoord[0].st).rgb - 1.0;
+void main(void) {
+	vec3 normal = 2.0 * texture2D(uNormal, gl_TexCoord[0].st).rgb - 1.0;
 
-	
-	//Load only the Alpha from the smoke texture
-	vec4 diffuseMaterial = texture2D (uSmoke, gl_TexCoord[0].st);
```
wolsen, line 11: See above, irrelevant change

wolsen (changes requested):
Overall, I'm think the changes are good. I just think we need to avoid including irrelevant changes with the commit. Subsequent patches/prs can address code style/formatting

core/src/net/sf/openrocket/rocketcomponent/RailButton.java
SiboVG: Although I agree with your point of view, I also think it's easier to refactor some things when you see them, instead of having to do everything at once in a later patch. But that's just an opinion, it does make it harder to see what has really changed in a commit when refactoring is included.

wolsen (approved):
LGTM

PR #951, opened 2021-06-16
Make spinner options in photo keyboard editable

### LINKED ISSUE #873: Photo Studio does not allow keyboard entry of values
Windows 10 OS, Photo Studio, all tabs, a double left click on a value field highlights the value, but a numeric value cannot be entered from the keyboard. Value changes can only be made using a mouse to left-click on the up and down arrows.

wolsen:
Spinner options in photo frame does not have an editor, so doesn't
allow for keyboard edits. Add an EditableSpinner class which
automatically sets the editor and use that instead of JSpinner.

Note: this EditableSpinner can be used throughout OR, but not including
everywhere as it is a large change at this point and I'd rather wait for
the refactor until after release, so change is small.

Fixes #873

### DIFF
--- a/swing/src/net/sf/openrocket/gui/components/EditableSpinner.java
+++ b/swing/src/net/sf/openrocket/gui/components/EditableSpinner.java
@@ -1,0 +1,28 @@
+package net.sf.openrocket.gui.components;
+
+import javax.swing.JSpinner;
+import javax.swing.SpinnerModel;
+
+import net.sf.openrocket.gui.SpinnerEditor;
+
+/**
+ * A JSpinner that allows edits.
+ * 
+ * @author Billy Olsen
+ */
+public class EditableSpinner extends JSpinner {
+
+	private static final long serialVersionUID = 1L;
+
+
+	/**
+	 * Creates a new EditableSpinner for the SpinnerModel
+	 * 
+	 * @param model
+	 */
+	public EditableSpinner(SpinnerModel model) {
+		super(model);
+		this.setEditor(new SpinnerEditor(this));
+	}
+
+}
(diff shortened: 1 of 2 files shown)


SiboVG:
Is there a reason why the other JSpinners (i.e. the ones in the home window) didn't have this problem, even though they are also JSpinners? I found that by removing the Spinnermodel from the JSpinner argument in PhotoSettingsConfig, the spinner was again editable. Is it possible that the problem is somewhat deeper rooted within the Spinnermodel that is given as an argument (e.g. pitchModel.getSpinnerModel()), instead of just having to replace the JSpinner with custom declarations?

neilweinstock:
Interesting.  What do all the other edit fields use?  Almost all of them support up/down and test entry.  The Photo Studio fields seemed to be the outliers.

SiboVG:
As far as I can see, the other spinners have an additional line of code to add the text editor (so an extra line for each new JSpinner).

wolsen:
@SiboVG to answer your (I believe now deleted) comment about the JSpinners and the model. If you check the [JSpinner javadocs](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/swing/JSpinner.html) you'll find that the JSpinner creates a default editor that's "appropriate for the model" (see the [createEditor method](https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/javax/swing/JSpinner.html#createEditor(javax.swing.SpinnerModel))), where there are 3 models defined and all others get a JSpinner.DefaultEditor, which is a read-only view of the model.

I had initially discarded overriding the createEditor method because the models return private implementations of the SpinnerModel which cannot be cleanly accessed. However, since I'm just blindly setting it in the constructor - overriding the createEditor method to return the SpinnerEditor is a viable option as well.

PR #953, opened 2021-06-20
[fixes 771] Fix simulation not running after motor configuration change + fix simulation table not updating

SiboVG:
This pull fixes the simulation not updating after having selected a different motor configuration. I just added updateExtras method to the rocket componentChangeListener in RocketPanel.

There was an additional problem that the simulation table would not update its state. So if the simulation was done, this would not update the simulation table. Only if you selected the row in the simulation table that should have updated, that row would update its information. I fixed this by adding a fireDocumentChangeEvent to the BackgroundSimulationWorker in RocketPanel once the simulation was done.

Here is a [jar](https://www.dropbox.com/s/63s84ubhjk7fej4/OpenRocket-771.jar?dl=0) file for testing.

### DIFF
--- a/core/src/net/sf/openrocket/document/OpenRocketDocument.java
+++ b/core/src/net/sf/openrocket/document/OpenRocketDocument.java
@@ -791,7 +791,7 @@ public class OpenRocketDocument implements ComponentChangeListener {
 		listeners.remove(listener);
 	}
 	
-	protected void fireDocumentChangeEvent(DocumentChangeEvent event) {
+	public void fireDocumentChangeEvent(DocumentChangeEvent event) {
 		DocumentChangeListener[] array = listeners.toArray(new DocumentChangeListener[0]);
 		for (DocumentChangeListener l : array) {
 			l.documentChanged(event);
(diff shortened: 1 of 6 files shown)


hcraigmiller:
Test .jar OpenRocket-771.jar: Did not correct; neither a change in an existing motor configuration nor a change in an existing recovery configuration causes the outdated simulation (or any other simulation) to be rerun when the "Flight simulation" tab is later selected.

SiboVG:
I'm sorry, could you give more information on which problem should be fixed then? I thought your problem was that when you change the flight configuration to a different motor, that the simulation of that motor should be auto-run, which was previously not the case.

hcraigmiller:
Using OpenRocket-927.jar, by way of example, after running all of the simulations from the _Flight configurations tab_, if you change a component attribute (such as the length of a nosecone) then, upon returning to the _Flight configurations tab_, **all of the flight configurations are automatically re-run**.

In that same vein, after running all of the simulations from the _Flight configurations tab_, if you change a motor attribute from the _Motors_ tab (such as the delay length, or even the motor itself) then, upon returning to the _Flight configurations tab_, all of the flight configurations **should, but do not** re-run automatically.

Similarly, after running all of the simulations from the _Flight configurations tab_, if you change a recovery attribute from the _Recovery_ tab (such as the deployment event type or the seconds after the deployment event) then, upon returning to the _Flight configurations tab_, all of the flight configurations **should, but do not** re-run automatically.

SiboVG:
Aha I see what you mean, thank you for the clarification!

PR #955, opened 2021-06-21
[fixes #954] Add real-time color change Photo Studio

### LINKED ISSUE #954: Real-time color change not working in Photo Studio
The real-time color change of #879 does not work in Photo Studio.

SiboVG:
Added real-time color changing in Photo Studio. Same method as #944.

[Jar file](https://www.dropbox.com/s/dth2cx3hc1nze6l/OpenRocket-954.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoSettingsConfig.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoSettingsConfig.java
@@ -20,6 +20,9 @@ import javax.swing.JSpinner;
 import javax.swing.JTabbedPane;
 import javax.swing.JTextArea;
 import javax.swing.SwingConstants;
+import javax.swing.colorchooser.ColorSelectionModel;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
 
 import net.miginfocom.swing.MigLayout;
 import net.sf.openrocket.gui.adaptors.BooleanModel;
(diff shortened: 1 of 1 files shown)


hcraigmiller:
No discernable difference in behavior between the current unstable and the test file **OpenRocket-954.jar**.

If a component configuration editor and Photo Studio are both open, when the _Chose color_ window is opened it is no longer possible to select Photo Studio. So, it is not possible to see whether real-time color changing is occurring in Photo Studio because you cannot select Photo Studio until the _Chose color_ window is actually closed.

SiboVG:
It was about being able to change e.g. the sky color or flame color in Photo Studio, which was not possible before.

hcraigmiller:
The functionality of the test file **OpenRocket-954.jar** in Photo Studio is only partially correct.

As background, it is my understanding that the intended correct function of the color-picker is that a newly selected color will show immediately when selected, and that expected functionality is:
1. Clicking "OK" applies the color change and exits the picker,
2. Clicking "Cancel" discards the color change and exits the picker, and 
3. Clicking "Reset" restores the original color and _does not_ exit exit the picker.

Checking the functionality of the test file **OpenRocket-954.jar** in Photo Studio against the above:
1. Clicking "OK" applies the color change and exits the picker [**result is as expected**],
2. Clicking "Cancel" applies the color change and exits the picker [**the expected result is to discard the color change and exits the picker**], and
3. Clicking "Reset" restores the original color and _does not_ exit exit the picker [**result is as expected**].

The "Cancel" function is not working correctly.

SiboVG:
Ah, yes, you're right, sorry, didn't see that, will fix it!

SiboVG:
The issue is now fixed (and amended to the original commit).

Here is the [jar file](https://www.dropbox.com/s/x92rfq42ozi4lhz/OpenRocket-954.jar?dl=0).

hcraigmiller:
The functionality of the amended test file **OpenRocket-954.jar** in Photo Studio is **correct** as to the "Flame" effect; no abnormal behavior detected. As to the "Smoke" effect, this feature could not be tested because the "Smoke" effect is non-functional.

wolsen (approved):
LGTM

PR #957, opened 2021-06-22
[fixes #859] Fix mass component radial offset in 3D view

### LINKED ISSUE #859: Mass component radial displacement not shown in 3D view
Side view:
image

3D view:
image

SiboVG:
This fixes issue #859 in which the radial offset of mass components were not getting added in any 3D view. Fixed it by adding the offsets xOffset and yOffset to the vertices in MassObjectRenderer.

Here is [jar file](https://www.dropbox.com/s/nqc91h4s15i73ft/OpenRocket-859.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/geometry/MassObjectRenderer.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/geometry/MassObjectRenderer.java
@@ -124,12 +124,20 @@ final class MassObjectRenderer {
 	
 	private MassObjectRenderer() {
 	}
-	
+
+	/**
+	 * Draw the mass object as a 3D figure.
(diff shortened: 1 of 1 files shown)


wolsen:
FWIW, I find it generally cleaner to rebase rather than merging the unstable branch into the local issue branch. I know github gives you that for free with a click of a button, but its a feature I personally don't like because it adds merge commits in the git history which can make it harder to do things like git bisect when finding commits which break a change. Its all a tradeoff of ease of use (merge unstable into local branch with a single click) vs git pull unstable, git rebase unstable.

SiboVG:
Ah, good remark, wasn't aware of the rebase option, thanks!

Issue #905, opened 2021-03-14
New feature request: Ability to select a different color for the outside and inside of a tube


hcraigmiller:
Many user desire to have the outside of a tube be a different color than the inside of that tube.  This is especially desired when the user paints the exposed outside surface of the motor mount,  the exposed surface of the aft centering ring, and the exposed inside surface of the body tube all the same color, such as dark gray or black, to protect against or conceal discoloration caused by exposure to motor thrust.

Currently, subject to Issue #904, the outside and inside color of tubes is the same.

PR #958, opened 2021-06-23
[fixes #904] Fix inside color tube different color from outside

### LINKED ISSUE #904: Outside/Edge/Inside color of a tube should be the same
A user commented on the inability to color the inside of a body tube:
[https://www.rocketryforum.com/threads/painting-the-inside-of-a-body-tube-with-openrocket.165220/#post-2114880](url)

Changing the appearance color from default to a specific color changes the outside and edge color of all tube types, but  _doing so only changes the **inside** color of **inner tubes**_. Doing so _should instead change the **inside** color of **all tube types**_, body tubes, inner tubes, tube fins, and launch lugs.

SiboVG:
This pull fixes issue #904 where the inside of a tube did not match the color of the outside. The previous implementation was deliberately programmed to have the default appearance (e.g. cardboard appearance) for the inside (unless you set the alpha of the appearance to a value lower than 255). 

As mentioned on the [forum](https://www.rocketryforum.com/threads/painting-the-inside-of-a-body-tube-with-openrocket.165220/#post-2114880), this might have be done to reflect the reality of not painting the inside, but only the outside of e.g. the body tube. Because of this, I will work on #905 which gives the user the control for different inside and outside colors.

Here is the [jar file](https://www.dropbox.com/s/5ty1uvgi266l5yv/OpenRocket-904.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/RealisticRenderer.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/RealisticRenderer.java
@@ -91,12 +91,7 @@ public class RealisticRenderer extends RocketRenderer {
 	@Override
 	public void renderComponent(final GL2 gl, Geometry geom, final float alpha) {
 	    Appearance app = getAppearance( geom.getComponent() );
-	    if (app.getPaint().getAlpha()<255){
-			// if transparent, draw inside the same as the outside so we dont get a cardboard interior on a clear payload bay
-			render(gl, geom, Surface.INSIDE, app, true, alpha);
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Works beautifully, and it improves the transparency effect for those wishing to hide a component.

wolsen:
@SiboVG - for future reference, if `Fixes #904` is includes in the git commit message, then github will auto link the issue for you :-)

wolsen:
lgtm

PR #959, opened 2021-06-24
[fixes #843] Added radial positioning for the shock cord component

### LINKED ISSUE #843: "Radial position" tab missing for shock cord component.
After further consideration, this issue seems to be an oversight, rather than a new feature.

Every other mass component has a radial position feature, including other recovery components such as parachutes and streamers. Many rockets use attachment points for which users prefer a radial offset display, as well. Adding this feature would also add consistency where the shock cord is the only component within the class that does not have the radial position feature.

SiboVG:
This pull fixes #843, where you could not change the radial position of the shock cord component.

The fix was to first add the radial position tab to the shock cord configuration panel into ShockCordConfig. These settings were taken from the MassComponentConfig, which does have radial positioning. I also had to add new translations for the shock cord radial positioning in message.properties.

Next, the changes in radial position settings of the shock cord had to be applied to the 2D figure of the shock cord component. This was done by porting some settings from the MassComponentShapes to the ShockCordShapes.

! Be aware, for the radial positioning of the shock cord to work in the 3D view, #957 first needs to be merged. I already included these changes for the jar file, so that you can still test the radial positioning in the 3D view, but left it out of the commit.

Here is the [jar file](https://www.dropbox.com/s/xpmgcntp53uzvxb/OpenRocket-843.jar?dl=0) for testing.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1066,6 +1066,11 @@ ShockCordCfg.lbl.Posrelativeto = Position relative to:
 ShockCordCfg.lbl.plus = plus
 ShockCordCfg.lbl.Packedlength = Packed length:
 ShockCordCfg.lbl.Packeddiam = Packed diameter:
+ShockCordCfg.tab.Radialpos = Radial position
+ShockCordCfg.tab.ttip.Radialpos = Radial position configuration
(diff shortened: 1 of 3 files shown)


wolsen (approved):
LGTM

PR #961, opened 2021-06-26
[fixes #816] Fix select rail button side view

### LINKED ISSUE #816: Can't click on rail button in side view to select
Just noticed this on my Mac.  If I click on a rail button in the side view, it doesn't select.  Every other component type seems to work normally.

SiboVG:
This PR fixes #816 where it was not possible to select the rail button from the side view.

The problem was the following:
How the RailButton figure is built up in the side view is as follows: the rail button shape itself is basically 3 cylinders stacked on top of each other. However, in the side view, this cylindrical 3D view is created by using ellipses and lines. This can be best seen if you rotate the view a bit. Each cylinder's top and bottom face is just an ellipse. These ellipses are then connected using straight lines (these are kind of the side faces of the 3D cylinder).

The problem why you can't select the rail button is because what you see is not one shape. By this I mean that if you are at 0° side view and try to select the middle part of the rail button, you aren't actually selecting any shapes, because to the left of the selection point and to the right of it, you just have a 2D line and to the top and bottom of it, you have a flattened 2D ellipse. In between those 2D shapes, there is just emptiness.

Explanation side view

The reason why you can select the flanges if you rotate the view a bit, is because you are then clicking on one of the 2D ellipses. So in that case, you ARE selecting a shape. Also note that if you try to select anything else than the ellipses (e.g. one of the "side faces"), it won't work because then you are back in the first situation, where you aren't selecting a shape, but just emptiness.

-------------

How I eventually solved it: I created 3 invisible rectangles in the rail button side view (in RailButtonShapes) that will kinda act as the side faces of the 3 cylinders. These rectangles fit the encompanied shape of 2 opposite 2D lines. If the user would now click on one of these rectangles, the user is indeed clicking on a shape and not just emptiness, thus the rail button can be selected.

To make the rectangles visible, I added a setter for the color in RocketComponentShape and set the color for each invisible rectangle to one with 0 alpha. 

Here is a [jar file](https://www.dropbox.com/s/9fs8rtdrksull22/OpenRocket-816.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
+++ b/swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
@@ -1,14 +1,17 @@
 package net.sf.openrocket.gui.rocketfigure;
 
 import java.awt.Shape;
-import java.awt.geom.*;
+import java.awt.geom.Ellipse2D;
+import java.awt.geom.Line2D;
+import java.awt.geom.Path2D;
+import java.awt.geom.Point2D;
+import java.awt.geom.Rectangle2D;
 import java.util.Arrays;
 
 import net.sf.openrocket.rocketcomponent.RailButton;
 import net.sf.openrocket.rocketcomponent.RocketComponent;
 import net.sf.openrocket.util.Color;
 import net.sf.openrocket.util.Coordinate;
-import net.sf.openrocket.util.LineStyle;
 import net.sf.openrocket.util.Transformation;
 
 
@@ -64,10 +67,12 @@ public class RailButtonShapes extends RocketComponentShape {
 
 			// Invisible rectangle
 			double y_invis;
-			if (baseHeightcos >= 0)
+			if (baseHeightcos >= 0) {
 				y_invis = center.y;
-			else
+			}
+			else {
 				y_invis = center.y + baseHeightcos;
+			}
 			pathInvis.append(new Rectangle2D.Double(center.x-outerRadius, y_invis, drawWidth, Math.abs(baseHeightcos)), false);
 		}
 		
@@ -85,10 +90,12 @@ public class RailButtonShapes extends RocketComponentShape {
 
 			// Invisible rectangle
 			double y_invis;
-			if (innerHeightcos >= 0)
+			if (innerHeightcos >= 0) {
 				y_invis = center.y;
-			else
+			}
+			else {
 				y_invis = center.y + innerHeightcos;
+			}
 			pathInvis.append(new Rectangle2D.Double(center.x-innerRadius, y_invis, drawWidth, Math.abs(innerHeightcos)), false);
 		}
 		{// outer flange
@@ -105,10 +112,12 @@ public class RailButtonShapes extends RocketComponentShape {
 
 			// Invisible rectangle
 			double y_invis;
-			if (flangeHeightcos >= 0)
+			if (flangeHeightcos >= 0) {
 				y_invis = center.y;
-			else
+			}
+			else {
 				y_invis = center.y + flangeHeightcos;
+			}
 			pathInvis.append(new Rectangle2D.Double(center.x-outerRadius, y_invis, drawWidth, Math.abs(flangeHeightcos)), false);
 		}
 
@@ -116,7 +125,7 @@ public class RailButtonShapes extends RocketComponentShape {
 		RocketComponentShape[] shapesInvis = RocketComponentShape.toArray(new Shape[]{ pathInvis }, component);
 
 		for (RocketComponentShape s : shapesInvis)
-			s.setColor(new Color(1, 1, 1, 0));
+			s.setColor(Color.INVISIBLE);
 
 		RocketComponentShape[] total = Arrays.copyOf(shapes, shapes.length + shapesInvis.length);
 		System.arraycopy(shapesInvis, 0, total, shapes.length, shapesInvis.length);
--- a/swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
+++ b/swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
@@ -1,19 +1,28 @@
 package net.sf.openrocket.gui.rocketfigure;
 
 import java.awt.Shape;
-import java.awt.geom.Ellipse2D;
-import java.awt.geom.Line2D;
-import java.awt.geom.Path2D;
-import java.awt.geom.Point2D;
+import java.awt.geom.*;
+import java.util.Arrays;
 
 import net.sf.openrocket.rocketcomponent.RailButton;
 import net.sf.openrocket.rocketcomponent.RocketComponent;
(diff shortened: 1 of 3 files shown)


wolsen:
Its an interesting approach to add the rectangles, we might need to add a cube instead. If you rotate the rocket by using the slider on the left, or by the rotation in the component dialog, you can no longer select the railbutton.

wolsen (changes requested):
See other comments - this "breaks" selection when rotated, though the selection wasn't really that accurate before.

SiboVG:
Adding a cube wouldn't really help; the combination of the ellipses and rectangles already cover the whole shape of the rail button, even when rotated (I checked this by making the rectangles visible and checking their shape change). I believe the problem lies somewhere else, because you also can't select the ellipses of the rail button in the rotated view, neither can you select the launch lug in that situation. So my guess is that the body tube shape is on top of all the other shapes in that situation, which makes it so that when you try to select anything behind the body tube, it won't work because you are selecting the body tube first because it is on the 'top' layer.

hcraigmiller:
The rail buttons and launch lugs are more consistent than before. When the guide is place over only the airframe, both the launch lug and rail button worked without fail. But, when the guide is place over the airframe and an inner tube, both the rail button and launch lug work sporadically; I haven't been able to determine why yet.

SiboVG:
The selection problem is something that is already present in the current unstable (maybe less frequently, but it is present), see #976. I will take a look at #976 first and see if that fixes the changes that @wolsen wanted.

SiboVG:
The problem mentioned of the rail button not selecting when it was behind the body tube is fixed in #979 and was not related to this PR.

Here is a [jar file](https://www.dropbox.com/s/pk7wc44ypwukelq/OpenRocket-816-2.jar?dl=0) where the fix of #979 is combined with this PR.

hcraigmiller:
No anomalies found... see comments to PR #979.

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -1,19 +1,28 @@
 package net.sf.openrocket.gui.rocketfigure;
 
 import java.awt.Shape;
-import java.awt.geom.Ellipse2D;
-import java.awt.geom.Line2D;
-import java.awt.geom.Path2D;
-import java.awt.geom.Point2D;
+import java.awt.geom.*;
```
wolsen, line 4: nit: its a matter of style (and your editor defaults like) - I personally think we should not use wildcard endpoints (nor does [google coding style guide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements) for that matter. It's a nit however as we don't have a standard defined

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -40,45 +49,78 @@
...
 			path.append( new Ellipse2D.Double( lowerLeft.x, lowerLeft.y, drawWidth, drawHeight), false);
 			
 			path.append( new Line2D.Double( lowerLeft.x,  center.y, lowerLeft.x, (center.y+baseHeightcos) ), false);
 			path.append( new Line2D.Double( (center.x+outerRadius),  center.y, (center.x+outerRadius), (center.y+baseHeightcos) ), false);
 			
 			path.append( new Ellipse2D.Double( lowerLeft.x, (lowerLeft.y+baseHeightcos), drawWidth, drawHeight), false);
+
+			// Invisible rectangle
+			double y_invis;
+			if (baseHeightcos >= 0)
```
wolsen, line 67: another nit: curly braces always, even when optional... why? The following is valid, but it is difficult to chase down those missing {}

```java
if (5 == bar)
    do_something();
    do_something_else();
```

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -40,45 +49,78 @@
...
 			
 			path.append( new Line2D.Double( lowerLeft.x,  center.y, lowerLeft.x, (center.y+innerHeightcos) ), false);
 			path.append( new Line2D.Double( (center.x+innerRadius),  center.y, (center.x+innerRadius), (center.y+innerHeightcos) ), false);
 			
 			path.append( new Ellipse2D.Double( lowerLeft.x, (lowerLeft.y+innerHeightcos), drawWidth, drawHeight), false);
+
+			// Invisible rectangle
+			double y_invis;
+			if (innerHeightcos >= 0)
+				y_invis = center.y;
```
wolsen, line 89: {}

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -40,45 +49,78 @@
...
 			path.append( new Ellipse2D.Double( lowerLeft.x, lowerLeft.y, drawWidth, drawHeight), false);
 			
 			path.append( new Line2D.Double( lowerLeft.x,  center.y, lowerLeft.x, (center.y+flangeHeightcos) ), false);
 			path.append( new Line2D.Double( (center.x+outerRadius),  center.y, (center.x+outerRadius), (center.y+flangeHeightcos) ), false);
 			
 			path.append( new Ellipse2D.Double( lowerLeft.x, (lowerLeft.y+flangeHeightcos), drawWidth, drawHeight), false);
+
+			// Invisible rectangle
+			double y_invis;
+			if (flangeHeightcos >= 0)
```
wolsen, line 108: {}

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -40,45 +49,78 @@
...
+				y_invis = center.y + flangeHeightcos;
+			pathInvis.append(new Rectangle2D.Double(center.x-outerRadius, y_invis, drawWidth, Math.abs(flangeHeightcos)), false);
 		}
-	
-		return RocketComponentShape.toArray( new Shape[]{ path }, component );
+
+		RocketComponentShape[] shapes = RocketComponentShape.toArray(new Shape[]{ path }, component);
+		RocketComponentShape[] shapesInvis = RocketComponentShape.toArray(new Shape[]{ pathInvis }, component);
+
+		for (RocketComponentShape s : shapesInvis)
```
wolsen, line 118: {}

swing/src/net/sf/openrocket/gui/rocketfigure/RailButtonShapes.java
```diff
@@ -40,45 +49,78 @@
...
+			s.setColor(new Color(1, 1, 1, 0));
```
wolsen, line 119: the color is probably worth a constant with a clear name like INVISIBLE

wolsen (changes requested):
In general, this looks great! I'd like to fix up the nits for style purposes. We don't officially have a style guide - but the braces should be fixed imo

PR #962, opened 2021-06-28
[fixes #905] Added ability to select a different color for the outside and inside of component

### LINKED ISSUE #905: New feature request: Ability to select a different color for the outside and inside of a tube

SiboVG:
This PR fixes #905 which was a feature request to be able to select a different outside and inside color of certain components. This is especially useful if, in real life, e.g. you would paint only the outside of the body tube, but leave the inside unpainted (so a cardboard look).

The general idea is: if you have a tube that can have a different outside and inside appearance, then it will have the option in the Appearance editor to change the appearance of both the inside and the outside (see images below). These are just two different tabs. For the inside, you can select whether you want the inside to just be the same as the outside (see checkbutton 'Same as outside') and change all the appearance parameters just like you would do in a normal appearance editor.

Outside panel
Inside panel

You also have the option to set the edge appearance the same as the inside appearance or the outside appearance (see checkbutton below the outside-inside tabbed pane).

EdgesSelection

For objects that do not have a custom inside-outside appearance, the appearance editor is just the same as normal.
RailButton

------------

How I fixed the problem: I created an interface (InsideColorComponent) for components that should exhibit this custom inside-outside behaviour. This interface contains getters and setter for the inside appearance, edgesSameAsInside and insideSameAsOutside. If you e.g. want the BodyTube to have custom inside-outside appearance, it should implement this interface.

The major changes are found in AppearancePanel. This file contains the code for the Appearance configuration panel. I extracted the code for the 'Appearance'-section in the Appearance panel and put it in a separate function 'appearanceSection', to avoid code duplication. If the RocketComponent is of type InsideColorComponent, then a tabbedPane will be created for the outside and inside appearance + a checkbutton is placed underneath the tabbedPane to select whether the edges should have the inside or outside appearance. If the component is not of type InsideColorComponent, then a normal appearance-section is created.

The appearanceSection-method is basically just the same code as the appearance-part in the original AppearancePanel constructor, but with some extra checks for when you have an InsideColorComponent. Also added to AppearancePanel is a separate AppearanceBuilder for the inside appearance.

Next, the renderComponent-method in RealisticRenderer had to be edited to be able to set a different inside appearance than the outside appearance. Here is also the logic for if you want the inside appearance to be the same as the outside appearance.

Finally, every RocketComponent that we want to act like an InsideColorComponent needs to implement this interface and the corresponding methods. Each of these components have the following fields added: insideAppearance (pretty obvious what this is), insideSameAsOutside (= flag to check whether the inside appearance should be the same as the outside one) and edgesSameAsInside (= flag to check whether the edges should have the inside or outside appearance).

The bodies that currently have the custom inside/outside appearance: Nose cone, Body tube, Transition, Tube fins and Launch lug.

I also had to add some extra translations in message.properties for the newly added features.

Here is the [jar file](https://www.dropbox.com/s/w1g00gdsmhth6me/OpenRocket-905.jar?dl=0) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
@@ -30,10 +30,7 @@ public class LaunchLug extends ExternalComponent implements AnglePositionable, B
 	private int instanceCount = 1;
 	private double instanceSeparation = 0; // front-front along the positive rocket axis. i.e. [1,0,0];
 
-	// Settings for inside/edge appearance
-	private Appearance insideAppearance = null;
-	private boolean insideSameAsOutside = true;
-	private boolean edgesSameAsInside = true;
+	private final InsideColorComponentHandler insideColorComponentHandler = new InsideColorComponentHandler(this);
 	
 	public LaunchLug() {
 		super(AxialMethod.MIDDLE);
@@ -289,47 +286,7 @@ public class LaunchLug extends ExternalComponent implements AnglePositionable, B
 	}
 
 	@Override
-	public Appearance getInsideAppearance() {
-		return this.insideAppearance;
-	}
-
-	@Override
-	public void setInsideAppearance(Appearance appearance) {
-		this.insideAppearance = appearance;
-		if (this.insideAppearance != null) {
-			Decal d = this.insideAppearance.getTexture();
-			if (d != null) {
-				d.getImage().addChangeListener(new StateChangeListener() {
-
-					@Override
-					public void stateChanged(EventObject e) {
-						fireComponentChangeEvent(ComponentChangeEvent.TEXTURE_CHANGE);
-					}
-
-				});
-			}
-		}
-		// CHECK - should this be a TEXTURE_CHANGE and not NONFUNCTIONAL_CHANGE?
-		fireComponentChangeEvent(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);
-	}
-
-	@Override
-	public boolean isEdgesSameAsInside() {
-		return this.edgesSameAsInside;
-	}
-
-	@Override
-	public void setEdgesSameAsInside(boolean newState) {
-		this.edgesSameAsInside = newState;
-	}
-
-	@Override
-	public boolean isInsideSameAsOutside() {
-		return this.insideSameAsOutside;
-	}
-
-	@Override
-	public void setInsideSameAsOutside(boolean newState) {
-		this.insideSameAsOutside = newState;
+	public InsideColorComponentHandler getInsideColorComponentHandler() {
+		return this.insideColorComponentHandler;
 	}
 }
--- a/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
@@ -2,7 +2,10 @@ package net.sf.openrocket.rocketcomponent;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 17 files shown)


SiboVG:
Just noticed that the inner appearance still needs to be added to RocketComponentSaver, so will put this as a draft for now.

hcraigmiller:
On construction, works great, no abnormalities found. But, after save and reopen...
- Outside Tab: Use inside appearance for edges - checked (saved without being checked)
- Inside Tab: Default - checked (saved without being checked)
- Inside Tab: Same as outside - checked (saved without being checked)
- Inside Tab: Texture - blank (saved texture was from a file; filename not preserved)
- Inside Tab: Color - black (saved color was red)
- Inside Tab: Use inside appearance for edges - checked (saved without being checked)
Once the save and reopen is fixed, this is going to be a great new feature, especially for those building ring fin designs. 

SiboVG:
I have added the ability to save the applied settings to the inside appearance in the latest commit.

To do this, I had to kinda copy the already present saving code for the normal appearance + change the decal saving to also include decals from the inside appearance.

Here is a [jar file](https://www.dropbox.com/s/kp86d0likhi557c/OpenRocket-905-2.jar?dl=0) for testing.

hcraigmiller:
No abnormalities found on construction or reopening saved file. Nice... 

core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
```diff
@@ -279,4 +287,49 @@ public AngleMethod getAngleMethod() {
...
+
+					@Override
+					public void stateChanged(EventObject e) {
+						fireComponentChangeEvent(ComponentChangeEvent.TEXTURE_CHANGE);
+					}
+
+				});
+			}
+		}
+		// CHECK - should this be a TEXTURE_CHANGE and not NONFUNCTIONAL_CHANGE?
```
wolsen, line 312: I think the current event type is fine.

wolsen (changes requested):
In general, love the change. Its clearly been thought through and implemented nicely.

However, I'm not in love with the copy-pasta between the different components - but I understand why you went this route and don't think its the worst thing in the world. I'd prefer to see something like an InsideAppearanceHandler or something to that effect that wraps all of the copy/paste for the setInsideAppearance. The reason is that in 3-6 months, we'll have a bug where it will get fixed one place but not the others.

core/src/net/sf/openrocket/rocketcomponent/LaunchLug.java
SiboVG: > In general, love the change. Its clearly been thought through and implemented nicely.
> 
> However, I'm not in love with the copy-pasta between the different components - but I understand why you went this route and don't think its the worst thing in the world. I'd prefer to see something like an InsideAppearanceHandler or something to that effect that wraps all of the copy/paste for the setInsideAppearance. The reason is that in 3-6 months, we'll have a bug where it will get fixed one place but not the others.

100% agree, I just didn't see any other possibility other than an interface at the time, but a handler is a good idea!

SiboVG:
Fixed code duplication with InsideColorComponentHandler in 744dbeb.

wolsen (approved):

PR #963, opened 2021-06-28
[fixes #950] Photo Studio sky image empty option

### LINKED ISSUE #950: Photo Studio sky image empty option should be custom color
When picking a sky color in Photo Studio, you can choose different sky images, or pick a custom color. However, the custom color option is just a blank option right now. It's just not that obvious now that you can change the sky color by picking that empty option. Maybe add a checkbox or something else to make it more clear?

image

SiboVG:
This PR fixes #950 where in PhotoStudio the sky image option would just have an empty option to use the sky color instead of an image.

In line with how the appearance in a component config is implemented, I made it so that the empty option now displays a "< none >"-message. This is done by adding a dummy sky object with as string value the "< none >"-message.

None option

@hcraigmiller suggested to also leave out the 'Sky'-terms out of 'Sky color' and 'Sky image' because it is already present as the bold 'Sky'-label. I decided not to change this because to be honest I don't really look at those title labels, plus it reduces having to continuously look at different locations to understand what's going on.

Here is a [jar file](https://www.dropbox.com/s/4ckavcgf78h73o2/OpenRocket-950.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoSettingsConfig.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoSettingsConfig.java
@@ -23,6 +23,7 @@ import javax.swing.colorchooser.ColorSelectionModel;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
 
+import com.jogamp.opengl.GL2;
 import net.miginfocom.swing.MigLayout;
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Works as expected... no abnormalities found; Photo Studio does not allow saved changes, restarts with defaults.

wolsen (approved):

PR #970, opened 2021-07-01
[fixes #651] Fix shine wrong init value

### LINKED ISSUE #651: "Shine" parameter misbehavior when unchecking "Use default" appearance
Sorry this one is a bit hard to explain.

Here's the Appearance tab of a newly added body tube:
image

Now I uncheck "Use default":
image

It says Shine = 30, but in fact internally it has Shine = 0.  How do I know?  Two ways.  First, if I close the edit dialog and then reopen it, it says Shine = 0.  Second if I click the "up arrow" next to 30, it goes to 1.

SiboVG:
This PR fixes #651 where the shine value would have the wrong initial value.

The problem was that the shineModel takes in the value of ab. However, ab's appearance is first set to the default appearance for that component. For example, the body tube's default appearance has a shine value of 30%, thus the spinner will be set to 30. When unchecking the default value, ab now has a new (non-default) appearance, with init value 0. The spinner is still at 30% at this time, but the shineModel value is already at 0. So when you change the slider or use the spinner up/down button, you suddenly jump from the 30%-value to the 0%-value.

So at the initialization of shineModel, I set its value equal to the reset value 0, found in resetToDefaults of AppearanceBuilder.

Here is the [jar file](https://www.dropbox.com/s/rxfcksfxkh4qi47/OpenRocket-651.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
@@ -389,6 +389,9 @@ public class AppearancePanel extends JPanel {
 			add(new JLabel(trans.get("AppearanceCfg.lbl.shine")));
 			DoubleModel shineModel = new DoubleModel(ab, "Shine",
 					UnitGroup.UNITS_RELATIVE);
+			// Set the initial value to the reset state, not the shine value of the default appearance of this component
+			if (mDefault.getValue() && previousUserSelectedAppearance != null)
+				shineModel.setValue(previousUserSelectedAppearance.getShine());
 			JSpinner spin = new JSpinner(shineModel.getSpinnerModel());
 			spin.setEditor(new SpinnerEditor(spin));
 			JSlider slide = new JSlider(shineModel.getSliderModel(0, 1));


hcraigmiller:
I found the following anomaly:

1. Open program
2. Add default nosecone component
3. Add default body tube component
4. Open body tube component appearance tab... **default appearance should be checked with an appearance of 30%**
5. Select "Use default" appearance... **shine shows "0" instead of 30%**
6. Select nosecone to open configuration panel
7. Open body tube configuration panel and select appearance tab... **shine now shows "30%"**

Shine 01

The changes are looking much better than 15.03. 

SiboVG:
Well Step 4 and 5 are correct to have shine value 0%, because this is the shine value for a newly created appearance (the 30% was wrongly adopted from the default appearance of that component; other components for instance had 0% as the shine for their default appearance, which would then also show 0% as the shine value instead of 30%). But I know what the problem might be. Thanks for testing!

hcraigmiller:
Got it. I'll that premise in future testing.

hcraigmiller:
Premise... default appearance should be active (box checked) and default shine percentage (shaded) should be 0% for newly created components. The following unexpected results occurred:

1. Open program, and select _3D Finished_ view
2. Add default nosecone component, and select the appearance tab -- as expected, _Use default_ appearance was checked, the default _Shine_ percentage of "0%" was shaded, and the nosecone _Color_ (white) matched the 3D view
3. Uncheck _Use default_ appearance -- **unexpected, default _Color_ changed to black (including 3D view); re-check _Use default_ appearance and _Color_ changes back to white (including 3D view**
4. With the _Use default_ appearance **checked**, close nosecone component configuration dialog
5. Open nosecone component configuration dialog, and select appearance tab -- **unexpected, _Use default_ appearance was NOT checked** and **the unshaded _Shine_ percentage is now "30%"**, although the nosecone _Color_ (white) matched the 3D view
6. Check _Use default_ appearance -- as expected, the default _Shine_ percentage of "30%" shaded, and the nosecone _Color_ (white) matched the 3D view

Shine 02

No other anomalies were found.
 
SiboVG:
Yes, I have also just noticed the problem of the default unchecking. Compiled using the latest unstable release and also get the same anomaly, so I guess it's a new issue that is already present in unstable?

hcraigmiller:
Other than the default _Shine_ of 30%, the same behavior is confirmed in the current unstable too.

SiboVG:
Could you make this a new issue @hcraigmiller ?

SiboVG:
Okay, so with the fix of #972, everything should work nominally.

The code of this PR has not changed, but here is a [new jar file](https://www.dropbox.com/s/pswhhaypzxicl1c/OpenRocket-651-2.jar?dl=0) for testing that also includes the changes of #972.

hcraigmiller:
Tested all components, no anomalies found.

wolsen:
lgtm

PR #973, opened 2021-07-01
[fixes #969] Added custom left and right appearance for fin set

### LINKED ISSUE #969: New feature request: Ability to select a different color for each side of a fin
Many users wish to have one side of a fin be a different color than the other side of that fin. This is especially true when users paint or use decals to simulate aircraft type features. The Estes Mega Der Red Max is a good example.

Adding the Issue #905 fix to fins is essentially what is needed.

SiboVG:
This PR fixes #969, which was a feature request for a custom left side and right side appearance for the fin sets.

This PR is based on #962, with some minor changes needed to implement it for the fin set.

Here is a [jar file](https://www.dropbox.com/s/k2visgcxjzr454o/OpenRocket-969.jar?dl=0) for testing.


hcraigmiller:
No anomalies found. K'Tesh will finally be able to stop using phantom fins to get the same effect.

neilweinstock:
This is spectacular.

I'd like to see a change to the UI.  Reasons: 

1) Right now, when looking at the default right side appearance, you don't know if there is a separate left side appearance.
2) Those who don't care about the right side/left side stuff still have to look at it in the UI, and possibly be confused by it.

To fix this, my ideal implementation would be:

1) a checkbox, always visible, to the effect of "Use separate left and right appearance".  Default unchecked.
2) When unchecked, there are no left and right tabs.  In other words, it looks exactly like 15.03, other than the additional checkbox.
3) When checked, the left/right tabs appear.

I can create a mockup if you like but I think what I'm describing is pretty obvious.

Feedback welcome.

SiboVG:
> 2. Those who don't care about the right side/left side stuff still have to look at it in the UI, and possibly be confused by it.

This is actually the reason why I didn't include a checkbox that is always visible. In the current implementation, the user just has to open the appearance-tab, change the right-side appearance and don't bother about the left-side appearance, as if nothing was changed. It is only once the user actively goes into the left-side appearance tab that he/she actively has to say whether he/she wants to use separate appearances.

I just don't really like the idea of a lot of extra checkboxes; putting the more advanced separate left-side appearance in a separate tab seems more like a form of information hiding for users who indeed don't need the added feature.

hcraigmiller:
This is the same format that was used for PR #962 (Issue #905) which was recently merged. For the sake of conformity, any changes to this PR should also be implemented in that PR. At this time, where PR #962 was implemented without a request for change, and this PR is such an extreme improvement, I would prefer merging this PR as, and let users get familiar with it. If changes/improvements need to be made, let's do it after that.

neilweinstock:
I missed that other PR, unfortunately.

I would like there to be _some_ kind of indication (I'm not sure what it is), when looking at the default Right side, whether the left side is different or not.  I shouldn't have to click on the Left side tab to figure it out.

Same comment for the Tube fins appearance.

neilweinstock:
I've been testing some more and there are a number of problems in this build.  Sorry for long explanation but I want to make sure it's clear.

[Editorial note: although I would be thrilled to have this feature, I'm concerned at the effort it requires considering that it is a non-critical feature for the release.  I think it is worth considering putting this on hold at least until after we get a beta release out.  *But* if we're gonna do it, we should make sure to get it right.  It's a little squirrelly.]

1) Putting text on fins (very common) was always one of the reasons for needing two separate images on the two sides of the fins: due to the way the textures are mapped on the fin surface, one side always came out with the text mirror-imaged.

Historically, the text was rendered correctly on the *left* side of the fin.  To get text to render correctly on the right side of the fin, you needed to mirror the text in your decal image.

If I am now allowed to specify separate right and left appearances, then I would want the *right* side decal to be handled better, which is to say: not requiring mirrored text.  This means that the whole orientation of the right side texture would be flipped horizontally.

2) Because the *left* side of the fin is where the texture has always worked "as expected", and requires, no change, that is the side that should be the default, rather than the right as currently implemented.  Also the tabs are backwards:
image
The "left" tab should be on the left, and the "right" should be on the right.  This further supports having the "left" appearance be the default.

More to come in next message, may take me a little while to put together.

neilweinstock:
Other problems I found seem to be unrelated to this PR, so the previous message is all I have for the moment.

SiboVG:
Latest commit has the following changes:

- left and right side panels switched
- 'same as outside'/'same as right side' checkbox moved to the outside/left side tab
- right side finset decal by default mirrored

Here is a [jar file](https://www.dropbox.com/s/9cvscutd3i38055/OpenRocket-969-2.jar?dl=0) for testing

hcraigmiller:
No anomalies found. However, the default for the "Use right side appearance for edges" should probably be unchecked so that, by default, the left side controls everything. And, the "Same as inside" text should probably be "Use for right side/inside" where the left side/outside is being applied to the right side/inside.

neilweinstock:
This is closer, but still there are UI issues, some of which Craig has pointed out.

As much as I would like to have this feature, and appreciate the effort into implementing it, I still strongly object to mergin this right now because we are supposed to be feature frozen, bug fixes only.  There are fiddly UI details to work out, and we should not be doing that right now.

BTW this applies to tube fin inside/outside appearance as well; I regret sleeping through that one because it should not have been merged either.  And body tubes and internal tubes should eventually have the same features.

SiboVG:
> No anomalies found. However, the default for the "Use right side appearance for edges" should probably be unchecked so that, by default, the left side controls everything. And, the "Same as inside" text should probably be "Use for right side/inside" where the left side/outside is being applied to the right side/inside.

Fixed in the latest commit.

SiboVG:
i appreciate your concern Neil, but are there any more issues that you see?

Also: about the freezing features and fixing bugs, developing OR, which is an open-source program, is also about motivation. I liked working on this issue, so I'm more prone to chipping in on this one in my free time instead of tackling more deeply rooted bugs. We are working hard on fixing bugs, which is also important, but I also like to see new features which are relatively easy to implement (which this issue kinda was; the UI issues that you speak of are all fixed in 5 minutes tops).

neilweinstock:
After rereading my previous comment, I realize I worded it poorly.  In fact, I was previously swept up in the excitement of having this feature and didn't really object too much until now.  I understand what you are saying about motivation and doing what's interesting, but after _more than six years_ without a release, I really think we need to be doing everything we can to push this over the finish line.

That said, here are my remaining comments on the UI:

1) The checkbox "Same as right side" should say "Use same appearance for both sides", and be checked by default.  To me that is much clearer.  I would further posit that when this box is checked, the Right side tab should be disabled.  Otherwise, you can go to the Right side tab, change a bunch of stuff, and it has no effect.  Basically, any time you're on the Right tab (or on tube fins, the "Inside" tab), that should mean that what you're doing matters.  On the tube fins, it should say "Use same appearance for inside and outside".

2) The "Use right side appearance for edges" is confusing to me, particular in where it's placed, off on its own at the bottom like that.  I'm not exactly sure how it should be done though.

One possibility is to have a checkbox that says "Use this appearance for edges", and put it in the main appearance box on *both* tabs, but showing the opposite indicator.  That is, if it's checked on the Left side tab, it'll be unchecked on the right side tab.

This is sort of a consistent approach that I would take: the tabs up top say Left/Right, but *inside* that tab content it shouldn't be necessary to think in terms of left and right.  Just "this" one that I'm editing.  Dunno if that makes sense, but it does to me. :)

Other than wanting to have it inside the main appearance fieldset, I'm not sure exactly where inside it it should go.  Probably either just above "Texture" or at the bottom.

neilweinstock:
Final note: if I seem highly opinionated about this feature, it's because it represents an increase in complexity that every user will be forced to see, and therefore I _really_ want to try as hard as possible to make it absolutely clear and minimize confusion.  I say this as someone who wrote the decals tutorial on TRF and have spent a lot of time answering questions from confused users (not necessarily about appearance, but other confusing things).

SiboVG:
Thanks for the feedback Neil! It should indeed not be ambiguous.

The latest commits disable the inside/right tab when the 'Use for...' checkbox is enabled. I also changed the checbox for the edges so that it is only visible when the 'Use for...' checkbox is disabled (so when the inside/left tab is activated). This way, the user doesn't have unneeded information, plus there is a visual cue that the edges checkbox is related to the separate inside/outside appearance. Please let me know if you'd like to try a different implementation.

Here is a [jar file](https://www.dropbox.com/s/rhw88yzrvez94xk/OpenRocket-969-3.jar?dl=0) for testing.

neilweinstock:
Great, I'll take a look tonight.

hcraigmiller:
No anomalies found.

With regard to standard fins, UI looks fine to me.

With regard to tube fins, "Use inside appearance for edges" default is checked. The default should probably be unchecked, the same as the standard fins.

Looks really good to me.

neilweinstock:
I'll have some more comments on this tomorrow morning; in the meantime one quickie:

Is it possible to *visually* disable the "Right" tab when using a single appearance for both?  Right now it looks like it should work, but then clicking on it does nothing, which makes it just seem broken.  Can a disabled tab be greyed out or something?

SiboVG:
> With regard to tube fins, "Use inside appearance for edges" default is checked. The default should probably be unchecked, the same as the standard fins.

Wouldn't it be more logical to have it checked by default because it is a tube and if I would paint a tube, I would use spray paint or something for the outside, leaving the edges and inside unchanged.

SiboVG:
> Is it possible to _visually_ disable the "Right" tab when using a single appearance for both? Right now it looks like it should work, but then clicking on it does nothing, which makes it just seem broken. Can a disabled tab be greyed out or something?

I'm afraid us Mac users are at a disadvantage once again :( a disabled tab should indeed be greyed out, but it seems like a Java Swing issue that this is not the case.

hcraigmiller:
With regard to tube fins, "Use inside appearance for edges". It has been my experience that the outside of the tubes and the tube edges are most often the same color, but others may have had different experiences. I'm just glad to have the option; whatever you decide works for me.

neilweinstock:
Sorry I am being slow on this, will try to get to it tomorrow.

neilweinstock:
After fiddling around quite a bit, here's my preferred implementation.  I understand that this is subjective.
image
image

Reasoning:
1) IMHO this is much less confusing.  All options are clearly understandable and have predictable effects.*
2) The two new checkboxes are kept *outside* the normal appearance area, because they are options that affect the entirety of the appearance, not just the left or right side.
3) In the default config with "Use separate appearances" unchecked, there is no extraneous left/right chrome to confuse the user.  Rather, it looks exactly like before.
4) Arguably, you could put the two new checkboxes at the bottom, a bit more out of the way, but I believe that they should appear *before* the area that they affect.  But I haven't tried mocking that up to see how  I like it.

*Actually I am not sure about the "Use right side appearance for edges checkbox".  I am wondering if that should be something like a pulldown with label "Use for edges:" and then two options "left side" and "right side".  I believe that is clearer than what's there now, but I'm not sure if it's better.

Anyway, that is what I would do.

neilweinstock:
Any reactions to my comments?  @SiboVG or @hcraigmiller?

SiboVG:
Am working on it :)

Outlook voor Android<https://aka.ms/AAb9ysg> downloaden

hcraigmiller:
Personally, the current iteration seems simpler than the proposed change. All of the appearance options are clearly within an appearance selection area, rather than appearing in between "Figure style" and "Appearance", looking more closely related to the former than the latter.

Should the text "Use for right side" be changed to something like "Use left side appearance for right side", that would be more clear (and more consistent with the edge text). However, I would not change the "Use right side appearance for edges" text, this language is clear and concise.

Moreover, I would not create a pulldown with "Use for edges" label.

hcraigmiller:
No anomalies found. Looks good to me. 

neilweinstock:
To me that looks good.  I would echo Craig’s comment that for tube fins, a better default would be to use outside appearance for edges. 

The jump when the separate appearance checkbox is changed is not ideal, but I understand why it’s there and I don’t think it’s a blocker. Possibly something for refinement in the future. 

I won’t be surprised to get some mixed feedback on this UI when it goes into wider distribution, but to me this is a good starting point, least likely to confuse users. 

The last remaining possible source of confusion is the inverted texture on the right side appearance.  That is definitely the correct implementation, but I have ideas to make it clearer.  But that will have to wait for a future feature request, definitely not for this release. 

I’m going to leave this one for a dev to merge.

SiboVG:
So basically every components needs to have the outside appearance for the edges by default, because they are all tubes?

neilweinstock:
> So basically every components needs to have the outside appearance for the edges by default, because they are all tubes?

Well, they're not all tubes, but I would say that does apply to tubes in general.  Firstly, I believe that edge appearance matches outside in most typical paint jobs I've eve seen (including my own).  Secondly, when you think about it, the edge is indeed "outside".

Ultimately, I would imagine the separate inside/outside appearance applying to body tubes, inner tubes, tube fins, and transitions.  I believe all other component types should be fine with a single appearance.

neilweinstock:
Problem: when a single common appearance is used for both fin sides (that is, the default behavior), it should look exactly like it did in 15.03, e.g.:
image

Instead, the decal is being inverted on the right side of the fin:
image

My previous request for the right side decal to be inverted in the case _when separate left/right appearances were being used_.

SiboVG:
Good catch @neilweinstock! Latest commits use outside/left side appearance as default option for the edge appearance + only mirrors the decal in separate appearance mode.

Here is a [jar file](https://www.dropbox.com/s/g25g2s7mcee28s3/OpenRocket-969-5.jar?dl=0) for testing.

hcraigmiller:
In Windows 10 Pro, Version 15.03 has the same visual look for the _Repeat_, _Repeat & Mirror_, _Clamp Edge Pixels_, and _Sticker_  options, being:
Fins 01

OpenRocket-969-4.jar has four separate looks, that I will post in a couple of hours.

neilweinstock:
Hmm, doesn't look to me like that decal image would have different looks in
the different settings.  Remember that 969-4 had significantly broken decal
rendering on fins, so tough to use that one as reference.

I'll look at this on my Mac tonight.

hcraigmiller:
When using one side for both, 15.03 and OpenRocket-969-5.jar function the same.

Windows 10 Pro.

neilweinstock:
Everything looks to be working correctly now.

swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
```diff
@@ -463,6 +465,8 @@ public void actionPerformed(ActionEvent e) {
 					c.fireComponentChangeEvent(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);
 					if (outsideInsidePane != null)
 						outsideInsidePane.setEnabledAt(1, !customInside.isSelected());
+					if (edgesCheckbox != null)
```
wolsen: nit: while style previously has no brackets, I think we should *always* use brackets, even for one line if statements. However, this is a style nit for now. In the future, we should add a linter of sorts to check a coding style we can agree upon.

swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
```diff
@@ -363,6 +364,9 @@ public void actionPerformed(ActionEvent e) {
 					c.fireComponentChangeEvent(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);
 				}
 			});
+
+			customInside.getActionListeners()[0].actionPerformed(null);
+			//edgesCheckbox.getActionListeners()[0].actionPerformed(null);
```
wolsen: nit: this should be removed rather than commented out

wolsen (approved):
Couple of minor nits inline, but I see no reason to hold this review up any longer. Technically, we shouldn't merge this one as we should be in feature freeze. However, given the amount of testing done by both hcraigmiller and neilweinstock AND that I have previously merged a related patch set, I'll now merge this.

PR #975, opened 2021-07-02
[fixes #697] Added Photo Studio settings saving

### LINKED ISSUE #697: Feature request: Photo Studio should remember settings
I use the Photo Studio quite a bit.  The default settings are pretty bad, and so usually I have to spend some time adjust everything to look good.  If I close the Photo Studio window, then everything is lost and I have to start over next time.  This is maddening.

It would be extremely useful if the Photo Studio could remember it's settings, preferably in the ORK file for each design.

Frankly, the Photo Studio settings panel could use an overhaul, but this one change would make it a lot less frustrating to use.

SiboVG:
This PR fixes #697 which was a feature request for saving the Photo Studio settings, not just in one session, but also on the .ork-file.

Commit 52fa955 adds the basic saving mechanism for PhotoSettings, very similar to how other objects/properties are saved.  Commit 886e300 just moves some files. Finally, I found that there were dependency issues because the core module was trying to use files from the swing module, causing circular dependencies and other nasty stuff. Big thanks to @wolsen for helping me understand and solve this problem. Finally, commit 568abec is the fix for the dependency problem. Basically, the PhotoSettings from the swing module is represented in the core module as a map of Strings that encapsulates the parameters of Photo Studio. Saving and opening the mapped PhotoSettings is done in PhotoStudioSaver and PhotoStudioHandler. The swing module's PhotoStudioGetter and PhotoStudioSetter handle the real PhotoSettings events. They can get and set the mapped PhotoSettings of the core module through theses classes.

Here is a [jar file](https://www.dropbox.com/s/5b7qgqic1x3ht1d/OpenRocket-697.jar?dl=0) for testing.


hcraigmiller:
The only anomaly that I noticed is that the default "smoke" opacity is now "4000" instead of "40." Otherwise, seems to work well in all respects.

SiboVG:
Ah yes, good catch! I should have set the default smoke opacity as a decimal number. Fixed it in the previous commit.

Here is a [jar file](https://www.dropbox.com/s/14o2f8jiw76u06w/OpenRocket-697-2.jar?dl=0) for testing.

hcraigmiller:
Retested all attributes, **no issue related anomalies detected**.

There is a new issue, however. As part of the testing, I created a design file for each environment sky image, and opened them all at once. All of the sky images worked without anomaly, except for _Stormy Days_, which began to malfunction when the third iteration was opened (opening an additional design simultaneously) and got progressively worse with each subsequent iteration, until it had a blank background (unrelated to the order in which the images were opened):

PhotoStudio Environment SkyImage StormyDays

Where only the _Stormy Days_ sky image is affected, this appears to be an entirely new issue, not related to Issue #697; unless there is an objection, I will create a new issue for this later this afternoon.

SiboVG:
Thanks for testing! Maybe first check if you can reproduce the issue in unstable to see if it is indeed a new issue.

Outlook voor Android<https://aka.ms/AAb9ysg> downloaden

hcraigmiller:
The anomaly with only the Stormy Days sky image described above occurs in the current unstable as well, confirming that the issue is unrelated to OpenRocket-697-2.jar being tested here.

neilweinstock:
Are we waiting for anything to merge this one?

hcraigmiller:
Only for someone to review and merge...

SiboVG:
@wolsen care to review this one?

wolsen:
This was merged, but due to conflicts it was merged offline and this PR doesn't reflect it properly. Closing, but again, this is merged.

PR #978, opened 2021-07-03
[fixes #977] Fix Storm sky image going blank

### LINKED ISSUE #977: Bug: Photo Studio Stormy Days sky image anomaly
Using _A simple rocket_, create a design file for each environment sky image, and open them all at once. All of the sky images work without anomaly, except for _Stormy Days_, which malfunctions when the third sky image iteration or more are opened (opening an additional design simultaneously) and the malfunction gets progressively worse with the opening of each subsequent iteration, until the _Stormy Days_ Photo Studio display has a blank background (this is unrelated to the order in which the ski images were opened):

PhotoStudio Environment SkyImage StormyDays

SiboVG:
This PR fixes #977 where the sky image for Stormy Days would go blank after selecting it, closing Photo Studio, reopening Photo Studio and trying Stormy Days again.

The problem was that the north, east, west... textures inside SkyBoxCross were being saved/loaded in the Storm singleton class, but the textures get somewhat bound to the gl. So, I added a variable to check what the previous gl was. If it does not equal the current gl, then reload the textures.

Here is a [jar file](https://www.dropbox.com/s/ir2qorsesc31vrq/OpenRocket-977.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/photo/sky/SkyBoxCross.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/photo/sky/SkyBoxCross.java
@@ -17,6 +17,7 @@ import com.jogamp.opengl.util.texture.awt.AWTTextureIO;
 public class SkyBoxCross extends Sky {
 	
 	private final URL imageURL;
+	private GL2 previousGL = null;	// GL that the textures were previously bound to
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Tested all sky images, no anomalies found.

PR #979, opened 2021-07-04
[fixes #976] Sometimes can't select components when behind body tube

### LINKED ISSUE #976: Sometimes can't select launch lug when "behind" body tube
Sometimes, when you add a launch lug and rotate the view a bit so that the launch lug is behind or in front of the body tube (the launch lug component is fully within the body tube component shape in the side view), I can't select the launch lug any more.

Steps to reproduce: make a new file, add a body tube, add a launch lug, rotate the view to e.g. 70°, deselect everything, see if I can select the launch lug. If so, delete the launch lug, add a new one, try again to select it and so on... After a couple of tries (sometimes the issue is from the start, sometimes you have to repeat the previously stated issue 10 times or more), the launch lug is not selectable anymore. In this state, if you rotate the view so that the lug is again somewhat or fully outside of the body tube shape, you can again select it. When it is back inside, selection doesn't work. After deleting that launch lug and adding again a new one, I was able to select that new lug.

SiboVG:
This PR fixes #976 where you could sometimes not select the launch lug when it was behind the body tube.

The problem was in RocketFigure. How drawing the shapes works is that you have a list of all the RocketComponentShapes, figureShapes, to be drawn. You then parse this list and draw each shape. However, this list was implemented as an ArrayList. ArrayLists don't have a fixed order when parsing, meaning that if you have an ArrayList with elements {A, B, C}, then the first parse, you might get {A, B, C}, but the next time, it might be {B, A, C}.

What finally caused the problem of not being able to select the launch lug when it is behind the body tube shape, is that sometimes the body tube would be drawn first and the launch lug drawn last. This had the effect that the body tube is drawn on the "top layer" and the launch lug on the "bottom layer". So when you want to select the launch lug, the first shape that you encounter is the body tube, because it is on top.

How I fixed it: I decided to implement a system where each RocketComponent gets an order number, displayOrder, in which it should be drawn. Having a higher displayOrder will cause the RocketComponent to be more on a top layer, so it will be selected first compared to a RocketComponent with a lower displayOrder. Instead of using an ArrayList, I used a PriorityQueue where the priority was determined based on the displayOrder of the RocketComponent of the RocketComponentShape.

So it is now possible to control which type of RocketComponent has priority to be selected. The priority of selection, from less priority to more priority to be selected first is:
0: BodyTube
1: NoseCone
2: Transition
3: TubeFinSet
4: FinSet (this includes all the different kinds of fins)
5: RailButton
6: LaunchLug
7: InnerTube
8: TubeCoupler
9: CenteringRing
10: Bulkhead
11: EngineBlock
12: Streamer
13: ShockCord
14: MassComponent

Here is a [jar file](https://www.dropbox.com/s/w3d40999b4s0del/OpenRocket-976.jar?dl=0) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
@@ -45,7 +45,8 @@ public class BodyTube extends SymmetricComponent implements BoxBounded, MotorMou
 		this.outerRadius = Math.max(radius, 0);
 		this.length = Math.max(length, 0);
 		motors = new MotorConfigurationSet(this);
-		super.displayOrder = 0;		// Order for displaying the component in the 2D view
+		super.displayOrder_side = 0;		// Order for displaying the component in the 2D side view
+		super.displayOrder_back = 0;		// Order for displaying the component in the 2D back view
 	}
 	
 	public BodyTube(double length, double radius, boolean filled) {
--- a/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
@@ -45,6 +45,7 @@ public class BodyTube extends SymmetricComponent implements BoxBounded, MotorMou
 		this.outerRadius = Math.max(radius, 0);
 		this.length = Math.max(length, 0);
 		motors = new MotorConfigurationSet(this);
+		super.displayOrder = 0;		// Order for displaying the component in the 2D view
 	}
 	
 	public BodyTube(double length, double radius, boolean filled) {
--- a/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/BodyTube.java
@@ -46,7 +46,7 @@ public class BodyTube extends SymmetricComponent implements BoxBounded, MotorMou
 		this.length = Math.max(length, 0);
 		motors = new MotorConfigurationSet(this);
 		super.displayOrder_side = 0;		// Order for displaying the component in the 2D side view
-		super.displayOrder_back = 0;		// Order for displaying the component in the 2D back view
+		super.displayOrder_back = 1;		// Order for displaying the component in the 2D back view
 	}
 	
 	public BodyTube(double length, double radius, boolean filled) {
--- a/core/src/net/sf/openrocket/rocketcomponent/Bulkhead.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Bulkhead.java
@@ -13,8 +13,8 @@ public class Bulkhead extends RadiusRingComponent {
 	public Bulkhead() {
 		setOuterRadiusAutomatic(true);
 		setLength(0.002);
-		super.displayOrder_side = 10;		// Order for displaying the component in the 2D side view
-		super.displayOrder_back = 10;		// Order for displaying the component in the 2D back view
+		super.displayOrder_side = 8;		// Order for displaying the component in the 2D side view
+		super.displayOrder_back = 13;		// Order for displaying the component in the 2D back view
 	}
 	
 	@Override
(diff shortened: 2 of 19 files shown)


hcraigmiller:
Beautiful concept and implementation... no anomalies found.

I would suggest one change in the order of priority. From a designers point of view, it is preferred to be able to select components that may be covered by other components in the side view. One example is placing a rail button over a bulkhead or centering ring. In that case, clicking on the rail button will currently select the bulkhead or centering ring even though that component has other areas exposed where the user can click to select that component. Another more stark example is placing a rail button or launch lug over an inner tube (this is likely as common as not in high power designs), in this situation, the rocket must be rotated until the launch guide is or is nearly outside the airframe in order to select it.

Where rail buttons are likely one of the smallest standard components (and launch lugs are of the same class), I believe that the launch lug and rail button priority should be 13 and 14, respectively, but certainly no lower than immediately after the bulkhead (currently 10). 

Again, great way of approaching the issue.

wolsen:
Agreed this is a nice approach, giving a priority to the selection components. You've essentially gone with the logic of inside components get higher priority than external components, and generally smaller components get higher priority than generally larger components. I'll review this in more depth and give it some more thought. I'd love to be able to generalize it based on *current* visual area of the component (to account for rotations and such) rather than static numbers, but the static numbers may be good enough. I'll have to give it a spin and some more thought.

SiboVG:
I have included a different display order in the side view and back view. Based on the suggestions of @hcraigmiller and @wolsen I have changed the order like this (priority of selection, from less priority to more priority to be selected first is):
**Side view:** _logic sort of 'smaller components get more priority'_
0: BodyTube
1: NoseCone
2: Transition
3: TubeFinSet
4: FinSet (this includes all the different kinds of fins)
5: InnerTube
6: TubeCoupler
7: CenteringRing
8: Bulkhead
9: EngineBlock
10: Streamer
11: Parachute
12: ShockCord
13: MassComponent
14: RailButton
15: LaunchLug

**Back view** _logic sort of 'components more to the back of the rocket get priority'_
0: NoseCone
1: BodyTube
2: Transition
3: TubeFinSet
4: FinSet (this includes all the different kinds of fins)
5: ShockCord
6: Streamer
7: Parachute
8: MassComponent
9: RailButton
10: LaunchLug
11: TubeCoupler
12: InnerTube
13: Bulkhead
14: CenteringRing
15: EngineBlock

Please notify me if you would like a different order or approach.

Here is a [jar file](https://www.dropbox.com/s/w3d40999b4s0del/OpenRocket-976.jar?dl=0) for testing.

hcraigmiller:
No anomalies found. As to the side view, no suggested changes. As to the back view I would suggest lowering the priority of the bulkhead and centering rings.

It appears that centering rings are treated as a disk and not as a doughnut when clicked, So, when the outer edge of a centering ring or bulkhead is against the inside of the airframe, the current priority does not allow the selection of any component except an engine block; you can only select the airframe, fins, or something else that "sticks out" beyond the inner edge of the airframe. Because of this, I would suggest **switching the priority of the centering rings and the bulkheads**, and then **putting these two component between the "FinSet" and the "ShockCord" components**. Doing so will allow the user to select most components, realizing that you will never be able to select all components from the rear view (there is just too much overlap).

SiboVG:
Once again, thank you for the feedback @hcraigmiller! The new back order is:
0: NoseCone
1: BodyTube
2: Transition
3: TubeFinSet
4: FinSet (this includes all the different kinds of fins)
5: CenteringRing
6: Bulkhead
7: ShockCord
8: Streamer
9: Parachute
10: MassComponent
11: RailButton
12: LaunchLug
13: TubeCoupler
14: InnerTube
15: EngineBlock

Here is a [jar file](https://www.dropbox.com/s/fcp0mivpyo37i25/OpenRocket-976-2.jar?dl=0) for testing

hcraigmiller:
No anomalies found.

OS Windows 10 Pro service pack 120.2212.3530.0 with Java version 11.0.9 2020-10-20 LTS.

PR #981, opened 2021-07-09
Improved plot appearance + tooltip

SiboVG:
In my opinion, the current plot appearance has a somewhat outdated look, not to mention that it is pretty hard to read the data. This PR updates the appearance to make it cleaner and more in sync with how e.g. MATLAB-plots look like. In addition, it also adds a tooltip for data points on the plot. Hovering on a line of the plot will give you a popup of the data value of that point.

Here is the old appearance:
Screenshot 2021-07-09 at 04 06 44
Here is the new one:
Screenshot 2021-07-09 at 03 39 51

Here is a [jar file](https://www.dropbox.com/s/qgbtvtsh757iyta/OpenRocket-better_plot.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/plot/SimulationChart.java
+++ b/swing/src/net/sf/openrocket/gui/plot/SimulationChart.java
@@ -57,7 +57,7 @@ public class SimulationChart extends ChartPanel {
 				/* tooltips */true);
 		this.setMouseWheelEnabled(true);
 		this.setEnforceFileExtensions(true);
-		this.setInitialDelay(500);
+		this.setInitialDelay(50);
 		this.setBorder(BorderFactory.createLineBorder(Color.GRAY, 1));
 	}
 	
--- a/swing/src/net/sf/openrocket/gui/plot/SimulationPlot.java
+++ b/swing/src/net/sf/openrocket/gui/plot/SimulationPlot.java
@@ -14,11 +14,9 @@ import java.awt.geom.Line2D;
 import java.awt.geom.Point2D;
 import java.awt.geom.Rectangle2D;
 import java.text.DecimalFormat;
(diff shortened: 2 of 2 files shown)


wolsen:
I think the plot data tooltip should show more useful information. While it does show details such as X and Y values, I think it would be better if it showed what the value in terms the user would want to see. From your example, the "Item: 231" is really only showing the 231st sample, which is interesting - but from a UX perspective, its not clear what it is. I personally think it would be better for the tooltip to provide information readily interpreted by the user, i.e. 284,95 m/s. This is especially helpful when there are lines that are near each other (which might mean we need the series label in the tooltip as well). Yes, the XY values provided are accurate, it just takes more work for the user to interpret them

123e443:
Nice! This Looks really good

SiboVG:
> I think the plot data tooltip should show more useful information. While it does show details such as X and Y values, I think it would be better if it showed what the value in terms the user would want to see. From your example, the "Item: 231" is really only showing the 231st sample, which is interesting - but from a UX perspective, its not clear what it is. I personally think it would be better for the tooltip to provide information readily interpreted by the user, i.e. 284,95 m/s. This is especially helpful when there are lines that are near each other (which might mean we need the series label in the tooltip as well). Yes, the XY values provided are accurate, it just takes more work for the user to interpret them

Agreed, I tried it already, but had trouble linking the units and variable names, but I will look into it again.

SiboVG:
The latest commit improves the information shown by the tooltip.

Improved tooltip information

Here is a [jar file](https://www.dropbox.com/s/o0939yzhm67bvzo/OpenRocket-better_plot-2.jar?dl=0) for testing.

wolsen:
Thanks for the changes!

PR #984, opened 2021-07-15
[fixes #948] Fix bad appearance of active tabs on Mac

### LINKED ISSUE #948: Active tabs on Mac look just awful
When a tab is selected and active in a window, on Mac OS this is displayed as a white tab background with white text with just a little bit of a drop shadow. The white text of the active tab should be replaced by e.g. a black color to give it more contrast compared to the background.

Screenshot 2021-06-15 at 20 31 13

SiboVG:
This PR fixes #948 where the foreground of active tabbed panes was white on OSX machines, making it very much unreadable. All credits for the solution go to @wolsen .

Here is a [jar file](https://www.dropbox.com/s/rtv7vj30nwfgr67/OpenRocket-948.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/util/GUIUtil.java
+++ b/swing/src/net/sf/openrocket/gui/util/GUIUtil.java
@@ -250,7 +250,6 @@ public class GUIUtil {
 		try {
 			// Set system L&F
 			UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
-			UIManager.put("TabbedPane.foreground", Color.black);
 			
 			// Check whether we have an ugly L&F
 			LookAndFeel laf = UIManager.getLookAndFeel();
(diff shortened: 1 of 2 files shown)


swing/src/net/sf/openrocket/gui/util/GUIUtil.java
```diff
@@ -258,6 +250,7 @@ public static void setBestLAF() {
 		try {
 			// Set system L&F
 			UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
+			UIManager.put("TabbedPane.foreground", Color.black);
```
wolsen, line 253: I think we should only set this in the OSX case. From SwingStartup there's a unique invocation to OSXSetup.setupOSX() when on Mac. It _might_ work just moving this into the OSXSetup.setupOSX() method, otherwise we'll want to wrap this with an if block for the platform. Also, add a comment on why we're doing this - someone will come along in the future and remove this thinking it does nothing, but a little breadcrumb about why will go a long way.

wolsen (changes requested):

PR #986, opened 2021-07-25
[fixes #968] Fix bad highlight colors

### LINKED ISSUE #968: Bad highlight colors on Mac
This problem was present in 15.03, never got around to reporting it.  Currently running on an M1 Mac on Big Sur.

Certain UI elements are colored with a very dark highlight color, rendering the contained black text unreadable.  E.g.:
image

Same on the delete confirmation dialog:
image

There may be others; these are the ones I was able to find quickly.

Elsewhere in the OS, when UI elements are highlighted with this color, the text is set to white to make it readable.

SiboVG:
This PR fixes #968 for the most part, where some highlight colors or button colors were hardly readable.

The tables from the motor configuration now get assigned the correct foreground color (white) upon selection. This was fixed by changing the foreground in setSelected in FlightConfigurablePanel.

Also toggle buttons, such as the stage selector buttons, are fixed by adding and using a custom JToggleButton implementation, SelectColorToggleButton, which sets the foreground color correctly.

Finally the color of JButtons that are pressed is also fixed by a custom JButton implementation, SelectColorButton.

The only thing that I could not fix were the buttons in pop-up dialogs. For this to work, you would have to create a custom implementation of the JOptionPane.showConfirmDialog with custom buttons (the SelectColorButtons).

Here is a [jar file](https://www.dropbox.com/s/e9m9mmm7uhslwqh/OpenRocket-968.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/components/ColorChooser.java
+++ b/swing/src/net/sf/openrocket/gui/components/ColorChooser.java
@@ -17,6 +17,8 @@ import java.awt.Color;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 
+import net.sf.openrocket.gui.widgets.SelectColorButton;
+
 /**
  * A panel implementation of a color chooser.  The panel has a label and a textfield.  The label identifies 
  * what the color is to be used for (the purpose), and the textfield is uneditable but has its background set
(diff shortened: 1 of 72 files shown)


neilweinstock:
This is a billion times better.  Would love to get a fix in there for the pop-up dialogs, but in the meantime this is a huge improvement.  Should we close #968 and open a new one for the dialog buttons, or leave 968 open with some additional notes?

SiboVG:
Well I would try to merge this PR already because the pop-up dialogs might take some more time. With that I would also close #968 and open a new ticket for the pop-up dialogs.

neilweinstock:
Would like appearance to be verified on Windows and Linux and then we'll merge.

Paging @wolsen and @hcraigmiller

hcraigmiller:
Testing OpenRocket-968.jar... current as compared to initial posted by @neilweinstock:

968 Compare 01

Looks good to me.

neilweinstock:
I'm just gonna merge it now.  If problems show up in Linux (highly unlikely) we'll fix them.

Issue #992, opened 2021-08-10
Very high CPU usage


ShadowNinja:
After a recent update I found the app consuming about 200% CPU (100% in the java process + 100% in the display server process).  I bisected the issue to commit ebcdd469b7678d1b96202634b97ad54077474af8 (Use SelectColorToggleButton).

It looks like `SelectColorToggleButton` resets the foreground style on every frame.  Changing the code to only call `setForground` when the selected state has actually changed resolved the issue, see below:

```diff
--- a/swing/src/net/sf/openrocket/gui/widgets/SelectColorToggleButton.java
+++ b/swing/src/net/sf/openrocket/gui/widgets/SelectColorToggleButton.java
@@ -45,13 +45,13 @@ public class SelectColorToggleButton extends JToggleButton {
     }
 
     @Override
-    public void paint(Graphics g) {
-        if (isSelected()) {
+    public void setSelected(boolean selected) {
+        if (selected) {
             setForeground(UIManager.getColor("ToggleButton.selectForeground"));
         }
         else {
             setForeground(UIManager.getColor("ToggleButton.foreground"));
         }
-        super.paint(g);
+        super.setSelected(selected);
     }
 }
```

SiboVG:
Indeed, the foreground setting should never have been inside the paint-method, my mistake. However, placing it inside the setSelected-method does not yield the desired functionality of setting a white foreground on the blue background on OSX-systems when the button is selected.

SiboVG:
The issue might be resolved in #993, can you check?

ShadowNinja:
Yep, #993 fixes the issue.

PR #993, opened 2021-08-10
[fixes #989 & #992] Fix 3D view going blank + high CPU usage

### LINKED ISSUE #989: 3D using Windows is intermittent
Since the last merge, I have lost all 3D functionality, both in the editor and in Photo Studio.

Latest merge:
3D No

OpenRocket-969-4.jar:
3D Yes

SiboVG:
This PR (I do very much hope) fixes #989 and #992 where the editor view went blank when switching to the 3D view + where a very high increase in CPU usage was noticed. Okay, so I did an oopsie in my PR of fixing bad highlight colors. The code written to change the foreground for the SelectColorButton and SelectColorToggleButton should never have been put inside the paint-method. Not only does this cause a very high CPU usage, it also messes up the init-method calling of the editor window (which caused it to go blank).

I added the foreground updating in a changelistener and only let that code execute on OSX-devices (i.e. devices that have a custom Button.selectForeground or ToggleButton.selectForeground defined).

Tested on Mac OSX Big Sur to check the functionality and testen on Windows 10 Home to test the 3D view going blank (CPU usage not tested).

Here is a [jar file](https://www.dropbox.com/s/ptur74vl7o5qwhl/OpenRocket-989.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/widgets/SelectColorButton.java
+++ b/swing/src/net/sf/openrocket/gui/widgets/SelectColorButton.java
@@ -5,38 +5,49 @@ import javax.swing.Action;
 import javax.swing.Icon;
 import javax.swing.JButton;
 import javax.swing.UIManager;
+import javax.swing.event.ChangeEvent;
+import javax.swing.event.ChangeListener;
 import java.awt.Graphics;
 
 public class SelectColorButton extends JButton {
     public SelectColorButton() {
+        addChangeListenerSelectColor();
     }
 
     public SelectColorButton(Icon icon) {
         super(icon);
+        addChangeListenerSelectColor();
     }
 
     public SelectColorButton(String text) {
         super(text);
+        addChangeListenerSelectColor();
     }
 
     public SelectColorButton(Action a) {
         super(a);
+        addChangeListenerSelectColor();
     }
 
     public SelectColorButton(String text, Icon icon) {
         super(text, icon);
+        addChangeListenerSelectColor();
     }
 
-
-
-    @Override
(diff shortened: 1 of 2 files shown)


hcraigmiller:
This appears to have fixed the issues... no anomalies found.

hcraigmiller:
This does not appear to be an issue in Windows (when resizing the window, the 3D display does not black out).

SiboVG:
Do you not have the problem that I've discussed [here](https://github.com/openrocket/openrocket/issues/989#issuecomment-896137410) @hcraigmiller?

hcraigmiller:
Issue #989 appears to have been fixed by the 989.jar insofar as I have been able to determine.

neilweinstock:
Outside of PR994, when did the blank buttons occur?  I need to know how to reproduce so I can verify the fix.

SiboVG:
> Outside of PR994, when did the blank buttons occur? I need to know how to reproduce so I can verify the fix.

Starting from the fix bad highlights commit.

neilweinstock:
I mean, how do I reproduce it (in a build before this one)?

SiboVG:
The main OR window needs to be out of focus, you can do this by either making a popup window appear inside OR itself (e.g. change the configuration of a component) or you can go to a different program, e.g. your browser and move the browser window out of the way so that you can see the OR main window.

neilweinstock:
OK, I'll try that tonight.

neilweinstock:
Looks good.

wolsen (approved):

PR #994, opened 2021-08-11
[fixes #897] Permanently enable add stage button

### LINKED ISSUE #897: Suggestion: keep the "Stage" assembly component button enabled all the time
Each component button is enabled only when a suitable parent component is selected in the tree.  For all component types that makes sense, except for one: the "Stage" button.  To get it active you need to select "Rocket" in the component tree, which is unusual and (IMHO) not obvious.  Since there's only ever one rocket, there's no ambiguity of where the new stage should go.  So I would just keep that button enabled all the time.

Recall that we removed the always-enabled "New Stage" button from the column of buttons in the middle, which is fine, but the new button should be just as easy to use as the old one.

SiboVG:
This PR fixes #897 which was a request to permanently enable the button to add a new stage.

The PR has the following functionality:

- If a stage component or any of its subcomponents is selected and there is already a stage after the selected stage or parent stage, then a popup window will ask whether the user wants to insert the stage at the end or between the two stages (same behaviour as with the body components).
- In any other case (last stage selected, rocket selected or no component selected), the stage will be added to the end.

Here is a [jar file](https://www.dropbox.com/s/ec61y7gaehq7xqk/OpenRocket-897.jar?dl=0) for testing.


hcraigmiller:
The only anomaly found relates to the ordering of the stage buttons for the display window when a stage is "inserted", as shown below.
Stage Buttons

neilweinstock:
Confirming what Craig saw with regard to the ordering of the stage buttons in the display area.

One additional anomaly: while the dialog is up to choose where to put the new stage, the existing stage buttons go blank, on Big Sur.  After the dialog is closed, the buttons return to normal.
image

SiboVG:
Fixed @hcraigmiller's remark in the latest commit. This was however an issue that was already present before this PR, as changing the stage order in the component tree did not change the button order in previous unstable versions. It does however work with the new commit.

Here is a [jar file](https://www.dropbox.com/s/e42h0r0j7ffii1j/OpenRocket-897-2.jar?dl=0) for testing.

SiboVG:
And yes @neilweinstock, I'm aware, this isn't a problem of this PR however, but a consequence of the 'fix bad highlight colors' fix. What you are seeing is the text color still being white because the toggle button is still selected, but the background color for some reason returns to the unselected background when you have your focus in another window. So this should really be a separate issue.

hcraigmiller:
No anomalies found.

SiboVG:
@neilweinstock the problem of the blank buttons should be fixed in the [latest commit of #993](https://github.com/openrocket/openrocket/pull/993#issuecomment-897595594).

neilweinstock:
Looks good.

neilweinstock:
@wolsen or @JoePfeiffer do you want to look at the code for this one before merge?

PR #998, opened 2021-08-22
[fixes #871] Automatic base/outer diameter unexpected results/warnings

### LINKED ISSUE #871: Automatic base/outer diameter unexpected results/warnings
Using the simple model rocket design, _with the nose cone automatic outer diameter off_, turning off the body tube automatic outer diameter causes a body tube discontinuity warning.

Error Discontinuity 02 Text

The same is true with the nose cone,  _with the body tube automatic outer diameter off_, turning off the nose cone automatic outer diameter causes a body tube discontinuity warning. This outcome can be changed by manually re-entering the outer diameter for both the nose cone, then the body tube. 

[Opening a new instance of A simple model rocket] Where turning off the automatic body tube outer diameter results in  A below:

Error Discontinuity 07 Text

Even though the nose cone shows a base diameter of .984 in. when the design file is opened (see B above), the nose cone may be retaining the default 1.969 in. base diameter when determining body tube discontinuity. This would explain why the body tube discontinuity warning remains after manually entering _the same outer diameter for the body tube as the nose cone base diameter_ (see C above). 

As part of correcting the above, two other suggestions:

1.  **Do not allow** an automatic base/outer diameter box to be checked (grey it out) if there is no other component base/outer diameter to reference. For example, when there is only a nose cone, there is nothing for the automatic base diameter to reference, and the automatic base diameter box should be greyed out. This will prevent A from accidentally becoming B, below:

Error Discontinuity 04 Nose Cone Text

2. **Do not allow** _all components_ to be set to automatic base/outer diameter; if this is allowed, there is no "fixed" component base/outer diameter to reference. And, this will prevent A from accidentally becoming B, below:

Error Discontinuity 05 Text

SiboVG:
This PR fixes #871 which had a number of problems and suggestions for the automatic base/outer diameter system, used in the Body Tube, Nose Cone and Transition.

These first commits fix 3 suggestions given by @hcraigmiller.

These are the current changes:

- Disable the auto checkbox when there is no previous/next reference component to take the diameter from.
- Disable the auto checkbox when the component that you want to get the diameter from already has its auto checkbox checked (so the current component already serves as the reference component for that component).
- For the body tube: disable the auto checkbox when there is only one component connected to it (left or right) and that component already has its auto checkbox checked.
- For the body tube: enable the auto checkbox when there is already a component connect to it (left or right) with its auto checkbox checked, but there is also another component on the other side of the body tube that does not have its auto checkbox checked yet.
  - With this change, the user can also determine which component to pick as a reference component for the outer diameter of the Body Tube. Previously, when you checked auto diameter of the Body Tube, the previous component to the Body Tube would be the first choice to get the reference diameter from. Only if there was no previous component, then the next component would be looked at. However, now you can decide the previous or next reference component by setting a certain sequence for the auto diameter. E.g. you have a nose cone, then a body tube and then a transition. If you want the nose cone to be the reference for the body tube, set auto on for the transition and then auto for the body tube. If you want the transition as a reference, first set the auto of the nose cone on and then auto of the body tube.
- Previously, when you entered a manual diameter (so auto diameter checked off), then checked auto diameter back on, saved the document and reopened the document, the manually entered diameter was reset to a default value, once you unchecked the auto diameter again. Manually entered diameters are now also saved in the OR file, even if auto diameter is turned on (previously, only the 'auto' state was saved in the OR file, but now both the auto state and its manual value are saved).

  - @hcraigmiller suggested to leave out the manual values and set the diameter to the auto diameter value, but I like the "memory" of having previous manually entered values that you can switch between by checking or unchecking the auto checkbox, so I decided to just save the manual values in the OR file

The solution for rounding of misconceptions is not yet implemented. I will check with @hcraigmiller and potentially others what the best solution for that might be.

Here is a [jar file](https://www.dropbox.com/s/j6wwjivvomtdpxw/OpenRocket-871.jar?dl=0) for testing.


hcraigmiller:
Works well, no anomalies found. 

One suggestion though, when the top component is selected (usually a nosecone, body tube, or transition), it would be helpful if the user were asked whether to put the component being added before the top component or at the end. A lot of times I start with body tubes, instead of nosecones, usually because I'm designing from parts I have on hand, and when I decide on a nosecone later by selecting the top component (the body tube) the nosecone is added to the bottom. Yes, you can just select the nosecone and move it up, but it would be more convenient if I were asked whether to insert it before or after the top component, or at the end of all components.

Just a thought.

SiboVG:
I agree, however, I think it's best to put this in a new issue?



SiboVG:
I will put this PR up for review because I believe it's suitable to put it up for merging, but the rounding off part of #871 is best issued for future release. @hcraigmiller would you be so kind as to make a new issue purely for the rounding off part of #871 (and also a new issue for your previous comment of the adding of top components).

JoePfeiffer:
How is it handling inner/outer components?  For instance, under what conditions are the outer diameter of a coupler and the diameter of a bulkplate allowed to be automatic? At first glance using a design of mine it appears to be doing the "right thing", so this is really just a curiosity question.

SiboVG:
> How is it handling inner/outer components? For instance, under what conditions are the outer diameter of a coupler and the diameter of a bulkplate allowed to be automatic? At first glance using a design of mine it appears to be doing the "right thing", so this is really just a curiosity question.

Well, outer components can not have dependencies on the dimensions of inner components. So a body tube doesn't care what the dimensions of its child coupler are, because it's merely a "child"/inner component. Inner components can however have a dependency on the dimensions of its parent outer component. So inner components can take automatic dimensions from outer components, but not the other way around.

PR #1006, opened 2021-09-01
[fixes #1002] Fix offset distances scaling on single component scaling

### LINKED ISSUE #1002: Bug: Scaling single component scales reference point distance
When scaling a single component, the distance between the reference point is not preserved, it is scaled as well; subcomponent reference distances are preserved when the parent is scaled. All components are affected (not applicable to assembly and nose cone, body tube, and transition components).

SiboVG:
This PR fixes #1002 where the offset distances (axial/radial) would also scale when a single component is scaled. This PR splits the scaler into a scaler with offset scaling and one without offset scaling. The right scaler parameters are then attributed to the right scaler. The scaler to be used is based upon the selection in the 'Scale :'-combobox from the scale dialog window ('Entire rocket', 'Selection and all subcomponents' or 'Only selected component').

Here is a [jar file](https://www.dropbox.com/s/egmczeay7qs4jki/OpenRocket-1002.jar?dl=0) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/ScaleDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/ScaleDialog.java
@@ -7,6 +7,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


hcraigmiller:
No anomalies found.

PR #1008, opened 2021-09-03
[fixes #1001 + others] Change default tab + separation event when adding new stage

### LINKED ISSUE #1001: New stage configuration is unexpected
When a new stage is added, the configuration dialog is open to the relatively advanced "Overrides" tabs.  Meanwhile, hidden in the "Separation" tab, the separation event is initially set to "never".  This could easily cause confusion for a new user -- the "Overrides" is sufficiently advanced (and typically not used until updating a design to match the constructed rocket) that a new user is apt to simply close the configuration dialog.  If this happens, the user is likely to be puzzled as to the stage's behavior.

I would suggest that the dialog have the "Separation" open when it first opens, and have the initial separation event be "ignition of next stage".

SiboVG:
This PR fixes #1001 which was a request to change the default tab of the stage configuration dialog, which is a more logical choice. The default separation event is also changed to ignition of next stage instead of the default 'never'.

While working on this PR I also noticed some other issues:

- When rearranging the stages in the component tree, the stage numbers of the stages would not update accordingly. E.g. in the 'Three-staged rocket' template project, if you rearrange the Sustainer-stage to be between the two Booster stages, you would expect the Sustainer and last Booster stage to now be visible in the stages table in 'Motors & Configuration > Stages'. This was not the case, because the stage numbers of the stages did not get updated when their position in the component tree would get updated.
  - Fixed this issue by implementing a method 'updateStageNumbers' in Rocket.java to update the stage numbers
- When changing the name of a stage, the name change would not be executed in the stage table. E.g. in the 'Three-staged rocket' if you rename the final booster stage to 'Booster 2', in the stage table this stage would still be called 'Booster 2'. Only if you would rearrange the stages in the component tree, the name change would get executed.
  - Fixed this by listening for a non-functional change (i.e. a name change) inside the stage table and then triggering a change event so that the table will update if the name has changed
- Same problem as the previous one, but now with the motor table. E.g. in 'A simple model rocket', when changing the name of the inner tube, the name would not be changed in the motor table.
  - Fixed in the same way as with the stage table

Here is a [jar file](https://www.dropbox.com/s/2laxz6ajnpw4q5k/OpenRocket-1001.jar?dl=0) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
@@ -496,10 +496,24 @@ public class Rocket extends ComponentAssembly {
 	
 	@Override
 	public void update(){
+		updateStageNumbers();
 		updateStageMap();
 		updateConfigurations();
 	}
-	
+
+	/**
+	 * Update all the stage numbers based on their position in the component tree
+	 */
+	private void updateStageNumbers() {
+		for (RocketComponent component : getChildren()) {
+			if (component instanceof AxialStage) {
+				AxialStage stage = (AxialStage) component;
+				forgetStage(stage);
+				stage.setStageNumber(getChildPosition(stage));
+			}
+		}
+	}
+
 	private void updateStageMap(){
 		for( RocketComponent component : getChildren() ){
 			if (component instanceof AxialStage) {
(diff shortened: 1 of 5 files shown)


hcraigmiller:
It seemed to work on all of the variations that I tried... no anomalies found.

PR #1010, opened 2021-09-04
[fixes #1009] Added more advanced DecimalFormat for plot hover

### LINKED ISSUE #1009: Change rounding off point at hoover tooltip in simulation plot for values < 0.10
Right now, when you hover over a data point in a simulation plot, a hover window shows up that displays information about that data point. However, the rounding off of the y-value is now done upto 2 decimals. This causes problems if your data set is always smaller than 0.005. In that case, you will always get '0.00' as a rounded of value. An example of this can be seen below where 'Position North of launch' ranges from -0.0005 to 0.0006.

The rounding off to 2 decimals is good for normal situations (for values >= 0.10), but below that, I would always make sure that
you have 2 decimals remaining of your raw result. E.g. if the value is 0.09213, then round off to 0.092. If the value is 0.0001345, then round off to 0.00013 etc.

image

SiboVG:
This PR fixes #1009 by adding a more advanced DecimalFormat for the hover window data values.

Here is a [jar file](https://drive.google.com/file/d/1hxySnlg9iv2HjxF3Ul__5JflBMn9uLai/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/plot/SimulationPlot.java
+++ b/swing/src/net/sf/openrocket/gui/plot/SimulationPlot.java
@@ -27,6 +27,7 @@ import net.sf.openrocket.unit.Unit;
 import net.sf.openrocket.unit.UnitGroup;
 import net.sf.openrocket.util.LinearInterpolator;
 
+import net.sf.openrocket.utils.DecimalFormatter;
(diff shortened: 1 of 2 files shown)


hcraigmiller:
It seemed to work on all of the variations that I tried... no anomalies found.

PR #1013, opened 2021-09-06
[fixes #899] Use only filename in save/export filename input

### LINKED ISSUE #899: Exporting Rocksim file seems broken, and OR gets very confused about the file extension
Here's what the "Export File" dialog looks like:
image
Notice it has given the file the extension ".ork.rkt", rather than simply ".rkt".  This is problem #1.

Then, if I hit "save", here's the file I get:
image

It is indeed an ORK file that OR can read in.  So it seems that Rocksim export is totally not working at the moment.

This is on my Mac running Mojave.  It's a pretty recent unstable build, minus the last couple of PRs.

SiboVG:
This PR fixes #899. I decided to follow @hcraigmiller's suggestion and only use the filename, without file extension, in the 'File name' text input. This not only fixes the issue that @neilweinstock experienced, but is also more logical imo and is often seen in more modern software (e.g. Adobe software).

Here is a [jar file](https://drive.google.com/file/d/1EsEPEzPkPm6bwYNiGW-N7x_bxJAs7TqV/view?usp=sharing) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/document/OpenRocketDocument.java
+++ b/core/src/net/sf/openrocket/document/OpenRocketDocument.java
@@ -34,7 +34,7 @@ import net.sf.openrocket.util.ArrayList;
  */
 public class OpenRocketDocument implements ComponentChangeListener {
 	private static final Logger log = LoggerFactory.getLogger(OpenRocketDocument.class);
-	
+	private final List<String> file_extensions = Arrays.asList("ork", "rkt");	// Possible extensions of an OpenRocket document
 	/**
 	 * The minimum number of undo levels that are stored.
 	 */
@@ -191,6 +191,23 @@ public class OpenRocketDocument implements ComponentChangeListener {
 	 * @return	the File handler object for the document
 	 */
 	public File getFile() {
+		return file;
+	}
+
+	/**
+	 * returns the File handler object for the document without the file extension (e.g. '.ork' removed)
+	 * @return the File handler object for the document without the file extension
+	 */
+	public File getFileNoExtension() {
+		if (file == null) return null;
+		int index = file.getAbsolutePath().lastIndexOf('.');
+		if (index > 0) {
+			String filename = file.getAbsolutePath().substring(0, index);
+			String extension = file.getAbsolutePath().substring(index + 1);
+			if (file_extensions.contains(extension)) {
+				return new File(filename);
+			}
+		}
 		return file;
 	}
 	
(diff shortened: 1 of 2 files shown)


hcraigmiller:
Using _A simple model rocket_, I could find no functional difference between OpenRocket-899.jar and the current unstable (on the initial export both stripped the motor files). On second export (with the same file name), both give the following warnings: (1) file exists, (2) Export not fully supported, and (3) file exists (for a second time), then the save function executes by overwriting the file.

Other than the file exists/overwrite warning coming up twice, no anomalies found, But, again, I found no functional difference between OpenRocket-899.jar and the current unstable. **Am I missing something.**

SiboVG:
On the current unstable, if you create a new file, let's say 'test.ork', if you then open that file in OR and export that file to .rkt, the file name text input will show 'test.ork.rkt'. Same goes for if you open test.rkt and try to save it as an ork-file, it will show 'test.rkt.ork'.

It might be that the jar file was in a haste not yet from the latest changed I made; I have updated the jar-file in the original comment in case that I did not have the right jar file.

neilweinstock:
Following applies to both the Export and Save As... dialogs:
1) when you first open them, no extension is shown
2) Switch to "All files", still no extension is shown
3) Switch back or original setting, extension is now shown

Is this intentional?

I don't think I'm on board with the "don't show the extension strategy".  Given that you could be going back and forth between RKT and ORK files, leaving the extension hidden is a possible recipe for mistakes and/or confusion.  Yes I know the file format selector at the bottom tells you what the extension will be, but....  I dunno, it doesn't feel right to me.

However, I don't feel strongly about this.

I know that some programs behave like this.  I just checked Pixelmator Pro here on my Mac, and note that it *does* show the extension, so I don't think any particular behavior is universal.

SiboVG:
Ah, just checked with Adobe Photoshop and it does show the file extension. On second thought, I understand @neilweinstock's suggestion of showing the extension, so I will change it. And no, the extension showing after switching after 'All Files' was not intentional 🙃 .

By the way: shouldn't the 'All files'-option in the save dialog and export dialog be removed? Because it makes no sense to have an 'All files' option for saving a file. It should only be visible for the import dialog, no?

neilweinstock:
I tend to agree about the "All Files" option in those dialogs.  I'm not sure what it even does.

I say get rid of it.

SiboVG:
Latest commit shows the file extension + removes the 'All Files' option.

Here is a [jar file](https://drive.google.com/file/d/1gJ-bHFZ98MrUPieWcVQX2EGtbiUb1pKr/view?usp=sharing) for testing.

hcraigmiller:
Other than the file exists/overwrite warning coming up twice (after the export as dialog, and again after not all features supported dialog), no anomalies found.

SiboVG:
Ah, good catch @hcraigmiller! The latest commit removes the redundant (and illogically placed) 'confirmwrite'-function in saveAsRocksim in BasicFrame, thus removing the extra overwrite-dialog.

Here is a [jar file](https://drive.google.com/file/d/1tak-ooOoe0EVXiq4KuKFRVWCY02vb2SK/view?usp=sharing) for testing.

PR #1016, opened 2021-09-12
[fixes #1003] Fix FinSet point clipping when further than parent body

### LINKED ISSUE #1003: Bug: Fin root edge not drawn correctly
When the root edge of a flat fin extends past the end of a body tube, the root edge is not drawn beyond the end of the body tube; tube fins draw correctly.

0000 Fins

SiboVG:
This PR fixes #1003 which was a bug that when a FinSet was to the right of the end of its parent body, the last point of the root of the FinSet would clip to the end of the parent body.

The problem was that the getMountPoints-method in FinSet.java contained some code that deliberately clipped the last point of the root of the FinSet to the end of the parent body, if the position of the last point was beyond the end position of the parent body.

I just removed that code block, since it seems illogical (the FinSet is allowed to go further than the parent body and the clipping only has effect on the root of the FinSet; e.g. in a freeform FinSet, all the points besides the root were allowed to go further than the parent body, so I don't see a reason why the root shouldn't be allowed to) and produces confusing results, as with #1003.

Here is a [jar file](https://drive.google.com/file/d/1NwVJbHUZDfXR8GAIHk8Tdgm48qyLOMYB/view?usp=sharing) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -1161,6 +1161,12 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 			xCur += increment;
 		}
 
+		// correct last point, if beyond a rounding error from body's end.
+		final int lastIndex = points.length - 1;
+		if (Math.abs(points[lastIndex].x - body.getLength()) < 0.000001) {
+			points[lastIndex] = points[lastIndex].setX(body.getLength()).setY(body.getAftRadius());
+		}
+
 		if( 0.0000001 < (Math.abs(xOffset) + Math.abs(yOffset))){
 			points = translatePoints(points, xOffset, yOffset);
 		}
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -1161,12 +1161,6 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 			xCur += increment;
 		}
 
-		// correct last point, if beyond a rounding error from body's end.
-		final int lastIndex = points.length - 1;
-		if( body.getLength()-0.000001 < points[lastIndex].x) {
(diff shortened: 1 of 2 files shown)


neilweinstock:
We have had some squirrelly issues with the clipping of the finset editor.  I will be testing this one carefully.

SiboVG:
Ah okay, sorry, wasn't aware, the latest commit re-enabled the clipping, but only clips around the edge of the parent body, not further than it, so still allowing the fin point to go further than the parent body, once it passes the edge clip threshold.

Here is a [new jar](https://drive.google.com/file/d/1zAkiivklvdQIwkBBkneAbx_2JnVw2jnR/view?usp=sharing) for testing.

neilweinstock:
After a brief test I'm not sure what this is doing.  Can you give detailed explanation of what I would do to see the clipping in action?

SiboVG:
In the current unstable, if you open 'A simple model rocket' and give the fin set an axial offset so that it goes beyond the body tube (the fin's parent component), you will see that the root edge of the fin is not drawn correctly (see picture below). This is because the right point of the root edge is clipped to the end of the body tube; it can't go further.
Screenshot 2021-09-16 at 02 27 15


This effect is especially visible if you increase the axial offset to an extreme. The code is written so that only the last point of the root edge is clipped, so the left point can go further than the body tube without clipping, while the "right" point is still stuck at the end of the body tube.
Screenshot 2021-09-16 at 02 27 59

hcraigmiller:
Using OpenRocket-1003-2.jar, no observable anomalies found as to fins attached to a body tube.

There continues to be an anomaly when fins are attached to transitions. Rather than referencing the end of the body tube as the set point when drawing the root edge beyond the body tube, **the first fin point is used**. So, you get A instead of B, as shown below.

0000 Fins

I am assuming that the same technique is being used when drawing the root edge beyond the body tube, it just doesn't create a visual anomaly because the root edge is flat instead of angled. 

**Should the transition anomaly be broken off as a separate issue, or is it sufficiently related to keep combined where the body tube and transition likely reference the first fin point, instead of the end of the component (body tube or transition in this case)?**

neilweinstock:
I would say that's a separate issue, and not a beta blocker.

PR #1017, opened 2021-09-12
[fixes #987] Use warning dialog for thrust curve import issues

### LINKED ISSUE #987: OpenRocket Crashes
I double-clicked the desktop shortcut, and a couple of seconds later, what you see in the screenshot is what happened.

I have attached the ORK file, user thrust curves and screenshot. The attached file is a 7zip file renamed to the .gif extension. I will be in and out this evening for about the next 4 hours. I am assuming that ya'll may want more info, I will get back with you.

JP

jpjohnston-openrocket-BUG-20210801 7z

SiboVG:
This PR fixes #987 where an issue in a user defined thrust curve file would throw an exception, thus crashing OR. This PR shows a warning dialog instead and ignores the faulty thrust curve file.

The warning dialog looks as follows:
image

Here is a [jar file](https://drive.google.com/file/d/1RGKhM87h_tnvKAoE2h72aREDLXgJyD-u/view?usp=sharing) for testing (note, you need to have [RATT.eng](https://github.com/openrocket/openrocket/files/7150168/RATT.eng.zip) in your user defined thrust curve directory to be able to test this issue, as stated in #987.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -792,6 +792,12 @@ FinSetConfig.lbl.relativeto = relative to
 !FinMarkingGuide
 FinMarkingGuide.lbl.Front = Front
 
+! MotorDatabaseLoaderDialog
+MotorDbLoaderDlg.title = Error upon thrust curve import
+MotorDbLoaderDlg.message1 = Thrust curve
+MotorDbLoaderDlg.message2 = will be ignored during import.
+MotorDbLoaderDlg.message3 = You can try replacing, deleting or manually editing the thrust curve file to fix this issue.
+
 ! MotorDatabaseLoadingDialog
 MotorDbLoadDlg.title = Loading motors
 MotorDbLoadDlg.Loadingmotors = Loading motors...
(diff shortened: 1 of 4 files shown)


JoePfeiffer:
Just tried this, with the RATT.eng file from above.  This leaves me with three popups open:  the "Loading motors..." popup's progress bar is oscillating, I've got a dialog showing progress on "Opening file" for my .ork, and I've got the warning dialog described above.  I'm only able to get focus on the "Opening file" dialog (so I can't press the "OK" on the warning dialog).

Once I press "Cancel" on the "Opening file" dialog I can press "OK" on the warning dialog; I now get a window for an empty rocket design, and the "Loading motors" popup is still oscillating.

I suspect an issue with window focus.  This is under Debian Linux

JoePfeiffer:
My apologies -- when I pushed that I was thinking it would land in my workspace, and I would construct a PR inviting you (@SiboVG ) to add it to your PR if you wished.  I didn't mean to step on your PR.  Let me know if you see a problem with it (or even if you see any new behavior, since you still weren't seeing the problem I had described!)

PR #1019, opened 2021-09-15
[fixes #136] Use warning dialog upon decal source file error

### LINKED ISSUE #136: Gentler errors for missing decals
Change this kind of error into a popup warning (so we don't get a bug for it).

Also, ensure that when saving, the ork file can be created with warnings if some decals are not found.

net.sf.openrocket.util.BugException: BUG: /home/user/Desktop/BodyStripe.png (No such file or directory)
    at net.sf.openrocket.document.DecalRegistry$DecalImageImpl.exportImage(DecalRegistry.java:174)
    at net.sf.openrocket.gui.ExportDecalDialog.export(ExportDecalDialog.java:103)
    at net.sf.openrocket.gui.ExportDecalDialog.access$300(ExportDecalDialog.java:28)

SiboVG:
This PR fixes #136 which was a request to have a sort of pop-up window show when a decal had a faulty or unexisting source file. In the current unstable issue, this would just create a bug and would even block you from saving your file.

Steps to reproduce the original problem:

1. Open 'A simple model rocket'
2. Go to 'File > Export Decal' and just save 'BodyStripe.png' somewhere
3. Next, delete the BodyStripe.png file that you just saved on your system
4. If you now again try to redo step 2, you would get a bug, also saving the OR file would create 

This PR fixes this issue by showing the following pop-up:
Yes-No Pop-up

After clicking 'Yes', the user can choose a replacement for the decal source file:
Replace source file

Also when saving, you will get the same pop-up window. If the user clicks 'No' in the first pop-up dialog or closes the FileChooser without selecting a file, instead of producing an error during the saving of the file, OR will just kinda ignore the faulty DecalImage, thus still enabling you to save your design.

From a functional view, this PR is relatively simple, from a codebase view it might seem more complex. This is because some steps needed to be taken to ensure that all the GUI-stuff went inside the swing module and the saving-part also took some more fiddling.

Here is a [jar file](https://drive.google.com/file/d/1ia42-3UZONTT56G1T10W0CF3DhVNYoCF/view?usp=sharing) for testing.


neilweinstock:
I'm a little confused by how this feature is even *supposed* to work.

My assumption was that exporting the decal very simply allowed one to save a copy of the decals, which are otherwise buried inside the ORK file.

But based on what your saying, it not only does that but it tells OR to look in that external location for the decal file in the future, rather than inside the ORK file.  Is that it?  If so, that is very weird and I wonder if *any* user would understand what it's doing.  Is that what users would *want* it to do?  It does seem to be the mode that Rocksim operates in by default (decals stored outside the RKT file).

SiboVG:
> But based on what your saying, it not only does that but it tells OR to look in that external location for the decal file in the future, rather than inside the ORK file. Is that it? If so, that is very weird and I wonder if _any_ user would understand what it's doing. Is that what users would _want_ it to do? It does seem to be the mode that Rocksim operates in by default (decals stored outside the RKT file).

I do believe that's indeed how it is programmed. It's not saving a copy, it's really moving the decal images to an external file location. Don't really know the logic behind it myself, but yeah

neilweinstock:
To me, that is not at all what the word "Export" means in common usage in software.  I object to this whole feature on principle. :)

But I also can't seem to reproduce it in 15.03.  I made a copy of the Simple Model Rocket example, and exported the body stripe decal to my Downloads folder.  Inside the ORK file, no visible changes were observed (including in the Appearance tab, where it still said the decal file was decals/BodyStripe.png.)  I renamed the file xBodyStripe.png, so the ORK file wouldn't find what it was looking for.  But when I saved and reopened the file, it worked fine, decal was still there, still showing, still presumably inside the ORK file.

How do I actually make this issue occur?

SiboVG:
Yeah so this is where things get a little funky. Say, you have 'A simple model rocket' open and you export decal 'decals/BodyStripe.png' to your downloads-folder. If you keep the design file open in OR and delete the exported BodyStripe.png, if you then again try to export 'decals/BodyStripe.png', then you will get the error that OR could not find the decal file (or in the case of this PR, you get a pop-up window to replace the file).

Now, if you try a different approach: again, open 'A simple model rocket', export the decal, now save your file to e.g. 'test.ork', and open 'test.ork' in a new window in OR. If you now again delete the exported 'BodyStripe.png' from your downloads-folder, OR doesn't give you the error because after saving and reloading the file, the exported decal is saved internally, in the ork-file.

So yes, I must admit, this is very unclear. In my next commit, I will remove the effect that OR picks the exported decal file as the new source file for that decal.

neilweinstock:
Although I agree with the change, given that this is a significant behavior change from 15.03 we should get some wider agreement on it.

JoePfeiffer:
This has been languishing since October.  I almost never use any of the 3D rendering features and have never put a decal on a rocket in OR so I don't feel I should be making comments or decisions.  I'm happy to merge it if other people like the PR.

SiboVG:
> This has been languishing since October. I almost never use any of the 3D rendering features and have never put a decal on a rocket in OR so I don't feel I should be making comments or decisions. I'm happy to merge it if other people like the PR.

Yeah, I tried to get a consensus a while back on Slack of whether the new decal behavior is good for merge, but never got a response. Either way, I think the behavior of this PR is the one that most users expect, so as far as @neilweinstock and I am concerned, this PR is good to go.

neilweinstock:
It's funny, I was just doing some work on a rocket and was thinking that I could make good use of the old export feature.... but I still believe 100% that that feature should not be called "export".  No idea what it should be called.

Would like input from Craig on this one (will ping him on Slack)

SiboVG:
@neilweinstock so you needed a way to externalize the decal image so that OR from then on always referenced the external decal image instead of storing it locally in the .ork-file and referencing it internally, right?

That seems like something that should be implemented in the Appearances-tab of a component; a way of the user selecting whether to use external referencing to the decal or not, like a checkbox or something.

neilweinstock:
> @neilweinstock so you needed a way to externalize the decal image so that OR from then on always referenced the external decal image instead of storing it locally in the .ork-file and referencing it internally, right?
> 
> That seems like something that should be implemented in the Appearances-tab of a component; a way of the user selecting whether to use external referencing to the decal or not, like a checkbox or something.

Ya, maybe.  I was iterating on decal design and what I ended up doing was writing a script to unzip the ORK file, insert the new decals, and re-zip it.  Not too bad actually, but that is the exact situation where the old export decal would have come in handy.

But I'm still not convinced the old functionality really makes sense, because it's too much effort to internalize the decals once they are finalized, and then you *don't* want them exported (or, you want to share the file).  So I'm not sure what the "right" way to do it really is.

hcraigmiller:
Frankly, I've never used the export decal feature because you can only export the first decal (of many) when exporting. My concern is also that, when exporting a decal image, it is not clear that you are doing away with the internal image, and I think it would be too easy for inexperienced users to accidentally lose/delete their decal images without knowing what happened. Can you explain the circumstances of when this feature would be used? The test .jar file does function without error, as you described what it should do.

neilweinstock:
Upon further reflection, I think there would be two possible use cases for the old export function.

1) Export the decal images for printing.  This seems like the most likely scenario for most people.  The fact that the old program only would export a single image at a time would be very annoying for this purpose.  I am thinking that this functionality is important, and if we remove it from one place (which I wholeheartedly support) then we should add it back elsewhere.  And it should support exporting all the decals at once.

2) Linking the decals to external files (what the old export actually does): this would be useful, on occasion, when iterating on the decal design.  The process can be cumbersome in the current OR right now, depending on how you do it.

I wonder how many folks have used the old functionality looking for (1), and getting (2) instead, maybe without even realizing it.  Impossible to know.

My thinking right at this moment is that we should support #1.  It's very simple.  A "File -> Export Decals" function (note the addition of the "s")  Select a folder and _all_ decals are saved into that folder.  No redirection of the decals within the rocket takes place.

No what we do, the change of functionality from 15.03 is going to bite someone.  I don't know how to avoid it.
Well I've never used it, but... my assumption is that you would do it when you want to be able to easily work on the decal designs outside of OR.  Image edit, save to file, see immediately in OR.  Without this, the workflow is tedious.
However, the way the feature works currently really makes no sense, and I have no idea how many people actually use it.  So repairing the meaning of what "Export" means is definitely good, the only question is whether the old functionality should be retained somewhere else.
I'll bet getting rid of it would affect very few people, although those people might be a bit angry...  There was an old issue (may still be open, dunno) where someone asked for the export function to apply to all decals, which suggests that at least someone is using it.
The other function that really is useful is the ability to actually export the decal images to files for printing.  That may be what the feature originally was for.  Maybe that is what we should replace.

hcraigmiller:
I've been playing with it, and the current iteration does neither in a meaningful or useful way.

1. I agree with the functionality you describe, but only if the internal graphic(s) are preserved. I believe that this is what users would expect.
2. I'm not seeing any actual linking. I exported a decal and modified it, and selected 3D view, but no change in the decal. I then exported a decal, modified it, then closed and reopened the design file and selected 3D, but no change in the decal.

It appears to me that the only thing that this feature is currently doing is blanking an exported decal if the file is deleted, but otherwise it is not doing anything.

I do support number 1 as the original intent, preserving the internal decal file.

SiboVG:
So the consensus is: implement this PR, but additionally, change the option of exporting each decal separately and instead export all the decals in one batch (e.g. save it as a zip-file or something)?

Maybe it's better to put the exporting of all decals as a new issue and merge this PR before the beta? Because exporting all decals is more of a new feature than something that needs fixing right now.

hcraigmiller:
I would have no objection to whatever others decide, but I believe that the internal decal file should be preserved.

hcraigmiller:
I agree that this should be a new feature issue.

SiboVG:
> I would have no objection to whatever others decide, but I believe that the internal decal file should be preserved.

This is the case with this PR, right? Or am I mistaken?

hcraigmiller:
Would you mind posting your current .jar for this issue. I want to re-test its function and I'm not sure I have the latest version.

SiboVG:
> jar file

Sure, [here it is](https://drive.google.com/file/d/17B7gxluezGkjmGHLB5OShKrOjC9btVtk/view?usp=sharing)

hcraigmiller:
Functions as expected, no anomalies found. It looks ready to merge to me.

PR #952, opened 2021-06-19
[fixes #927 & #771] Fix auto-run simulation NA values

SiboVG:
Okay, so this was a pretty deep problem to solve.

I noticed that the problem only occurred on velocity at deployment, ground hit velocity and optimum delay. After a bit of searching, I found that the simulation events of recovery device hit and ground hit in BasicSimulationEngine weren't getting triggered. To see what happens, I printed some log info about the simulation events when the problem arose and this was the output. The following shows in chronological order the different simulation events that occur after I have changed a parameter of a rocket component (event of altitude change not included, because there were a lot of them):
    Event - Launch
    Event - Motor ignition
    Event - Lift-off
    Event - Launch rod clearance
    Event - Motor burnout
    Event - Ejection charge
    Event - Apogee
    Event - Simulation end

For reference, this is the desired event log, from a normal simulation run:
    Event - Launch
    Event - Motor ignition
    Event - Lift-off
    Event - Launch rod clearance
    Event - Motor burnout
    Event - Ejection charge
    Event - Apogee
    Event - Simulation end
    Event - Recovery device deployment
    Event - Recovery device deployment
    Event - Ground hit
    Event - Simulation end

You can see that the simulation just ends after apogee. The culprit for this was the handleFlightEvent method in ApogeeEndListener. The ApogeeEndListener is instantiated in getExtraListeners in RocketPanel in the SimulationListener interface. This basically makes it so that when you change a parameter of the rocket and a simulation is auto run, when that simulation hits apogee, the next event of the simulation will be SIMULATION_END, as can be seen in ApogeeEndListener. This is not what we want, we want the simulation to stop after ground hit.

Therefor, I created a new listener, GroundHitListener, which will end the simulation after ground hit. This solves the issue.

I have to admit that I am not sure why the choice was made to stop the simulation after apogee in the first place, but I believe that ground hit should be the right way to end your simulation?

On top of that, I found a secondary problem. When you just changed a component value and change a value again, whilst the simulation is still calculating the first change (you can see it in the rocket preview window on the bottom left 'Calculating…'), then it also triggers a simulation end after apogee. This is a log of the events of first changing a component value and then doing it again during calculation. Note the 'simulation end' after apogee.
using pre-existing simulation
    Event - Launch
    Event - Motor ignition
    Event - Lift-off
    Event - Launch rod clearance
    using pre-existing simulation
    Event - Launch
    Event - Motor ignition
    Event - Lift-off
    Event - Launch rod clearance
    Event - Motor burnout
[...]

PR #1033, opened 2021-09-25
Remove debug printf's left over from PR 672

JoePfeiffer:
There was still some debug code left in #672; this removes it.


SiboVG:
The removal of the debug code is already fixed in #952. I replaced it with a log-message, but if you would like the debug text to be deleted all together, that's also possible.

JoePfeiffer:
Ah, OK -- upon reflection, log message is probably the better way to do it.

PR #1035, opened 2021-09-26
Use correct zero point for center of gravity calculation

### LINKED ISSUE #1034: Bug: Center of Gravity Issues when Position is Overridden
On the latest unstable the position of the center of gravity is wrong when the position of an individual component is overridden.

For example, here is the unmodified simple rocket example:
initial

After the mass of a part is greatly increased, the center of gravity of the rocket moves to the center of that part, as expected:
modified-mass

When the center of mass is modified, the center of gravity moves to the top of the body tube, which I believe is erroneous:
modified-both

Adjusting the position slider to the maximum moves the center of gravity of the rocket left by the length of the part (the expected behavior if zero was actually the top of the part instead of the body).

rsammelson:
Fixes #1034. The reference point when overriding the center of gravity is the front of the component.

### DIFF
--- a/core/src/net/sf/openrocket/masscalc/MassCalculation.java
+++ b/core/src/net/sf/openrocket/masscalc/MassCalculation.java
@@ -343,6 +343,10 @@ public class MassCalculation {
 			
 			// mass data for *this component only* in the rocket-frame
 			compCM = parentTransform.transform( compCM.add(component.getPosition()) );
+
+			// setting zero as the CG position means the top of the component, which is component.getPosition()
+			final Coordinate compZero = parentTransform.transform( component.getPosition() );
+
 			if (component.getOverrideSubcomponents()) {
 				if( component.isMassive() ){
 					// if this component mass, merge it in before overriding:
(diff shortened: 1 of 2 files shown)


rsammelson:
@neilweinstock @teyrana Are we sure that the value in the unit test is correct?

The bug I fixed was with the center of mass position, and that is the unit test that is failing. It seems like the test value might be wrong. I am expecting a small behavior change because the previous behavior was broken. If the test is the issue I can update it.

neilweinstock:
It's very possible the test value might be wrong, but I'll defer the question to @JoePfeiffer or @SiboVG.

SiboVG:
I'll have a look at it

SiboVG:
First of all, testing and the code of this PR LGTM.

The unit test fails here:
image

With this code:
```Java
final double expCMx = 1.0446069131149498;
Coordinate expCM = new Coordinate(expCMx, 0, 0, expMass);
assertEquals(" Booster Launch CM.x is incorrect: ", expCM.x, structure.getCM().x, EPSILON);
```

This is indeed a test of changing the CG override. Since I don't know how expCMx was input in the first place (looks like an empirical value) and the implementation of this PR looks about right, I would propose editing the unit test with setting expCMx to '1.1191303646438673'.

SiboVG:
There is an additional edit to the unit testing that should be made
image

In this code:
```Java
final double expMOI_tr = 0.036243133045;
double boosterMOI_tr = structure.getLongitudinalInertia();
assertEquals(" Booster transverse MOI is incorrect: ", expMOI_tr, boosterMOI_tr, EPSILON);
```

Change expMOI_tr to 0.040989095911.

After those two changes to the unit testing, the unit testing for this PR passed.

rsammelson:
It should be fixed now.

neilweinstock:
@SiboVG are we good to merge this one now?

SiboVG:
LGTM yeah

> @SiboVG are we good to merge this one now?

neilweinstock:
Thanks Robert!

PR #1057, opened 2021-11-04
[fixes #1044] Correct cdMach1 for ogive nose interpolator

### LINKED ISSUE #1044: calculateOgiveNoseInterpolator documentation mismatch?
According to the documentation, in order to calculate the nose pressure drag OpenRocket performs a polynomial interpolation for the [1.0, 1.3] Mach range, then finds the constants "a" and "b" for the subsonic region. From what I can see, ```PolyInterpolator``` takes a set of y values and then their derivatives. In this case, instead of passing the CD value for M = 1 and then passing the derivative for M = 1, the code seems to be using the derivative for M = 1.3. 

image
Here, the term in red should be equal to the CD at M = 1 (sine of epsilon)

```// In the range M = 1 ... 1.3 use polynomial approximation
		double cdMach1 = 2.1 * pow2(sinphi) + 0.6019 * sinphi;
		
		double[] poly = conicalPolyInterpolator.interpolator(
				1.0 * sinphi, cdMach1,
				4 / (GAMMA + 1) * (1 - 0.5 * cdMach1), -1.1341 * sinphi
				);
```
However, in here, the CD at mach 1 (cdMach1) is actually the CD at mach 1.3, at least according to the documentation. Therefore, the first value being passed as y to the interpolator is correct (1.0 * sinphi), but not its derivative (third parameter). Am I missing something?

SiboVG:
This PR fixes #1044, where in the calculateOgiveNoseInterpolator-method, cdMach1 was implemented as being the cd at Mach = 1.3, instead of the intended cd at Mach = 1.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/barrowman/SymmetricComponentCalc.java
+++ b/core/src/net/sf/openrocket/aerodynamics/barrowman/SymmetricComponentCalc.java
@@ -415,10 +415,11 @@ public class SymmetricComponentCalc extends RocketComponentCalc {
 		LinearInterpolator interpolator = new LinearInterpolator();
 		
 		// In the range M = 1 ... 1.3 use polynomial approximation
-		double cdMach1 = 2.1 * pow2(sinphi) + 0.6019 * sinphi;
+		double cdMach1 = sinphi;
+		double cdMach1_3 = 2.1 * pow2(sinphi) + 0.6019 * sinphi;
 		
 		double[] poly = conicalPolyInterpolator.interpolator(
-				1.0 * sinphi, cdMach1,
+				cdMach1, cdMach1_3,
 				4 / (GAMMA + 1) * (1 - 0.5 * cdMach1), -1.1341 * sinphi
 				);
 		


JoePfeiffer:
Hmmm....  I thought what we needed to pass as the first two parameters are Cd at Mach1 and Cd at Mach 1.3.  Doesn't this end up passing Cd at Mach 1 in twice?

Mudrost:
> Hmmm.... I thought what we needed to pass as the first two parameters are Cd at Mach1 and Cd at Mach 1.3. Doesn't this end up passing Cd at Mach 1 in twice?

Indeed. The first 2 parameters are Cd values, and the other two are their derivatives. I think having a cdMach1_3 or cdMach13 variable to store the old cdMach1 is the best approach for readability:
```
double cdMach1 = sinphi
double cdMach1_3 = 2.1 * pow2(sinphi) + 0.6019 * sinphi;
		
		double[] poly = conicalPolyInterpolator.interpolator(
				cdMach1, cdMach1_3,
				4 / (GAMMA + 1) * (1 - 0.5 * cdMach1), -1.1341 * sinphi
				);
```

SiboVG:
Whoops, sorry, was too hasty in the edit, will change it later

SiboVG:
Changed it in the last commit. Looking back at it, isn't this PR net exactly the same as the latest unstable? Because previously you just input 1.0 * sinphi as first parameter and as second parameter cdMach1, which was actually the cd at Mach 1.3?

Mudrost:
> Changed it in the last commit. Looking back at it, isn't this PR net exactly the same as the latest unstable? Because previously you just input 1.0 * sinphi as first parameter and as second parameter cdMach1, which was actually the cd at Mach 1.3?

Yep, but the third parameter which is supposed to be the derivative for M = 1 was using the cd value for M = 1.3 (since it is calling cdMach1).

SiboVG:
Ah, right, got it, thanks!

> > Changed it in the last commit. Looking back at it, isn't this PR net exactly the same as the latest unstable? Because previously you just input 1.0 * sinphi as first parameter and as second parameter cdMach1, which was actually the cd at Mach 1.3?
> 
> First two parameters are the same, but the third parameter which is supposed to be the derivative for M = 1 was using the cd value for M = 1.3 (since it is calling cdMach1).

PR #1064, opened 2021-12-09
Copy stageMap when loading a rocket's parameters from another rocket

### LINKED ISSUE #1007: Bug: NullPointerException after undoing 'Add Stage'
In the current unstable version, if you add a new stage and then undo add adding that new stage by hitting 'ctr/cmd + z' or going in 'Edit > Undo (Add Stage)', if you then try to do edit, delete or add any new component or stage, you get a NullPointerException.

Here is the full [Bug report](https://github.com/openrocket/openrocket/files/7094378/Bug.report.txt).

You can recreate the issue by making a new file, adding a stage, undoing adding the stage and then e.g. adding a new body tube, it will immediately produce the error. Also in existing file, like e.g. the 'A simple model rocket' template, adding a new stage and undoing it will produce the exception even without editing anything else.

JoePfeiffer:
Fixes #1007 

Also removed some superfluous copying of the rocket in OpenRocketDocument:undo()

### DIFF
--- a/core/src/net/sf/openrocket/document/OpenRocketDocument.java
+++ b/core/src/net/sf/openrocket/document/OpenRocketDocument.java
@@ -727,9 +727,7 @@ public class OpenRocketDocument implements ComponentChangeListener {
 		}
 		
 		rocket.checkComponentStructure();
-		undoHistory.get(undoPosition).checkComponentStructure();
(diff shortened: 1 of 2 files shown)


SiboVG:
This fixes the issue and good LGTM.

PR #1069, opened 2022-01-11
[fixes #997] Fix Photo Studio exhaust effects starting at top stage

### LINKED ISSUE #997: Bug: Photo Studio flame and spark Effect no longer display correctly in Windows 10 Pro
Beginning with OpenRocket-954-jkl1.jar dated 07/02/2021, and post 07/11/2021 merge, the Photo Studio flame and spark effects no longer display correctly in Windows 10 Pro.
0000 Photo Studio Flame

A similar error may be what is causing the smoke effect not to function as well (see Issue #874).

SiboVG:
This PR fixes #997. The issue was that the exhaust effects' start point was at the top stage, as can be [seen here](https://github.com/openrocket/openrocket/issues/997#issuecomment-1010332110), instead of on the bottom stage. The fix was very simple, because there was already logic implemented to use the bottom stage motor for the exhaust effects, but the problem was that there was a bottomStage.getStageNumber() method that was never assigned to the bottomStageNumber variable, causing it to always remain 0. This PR just simply assigns bottomStageNumber _et voila_.

Here is a [jar file](https://drive.google.com/file/d/1hB3dAN6-dkGASg09Qa_YFd-YLpUdWWq3/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoPanel.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/photo/PhotoPanel.java
@@ -422,7 +422,7 @@ public class PhotoPanel extends JPanel implements GLEventListener {
 		AxialStage bottomStage = configuration.getBottomStage();
 		int bottomStageNumber = 0;
 		if (bottomStage != null)
-			bottomStage.getStageNumber();
+			bottomStageNumber = bottomStage.getStageNumber();
 		//final int currentStageNumber = configuration.getActiveStages()[configuration.getActiveStages().length-1];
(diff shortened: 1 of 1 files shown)


hcraigmiller:
This test .jar file PR fixes the Flame and Sparks effects, but does not change the non-functioning smoke effect. I want to do some more in-depth testing on the smoke effect. In the mean time, I believe that this fix should be merged and that the smoke effect be made its own separate issue, tagged as a _release blocker_ and _not as a beta blocker_.

[Windows 11 Pro, Build 22000.376, Feature Experience Pack 1000.22000.376.0]

Issue #1076, opened 2022-01-25
"favorite" preset components dropdown doesn't update until edit window is reopened.


JoePfeiffer:
To duplidate:
1) Select a component to edit, and open the edit window
2) Open the dropdown in the upper right corner. Observe the selected presets
3) Select "From database..." to open the component preset chooser dialog
4) Change the set of selected components, either adding or deleting, or both
5) Click "OK" to close the dialog, and open the dropdown again. Observe that the selected presets is unchanged
6) Close and reopen the edit window.  When the dropdown is reopened, the changes have taken place.

hcraigmiller:
I am unable to reproduce this problem using _A simple model rocket_. If I change the nose cone to one from the database, the _Select preset_ window immediately changes to the part number of the database component when the _Okay_ button in pushed.

[Windows 11 Pro, Version 21H2, Build 22000.434, Windows Feature Experience Pack 1000.22000.434.0]

JoePfeiffer:
Here's what I'm seeing.

First photo, dropdown is open and I'm about to select "From database..."
![photo--1](https://user-images.githubusercontent.com/3065196/151008634-738ec444-3cfa-42ca-aecb-6ab31380a6de.png)

Second photo, I'm in the component preset choose dialog.  I've sorted by the first column, deselected, the BNC-20B, and selected the BNC20L (BNC-20B doesn't appear in the image due to sorting).
![photo-2](https://user-images.githubusercontent.com/3065196/151009058-1bd266d8-2c38-45db-84d7-dca45a2a7da5.png)

Third photo, I've clicked OK and reopened the dropdown. BNC-20B still appears, BNC20L does not.
![photo-3](https://user-images.githubusercontent.com/3065196/151009184-96015d5b-18d2-4234-87e6-fd8e46214fe7.png)

Fourth photo, I've closed and reopened the edit window, and opened the dropdown.  Correct presets are present.
![photo-4](https://user-images.githubusercontent.com/3065196/151009552-a6f12a23-0a06-44f1-9fa0-ddc68cfa27a4.png)

hcraigmiller:
Error replicated in Windows 11.

[Windows 11 Pro, Version 21H2, Build 22000.434, Windows Feature Experience Pack 1000.22000.434.0]

SiboVG:
[...]

Issue #1073, opened 2022-01-19
Get rid of serialization of parts and motor databases


JoePfeiffer:
Both these databases use Java's serialization to create files to import.  Oracle announced several years ago they intend to remove this feature (though I *think* it's actually around, even as late as Java 17).  From our perspective it just adds extra code to maintain, and makes manipulation of the databases unnecessarily difficult.  It needs to go.

SiboVG:
Motor database serialization already fixed by #2991. I would like to implement something similar for the parts database.

PR #1077, opened 2022-01-25
Fix 713: Include legacy and new databases

JoePfeiffer:
1) included Dave Cook's parts database repository as a subrepository at .../swing/resources-src/datafiles/components/.
2) moved .../presets to .../legacy_components
3) modified serialize-presets.java to read several directories instead of just one
4) modified jaxb code to use the units assigned to components (needed for Dave's database).
5) added an optional <Legacy /> element near the top of component files.  All of the components in a file marked as <Legacy /> are regarded as legacy components. Marked all of the legacy component files as <Legacy />
6) added a "Show Legacy Components" checkbox to the component preset chooser dialog.  When the checkbox is checked, legacy components are visible and a column identifying them as legacy is added to the chooser table.

Note:  there is an issue (#1076) with changing the selected "favorite" components not resulting in an immediate change in the dropdown.  That issue exists in the unstable branch already, so it's not a result of this PR.


JoePfeiffer:
Adding Dave's repository as a subrepository locally worked, but pushing my changes didn't push it as well.  Anybody with better knowledge of github than mine know how to get that to happen? If not I'll be looking at it in the morning...

JoePfeiffer:
jar file for testing [https://drive.google.com/file/d/1lIF-X7Mpi_n0ObdVOgx5mJNJhVI2szI3/view?usp=sharing](https://drive.google.com/file/d/1lIF-X7Mpi_n0ObdVOgx5mJNJhVI2szI3/view?usp=sharing)

hcraigmiller:
Is this intended:

- If legacy components are selected as presets, but show legacy components is false, the legacy components still show as presets.

No abnormalities found.

JoePfeiffer:
> Is this intended:
> 
>     * If legacy components are selected as presets, but show legacy components is false, the legacy components still show as presets.

Yes. It lets the user choose from either the current components or the current+legacy, but once chosen it doesn't change depending on whether both are shown in the chooser.  I expect if we want to change this behavior the right place will be in the fix for #1076

hcraigmiller:
Is this intended:

 - There are a host of missing legacy components, such as nose cones manufactured by Balsa Machine Service and Fliskits, and others, that are part of the 15.03 component database.

If so, all is working well for me.

hcraigmiller:
Never mind for now, the components that I was looking for are classified as legacy.

Question, would it be helpful for me to verify whether a legacy component is currently in production so that the classification can be changed, or does legacy have a different meaning here?

JoePfeiffer:
> Question, would it be helpful for me to verify whether a legacy component is currently in production so that the classification can be changed, or does legacy have a different meaning here?

Different meaning.  Legacy components are components from the legacy, ie 15.03, database (and they should all be in there!).  The non-legacy components are from Dave Cook's database which we are planning to use going forward.  The intent here is for users to migrate to the new database, but let designs developed in 15.03 not bring up a warning because of presets not being found.

hcraigmiller:
That being the case, all looks good to me.

JoePfeiffer:
Ah.  Just heard from Dave; it appears baking our old legacy database in with his database runs afoul of the GPL's requirements regarding compilations.  Taking this to draft status while I think about what to do.

hcraigmiller:
Essentially, there needs to be three different component libraries,
1. the primary component database (Dave Cook), 
2. the legacy component database, and 
3. the user's custom component database. 
And, the question is what approach to take to achieve a functional combining of these libraries.

The concern raised by Dave is likely that, under the language of the GPL, a compilation of these libraries into a single stored database could result in the loss of the Apache license restrictions on the use of his work. 

Based upon the foregoing, the question, then, is whether there is a method of merging/earmarking/sorting these libraries within the running program, but without the ability to save a single combined library. In other words, when the program is started, the three component libraries are combined (marking each component with the library type: "OpenRocket", "Legacy", "Custom") and then sorting for internal use as a single library. The OpenRocket Library (Dave's) and the Legacy library would be non-modifiable. The Custom library would be initialized, but empty until a component is added (and saved at that time). The libraries would never be aggregated in a way that could be saved.

Would this approach be feasible and acceptable to Dave?

JoePfeiffer:
The issue, as I understand it from Dave, is that the serialization (which combines the legacy and new component databases into a single file) becomes a compilation that runs into GPLv3 vs Apache2 problems.  The possible fixes appear to be either serialize into two files and read them both at startup time, or get rid of serialization.  For this release, and in particular for this Beta, I'm going to take a look and see whichever is easier and go with it.

I'm not really concerned about adding a user preset library at this point.  We'll want to make that possible, but again I don't see it as necessary for now.  If whatever I do about separating legacy and new libraries makes it easy to throw it in, I will.  There is no real way to edit, let alone save, a preset from inside OR.  A user preset library will be as awkward to create as a motor thrustcurve; I think we want to make both of these much easier -- eventually.

JoePfeiffer:
It was easy enough to just eliminate the serialization that I did it (so this is a partial resolution of #1073 ).  It turns out we are already able to load user component libraries!  Under Linux, they're in ~/.openrocket/Components/.  This location cannot be changed in our current preferences editor; I'll file an issue on that in a moment but I don't see it as something we need to do right away.

JoePfeiffer:
New jar for testing at https://drive.google.com/file/d/1cdLeBbnvxcvCCmW-7vDY3F4_ztYyqs-B/view?usp=sharing

hcraigmiller:
As to those components that do not encounter an uncaught exception, the database selection functionality worked very well. But, as you select "from database" in the configuration window, the following components encounter an uncaught exception:

- Tube fins
- Rail button
- Launch lug
- Inner tube
- Coupler
- Bulkhead
- Engine block
- Parachute
- Streamer

JoePfeiffer:
Oh, this is interesting....  it turns out there's been a bug in current unstable that would trigger that null pointer exception if you tried to enter any sort of text filter in the database for a bunch of component types -- quite possibly all the ones you identified above.  My changes have wound up triggering it without the need for a text filter.  I guess text filters aren't used much!

I expect to have a fix in a moment.

JoePfeiffer:
jar for testing https://drive.google.com/file/d/1E0b5ZSXnB2VyfFptaBmZt2fUEVsvhH0S/view?usp=sharing

hcraigmiller:
No abnormalities found. Looks good to me.

SiboVG:
Code LGTM, very welcome update, thanks @JoePfeiffer!

PR #1082, opened 2022-01-28
[fixes #825] Rewrite Software Updater

### LINKED ISSUE #825: Update checker still checking SourceForge
Noted that on startup, OpenRocket is still checking SourceForge for updates. The update URL is apparently specified in net.sf.openrocket/communication (not a Java coder; please forgive if this path to the trouble isn't well-specified).

I'm assuming it should be checking github.

Also noting that "sourceforge" appears a few times in the code elsewhere, and many times in "web" documentation that's in the repo.

```
public abstract class Communicator {

	protected static final String BUG_REPORT_URL;

	protected static final String UPDATE_INFO_URL;
	
	static {
		String url;
		url = System.getProperty("openrocket.debug.bugurl");
		if (url == null)
			url = "http://openrocket.sourceforge.net/actions/reportbug";
		BUG_REPORT_URL = url;
		
		url = System.getProperty("openrocket.debug.updateurl");
		if (url == null)
			url = "http://openrocket.sourceforge.net/actions/updates";
		UPDATE_INFO_URL = url;
	}

```

SiboVG:
This issue fixes #825 by rewriting OR's software update checker to use GitHub's release page.

**[Practical implementation]**
There are 4 different situations when checking for updates:

1. Something went wrong (e.g. bad internet connection) => a warning message will be shown:
Something went wrong

2. Nothing went wrong and the current build version is the latest version, then the user will get a popup like this:
Latest version

3. The current version is "newer" than the latest official release. This can be in the case of a beta version like we have now (the unstable build version is 20.11.alpha.16):
Newer version

4. The current version is older than the latest official release:
Older version

As you can see, the user gets to see all the release info from the GitHub release page, nicely formatted in HTML (for which I used the [commonmark-java library](https://github.com/commonmark/commonmark-java)). To facilitate downloading the new update, on the bottom you have the ability to download the new update for whatever operating system you're running:
Older version - platform choice

By default, OR will select your operating system as the update platform, but when you e.g. select 'JAR', it gets stored in the preferences. The different update platforms are by the way extracted from the different assets available on the GitHub release page.

**[Technical implementation]**
GitHub has an API for fetching the releases from a repository.  E.g. [this is the URL](https://api.github.com/repos/openrocket/openrocket/releases/latest) to get the latest release. So the release objects are formatted in a JSON object. The update checker just fetches those release objects from two different URLs, [this one](https://api.github.com/repos/openrocket/openrocket/releases/latest) (to get the latest release) and [this one](https://api.github.com/repos/openrocket/openrocket/releases) (to get all other releases - mostly not important to us, but can sometimes be important).

I also changed the build version formatting to use a dot-separator instead of a dash-separator. This is in line with how the releases on GitHub are formed. We should stick to one standard as to not overcomplicate the update checker.

There was a unit test from the previous update checker, UpdateInfoTest, which I for now just commented out instead of rewriting it to the new implementation. It would just have taken a lot of extra time to tie a unit test to the update checker, which I saved myself from for now by bypassing the unit test.

----

I have created 3 separate jar files for testing, each to test the different update scenarios:
- [OpenRocket-825-v14.03.jar](https://drive.google.com/file/d/1CcFeR9L-WY_4Zn6Bd5Fv1fVOTYF7bePl/view?usp=sharing) to test when an update is available (current build is older than latest release)
- [OpenRocket-825-v15.03.jar](https://drive.google.com/file/d/14Us_3TWR4VM99-tID7SS03AkEHpXCodw/view?usp=sharing) to test when the current build is the latest
- [OpenRocket-825-v20.11.alpha.16.jar](https://drive.google.com/file/d/1FsN8CmeSChPfsPR60MWYBG52xHHVNu3s/view?usp=sharing) to test when the current build is newer than the latest release


hcraigmiller:

JoePfeiffer:
OpenRocket-825-v20.11.alpha.16.jar to test when the current build is newer than the latest release
    works as expected (got time travel popup)

OpenRocket-825-v15.03.jar to test when the current build is the latest
    works as expected (no popup) under Java 8
    
OpenRocket-825-v14.03.jar to test when an update is available (current build is older than latest release)
    gave me a no such method exception (tried under both java 8 and java 11)
exeption

Does 14.03 require java 7?

JoePfeiffer:
My question wasn't whether 14.03 required a newer version, it was whether it required an older one.

15.03 didn't work on anything newer than 8 (and when I tried OpenRocket-825-v15.03.jar with 11 I got the exception there too; I got the expected behavior under 8).  I got the exception with OpenRocket-825-v14.03.jar running both 8 and 11. So I was wondering if 8 was too new for it.

We've been saying we're targeting 11 with the new release; my experience is it works on 11-14, 15 fails a unit test involving scripting but otherwise works, I've never tried 16 and other people are saying it doesn't work with 17.

SiboVG:
My bad, I thought the NoSuchMethodError was from an older Java version not yet having that method implemented. I'll take a look at it.

JoePfeiffer:
Hmmm, I'll recheck my work.  OK, here's the behavior of those old versions vs. Java version
Java     OR Result
     8  14.03 Exception
     8  15.03 Exception
   11  14.03 Exception
   11  15.03 OK

So I remembered 15.03's behavior with 8 vs. 11 exactly backwards (I've got no idea what 15.03 is doing running under 11, but that's unrelated!).

Stepping back a minute, does it matter if 14.03 and 15.03 are getting exceptions with the back-ported PR? We're never going to publicly release them anyway. Since it works against current unstable, should we call it OK?

hcraigmiller:
My vote is to call it okay.

neilbalch:
> > Does 14.03 require java 7?
> 
> Yes, I don't know exactly which version is needed, but older versions of Java are not supported. What is the Java version that the unstable build is supposed to ship with?

I'm pretty sure that a patch I submitted a while back makes it so that our project can be built and distributed with any of the newest JDK revisions. I've recently tested it with JDK 17.0.1 and had no issues. _(my vote is to ship with the newest possible JDK that is properly supported and tested)_

neilweinstock:
> _(my vote is to ship with the newest possible JDK that is properly supported and tested)_

We've been primarily focused on 11 up to this point, and it is solid.  Java 17 is the next LTS version, so it should be either 11 or 17.

If we're going to think about using 17, then we need to do more testing with it before going public beta, probably including create packaged installers for it.

17 has a greater support lifespan than 11, but other than that is there any real advantage for us?

neilbalch:
I think it's mostly just the lifespan. I'm also taking into mental account the possibility that the release following this one will likely take a long time to put together. It sometimes gets harder to find the older JDK versions on Oracle's site too, as [their page really tries to make you download JDK 17](https://www.oracle.com/java/technologies/downloads).

JoePfeiffer:
Something I *really* want us to do after this release is make a commitment to have small, reasonably frequent releases.  In particular, I'd like us to decide which features or issues we want to address, address them, and release.  The size of this release has nearly killed the project several times, and I don't want to see that happen again.

Until you said we now run under 17, my proposal for the next release would have been (1) get running on newest Java, (2) finish getting rid of serialization, (3) fix the 3D part selector.  If it turns out we can already run under 17 we can think of something else in its place.

hcraigmiller:
> Something I _really_ want us to do after this release is make a commitment to have small, reasonably frequent releases. In particular, I'd like us to decide which features or issues we want to address, address them, and release. The size of this release has nearly killed the project several times, and I don't want to see that happen again.
> 
> Until you said we now run under 17, my proposal for the next release would have been (1) get running on newest Java, (2) finish getting rid of serialization, (3) fix the 3D part selector. If it turns out we can already run under 17 we can think of something else in its place.

I agree with this approach. I can go through the outstanding issues and recommend an order for consideration, if that would be helpful.

neilweinstock:
More frequent smaller releases is more than just good, it is *necessary*.
Choice of which features to put in the next release.... for future
discussion. :)

I have not given up hope of getting the 3D part selection fixed for this
release, although that may be unlikely.  It's a pretty bad breakage,
although it is something that can be worked around.

Anyway... one step at a time.

PR #1083, opened 2022-01-28
[fixes #926] Fix motor selection window layout issues

### LINKED ISSUE #926: Motor selection dialog responsive layout issues
When the motor selection dialog is minimum size:
image

When the dialog is made taller:
image

Note that the "Select thrust curve" and "Ejection charge delay" pickers and the motor list are have spread apart from each other.  The way this should work is that those three elements remain close together (as in the first screenshot), and all addition window height is consumed by making the motor list are taller.  That is, it should look more like this:
image

Similarly, the Manufacturer list stops expanding after a while and the entire right column moves down from the top.  It seems to me like right column should stay in place (top-justified), and the Manufacturer list area should expand to consume all additional space.

SiboVG:
This PR fixes the layout issues in the motor selection window from #926. 

I also made it so that in the 'Show details' section, the motor comment text area takes over the rest of the horizontal space left. This has the side-effect that the labels on the top are more squished together than you may be used to. If you want me to add a little space between them, just let me know.
Screenshot 2022-01-28 at 01 59 32

Finally, I noticed that the 'Compatible Cases' label in the 'Show details' section did not have a colon on the end, which mildly infuriated me, so I fixed it here. 

Here is a [jar file](https://drive.google.com/file/d/1POL1Gdag57Xp5tQztmTM9NNuObRh_oo1/view?usp=sharing) for testing.

PS: For any future developer having to endure Java Swing MigLayouts, here are two very good resources:
http://www.miglayout.com/QuickStart.pdf and http://www.migcalendar.com/miglayout/mavensite/docs/cheatsheet.pdf.


hcraigmiller:

neilweinstock:
Screenshot looks great, I will update my version tonight.  One day I'll try messing around with the layouts, so I can do some of these little fixes myself.

SiboVG:
> Screenshot looks great, I will update my version tonight. One day I'll try messing around with the layouts, so I can do some of these little fixes myself.

Well prepare yourself for a lot of trial-and-error 😉 but once again, cannot recommend the two links I sent enough.

PR #1087, opened 2022-01-28
Improve Bug Report formatting

SiboVG:
This PR does not have an issue tied to it, but it is a quick fix to improving bug reports. This PR puts the lines "_Include detailed steps on how to trigger the bug:_" and "_What does the software do and what in your opinion should it do in the case described above:_" in bold and adds a line "_(You can edit text directly in this window)_" because I was shamefully unaware this was possible until recently.

Old bug report:
Screenshot 2022-01-28 at 22 55 58

New bug report:
Screenshot 2022-01-28 at 22 32 28

I thought this would be useful, because I have seen in the past that people just post their bug report without any indication of how the bug was created, after which we often have to close the issue due to lack of information. My goal with this PR is to guide users to elaborating their bug report.

Oh yeah and I've also edited the LogLine number formatting to use as default the English LOCALE. This makes it so that the timestamp number (the second number in the 'Error log' lines, e.g. for line '13 4.070 DEBUG (JarUtil.java:40) Found jar file using codeSource', '4.070' is the timestamp number) will be formatted the same for all users. I've encountered bug reports that had a dot as a decimal separator, but on my machine, a comma was used instead. The English LOCALE uses a dot separator by default.

Here is a [jar file](https://drive.google.com/file/d/1okdw97oTxezoTOpao0z8WLEnD6dfxBRD/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/BugReportDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/BugReportDialog.java
@@ -71,7 +71,7 @@ public class BugReportDialog extends JDialog {
 				  "gapleft para, split 2, gapright rel");
 		panel.add(new URLLabel(REPORT_EMAIL_URL, REPORT_EMAIL), "growx, wrap para");
 
-		final JEditorPane editorPane = new JEditorPane("text/html", message.replace("\n", "<br>"));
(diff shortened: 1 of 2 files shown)


hcraigmiller:
I like it.

PR #1094, opened 2022-01-29
[fixes #1092] Draw root edge of fin tab

### LINKED ISSUE #1092: Bug: Fin tab edge not drawn correctly
This is essentially the same issue as #1003, but now for the tab of a fin, instead of the fin itself. If the tab is further than the parent component, the right edge is not drawn correctly.

image

[EDIT] Well it's not exactly the same as #1003, because the root edge of a fin tab is just simply not drawn

SiboVG:
This PR fixes the root edge of the fin tab not being drawn, as stated in #1092. It's a simple fix: just append the start coordinate point of the fin tab to the coordinate list so that a final line is drawn from the last point to the starting point.

Result:
image

Here is a [jar file](https://drive.google.com/file/d/1E9MXXx5Beghgu31Zyjt2TjfjZc_IDJgm/view?usp=sharing) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -909,7 +909,7 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 			return new Coordinate[]{};
 		}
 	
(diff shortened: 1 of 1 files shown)


hcraigmiller:
No anomalies found.

PR #1097, opened 2022-01-30
[fixes #1088] Select new motor config + others

### LINKED ISSUE #1088: Newly created motor config still not automatically selected properly
This was fixed a long time ago, but it has come to my attention that it still doesn't work right.

After creating a new motor configuration, the motor is highlighted, but it is not properly "selected", e.g.:
image
None of the buttons at the bottom are enabled, so you still need to click on the new one before going in to select the motor.

After creating a new motor config, it should look like this:
image

As a bonus, I would love the "Select Motor" button to be the default action, so you'd only need to hit <return> to select motor after creating the new config.  Or just go and open the motor selection dialog immediately.  But it's possible that could be slightly confusing with multi-stage designs.  

Most important thing: properly auto-select the new config.

SiboVG:
This PR fixes #1088, with the following changes:

1. When creating a new configuration or copying an existing one, a selection event is triggered so that the 'Select motor' button etc. become enabled
2. When creating/copying a configuration, the Select motor dialog automatically opens. I found this was a good suggestion from @neilweinstock, as this is the action that you almost always want
3. When a motor configuration in the motor configuration table is selected, hitting the 'Enter' key opens the Select motor dialog. (be aware: e.g. just switching to the 'Motors & Configuration' tab and then hitting 'Enter' is not enough, you really have to click on the configuration in the table and only then is the Enter hotkey applicable)

Here is a [jar file](https://drive.google.com/file/d/1UwDfazRc6PjIsbYZAEkILyMNX2fM5CNu/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/flightconfigpanel/FlightConfigurationPanel.java
+++ b/swing/src/net/sf/openrocket/gui/main/flightconfigpanel/FlightConfigurationPanel.java
@@ -82,6 +82,7 @@ public class FlightConfigurationPanel extends JPanel implements StateChangeListe
 				motorConfigurationPanel.table.setColumnSelectionInterval(lastCol, lastCol);
 				configurationChanged(ComponentChangeEvent.MOTOR_CHANGE);
 				configurationChanged(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);	// Trigger select
+				motorConfigurationPanel.selectMotor();
 			}
 			
 		});
@@ -115,6 +116,7 @@ public class FlightConfigurationPanel extends JPanel implements StateChangeListe
 				addOrCopyConfiguration(true);
 				configurationChanged(ComponentChangeEvent.MOTOR_CHANGE);
 				configurationChanged(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);	// Trigger select
+				motorConfigurationPanel.selectMotor();
 			}
 		});
 		this.add(copyConfButton, "wrap");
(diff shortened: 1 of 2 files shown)


neilweinstock:
I officially love this.

It's a bit unclear when working with a multi-stage rocket, which motor gets asked for selection first... I was able to figure it out by moving the motor selection window out of the way so I could see which motor was highlighted in the main window.  To me this is a very small penalty for what is by far preferred functionality.

neilweinstock:
One question, and to be clear, I don't particularly care about this.  BUT: from a standpoint of UI correctness, if hitting "enter" triggers the "select motor" action, then shouldn't that button be highlighted to show that it is the default action?

hcraigmiller:
Confirm intent: When creating a new configuration, the Select motor dialog automatically opens **to the upper stage motor**.

No anomalies found; subject to comment of @neilweinstock concerning highlighting.

SiboVG:
@neilweinstock I wasn't able to highlight the 'Select motor' button. I tried, but if I set focus on that button, the focus from the table would be gone, causing the configuration to be deselected and causing a lot of other problems. I tried doing other stuff such as trying to just get the focussed look on the button without actually focussing it, but that didn't work out. So I guess we'll just leave it as is.

I have made some additional changes:
1. There was an issue that when switching over to the recovery or separation tab the recovery and separation buttons were disabled. This is now fixed.
2. When you now make a new/copy configuration, the dialog corresponding to the active tab opens up. So if you are in the motor tab, creating a new configuration will automatically open up the select motor dialog, in the recovery tab the deployment configuration dialog opens, and in the stage tab, the separation configuration dialog.
3. I also implemented the 'Enter' hotkey in the recovery and stages table

Here is a [jar file](https://drive.google.com/file/d/1MrSrtLLHsEIdsdDRVsIzUP7cN25cYqMo/view?usp=sharing) for testing.

neilweinstock:
I agree it is OK if the button doesn't highlight; it was worth a try but not worth going crazy.
One final suggestion: in a multistage rocket, you should initially select the first booster motor.  That is the motor that you would typically select first when putting together a motor configuration (at least that's the way I always do it).

SiboVG:
Alright, final commit select the booster motor by default instead of the upper stage motor.

Here is a [jar file](https://drive.google.com/file/d/16z9reilor3Q2mMwry2LXZgDxWiqkp3XD/view?usp=sharing) for testing.

hcraigmiller:
Nicely done... I think I like the lower stage motor default as well. 

No anomalies found.

neilweinstock:
Looks good.

JoePfeiffer:
A very welcome change!

PR #1102, opened 2022-01-31
[fixes #1024] Rocket optimization window too small + others

### LINKED ISSUE #1024: Bug: Rocket optimization window too small upon startup
In 'A simple model rocket', when opening the rocket optimization (application ribbon --> 'Tools > Rocket optimization'), the window height is initially so small that a large part of the panel cannot be seen:
Screenshot 2021-09-16 at 23 33 37
Screenshot 2021-09-16 at 23 34 28

This can be very confusing for new users that don't know that there is still more to see about the panel.

Tested on Mac OS Big Sur with the latest unstable release.

SiboVG:
This PR fixes issue #1024 where the Rocket optimization window was too small at startup, thus clipping of the bottom of the window. While I was at it, I also saw some layout issues in the custom expression window and component analysis window (there still are numerous after this PR, but those are headaches for the future).

Rocket optimization window:
Rocket optimization

Custom expressions window:
Custom expressions

Component analysis window:
Component analysis

Here is a [jar file](https://drive.google.com/file/d/1JDgxOBS2XKYja3hBk--91yNcaShL0Z_G/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/customexpression/CustomExpressionPanel.java
+++ b/swing/src/net/sf/openrocket/gui/customexpression/CustomExpressionPanel.java
@@ -56,7 +56,7 @@ public class CustomExpressionPanel extends JPanel {
 		//expressionSelectorPanel.add(scroll);
 		
 		//this.add(expressionSelectorPanel, "spany 1, height 10px, wmin 600lp, grow 100, gapright para");
-		this.add(scroll, "hmin 200lp, wmin 700lp, grow 100, wrap");
+		this.add(scroll, "hmin 200lp, wmin 700lp, grow 100, pushy, wrap");
 		
 		//DescriptionArea desc = new DescriptionArea(trans.get("customExpressionPanel.lbl.UpdateNote")+"\n\n"+trans.get("customExpressionPanel.lbl.CalcNote"), 8, -2f);
 		//desc.setViewportBorder(BorderFactory.createEmptyBorder());
--- a/swing/src/net/sf/openrocket/gui/customexpression/CustomExpressionPanel.java
+++ b/swing/src/net/sf/openrocket/gui/customexpression/CustomExpressionPanel.java
@@ -56,7 +56,7 @@ public class CustomExpressionPanel extends JPanel {
 		//expressionSelectorPanel.add(scroll);
 		
 		//this.add(expressionSelectorPanel, "spany 1, height 10px, wmin 600lp, grow 100, gapright para");
-		this.add(scroll, "hmin 200lp, wmin 700lp, grow 100, pushy, wrap");
+		this.add(scroll, "hmin 200lp, wmin 700lp, grow, pushy, wrap");
 		
 		//DescriptionArea desc = new DescriptionArea(trans.get("customExpressionPanel.lbl.UpdateNote")+"\n\n"+trans.get("customExpressionPanel.lbl.CalcNote"), 8, -2f);
 		//desc.setViewportBorder(BorderFactory.createEmptyBorder());
--- a/swing/src/net/sf/openrocket/gui/dialogs/ComponentAnalysisDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/ComponentAnalysisDialog.java
@@ -148,7 +148,7 @@ public class ComponentAnalysisDialog extends JDialog implements StateChangeListe
 		JScrollPane scrollPane = new JScrollPane(warningList);
 		////Warnings:
 		scrollPane.setBorder(BorderFactory.createTitledBorder(trans.get("componentanalysisdlg.TitledBorder.warnings")));
-		panel.add(scrollPane, "gap paragraph, spany 4, width 300lp!, growy 1, height :100lp:, wrap");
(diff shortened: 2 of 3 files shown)


hcraigmiller:
The following screenshots relate to Windows:

Optimum All

1. The changes to the optimization pop-up do not work well for Windows users. The pop-up is nearly as large as the full screen and would be as large as any smaller window. In addition, the selection buttons at the bottom right of the screen are nearly completely off of the screen.
2. There is no discernable change in the _Custom Expressions_ pop-up.
3. Opening the _Component Analysis_ pop-up caused the rocket to rotate (not just the fins), with no warning or way to undo the result; the number of degrees of the rotation varies.

As to all three, if the user adjustment of the pop-up placement and size could be preserved (when reopened and when the program is reopened) when the user exits the pop-up, that would minimize my objection to the changes.

As to the _Component Analysis_, rotating the rocket without notice and without a way to undo the change is unacceptable from a user point of view. Moreover, merely rotating the rocket should have absolutely no effect upon the analysis.

SiboVG:
> 1. The changes to the optimization pop-up do not work well for Windows users. The pop-up is nearly as large as the full screen and would be as large as any smaller window. In addition, the selection buttons at the bottom right of the screen are nearly completely off of the screen.

Looking at your screenshots, on Windows there was never the problem of the window being cut off on the bottom? 

> 2. There is no discernable change in the _Custom Expressions_ pop-up.

The changes are only visible when you enlarge the window, then you will see that the expression window takes up the rest of the space, without growing gaps between the buttons and the expression window

> 3. Opening the _Component Analysis_ pop-up caused the rocket to rotate (not just the fins), with no warning or way to undo the result; the number of degrees of the rotation varies.

That's very strange since I only applied UX changes... will look deeper into it

SiboVG:
@hcraigmiller I see that the wind direction in the component analysis window is changed to 48°, is this something you may have done by accident or did OR do this by itself (i.e. you just hit the 'Component analysis' button and didn't touch your computer thereafter)? Also sidenote: in the screenshot of the component analysis with the old behavior, the wind direction is also already set at 2°, instead of 0°, so the rocket also rotated with the old behavior, it's just less noticeable.

So a question: does the rotate also rotate in the current unstable when you change the wind direction? If so, and this is not the behavior that we want (I guess it's not?) then this is a pre-existing issue that should be addressed in a separate issue. Then I would just want to know whether the wind direction slider moves to 48° by itself if you don't touch your computer after hitting the component analysis button.

SiboVG:
@hcraigmiller is the problem 1 alleviated in [this test file](https://drive.google.com/file/d/1jFJItU-HuHwYzsimMzuMglqCxPfyNM6M/view?usp=sharing)? I set a maximum height on the window, based on the screen size + I put the whole Rocket Optimizer inside a scroll pane, because the optimizer window is large to begin with, so if users want to make it smaller, they can now without losing the bottom half of the window.
(I also put a minimum width on the description box underneath the 'parameters to optimize' text box, because it went very small in height when I scaled the window down)

hcraigmiller:
This much worse, at least before I could shrink the windows and still see all of the elements.

Optimum_Rev02

SiboVG:
*sigh* back to the drawing board

> Using 1020-1, isn't there a way to just have the pop-up be smaller when it opens?

Yeah I've tried that, but wasn't able to get it to work. Can try again later...

SiboVG:
> @hcraigmiller I see that the wind direction in the component analysis window is changed to 48°, is this something you may have done by accident or did OR do this by itself (i.e. you just hit the 'Component analysis' button and didn't touch your computer thereafter)? Also sidenote: in the screenshot of the component analysis with the old behavior, the wind direction is also already set at 2°, instead of 0°, so the rocket also rotated with the old behavior, it's just less noticeable.
> 
> So a question: does the rotate also rotate in the current unstable when you change the wind direction? If so, and this is not the behavior that we want (I guess it's not?) then this is a pre-existing issue that should be addressed in a separate issue. Then I would just want to know whether the wind direction slider moves to 48° by itself if you don't touch your computer after hitting the component analysis button.

@hcraigmiller do you have an update on this?

SiboVG:
Hm okay, does this jar fix the [custom expression](https://drive.google.com/file/d/1-r8vBZfDuPNUVt01HFQaHPxrPeBxNETO/view?usp=sharing) dialog?

SiboVG:
By the way, I have found why in the Component Analysis dialog, the wind direction jumps to a certain value, it's because the 'Worst' toggle button to the right of the slider is selected by default, which sets the wind direction to the worst possible value.

I think that's not a bad default behavior, but my concern is with whether we want the rocket to rotate in the rocket preview when you change the wind direction? If not, then I'd suggest to make this a new issue.

hcraigmiller:
Yes, the custom expressions box is now fixed.

I will open a new issue for the rotation during component analysis.

FYI, the test file has the fin tab position issue; the fin tab shifts position when the file is opened, but the current unstable does not,

SiboVG:
Okay, in that case I'm gonna merge this one personally because I have grown very sick of it 😅

And yes, the fin tab issue is solved, just has to do with the test file

PR #1106, opened 2022-02-06
[fixes #1055] Improve OR's simulation behavior

### LINKED ISSUE #1055: Weirdness with saved simulation results upon opening a file
Sorry I couldn't think of a more precise title for this one.  Testing done with Unstable build from 27-Oct-2021.

Consider the following file.
[alcubierre_cranberry.ork.zip](https://github.com/openrocket/openrocket/files/7464841/alcubierre_cranberry.ork.zip)
When I open it up and then switch to the "simulations" tab, all simulations run except the last one:
image
This looks a lot like #915 except I've only been able to reproduce it in this one instance.  I'm not sure why any of the simulations re-run automatically in this case, perhaps something is wrong with the saved results in the ORK file.  But it definitely shouldn't do what it's doing, which is run all the sims except one.

Now consider this one.
[Firetail_update.ork.zip](https://github.com/openrocket/openrocket/files/7464847/Firetail_update.ork.zip)
When you open it up and switch to the simulations tab, all three show yellow (saved results):
image
None of them auto-run.

Now notice that in the rocket display, the current sim results all show N/A, despite the fact that the A3-4 is selected, and that motor *does* have a saved sim result.
image
In fact, if I change the motor selection, the display results stay at N/A.
image

I continue to be confused by the behavior of saved simulation results, and don't really have a clear idea of how they *should* work.  What's seen here is definitely wrong, though.

Have fun @SiboVG. :)

SiboVG:
This PR improves simulation update handling of OR, as stated in #1055.

It fixes the following issues:

1. Previously, if you saved a file with the option 'Simulated data to store: Only primary figures', after reopening the file, the sims still got the status 'loaded' even though no sim data was stored. This also caused N/A values in the sim results, as shown in the examples of #1055 by @neilweinstock.
Now, the sims get the status 'not yet simulated' when you save it with 'only primary figures'
2. For some of the example designs _(A simple model rocket, Boosted Dart, Clustered rocket design, Hybrid rocket with dual parachute deployment and Three-stage rocket)_, there was an issue where if you loaded the file, went to the simulations tab and edited one of the simulations, the simulation status turned to 'out of date'. The same behavior occurred when you saved the file with 'only primary figures'. The problem was that the launch rod direction was wrong in the .ork files. All these files had the option 'Always launch directly up-wind or down-wind' enabled, which forces the launch rod direction to be the same as the wind direction. But in the mentioned example designs, the launch rod directions had a different value than the wind direction.
I just replaced the example files, but now with the correct rod direction. In a normal case, the launch rod direction is set to the right value and you don't have this issue, so just fixing the issue in the example .ork files should be sufficient.

It makes the following change:

3. The simulation status look 'loaded' is now replaced by the 'up to date' status look (both the simulation status icon and the tooltip text). The 'loaded' status offered no real beneficial information for a user and could even provoke confusion because a yellow symbol often denotes a warning and that something would be wrong with that simulation.
I left all the internals of the internal simulation status as is and instead only changed the look of the loaded simulation status to match that of the up to date status.

Here is a [jar file](https://drive.google.com/file/d/1wWsr1ZAcMJXWVWeBNXwTvW_1odVgjeKf/view?usp=sharing) for testing.


hcraigmiller:
I have not completed my testing, but two significant problems have appeared, one related to motors and one related to fins. 
1. When a new configuration is selected, the motor dialog box is no longer opened.
2. The fin tab position is changing.
In the design file attached, the fin tab is saved at -1.625 from the bottom, but, reopens at -22.844 from the bottom. If the tab position is saved at 7.531 from the center, it reopens at -3.078 from the center. _If the tab position is saved at 16.688 from the top, it reopens **correctly** at 16.688.

[1055_New.Fins_Reaper.001.Save.txt](https://github.com/openrocket/openrocket/files/8011045/1055_New.Fins_Reaper.001.Save.txt)

SiboVG:
> 1. When a new configuration is selected, the motor dialog box is no longer opened.

Apologies for that, that's simply because I hadn't rebased my branch with the new changes of unstable, so it's only in the test file that that issues arises because it uses an older version of unstable. Will do a rebase and update the JAR file.

SiboVG:
@hcraigmiller I have updated the JAR in the first comment with the latest version of unstable, so you should continue with that new jar for testing.

hcraigmiller:
Whatever you did, it fixed the fin issue as well. Wow.

neilweinstock:
In my quick tests, I am seeing all my simulations run as soon as I switch to the Simulations tab for the first time, so whatever state those sims are in after loading the file, I'm not seeing it for more than a split second.

SiboVG:
> In my quick tests, I am seeing all my simulations run as soon as I switch to the Simulations tab for the first time, so whatever state those sims are in after loading the file, I'm not seeing it for more than a split second.

Yes, that's because you have the option 'run all sims when in simulation tab' enabled in your preferences. If you disable this, you can see the previous state. Your mentioned behavior is not an issue, right?

hcraigmiller:
No anomalies found. As to the comment of @neilweinstock above, if I am watching the lower right flight data as the design loads, you can watch it go from NA to the correct data with the same speed as if a component change is made; it looks like normal behavior to me.

Well done.

hcraigmiller:
Is this the intended behavior?
If the run outdated preference is off, when a design is opened it updates the loaded motor, but not the others.

SiboVG:
> Is this the intended behavior? If the run outdated preference is off, when a design is opened it updates the loaded motor, but not the others. To me, this makes sense, although somewhat inconsistent with the setting.

Yes, that's intended as far as I know

neilweinstock:
> > In my quick tests, I am seeing all my simulations run as soon as I switch to the Simulations tab for the first time, so whatever state those sims are in after loading the file, I'm not seeing it for more than a split second.
> 
> Yes, that's because you have the option 'run all sims when in simulation tab' enabled in your preferences. If you disable this, you can see the previous state. Your mentioned behavior is not an issue, right?

Not an issue, that's how I *want* it to behave, and why saving the sim results at all makes no sense to me.  I can imagine, though, that for a high-flying rocket and/or one with many sims defined, it could take quite a while for them to run, and in those cases not running automatically could make sense.

I tried turning off the automatically run sims option, and the behavior I'm seeing with this build looks mostly good to me.  One question though: if I save the file with "Save primary figures" selected, when I reopen the file I see *no* figures.  Shouldn't I see *something*?

image

hcraigmiller:
Is this the intended behavior?
If the run outdated preference is off, when a design component is changed, only the loaded motor is updated, but not the others.
To me, this would be the expected behavior.

neilweinstock:
> Is this the intended behavior? If the run outdated preference is off, when a design is opened it updates the loaded motor, but not the others. To me, this makes sense, although somewhat inconsistent with the setting.

When the design view behavior was updated so that the currently selected motor config is re-simulated after each design change, we did not also add a separate option to control it.  And so, right now the "automatically run outdated sims" only applies to the Simulations tab, and the Rocket Design tab always auto-updates.  Personally I quite like the current behavior, and think that yet another checkbox would just add confusion.

But right now the preference says "Run out-of-date simulations when opening the Simulations tab".  To me, that is accurate and specific.

hcraigmiller:
> Personally I quite like the current behavior, and think that yet another checkbox would just add confusion.

I am in no way suggesting a change. I too like it the way it now works, I just wanted to make sure that what I was seeing was the intended function.

hcraigmiller:
No anomalies found, unless the functionality described above is not intended. Looks good to me.

neilweinstock:
Looks good to me as well.

hcraigmiller:
Another one bites the dust...

SiboVG:
Hm, I found a bug, when opening the example 'Simulation extensions and scripting' and running the sims, it throws an internal exception... Will take a look at it.

PR #1110, opened 2022-02-09
[fixes #1070] Add backward compatibility OR 15.03

### LINKED ISSUE #1070: Change <axial-offset> in ORK file back to <position>
As debated on Slack, here is my official request to change <axial-offset> back to <position> in the ORK file format.  This removes an unnecessary incompatibility between files saved by the old and new versions of OR.

As part of the fix, the new version should always _save_ using <position> but should _accept_ either <axial-offset> *or* <position> when reading files, so as not to break all the files that have been created with dev versions up until now.

SiboVG:
This PR fixes #1070 which was a call to increase backward compatibility to OR 15.03.

I added duplicate tags to the .ork file so that .ork files in the new release contain both the tags of the new release and tags of OR 15.03 (so that it can still read in the files). The duplicated tags:

- `axialoffset` (added the `position` tag)
- `angleoffset` (added `rotation` for FinSet and TubeFinSet and `radialdirection` for other components)
- `instancecount` (added `fincount` for FinSet and TubeFinSet)

radiusoffset and instanceseparation were no issue in the backward compatibility, as the features they are referencing to were not yet implemented in OR 15.03 (as far as I could see in the code, correct me if I'm wrong).

_I also added comments to delete the duplicate tags once we want to break compatibility with 15.03, which I think we should at one point, because having those duplicate tags permanently is messy._

Here is a [jar file](https://drive.google.com/file/d/1byA0O4mQYW-kocaqmRObM6tqUOBz5i04/view?usp=sharing) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
+++ b/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
@@ -83,6 +83,10 @@ public class RocketComponentSaver {
 				// no-op.  Instance counts are set via named cluster configurations
 			} else {
 				emitInteger(elements, "instancecount", c.getInstanceCount());
+				// TODO: delete this when no backward compatibility with OR 15.03 is needed anymore
+				if (c instanceof FinSet || c instanceof TubeFinSet) {
+					emitInteger(elements, "fincount", c.getInstanceCount());
+				}
 			}
 			
 			if (c instanceof LineInstanceable) {
@@ -103,6 +107,13 @@ public class RocketComponentSaver {
 			final String angleMethod = anglePos.getAngleMethod().name().toLowerCase(Locale.ENGLISH);
 			final double angleOffset = anglePos.getAngleOffset()*180.0/Math.PI;
 			elements.add("<angleoffset method=\"" + angleMethod + "\">" + angleOffset + "</angleoffset>");
+			// TODO: delete this when no backward compatibility with OR 15.03 is needed anymore
+			if (c instanceof FinSet || c instanceof TubeFinSet) {
+				elements.add("<rotation>" + angleOffset + "</rotation>");
+			}
+			else {
+				elements.add("<radialdirection>" + angleOffset + "</radialdirection>");
+			}
 		}
 		
 		// Save position unless "AFTER"
@@ -110,6 +121,8 @@ public class RocketComponentSaver {
 			// The type names are currently equivalent to the enum names except for case.
 			String axialMethod = c.getAxialMethod().name().toLowerCase(Locale.ENGLISH);
 			elements.add("<axialoffset method=\"" + axialMethod + "\">" + c.getAxialOffset() + "</axialoffset>");
+			// TODO: delete this when no backward compatibility with OR 15.03 is needed anymore
+			elements.add("<position type=\"" + axialMethod + "\">" + c.getAxialOffset() + "</position>");
 		}
 		
 		// Overrides


hcraigmiller:
If I open a design with rail buttons (with the test file), save it under a different name, then reopen it, I get an error "Unknown parameter type 'radialdirection' for Rail Button, ignoring". If I click okay, the design appears complete with no anomalies.

Here is the design file.
[1055.FinTab_Reaper.001_Save.Rev_001.txt](https://github.com/openrocket/openrocket/files/8029238/1055.FinTab_Reaper.001_Save.Rev_001.txt)

After multiple saves, it threw this error:

Backward_01

SiboVG:
Right, good catch @hcraigmiller, should be fixed in the latest commit.

Here is a [jar file](https://drive.google.com/file/d/1emrNJ0iLEYXaxVSn4EFOTD-k2NjBSPBC/view?usp=sharing) for testing.

hcraigmiller:
Abnormal fin tab behavior.

1. Saved design file using test jar (**saved fin tabs relative to bottom -1.625**): 
[1055.FinTab_Reaper.001_Save.Backward_001.txt](https://github.com/openrocket/openrocket/files/8034197/1055.FinTab_Reaper.001_Save.Backward_001.txt)

2. Opened in 15.03. Errors displayed.
Backward_02 error

Motor removed (normal). Fin tabs had moved forward, **opened relative to midpoint -1.625** (abnormal).
Saved design file using 15.03 (saved fin tabs relative to midpoint). 
[1055.FinTab_Reaper.001_Save.Backward_002.txt](https://github.com/openrocket/openrocket/files/8034380/1055.FinTab_Reaper.001_Save.Backward_002.txt)

3. Opened in test jar. Fin tabs stayed the same (**opened relative to midpoint -1.625**).

Backward_03 error

4. In 15.03, repositioned fin tab relative to bottom -1.625.
5. Opened in test jar, fin tab remained relative to bottom -1.625 (normal).

SiboVG:
Right, latest commit adds backward compatibility for the fintab positions, **except for 'Tip of the nose cone' and 'After the sibling component' as reference points**, because those aren't available in OR 15.03 and trying to port it to 15.03 caused a whole lot of other issues... So you **will** get issues when saving a file with either one of those two reference points, we simply can't ensure full backward compatibility on that part.

Here is a [jar file](https://drive.google.com/file/d/1ZY4u9vuf8ujM00U4gGQ1IZhGQ4CDdpFl/view?usp=sharing) for testing.

hcraigmiller:
> Right, latest commit adds backward compatibility for the fintab positions, except for 'Tip of the nose cone' and 'After the sibling component' as reference points, ... So you will get issues when saving a file with either one of those two reference points.

Got it. I will add that to the known issues list.

I have a design file that will not open in 15.03 (it warns then goes to a blank design); it uses components and positions not available in 15.03. Is that normal behavior or should it always convert some parts.

Backward_01

I tried multiple designs and, to the extent that it uses components and features that 15.03 has, the backward compatibility appears fully functional (both backward then forward). Designs that use features not available in 15.03 vary in backward compatibility from clipping or repositioning parts, to not opening the design file. If this is the expected behavior, then no anomalies found; otherwise, I will create and post the design files.

SiboVG:
> I have a design file that will not open in 15.03 (it warns then goes to a blank design); it uses components and positions not available in 15.03. Is that normal behavior or should it always convert some parts.

Hm, I think that if it's to the point that it just loads a blank design, it's a more fundamental backward compatibility break that we shouldn't focus our time and resources to :/

hcraigmiller:
I agree. As long as the same features are backward compatible from current to 15.03, I think it's good.

neilweinstock:
Does what I expect in my limited testing.

hcraigmiller:
I noticed another fin issue. Saving a 4-fin design in the current unstable, when opened in 15.03, the design only has 3 fins.

neilweinstock:
How come I got an email notification regarding a comment from @hcraigmiller on 3/4 fin behavior, but it doesn't show up here?  Was it deleted?

SiboVG:
> How come I got an email notification regarding a comment from @hcraigmiller on 3/4 fin behavior, but it doesn't show up here? Was it deleted?

I hope so, less work for me hehe

hcraigmiller:
> How come I got an email notification regarding a comment from @hcraigmiller on 3/4 fin behavior, but it doesn't show up here? Was it deleted?

Yes, it was deleted; used error.

SiboVG:
So it's ready for a code review?

hcraigmiller:
I believe so.

JoePfeiffer:
LGTM

PR #1116, opened 2022-02-10
Fix layout issues license dialog

SiboVG:
This PR fixes some layout issues that are present in the new license dialog. It concerns changing the font to match the rest of OR, making the text pane more response when enlarging the dialog, moving the Close-button to the bottom right, centering the title label even when resizing the window and making the dialog a bit bigger by default.

Screenshot 2022-02-10 at 17 54 36

Whilst working on this issue, I also noticed some weirdness in the new bug report layout. I had changed the layout to HTML in a previous PR, but this caused weirdness when one of the bug report lines contained HTML-like text. In the image below you can see e.g. stack trace line '`> at java.desktop/javax.swing.JScrollPane.<init> (JScrollPane.java:306)`' that contains the text '`<init>`'. The HTML formatter interpreted this as an HTML tag, causing this awkward layout. This PR makes sure that bug report lines (stack trace, error log etc.) are not formatted like HTML, but as plaintext.
Screenshot 2022-02-10 at 18 04 09

Here is a [jar file](https://drive.google.com/file/d/1FUrepZArcyhUy7sA8Bf11IclH_O-0-FN/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/components/DescriptionArea.java
+++ b/swing/src/net/sf/openrocket/gui/components/DescriptionArea.java
@@ -12,18 +12,15 @@ import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
-import java.io.IOException;
 
+import javax.swing.JTextPane;
 import javax.swing.event.HyperlinkEvent;
 import javax.swing.event.HyperlinkListener;
 import javax.swing.JEditorPane;
-import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.ScrollPaneConstants;
 import javax.swing.SwingUtilities;
-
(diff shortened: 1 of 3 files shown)


hcraigmiller:
As to the License dialog, I like the bigger box, but the text is now too way small.
License

The text in the bug report is a little small, but I can live with it there.

SiboVG:
Hm, does [this jar](https://drive.google.com/file/d/1XPQdqw97MqZdE3C6ChdYm__s2Q2R8vMQ/view?usp=sharing) fix the font size issue @hcraigmiller?

hcraigmiller:
> Hm, does [this jar](https://drive.google.com/file/d/1XPQdqw97MqZdE3C6ChdYm__s2Q2R8vMQ/view?usp=sharing) fix the font size issue @hcraigmiller?

Yes, much better:
License_2

Looks good to me.

BadDadio:
Unsubscribe

PR #1118, opened 2022-02-11
Fix about dialog layout issues

SiboVG:
This PR fixes some small layout issues in the about dialog. It just bothered me that you couldn't rescale the window and that the font was not the same as the rest of OR on Mac. Those two issues are now fixed + I made the dialog a bit bigger.
combined

Here is a [jar file](https://drive.google.com/file/d/1ud-aCOvR0lOmKwe424wz0yjZikY0X7HH/view?usp=sharing) for testing.


hcraigmiller:
Looks great, no anomalies found.

PR #1120, opened 2022-02-11
[fixes #1119] Fix rocket rotation after component analysis

### LINKED ISSUE #1119: [Bug] Rocket rotation does not reset in the preview when you exit the Component analysis dialog
When the _Component analysis_ dialog box is open, adjustments in the wind direction cause the rocket to rotate in the preview. But, unlike adjustments that affect CP (which reset on exit), the rotation does not reset on exit, resulting in the preview rotation being different that the rotation shown on the left side bar.

When closing the _Component analysis_ dialog box, all adjustments should reset, including the rocket rotation shown in the preview.

SiboVG:
This PR fixes #1119 by switching back the rocket's orientation after closing the component analysis window.

Here is a [jar file](https://drive.google.com/file/d/15myAqcD7KlMq22n3eX8GXJIHoSp3l0By/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/ComponentAnalysisDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/ComponentAnalysisDialog.java
@@ -11,6 +11,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.awt.event.WindowListener;
 import java.util.ArrayList;
 import java.util.EventObject;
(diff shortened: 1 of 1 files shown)


hcraigmiller:
New problem. When starting, Component analysis dialog no longer initializes to worst case wind, and the worst button is inoperable.

SiboVG:
Could you elaborate @hcraigmiller? Because it works on my machine

hcraigmiller:
I redownloaded and ran again, no anomalies found. Works as expected (the difference in the "Worst" button may be a little subtle, but whatever is good for you is fine with me).

Component

PR #1124, opened 2022-02-12
[fixes #311] Fix undo selection after component edit

### LINKED ISSUE #311: Component selection removed when editing (regression)
When you edit any component, the component selection is removed.  You can still edit the component, but no component is no longer selected.

This bug occurs on 15.03 and 14.11, but works correctly on 13.09.1.  I haven't binary-searched the releases in between.

Steps to reproduce:

1. Open "A simple model rocket" example design
2. Double-click the body tube.  -->  The component is selected and edit dialog opens
3. Change the body tube length

What happens:  The component length changes and you can continue editing it, but the component is no longer selected.
What is expected:  The component selection remains.

SiboVG:
This PR fixes #311 where the selected component in the component tree would deselect after editing it + newly added components would not be selected by default.

I believe the error laid in `ComponentTreeModel.java` where in the `componentChanged()` method the component tree's structure was updated after a mass change event (so when a component was edited/added). A tree structure change triggered the `clearSelection()` method in the `TreeSelectionModel`. The mass change condition triggering a tree structure change doesn't seem logical in the first place since a mass change does not affect the tree structure, so I just removed the `e.isMassChange()` condition for the tree structure change.

Here is a [jar file](https://drive.google.com/file/d/1S60ITDdWgnnDp9Lo0EFOP5bIY3mk2GXY/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/componenttree/ComponentTreeModel.java
+++ b/swing/src/net/sf/openrocket/gui/main/componenttree/ComponentTreeModel.java
@@ -138,7 +138,7 @@ public class ComponentTreeModel implements TreeModel, ComponentChangeListener {
 	
 	@Override
 	public void componentChanged(ComponentChangeEvent e) {
-		if (e.isTreeChange() || e.isUndoChange() || e.isMassChange()) {
+		if (e.isTreeChange() || e.isUndoChange()) {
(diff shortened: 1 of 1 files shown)


hcraigmiller:

PR #1128, opened 2022-02-13
Add "package" build target

JoePfeiffer:
This adds a new ant build target, "package".  This target builds a .exe (on Windows), .dmg (on Mac), or appImage (on Linux), in the toplevel directory of the source tree.  This is now the default build target (so just 'ant' will build the appropriate package).  The jar file build along the way to the package is left in swing/build/jar/OpenRocket.jar, as before.  Alternatively, it is possible to run 'ant jar' and just create the jar file.

The Linux appImage target requires an extra program, appimagetool.AppImage, to execute.  Attempting the build under Linux without obtaining the the tool will cause instructions for obtaining it to be printed.

It has only been tested under Linux so far.  An appimage created using the tool can be found at https://drive.google.com/file/d/1oEDHbFWmM_SvQMGaSkUZ9qBmotNYerVN/view?usp=sharing

The Windows and Mac versions should (I hope!) be ready to go.  The Linux version requires some extra massaging of some metadata, but I wanted to make it available for testing early.


neilweinstock:
Is jpackage being used for all targets?  we should set file-associations to ORK at least, and... maybe RKT?  That's desirable for those who don't also have Rocksim, very *undesirable* for those who do.

Also: jpackage says that Windows and Mac targets must be generated on the appropriate systems.  How are we going to manage that?

@itshanney 's repo (https://github.com/itshanney/openrocket-installer-files) has config files for Install4j, generated Windows and Mac versions. The tool supports Linux, but I don't know details.  I am also not sure how the Install4j packages compare to those of jpackage, or whether it can generate them all from Linux, or if it too needs each target to be generated on the same OS.

Install4j *does* have some autoupdate capabilities, but again I have no idea of the details.  I only know that it was the tool that Justin had worked with, and I've browsed the product pages.

neilweinstock:
I vaguely recall it had a free license for open source projects, or something like that.  Need to check back on the details.

JoePfeiffer:
You're right:  https://www.ej-technologies.com/buy/install4j/openSource

JoePfeiffer:
Yes, with Justin's install4j configuration out there, this isn't needed

PR #1131, opened 2022-02-14
Update version string to 22.xx.BETA, add Justin Hanney to developers list in About dialog

JoePfeiffer:
Test jar at https://drive.google.com/file/d/1k5IGTelyPsUqeVsYCRTRWrrg1DRVQ8Y7/view?usp=sharing

### DIFF
--- a/core/resources/build.properties
+++ b/core/resources/build.properties
@@ -1,6 +1,6 @@
 
 # The OpenRocket build version
-build.version=22.xx.beta.01
+build.version=22.00.beta.01
 
 # The copyright year for the build. Displayed in the about dialog.
 # Will show as Copyright 2013-${build.copyright}
--- a/core/resources/build.properties
+++ b/core/resources/build.properties
@@ -1,6 +1,6 @@
 
 # The OpenRocket build version
-build.version=22.xx.BETA
+build.version=22.xx.beta.01
 
 # The copyright year for the build. Displayed in the about dialog.
 # Will show as Copyright 2013-${build.copyright}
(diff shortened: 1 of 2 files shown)


hcraigmiller:

core/resources/build.properties
```diff
@@ -1,6 +1,6 @@
 
 # The OpenRocket build version
-build.version=20.11.alpha.16
+build.version=22.xx.BETA
```
itshanney, line 3: Might as well commit to 22.0 as the release number. The .xx is too ambiguous, so hammer it home :)

JoePfeiffer: The plan is to replace the xx with the month number when this release is final.

SiboVG: Also: don't use capitals for beta, but make it e.g. 22.02.beta.01, so it's in sync with how the software update checker works

JoePfeiffer:
Wait a minute -- will the updater recognize version 22.04 is later than 22.xx.beta.01? Letters come after numbers in the sort order.

SiboVG:
> Wait a minute -- will the updater recognize version 22.04 is later than 22.xx.beta.01? Letters come after numbers in the sort order.

If xx is smaller than than or equal to 04, then yes, 22.04 should be recognized as being later 22.xx.beta.01

JoePfeiffer:
I meant literally the string "xx", as it currently appears in the version string.

SiboVG:
> I meant literally the string "xx", as it currently appears in the version string.

Then it would simply give a warning dialog that it cannot recognized the version tag, because 'xx' is an unrecognized string.

Issue #1101, opened 2022-01-30
[Feature request] Select motor mount when selecting motor configuration


SiboVG:
It could be visually useful that when you select a motor configuration in the 'Motors & Configuration > 'Motors' window, that the corresponding motor mount is selected in the rocket view on the bottom. This is especially helpful for multi-stage rockets.

Expected result:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/11031519/151712369-0cc06a0d-300b-49ce-8fb7-4af940de3f5a.gif)

Issue #358, opened 2017-06-30
Allow multi-selection of rocket components


SpamWall:
Allow multiple rocket components to be selected in the rocket component Tree View and the graphic rocket view for scaling, deleting, etc. by holding down the control key while selecting.

PR #1132, opened 2022-02-14
[fixes #244] Multi-selection in Motors, Recovery & Stages panels

### LINKED ISSUE #244: Multi-Select Motor & Configurations table for Removal/Deletion

SiboVG:
This PR fixes #244 and adds the ability to have multi-selections in the motors, recovery and stage panels. You can perform almost every action on multiple cells: selecting a motor (selects the same motor for all selected cells), remove motor, select/reset ignition (again uniform ignition selection for all selected cells), rename configuration, remove configuration, copy configuration, select/reset deployment, and select/reset separation.

(You can do multi-selection with holding the shift-key and clicking or ctrl/command + click)

Demo:

https://user-images.githubusercontent.com/11031519/153920756-fb7369cb-98e8-4221-8d93-d44fd5d9b82a.mp4

Here is a [jar file](https://drive.google.com/file/d/14iyFTu54tBjeri9CA8ULhI3O62T__Qro/view?usp=sharing) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
+++ b/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
@@ -1694,6 +1694,11 @@ public class BasicFrame extends JFrame {
 		this.selectionModel.setSelectedComponent(component);
 	}
 
+	public void setSelectedComponents(List<RocketComponent> components) {
+		this.selectionModel.setSelectedComponents(components);
+	}
+
+
 	public void stateChanged(ChangeEvent e) {
 		JTabbedPane tabSource = (JTabbedPane) e.getSource();
 		int tab = tabSource.getSelectedIndex();
--- a/swing/src/net/sf/openrocket/gui/main/DocumentSelectionModel.java
+++ b/swing/src/net/sf/openrocket/gui/main/DocumentSelectionModel.java
@@ -2,6 +2,7 @@ package net.sf.openrocket.gui.main;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 
 import javax.swing.ListSelectionModel;
@@ -29,7 +30,7 @@ public class DocumentSelectionModel {
 	
 	private final OpenRocketDocument document;
 	
-	private RocketComponent componentSelection = null;
+	private final List<RocketComponent> componentSelection = new LinkedList<>();
 	private Simulation[] simulationSelection = NO_SIMULATION;
 
 	private TreeSelectionModel componentTreeSelectionModel = null; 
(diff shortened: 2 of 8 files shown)


hcraigmiller:
Very nicely done, but the Issue #1101 functionality has been lost.

SiboVG:
> Very nicely done, but the Issue #1101 functionality has been lost. If the two are incompatible, I prefer this functionality. [Windows 11 Pro, Version 21H2, Build 22000.434, Windows Feature Experience Pack 1000.22000.434.0]

Yes this is because I worked on this PR before the merge with unstable.

Here is a [corrected jar file](https://drive.google.com/file/d/1qk47m7EKUCT0ug2G2GngrA5k5uEecYBk/view?usp=sharing) for testing. Do note: it only selects a single motor mount, even though you can have multiple selected. In the latest commit, I have implemented the code for allowing multiple motor mount selection, but this does not yet have an effect because multi-component in selection in the component tree is not yet supported. Once #358 is implemented, multiple selections for the motor mounts will automatically take its effect.

hcraigmiller:
Looks good, functions as described, no anomalies found.

neilweinstock:
It appears that the highlight thickens the line in both directions.  The "inner" thickening is almost completely masked by the motor outline.  Looking at the picture I attached above, it's *really* hard to tell which one is highlighted.  Here's the same image with no motors; here it's very easy to see:
image

I can think of three possible approaches:
1) Ignore it.  What you have there is still better than nothing.
2) Make the highlighting more prominent for all components, all the time.
3) Make the highlighting more prominent just for motor mount tubes when motors are present.

#3 is probably best, but may not be practical.  #2 is pretty easy: just make the outline thicker when selected.  Whatever it is now, add more to it.  You'd have to see how it looks, both on the MMT and on other components.  

If you point me to the code that does the highlighting, I can play with it myself as well.

SiboVG:
You can change the global variable `SELECTED_WIDTH` on line 64 of `swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java`. I found that setting it to 2.5 give selecting a component a nice contrast, better than with the original 2.0 value. Setting it to 3.0 looked a bit cartoonish IMO.

Maybe it's also an idea to lighten the color of a motor? A bit lighter grey so that the contrast with the motor mount line is better.

neilweinstock:
> You can change the global variable `SELECTED_WIDTH` on line 64 of `swing/src/net/sf/openrocket/gui/scalefigure/RocketFigure.java`. I found that setting it to 2.5 give selecting a component a nice contrast, better than with the original 2.0 value. Setting it to 3.0 looked a bit cartoonish IMO.

I'll play with it tonight.
 
> Maybe it's also an idea to lighten the color of a motor? A bit lighter grey so that the contrast with the motor mount line is better.
I think the bigger problem (at least as far as this PR goes) is the black outline of the motor, which swallows up the "inner" portion of the Selected highlight.  I'm not sure what would be a good alternative there.

neilweinstock:
2.5 is an improvement, but I gotta say I like it just fine set at 3, don't find it cartoonish at all.  I'm a fan of making things you're supposed to see *easy* to see.
image
image

hcraigmiller:
> But I like the selected weight setting at 3. @hcraigmiller what do you think?

The increased line thickness makes it more difficult to do fine adjustments without having to deselect the component to see whether it is correctly positioned, I think 3 is too thick. I prefer the current unstable. But, I do have a suggestion, darken the selected motor in addition to increasing the MMT line density. In that way, when the motor tube is empty, it is easier to see, and, when loaded, it will be much more visible, even more so that with a line density of 3.

neilweinstock:
In that case my suggestion is to leave it as-is, and file an issue for future consideration.  This doesn't sound like something for which we can get a quick and optimal solution.

hcraigmiller:
> In that case my suggestion is to leave it as-is, and file an issue for future consideration. This doesn't sound like something for which we can get a quick and optimal solution.

I agree. Especially when playing with global effects that go beyond the current issue. Let's get this issue merged and give some thought to all of the options proposed, and perhaps other options as well.

PR #1138, opened 2022-02-19
[fixes #358] Implement multi-component selection

### LINKED ISSUE #358: Allow multi-selection of rocket components

SiboVG:
This PR fixes #358 by allowing multi-component selection. This is implemented in a number of ways:

1. **Move up/down**
You can now move multiple components up or down at the same time
2. **Delete**
You can delete multiple components simultaneously
3. **Cut/Copy/Paste**
Cut/Copy/Paste a group of components or subparts of a nested group (more explanation on how this works below)
4. **Edit**
You can now edit same-type components simultaneously. E.g. if you select two body tubes, click the edit button, and change the length, then the length of both body tubes will be changed to the same value.
5. **Scale**
Scaling of multiple selected components is now possible + I added a checkbox to let you choose whether you want to scale component offsets.

Small demo:

https://user-images.githubusercontent.com/11031519/154801291-9dcb9c46-d732-44b5-9b3d-aed364f1dbc0.mp4

Now some explanation on how cut/copy/paste works for nested groups:
1. If you select the parent of a nested group and none of its children, then a cut/copy action will also include all of the children. 
2. If you select the parent of a nested group and all of its children, then a cut/copy action will include all of the children.
3. If you select the parent of a nested group and e.g. only one of two children, then only the selected child is taken into account.

Visual interpretation:
Visuals

This way of working is taken from the idea: "If a user deliberately selects specific components, then he/she/they does not want the unselected components to be in the pasted group. The rule of including all children if only the parent component is selected is because this is the implementation of how it currently works on unstable + it's also a common implementation in other software."

Here is a [jar file](https://drive.google.com/file/d/1ymJR43BBmOdzhCCIvcIPlPEAECoiR0D3/view?usp=sharing) for testing.

(note: this PR is a subbranch of PR #1132, so best to merge that one first)

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1757,7 +1757,7 @@ Warning.ZERO_RADIUS_BODY = Zero length bodies may not result in accurate simulat
 ! Scale dialog
 ScaleDialog.lbl.scaleRocket = Entire rocket
 ScaleDialog.lbl.scaleSubselection = Selection and all subcomponents
-ScaleDialog.lbl.scaleSelection = Only selected component
+ScaleDialog.lbl.scaleSelection = Only selected component(s)
 ScaleDialog.title = Scale design
 ScaleDialog.lbl.scale = Scale:
 ScaleDialog.lbl.scale.ttip = Select whether to scale the entire design or only the selected component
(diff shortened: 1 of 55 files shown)


hcraigmiller:
Awesome functionality, I want this in the beta. Functions as described; no anomalies found.

neilweinstock:
This is fantastic... but one anomaly found.  It's not super important, but I do believe it should be fixed.  Consider the following tree:
image
I've select two items underneath the body tube.  Now, if I drag both of them onto the nose cone, I would expect both of them to be, well, under the nose cone.  But instead I get this:
image
This behavior is fairly consistent when dragging multiple items, the end up split across the target.

SiboVG:
Ah right, forgot about click-and-drag. Should be fixed in the latest commit.

Here is a [jar file](https://drive.google.com/file/d/1vA53fbLJMLa0r_lJbgLyZPQJIUoNmNk2/view?usp=sharing) for testing.

hcraigmiller:
Drag-n-drop functions as expected; no anomalies found.

neilweinstock:
Looks good to me.

PR #1142, opened 2022-02-19
[fixes #1137] Use capitalized component names

### LINKED ISSUE #1137: Homogenize capitalization of rocket component names
This is a very nitpicky thing, I know, but it bothers me slightly that there are some rocket components that by default have all the words in their name capitalized, while others don't.

More specifically, I would make the following changes:

- Inner Tube --> Inner tube (in the component tree)
- Launch Lug --> Launch lug (both in the component tree and the create new component button)
- Rail Button --> Rail button (both in the component tree and the create new component button)
- Mass Component --> Mass component (both in the component tree and the create new component button)

This would just tidy things up a bit.

Or we could of course do it the other way around, changing the other component names to capitalize everything (e.g. 'Body tube' --> 'Body Tube'). @hcraigmiller @neilweinstock any opinions on this?

SiboVG:
This PR fixes #1137 by capitalizing rocket component names. Additionally, I noticed that the Body Tube, Nose Cone and Transition had a specific name in their config dialog to change the length ('Body tube length', 'Nose cone length' and 'Transition length'). Since these are redundant and other components also just use 'Length', I changed the translations of those labels to just 'Length'.


neilweinstock:
Why on line 1573 is "Bulk Head" listed as two words?  Everywhere else it is (correctly) a single word, "Bulkhead".

I can't find where in the UI it might appear as two words either, which makes me puzzled what is the meaning of that one reference.

SiboVG:
Good point, fixed in the latest commit

PR #1148, opened 2022-02-21
Fix nose cone subsonic pressure drag computation

plaa:
Previously the nose cone pressure drag coefficient subsonic range was interpolated in the Mach range 0...minY value, while it should be interpolated in the range 0...minX.

minY is typically 0.1 - 0.3, while minX is typically ~0.9, meaning there was a large range in which the pressure drag was interpolated linearly instead of exponentially.

Also the value `minValue - 0.01` could become negative in some cases (e.g. tangent ogive) resulting in a constant ~0.01 pressure drag. This might have a minor impact on a flight simulation, so one more zero was added.

The below image shows the impact before vs after:
Conical nose cone pressure drag

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/barrowman/SymmetricComponentCalc.java
+++ b/core/src/net/sf/openrocket/aerodynamics/barrowman/SymmetricComponentCalc.java
@@ -392,7 +392,7 @@ public class SymmetricComponentCalc extends RocketComponentCalc {
 		double minDeriv = (interpolator.getValue(min + 0.01) - minValue) / 0.01;
 		
 		// These should not occur, but might cause havoc for the interpolation
-		if ((cdMach0 >= minValue - 0.01) || (minDeriv <= 0.01)) {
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
Did you mean to merge this into unstable?  That's what we're prepping for the beta release.

SiboVG:
keep-calm-the-legend-is-back

plaa:
Changed to unstable.  Was not sure which one was in active use.

JoePfeiffer:
I'm really puzzled by how github is displaying this -- it's only showing the two lines you changed as changing, but it's ignoring many other changes made to the file recently (including the change in the for-loop).  It's as if you forked from master, made your changes, filed a PR against unstable -- and github's conflict check isn't noticing all the changes that have been made in the mean time.

plaa:
Yeah, that's what I did.  I noticed the bug on master, fixed it, and wasn't aware unstable was still the main branch.  I then changed the PR to unstable assuming no changes had been made to this class.  Apparently the bug has already been fixed in unstable, so this PR can be closed.

PR #1151, opened 2022-02-22
Don't need to save <radialdirection> for pod sets, since there's no backward compatibility issue with them

### LINKED ISSUE #1150: Reading file with pods generates warning
This just cropped up.  Create a rocket with pods, save, and then open again.  I get this warning: 
image

Here's a file that demonstrates it.
[pods demo rocket.ork.zip](https://github.com/openrocket/openrocket/files/8113356/pods.demo.rocket.ork.zip)

JoePfeiffer:
closes #1150

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
+++ b/core/src/net/sf/openrocket/file/openrocket/savers/RocketComponentSaver.java
@@ -111,7 +111,7 @@ public class RocketComponentSaver {
 			if (c instanceof FinSet || c instanceof TubeFinSet) {
 				elements.add("<rotation>" + angleOffset + "</rotation>");
 			}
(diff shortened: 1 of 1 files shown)


neilweinstock:
Ready to merge?

JoePfeiffer:
I think so, but I always prefer somebody to agree with me before we do it.  Test jar is [here](https://github.com/openrocket/openrocket/suites/5404970844/artifacts/170555202) (that automated artifact upload is cool!).

Note that rockets with podsets that have already been saved since the backward compatibility patch will show the warning, but saving and loading them again should get rid of it.

SiboVG:
> I think so, but I always prefer somebody to agree with me before we do it. 

Yup, as noted in [my comment](https://github.com/openrocket/openrocket/issues/1150#issuecomment-1047768738), I agree

PR #1155, opened 2022-02-22
Update build version to March

SiboVG:
Updates the build version to 22.03.beta.01. I think it's fitting if we would do the release in March (considering OR 15.03).

### DIFF
--- a/core/resources/build.properties
+++ b/core/resources/build.properties
@@ -1,6 +1,6 @@
 
 # The OpenRocket build version
-build.version=22.00.beta.01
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
I'm more optimistic than you: I think 22.02.beta.01 is realistic.

SiboVG:
Wanna place bets on it @JoePfeiffer? 😉 Yeah okay, I'll change it to 02

JoePfeiffer:
You're on for a virtual beer!

PR #1169, opened 2022-02-26
Ignore the resources\datafiles\presets folder as it is being built by the copy-orc-files build script

thzero:
This PR adds the resources\datafiles\presets to the .gitignore since the presets are being copied into this folder by the copy-orc-files build command.

A fresh fork of openrocket unstable branch, after executing the following:

git submodule init
git submodule update

ant (to build)

results in the presets folder being populated.  In any subsequent 'git status' the presets folder now shows up as changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,4 @@ openrocket.log
 *.snap
 prime/*
 swing/resources/datafiles/presets/system.ser
+swing/resources/datafiles/presets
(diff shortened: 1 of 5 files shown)


SiboVG:
Thanks for the change.

Just a tip: since your last commit basically undos the changes in your second commit, maybe in the future just try to undo the second commit so we don't have to code-check both of them :)

SiboVG:
Ah I see now, you eventually moved the changes to PR #1170. All the more reasons to undo the 2nd and 3rd commit 🙃

thzero:
Yeah...  sorry about that; meant to go undo it after I realized what it was, then didn't get a chance to.

PR #1185, opened 2022-02-27
Add install4j configuration files

JoePfeiffer:
This combines Justin Hanney's configuration files from https://github.com/itshanney/openrocket-installer-files.git with my changes for 22.02.beta.01 and brings them into the OR source tree.


SiboVG:
The install4j/README.md seems to be a bit out of date... (e.g. for which platforms it builds)

PR #1186, opened 2022-02-27
[fixes #1164] Account for different RockSim FinSet location directions

### LINKED ISSUE #1164: Sign flip in fin component offset when opening .rkt file
Opening "Estes Argent.rkt", things generally look good, but the fins are in the middle of the body. I initially wondered whether this is due to the model having two coupled tubes in the main body. Apparently not.

Screen Shot 2022-02-25 at 9 26 12 PM

However, intuition said that flipping the sign on the fin offset would put the fins where they belong, and this appears to work.

Screen Shot 2022-02-25 at 9 26 37 PM

Not sure whether this occurs with other components, but a quick check shows the sign-flip on fins in other .rkt files.

I'm using 22.02.beta.01, MacOS Catalina, installed from .dmg package, Intel machine

SiboVG:
This PR fixes #1164 where the location of FinSets from a Rocksim import would be inverted for locations that were referenced from the tip of the nose cone and the top of the parent component.

Here is a [jar file](https://github.com/openrocket/openrocket/suites/5467350835/artifacts/174098659) for testing.

### DIFF
--- a/core/src/net/sf/openrocket/file/rocksim/importt/FinSetHandler.java
+++ b/core/src/net/sf/openrocket/file/rocksim/importt/FinSetHandler.java
@@ -147,6 +147,11 @@ class FinSetHandler extends AbstractElementHandler {
 	private final RockSimAppearanceBuilder appearanceBuilder;
 
 	/**
+	 * Checks whether the location is already loaded in or not
+	 */
+	private boolean locationLoaded = false;
+
+	/**
 	 * Constructor.
 	 *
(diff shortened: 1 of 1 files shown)


SiboVG:
Is there someone with a RockSim license that can check that exporting back to RockSim also works for all the different finset location reference methods?

JoePfeiffer:
Note @hcraigmiller verified in a comment on #1164 that this handles writing to RockSim, too.

PR #1194, opened 2022-02-28
[fixes #1190] Fix ConfigDialog Close-button cut off

### LINKED ISSUE #1190: "Close" button in configuration dialogs cut off on Linux
vis:
image

Testing Ubuntu LTS 20.04 on AMD Ryzen 3400G (integrated Vega graphics), resolution set to HD (1920x1080)

SiboVG:
This PR fixes #1190 where the Close-button in a component config dialog got cut off on some OS'es.

As stated by [Joe's comment](https://github.com/openrocket/openrocket/issues/1190#issuecomment-1054414042), the fix works + I tested it on macOS, Ubuntu and Windows 11 and they are not (negatively) affected by this PR.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/RocketComponentConfig.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/RocketComponentConfig.java
@@ -149,7 +149,7 @@ public class RocketComponentConfig extends JPanel {
 		
 		updateFields();
 		
-		this.add(buttonPanel, "newline, spanx, growx, height 32!");
+		this.add(buttonPanel, "newline, spanx, growx");
 	}
 	
 	


JoePfeiffer:
Almost seems like a workaround for a bug in swing, doesn't it?

SiboVG:
> Almost seems like a workaround for a bug in swing, doesn't it?

Hm well I wouldn't say it's a swing bug, before this PR the bottom row containing the Close-button just used a fixed height and I guess that the Close-button in Linux is a bigger height than the fixed height that was set up for the row. So in short: just don't set fixed heights :)

JoePfeiffer:
I read too fast -- for some reason I was thinking you'd added the fixed height, not removed it. Yes, fixed heights are indeed Bad.

PR #1200, opened 2022-02-28
[fixes #1192] Use default 3D component color mid-grey + 30% shine

### LINKED ISSUE #1192: Request: default component color something other than black
Suggest we make default color something like #7F7F7F so that the black and white boxes in the preview area initially make some intuitive sense to users. They've never made sense to me until just a moment ago when I opened one of the sample files that was colored up (something I've never really used OR for).

The black-and-white preview boxes aren't labeled (other than being in the preview area), and unless the color chosen is not black (which is one of the colors in the boxes that's being used for preview against your chosen colors), they just look like a bunch of weird black and white boxes, and it's not clear that you can see your text on the sample backgrounds (because black-text-on-black and black-text-on-background-and-white don't give you the sense that you're looking at your chosen color).

SiboVG:
This PR fixes #1192 by setting the default 3D component color to a mid-grey #7F7F7F + default 30% shine.

Preview:
image

### DIFF
--- a/core/src/net/sf/openrocket/appearance/AppearanceBuilder.java
+++ b/core/src/net/sf/openrocket/appearance/AppearanceBuilder.java
@@ -51,7 +51,7 @@ public class AppearanceBuilder extends AbstractChangeSource {
 	*	Clears the builder cache and set to build blank appearances
 	*/
 	private void resetToDefaults() {
-		paint = new Color(127, 127, 127);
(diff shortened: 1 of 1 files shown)


hcraigmiller:
I did not like black, but I'll leave that preference to the rest of you.

SiboVG:
Updated color to #BBBBBB:
image

Issue #826, opened 2020-11-18
Upgrading to Java 15 causes unit test failures


JoePfeiffer:
Updating to either
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 15.0.1+9, mixed mode, sharing)
or
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.23.0, JRE 15 Linux amd64-64-Bit Compressed References 20201022_81 (JIT enabled, AOT enabled)
under Linux results in the following two unit test failures:

TestScriptingUtil	testGetLanguages	Failure	N/A
junit.framework.AssertionFailedError
at net.sf.openrocket.simulation.extension.impl.TestScriptingUtil.testGetLanguages(TestScriptingUtil.java:41)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

TestScriptingUtil	testGetLanguage	Failure	expected:<JavaScript> but was:<null>
junit.framework.AssertionFailedError: expected:<JavaScript> but was:<null>
at net.sf.openrocket.simulation.extension.impl.TestScriptingUtil.testGetLanguage(TestScriptingUtil.java:32)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

I'm not getting the unit test failures with Java 14.

hcraigmiller:
Where we will now be looking at JAVA 17 LTS compatibility, I recommend closing this issue.

JoePfeiffer:
Unless the problem still exists under 17, of course. Needs to be checked.

SiboVG:
This issue still exists under Java 17

Issue #799, opened 2020-10-19
Simulation crashes openjdk-13 and -14


JoePfeiffer:
The attached ork file is causing JRE exceptions with several recent versions of Java.  In all cases, the fatal error is happening in

Current CompileTask:
C2:  11710 5947       4       net.sf.openrocket.simulation.listeners.SimulationListenerHelper::firePostAerodynamicCalculation (119 bytes)
(there are differences in the other information on the line, but all of them point to firePostAerodynamicCalculation)

Note: this file was shared to The Rocketry Forum in the thread
https://www.rocketryforum.com/threads/open-rocket-doubt-flight-event-exception.162328/
The author appears to recognize it's a completely impractical design, but it shouldn't be crashing the JRE!

I'm attaching the original .ork (filename extension changed to .txt to pacify github's uploader), and all the error dumps I've got.
[rocket-2020-10-17.txt](https://github.com/openrocket/openrocket/files/5405321/rocket-2020-10-17.txt)
[hs_err_pid1744172.log](https://github.com/openrocket/openrocket/files/5405324/hs_err_pid1744172.log)
[hs_err_pid1744540.log](https://github.com/openrocket/openrocket/files/5405325/hs_err_pid1744540.log)
[hs_err_pid1745334.log](https://github.com/openrocket/openrocket/files/5405326/hs_err_pid1745334.log)
[replay_pid1744540.log](https://github.com/openrocket/openrocket/files/5405334/replay_pid1744540.log)

ore13:
I've had the same error, but occurring with some of the example rockets. I've attached the error dumps I got here too. fatal error is still happening in the firePostAerodynamicCalculation compile task.

A simple model rocket: 
[hs_err_pid16748.log](https://github.com/openrocket/openrocket/files/5427467/hs_err_pid16748.log)
[replay_pid16748.log](https://github.com/openrocket/openrocket/files/5427468/replay_pid16748.log)

3 stage example rocket:
[hs_err_pid15296.log](https://github.com/openrocket/openrocket/files/5427470/hs_err_pid15296.log)
[replay_pid15296.log](https://github.com/openrocket/openrocket/files/5427471/replay_pid15296.log)

Typically takes 2 simulationOption changes (then run) to crash when using the 3 stage example rocket, and  4 - 10 simulationOption changes to crash the simple model rocket.

JoePfeiffer:
[...]

PR #1202, opened 2022-03-01
[fixes #1176] Add JRE version checking at startup

### LINKED ISSUE #1176: App should check JRE at startup and report to user if it's no good
To the extent possible (I don't know what that is), the app (.jar file, in particular) should try to check the JRE version at startup and report to the user if it is unsupported, and refuse to proceed any further in that case.  Maybe this would save us some unnecessary bug reports.

SiboVG:
This PR fixes #1176 by adding JRE version checking at startup. If an unsupported Java version is detected, the following popup will show and the application is closed afterwards:
image

Currently the following Java versions are included as being supported: 11, 12 & 13 (I think these are the supported versions?)

Note: the packaged installers already have a fixed (and correct) JRE versions, e.g. 22.02.beta.01 will show JRE version 11 if you check in the Debug Report.

Here is a [jar file](https://github.com/openrocket/openrocket/suites/5482436783/artifacts/174946436) for testing. (Note: if you wanna test different JRE versions and are running on Linux or macOS, I recommend [jEnv](https://www.jenv.be/))

### DIFF
--- a/core/src/net/sf/openrocket/startup/Application.java
+++ b/core/src/net/sf/openrocket/startup/Application.java
@@ -22,7 +22,7 @@ public final class Application {
 	private static Injector injector;
 
 	// Supported Java Runtime Environment versions in which OR is allowed to run (e.g. '11' for Java 11)
-	public static int[] SUPPORTED_JRE_VERSIONS = {11, 12, 13};
+	public static int[] SUPPORTED_JRE_VERSIONS = {11};
 	
 	/**
 	 * Return whether to use additional safety code checks.
--- a/core/src/net/sf/openrocket/startup/Application.java
+++ b/core/src/net/sf/openrocket/startup/Application.java
@@ -20,6 +20,9 @@ public final class Application {
 	private static ExceptionHandler exceptionHandler;
 	
 	private static Injector injector;
+
+	// Supported Java Runtime Environment versions in which OR is allowed to run (e.g. '11' for Java 11)
+	public static int[] SUPPORTED_JRE_VERSIONS = {11, 12, 13};
 	
 	/**
 	 * Return whether to use additional safety code checks.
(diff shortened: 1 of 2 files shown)


neilweinstock:
I thought support was 9-13 (although I'm not sure because I've been running it exclusively on 11 for a long time).

SiboVG:
> I thought support was 9-13 (although I'm not sure because I've been running it exclusively on 11 for a long time).

I have no idea which versions are actually supported, but we mention on every install instruction to use Java 11, so I guess that's the minimum?

neilweinstock:
Java 11 was chosen because of long-term support.  I *think* 9 has always worked as well, although I'm not sure how much it was tested in recent days (same for 13, for that matter).

@JoePfeiffer is there an "official" answer on this?

JoePfeiffer:
At one point I was agreeing about having a range in versions in the popup, but I've reconsidered my position -- I really don't want to hear from a user who discovers something that only fails on Java 12 or something.  Pop up the warning for any version other than 11, and let it be.

That said, I don't know that I ever tried it under 9 and 10 so I don't know if it will run under those versions (I suspect it will). I've run it successfully on 11-12. 13-14 crash (see #799 -- I misremembered this as only being on 14, but the issue says it was also happening under 13). 15-up run successfully, but fail unit tests related to scripting (see #826).

neilweinstock:
In that case I would agree.  Plus, those who are running the Jar are in many (most?  all?) downloading a dedicated JRE and using command line to invoke... so those folks ought to be able to get Java 11 to run this.

It will be _very nice_ to stop getting weird error reports of problems from those running the wrong Java version.

SiboVG:
Should it be a soft warning message then or an error that quits OR when it is run from a non-'Java 11' version?

hcraigmiller:
I would vote for quitting OR… reduce bad bug reports.

JoePfeiffer:
My earlier comment said "warning", but yeah. Let's make it 11 or bust.

SiboVG:
(btw, this PR is ready for review, the first commit already quit OR when the version didn't match)

neilweinstock:
Seemed to work correctly in my tests.  I don't have a Java 12 or 13 install to verify that it will reject those, but it did definitely accept my 11 install, and reject my 8 install.

PR #1203, opened 2022-03-01
[fixes #1160] Update Mac installer to single bundle with Applications folder target

### LINKED ISSUE #1160: Update Mac installer to single bundle with Applications folder target
Something like this:
image

SiboVG:
This PR updates the install4j configuration for macOS to use an installer with application folder target:
image

The text is there, but invisible due to the black background (although I must say it doesn't really bother me):
image

PS: yes I'm aware that the two items aren't vertically aligned, but moving the OR-icon up a bit looks more natural to me.


SiboVG:
Yeah I was maybe thinking of something like this, so you have sort of gas cloud/nebula in the background:

image

neilweinstock:
I'll fiddle a bit.  If the goal is to make the text visible, then I'm going to make it *visible*.  Will see if I can come with something quickly that looks decent.

SiboVG:
Another concept if you want clear text:
image

I'll just wait until you have alternative ideas.

JoePfeiffer:
FWIW, on vertical alignment -- I think it would look best to me if the text lines were aligned. The icons themselves look fine to me.

neilweinstock:
I agree, if the text is going to be shown, then it should be aligned.

In any case: I'll need a "final" layout screenshot before I can generate the background image.

SiboVG:
Yes it's only after I saw the visible text that I agree the icons should be vertically aligned.

neilweinstock:
Final layout screenshot should include the text (selecting the icons will be fine; I need to know the exact position of the text.)

SiboVG:
Here it is
image

SiboVG:
Btw @neilweinstock don't use the background image you sent in #1160 because I had to create a new one to match the .dmg window dimensions, so pick the backdrop from this PR.

neilweinstock:
OK I'll get something for this tonight.

neilweinstock:
Here's my proposal: 
image

And here's the background image that will produce it:
mac_installer_bg_stars

SiboVG:
Final results works great in the actual installer:

image

JoePfeiffer:
I've built a bundle and put it on [my Drive.](https://drive.google.com/file/d/1jlImdXVzvTzpgu3sIatsbJlTjxW03TD0/view?usp=sharing). Give it a try!

neilweinstock:
Awesome, I will try it tonight.

neilweinstock:
Functions fine, but the background is not showing up:
image

I also note that the text wrapped onto two lines, which means that the white background I provide for the text will not fit correctly.  I wonder if the star background is worth the trouble, although it looks nice.  Maybe just an arrow image would be sufficient for the short term (but the arrow image is very useful IMHO).

SiboVG:
Strange, does [this package](https://drive.google.com/file/d/1q1zF3BC_JFi_oKaxANNTuL2QlJu_PmH2/view?usp=sharing) work for you @neilweinstock? Works fine on my machine.

SiboVG:
Does [this package](https://drive.google.com/file/d/1q1zF3BC_JFi_oKaxANNTuL2QlJu_PmH2/view?usp=sharing) work for you @neilweinstock?

neilweinstock:
That one works better (shows the background image), although the text under the OR icon is still wrapped to two lines.  Does it not show that way for you?
image

SiboVG:
Nope, shows up the way it's intended on my machine:
image

SiboVG:
If I enable 'Show all extensions, I run into the same issue'
image

JoePfeiffer:
If a shorter filename were picked, would it stop the wrap?

neilweinstock:
Yes.  As Sibo correctly points out, this will be true when the final
release goes out and the word "beta" isn't in there anymore.  I wouldn't be
inclined to change the beta name just for that reason, though (don't need
another wrinkle in the beta-naming dispute :)).

SiboVG:
> If a shorter filename were picked, would it stop the wrap?

Yes, I've tested it already with an official release name ('22.03.01.app') and that doesn't cause any problems. So for the moment, the current design is fine.

PR #1205, opened 2022-03-01
Make motor database search more robust, especially when loading files

### LINKED ISSUE #1174: Motors not recognized properly when opening file
It seems some motors are not being recognized, *especially* CTI.  See attached.
[Composite-Warehouse-Air-Strike-3829.ork.zip](https://github.com/openrocket/openrocket/files/8147382/Composite-Warehouse-Air-Strike-3829.ork.zip)

JoePfeiffer:
Instead of requiring motor "designation" to match either the actual designation or the common name, require it to be a substring of the actual designation or that the common name be a substring of the actual designation.  Many motors in stored rocket files have "designations" that resemble, but do not match, actual designations and common names.

Don't match on diameter or length when loading files.  Motors with errors in the old database (eg M1850) would fail to be recognized.

Fixes #1174


SiboVG:
Seems to be working just fine

PR #1206, opened 2022-03-02
Release notes for Beta 2

neilweinstock:
Will update as we go.

What's the version number going to be... is it now going to be 22.03.beta.02, or is it officially going to stay 22.02?


JoePfeiffer:
My assumption has been that as the first beta in March, it will be 22.03.beta.01. If we get a second beta out in March, it becomes 22.03.beta.02

neilweinstock:
That seems very confusing to me.  Folks (including me) will be thinking of beta 1, beta 2, etc., not trying to keep the whole version number in their head.  I can't think of another app where I've seen version number change from beta to beta.  This was why I was initially reluctant to even put the "02" in the version number, because it's going to be changing.

Suggest you reconsider this.

SiboVG:
> I can't think of another app where I've seen version number change from beta to beta.

That's because few other apps use year-month numbering for their versions (something I don't like, but it's what we're stuck with now). If we were to user 22.02.beta.02, we would be switching between different systems, i.e. not sticking to the year-month numbering. That seems even more confusing to me.

neilweinstock:
I believe the mistake is including the month number in the beta release version.  Normally, the release is numbered for when it will be finally released.  Which is why putting an xx or 00 in there to me makes more sense.

Having beta1 be followed by beta1 strikes me as bizarre.  And then later on, you could have a beta 1 that comes after a beta 2.  Makes no sense.

I mean, it's not our most pressing issue, but really this is some wackadoodle numbering IMHO.

SiboVG:
I mean I like the idea of using 00 or xx instead of months for beta releases, but I don't see how you would adopt that to betas of next releases.

neilweinstock:
Then leave it as 02 for the duration of the beta, and we can update the release number when it comes time for final release.  Changing the release number and resetting the beta count for each new beta just makes no sense.  _These are all successive betas for a single release._

SiboVG:
Any reason to have pushed a commit to delete ReleaseNotes.md @neilweinstock?

neilweinstock:
I am trying in vain to clear up this conflict, but nothing I do helps.  I thought I could delete it everywhere except my original copy at home, and then push it back through the chain.  So far it is not working. :(

SiboVG:
Could you undo your latest commit, then I can see if I can resolve the conflicts (don't know if GitHub allows me to do so, but we'll see)?

You can run `git reset --hard 9b11ac535778742c63a95896e73b8ec5097c7caa` to go back to the 'update dates on releases' commit and then do a `git push --force`

neilweinstock:
I'll try to get to this this weekend.

PR #1221, opened 2022-03-05
Correctly identify when time is past end of thrustcurve so thrust is 0 in getAverageThrust

### LINKED ISSUE #1210: Separating stages at booster ignition causes numerical instability

JoePfeiffer:
What looked like numerical instability in simulations with side boosters turned out to be two bugs involving thrust calculations:

1. When calculating motor thrust, the end of the thrustcurve was not being correctly identified, so the motor thrust was effectively random for times after that..

2. If sustainer and side boosters were firing after stage separation (requires a bad configuration, but we should still do the right thing with bad configurations), thrust for all motors was used for calculation of acceleration for each stage.

Fixes #1210


SiboVG:
Appears to be working just fine and code LGTM. Thanks @JoePfeiffer!

PR #1223, opened 2022-03-05
[fixes #1204] Fix stage CG override incorrect location

### LINKED ISSUE #1204: [Bug] Stage CG override not correct in design window, stability not correct
The Stage override is displaying a different value than that entered, and both the CG distance and stability values in the design window (upper-right corner) are incorrect.  It is 1.076" too far from the tip on this design. For example, to get the CG at 9.414", you have to enter 8.127"

[Cone_Rocket.Tip_Composite.txt](https://github.com/openrocket/openrocket/files/8159086/Cone_Rocket.Tip_Composite.txt)

SiboVG:
This PR fixes #1204 by correcting for the CG location when it is overridden in a stage. The problem was that the CG overriding had no effect on the motor calculation, which this PR now adds.

https://user-images.githubusercontent.com/11031519/156888505-34976113-f784-462f-be64-fa6ae7c62252.mov


SiboVG:
Whoops, unit test are failing... Will post a fix and test JAR soon.

SiboVG:
Okay, it's fixed, here is a [jar file](https://github.com/openrocket/openrocket/suites/5546716294/artifacts/178693452) for testing.

JoePfeiffer:
I made a comment on #1204 a moment ago; near as I can tell, it's working as intended.  The CG override is for an empty stage, the weight of the motor then changes it.  This way we don't need a different override for every different motor weight.

SiboVG:
> I made a comment on #1204 a moment ago; near as I can tell, it's working as intended. The CG override is for an empty stage, the weight of the motor then changes it. This way we don't need a different override for every different motor weight.

Tbh I like the implementation of this PR more, doesn't seem logical to override the CG of your whole stage, only to have it partially apply. Also as noted in #1204, the stability value was incorrect because of the previous implementation.

JoePfeiffer:
The current behavior matches how people (I, anyway) use the program -- I build the rocket, I weigh it and measure the CG. I typically haven't even bought the motor reload yet.  I run a bunch of sims with different motors; the sims calculate the "as flown" configurations.  The new behavior requires me to recalculate the CG and weight for every motor.

I haven't checked the stability calculation -- was it correctly calculating the stability, with the CG moved by the motor?  Then it's correct.

Please at least give me the checkbox. Very, very few people will ever choose to use the new behavior.

hcraigmiller:
After reading the comments of @JoePfeiffer, I believe that he is correct, it is functioning properly. I didn't realize at the time that the motor was causing the disparity. The override should not override the effect of the motor mass.

If a checkbox is added, rather than restoring the old behavior, the default should be not to override the motor.

SiboVG:
Okay, in that case I believe the text message should also include that the center of gravity does not include motors. The latest commits undo the unwanted behavior + update the message.
image

JoePfeiffer:
Looks good.  One tiny nit to pick in the code -- should the method names use Mass instead of Weight?

SiboVG:
> Looks good. One tiny nit to pick in the code -- should the method names use Mass instead of Weight?

The rest of the code used Mass, so I just went with that... From a scientific point of view, mass is also more correct than weight

JoePfeiffer:
I was misreading and thinking it was a new method -- Coordinate.java has a setWeight() whose name should be changed at some point.

PR #1230, opened 2022-03-07
Fix tooltip exception when 1 plot on right axis

SiboVG:
This PR fixes an issue with the plot tooltip text generator which previously threw an exception when there was only one data plot and it was on the right axis.


neilweinstock:
Was there a Github issue to to with this?

SiboVG:
> Was there a Github issue to to with this?

No, just something I noticed and wanted to quickly fix

Issue #1207, opened 2022-03-02
[Bug] Tube Fin flight profile significantly differs from 15.03


hcraigmiller:
TRF user **Neutron95** reported a disparity in flight data for tube fin rockets between 15.03 and the beta release: (https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-is-underway-download-it-now.171295/page-6#post-2248246)

Using the example tube fin rocket included with OR, I ran flight comparisons and confirmed the report

![Error Tube_Fin Flight_Profile](https://user-images.githubusercontent.com/68821492/156377630-6b99ed74-76ec-441c-8f6c-e570d3f73332.png)
.
The flight profiles are significantly different; the same disparity exists with the current unstable.

@neilweinstock ran a comparison using macOS and found no material difference: (https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-is-underway-download-it-now.171295/page-6#post-2248323)

[Windows 11 Pro, Version 21H2, Build 22000.434, Windows Feature Experience Pack 1000.22000.434.0]

neilweinstock:
My test was on Windows, for what it's worth.

SiboVG:
Same behavior noted on macOS (note: you should re-run the simulations on the unstable version, because the present simulation results are cached results)

neilweinstock:
Yes, I was looking at cached results, oops.  I confirm this issue as well.

SiboVG:
FYI: This issue is a result of commit [74ec0be340baf06d63cdc0e81cfd45e79bc2fa48](https://github.com/openrocket/openrocket/commit/74ec0be340baf06d63cdc0e81cfd45e79bc2fa48)

hcraigmiller:
To begin, 15.03 does not support tube fins. The user used a hack to simulate this feature, using launch lugs. The 15.03 results are at least 2x as high as both the unstable and Rocksim 10, using the user’s .ork file.

![Comparison](https://user-images.githubusercontent.com/68821492/157167351-e819e839-db16-4fa2-ac2e-63ef054646fe.png)
 
This does not look like an issue with the current unstable.

SiboVG:
Then I would suggest to update the simulation of the Tube Fins example design with unstable's results.

neilweinstock:
Wait, now I'm confused on multiple fronts.

[...]

PR #1238, opened 2022-03-10
Calculate friction drag from TubeFinSet

JoePfeiffer:
beginning of a fix for #1207.  Note that after looking at the code, I don't regard 15.03's results as any sort of "ground truth"; we really need comparisons with actual data (or at least CFD) to know what's more accurate.

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -19,6 +19,7 @@ import net.sf.openrocket.rocketcomponent.InstanceMap;
 import net.sf.openrocket.rocketcomponent.Rocket;
 import net.sf.openrocket.rocketcomponent.RocketComponent;
 import net.sf.openrocket.rocketcomponent.SymmetricComponent;
+import net.sf.openrocket.rocketcomponent.TubeFinSet;
 import net.sf.openrocket.util.Coordinate;
 import net.sf.openrocket.util.MathUtil;
 import net.sf.openrocket.util.PolyInterpolator;
@@ -448,7 +449,8 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 			
 			// Consider only SymmetricComponents and FinSets:
 			if (!(c instanceof SymmetricComponent) &&
-					!(c instanceof FinSet))
+				!(c instanceof FinSet) &&
+				!(c instanceof TubeFinSet))
 				continue;
 
 			// iterate across component instances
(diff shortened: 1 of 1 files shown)


neilweinstock:
Regrettably, I don't think we're likely to get test that will answer this question for us.  CFD is not out of the question, I suppose.  What would we want from a CFD model, would it just be a Cd number for an assortment of configurations?

in the near term I think if we can get results that are close to what Rocksim is yielding then it is an indicator that we're probably on the right track.

hcraigmiller:
Would it be helpful to generate some comparisons between OpenRocket and Rocksim 10? If so, any particular specifications or just a variety.

neilweinstock:
I do think a set of test files, along with the Rocksim results for us to compare against, would be very helpful indeed.

JoePfeiffer:
It would be very helpful to see RockSim and RASAero (does RASAero do tube fins?) results for the attached rocket
[tubular.zip](https://github.com/openrocket/openrocket/files/8311355/tubular.zip)
Even before seeing the comparison that C6-7 result doesn't look reasonable.

hcraigmiller:
Here is the comparison using Jean Nevada as the launch site for both.

Tube_Fins Compare Rev_01

neilweinstock:
> Even before seeing the comparison that C6-7 result doesn't look reasonable.

Yeah that number is nuts... suggests frictionless fins.

Also noteworthy is the very significant difference in CP between OR and Rocksim, as shown in Craig's screenshot above.  And OR is showing *double* the speed off the rod, which makes no sense because max acceleration is almost identical between the two.  Maybe Rocksim is simulating a longer rod?

JoePfeiffer:
Commented before my coffee. I haven't added the pressure drag calculation in yet. It does sound like there's a difference in launch rod length, also.

hcraigmiller:
Launch parameters.

Tube_Fins Compare Rev_02

neilweinstock:
Where's the rod length in Rocksim?

hcraigmiller:
This is with 36" rods... all other conditions unchanged.

Tube_Fins Compare Rev_04

hcraigmiller:
I went back and re-ran everything to ensure the launch rod length was the same.  Here are the results.

Tube_Fins Compare Rev_05

hcraigmiller:
Oh, I forgot... RASAero II does not allow tube fins. I'm working on a hack to do a ring fin in RASAero II for comparison in the event that OpenRocket adds that feature.

neilweinstock:
Interesting.  The rod speed numbers are closer now, but still much more different than I would expect, unless drag is making that much of a difference even at the relatively low speeds on the rod.

Issue #1229, opened 2022-03-07
Parallel staging boosters have no drag.


JoePfeiffer:
Simulation data export (tab-separated-values format) 
of booster set for rocket with parallel staging.  Note the "Drag" columns
[Pod-Boosters.boosterset.csv](https://github.com/openrocket/openrocket/files/8199646/Pod-Boosters.boosterset.csv)
[Pod-Boosters.ork.txt](https://github.com/openrocket/openrocket/files/8199653/Pod-Boosters.ork.txt)

PR #1241, opened 2022-03-11
Process entire part tree when finding active contexts

JoePfeiffer:
When processing the component tree to find active instances of components, it wasn't processing subtrees of inactive components.  The result was that in finding the drag for detached boosters, the drag was calculated as 0.
Notice that if you run the .ork that resulted in filing the issue after this applying this fix, you get a truly bizarre simulation result -- that's actually what we'd expect, since the boosters are separating at launch and have no fins so they're unstable. The attached ork is the same as that one, but uses A8-0's in the boosters and separates at motor ejection charge.
[Pod-Boosters-2.ork.txt](https://github.com/openrocket/openrocket/files/8234582/Pod-Boosters-2.ork.txt)

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
@@ -604,35 +604,12 @@ public class FlightConfiguration implements FlightConfigurableParameter<FlightCo
 				}
 
 				for (InstanceContext context : contexts) {
-					/*
-					 * If the instance is not active in the current context, then
-					 * skip the bound calculations. This is mildly confusing since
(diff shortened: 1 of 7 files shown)


JoePfeiffer:
Also some improvements to handling InstanceMap.

SiboVG:
Good work done here @JoePfeiffer, thanks! Is this PR related to #1229?

JoePfeiffer:
> Good work done here @JoePfeiffer, thanks! Is this PR related to #1229?

Yes, I should have marked it as fixing #1229.

neilweinstock:
This is a major fix.  Beta 2 is going to be really good.

PR #1255, opened 2022-03-22
[fixes #1254] Fix simulation plot element stretching

### LINKED ISSUE #1254: Fix simulation plot label stretching when resizing plot window
When you resize the simulation plot window, the labels, axes and basically everything except for the actual plot stretches with the plot window. They should retain their normal height-width ratio, so not like this:
image

But like this:
test

SiboVG:
This PR fixes #1254: stretching of simulation plot elements when resizing the window (or when the window was too large to begin with). The elements stretched because the panel dimensions were greater than the chartPanel maximum draw dimensions.

New behavior:
image

Here is a [jar file](https://github.com/openrocket/openrocket/suites/5747525932/artifacts/190744045) for testing.


hcraigmiller:

PR #1272, opened 2022-03-30
Allow execution with Java 17

JoePfeiffer:
Let's allow OR to run with either 11 or 17, so we can see whether we can migrate it.


hcraigmiller:
I would like to be able to test the unstable with Java 17 as well.

PR #1282, opened 2022-04-06
generalize getOuterRadius() to set radius so tubes touch for number of tubes greater than 2 (ie whenever possible)

JoePfeiffer:
create getTubesTouching() method to determine whether tubes are touching (will also return true if they overlap)

fix several places where outerRadius was used instead of getOuterRadius()


JoePfeiffer:
Unit test failures are due to outstanding scripting issues

PR #1298, opened 2022-04-14
Augment parachute manufactures and parachute components

hcraigmiller:
Adds Front Range and FruityChutes parachutes to preset components list


hcraigmiller:
Revising to include additional data.

PR #1306, opened 2022-04-19
Parachute configuration preset input enhancements

hcraigmiller:
Presets for selected parachute changes the Component Name to the parachute Description; input the parachute Cd; and input the override mass, checking the mass override box. The parachute packed values are also input, with the diameter of the parachute sized to the inner diameter of the parent component (where nose cones and transitions have inconsistent inner diameters, these components are sized to the largest diameter). Parachutes without a length and diameter (or without those fields) are sized using the default length and diameter. Once created, the parachute length and diameter do not automatically change if the parent diameter is changed, or the parachute is moved to a different component.


SiboVG:
Converted this to draft as I will be pushing some commits to improve the auto packed diameter

SiboVG:
My commits add an 'Auto' checkbox for the packed diameter of a Parachute, Streamer, Shock Cord and Mass Component, which sets the packed diameter to the inner diameter of its parent.

PR #1322, opened 2022-05-01
NEW Changed File > Export menu format to accommodate multiple file format

hcraigmiller:
Adds 4 custom icons to the File menu, "Import" and "Export as..." icons, and Rocksim and RASAero icons. Changes the File menu to have a line entry and icon for "Import" and a line entry and icon for the "Export as..." menu that opens a submenu for the export options, Rocksim with an icon (which is fully functional) and RASAero with an icon (denoted "RASAero - Future). The submenu format will readily allow future expansion of the export options. 

Here's a [test jar](https://www.dropbox.com/s/rnboymcw3qf10r1/OpenRocket.2022-05-01.Export.jar?dl=0)


SiboVG:
I like the change, but I would comment out the RASAero option code until the actual functionality is included.

Also would like to know @neilweinstock's opinion on the UI change
image

hcraigmiller:
I'm wondering whether showing RASAero as shaded with an "Unavailable" label would be acceptable during the beta period, while the translator is being developed. If not complete by the actual "release" date, the option would be commented out until fully developed. Until then, the submenu would look like this:

RASAero_Unavailable Rev_02

Thoughts, objections?

neilweinstock:
> Also would like to know @neilweinstock's opinion on the UI change
LOL are you sure?  If so, here goes:

1) I have thought about the Export function a lot.  Something like what Craig implemented there was definitely one of the possibilities I considered, but I couldn't decide if I liked it because it's a pretty non-standard way of doing an export to different formats.  A save dialog with a selector for output format is much more common.  

However, this implementation makes the options much more visible to the user, which is usually a good thing.  And so.... I guess in principle I approve.

2) Keeping all the different kinds of export straight is something we need to keep in mind.  We have, right now, three different export menu items, and a fourth is coming eventually (export file for 3D printing).  It should be as clear as possible which one does what.  I'd be inclined to change the menu item to something like "Export Design as".  The "Print/Export as PDF" item is very problematic, but I haven't come up with something good for it yet.

3) Menu items that lead to submenus should not have a trailing ellipsis.  Ellipses are for items that open up a dialog.  I now notice that we have that wrong on multiple items in the File menu (Open Recent and Open Example).  Whereas, "Export Decal" (which I would probably rename "Export Decal Image" if I had my druthers) *should* have an ellipsis.

4) The wireframe rocket in the "Export as" icon is very hard to make out.  This is a subject for future improvement I think.  It's fine as a starting point.

5) I would absolutely not include the RASAero submenu entry at all until it is implemented.

6) We should not be using the Rocksim icon without explicit permission, and I wouldn't bother to try.  Just get rid of the icons in the submenu.  Icons on every menu item are not necessary.  I guess OR has a de facto policy (?) of having them on all the File menu items (except the aforementioned "Export decal"), but we definitely don't need them on the submenus (and we don't on any of the others).

That's all I got for now. :)

hcraigmiller:
More to your liking @neilweinstock?

RASAero_Unavailable Rev_03

neilweinstock:
Fixes to Open recent and Open example: good

Change to Print/Export as PDF: doesn't really fix the problem that *I* have with it.  That function creates a set of documentation for the design which is IMHO not the same as just "printing it".  I'm not sure if that's really exporting either, I don't know what it is.

Export decal as .png: incorrect, because export merely saves out the existing file, which might not be .PNG.  And it needs a "..." because it opens a dialog.  Possibilities for this one are "Export decal..." (what's there now), "Export decal file...", "Export decal image...".  Personally I prefer one of the last two.

Export model submenu:
1) The submenu items should go pretty much to the left edge of the menu (see other submenus).  Not sure if the extra whitespace on the left is a relic from when there were icons there.
2) RASAero should not be included in there until it is implemented.  Period.  If and when we implement that feature, then it can go where it is shown.
3) 3D printer export probably should not go in this submenu, because more likely we will want to offer the ability to export individual components, not the whole rocket.  But since we haven't defined this feature yet (beyond a hand-wave), there is no reason to try to plan out the menus yet.

JoePfeiffer:
> Change to Print/Export as PDF: doesn't really fix the problem that _I_ have with it. That function creates a set of documentation for the design which is IMHO not the same as just "printing it". I'm not sure if that's really exporting either, I don't know what it is.
"Print Documentation/Templates"? At this point I'm pretty accustomed to print menu items then having the option to just save as a PDF.  Do we actually need to support saving to PDF ourselves, or do all system print dialogs do that now?

> Export decal as .png: incorrect, because export merely saves out the existing file, which might not be .PNG. And it needs a "..." because it opens a dialog. Possibilities for this one are "Export decal..." (what's there now), "Export decal file...", "Export decal image...". Personally I prefer one of the last two.

I sort of don't really like using "export" for anything except exporting the whole model in some other format.  But every other option I think of ("Save decals..." or something) doesn't look any better to me.

> Export model submenu:
> 
>     1. The submenu items should go pretty much to the left edge of the menu (see other submenus).  Not sure if the extra whitespace on the left is a relic from when there were icons there.
> 
>     2. RASAero should not be included in there until it is implemented.  Period.  If and when we implement that feature, then it can go where it is shown.

Agree.  We're not commercial, we don't need to preview coming attractions in our interface.

>     3. 3D printer export probably should not go in this submenu, because more likely we will want to offer the ability to export individual components, not the whole rocket.  But since we haven't defined this feature yet (beyond a hand-wave), there is no reason to try to plan out the menus yet.

Also shouldn't be there until it's implemented.  When it is, rather than saying "3D Printer format" we should just say "STL" or whatever ends up getting implemented.

neilweinstock:
> > Change to Print/Export as PDF: doesn't really fix the problem that _I_ have with it. That function creates a set of documentation for the design which is IMHO not the same as just "printing it". I'm not sure if that's really exporting either, I don't know what it is.
> "Print Documentation/Templates"?

I dunno, maybe?  That's better than what's there now.

>At this point I'm pretty accustomed to print menu items then having the option to just save as a PDF.  Do we actually need to support saving to PDF ourselves, or do all system print dialogs do that now?

Well, it's definitely not needed on the Mac (and has never been needed there).  The Windows machines I've used lately all have PDF generators available as printers.  Can't speak for Linux.

One thing for sure though: many people run OR on *old* Windows machines.  Don't want to abandon them.

I certainly wouldn't object to removing the built-in PDF functionality once we can confirm that no one will miss it.

> > Export decal as .png: incorrect, because export merely saves out the existing file, which might not be .PNG. And it needs a "..." because it opens a dialog. Possibilities for this one are "Export decal..." (what's there now), "Export decal file...", "Export decal image...". Personally I prefer one of the last two.
> 
> I sort of don't really like using "export" for anything except exporting the whole model in some other format. But every other option I think of ("Save decals..." or something) doesn't look any better to me.

I'm sort of in the same boat.  We're abusing the term "export" a bit, but I don't have anything better.  The *new* functionality of Export decal (the one in the current beta) is at least in the neighborhood of what "export" should mean.

hcraigmiller:
Based upon the comments given, and a review of generally accepted menu formatting and labeling guidelines and protocols, I propose the following:

01 03 File_Submenu

**New items shown in light grey would only be included as a submenu item when _actually available_ . These items are included here simply for nomenclature and appearance review.**

Moving from top down, I agree with the reservations expressed concerning the use of the term "export." As defined below, OpenRocket does not export data. The term export should be replaced with "generate," "convert," and "encode'" as explained below.

Next, I agree with @neilweinstock that "**save** decal image" (using save instead of export) is the best phrase to describe the File menu decal function.

Next, I also agree with @neilweinstock that the term "print" inappropriately describes what occurs when the user selects the  print/export PDF feature. Although the ability to print information about a design becomes available when this feature is selected, the primary function of the feature is to take the design data and "**generate**" a document, is this instance that document is the selected specifications for the design. 

Next, as stated above and below, changing data from one type to another is not an export function, it is "converting" the design from one data set to another data set, for a similar use. When selected, OpenRocket will "**convert**" a design to Rocksim (and hopefully soon to RASAero). It has a new icon too.

Next, when 3D printer functionality is added, OpenRocket will not be doing the printing, so the use of that term would not be correct. And, the 3D printer data put out by OpenRocket is not a save or export feature either. What is being done is "encoding" data for use by the 3D printer (for an entirely different use of the data). When implemented, the feature will "**encode**" a subset of the design data for use by a 3D printer, and, most likely, users who desire 3D printer functionality will either be using it to 3D print the _entire external airframe_ or an individual _component_; the file folder menu is probably the best place to put the file extension selections.

And, last, as can be seen below, the submenu items are flush left. The space between the left edge and the left side of the text is not an artifact, it is the way it has been all the way back to and including 15.03.

01 04 File_Submenu

The foregoing terminology is derived from the following definitions:
- **Export** is to take data from one program or computer to another, with examples being to backup, move, save, or save portions of a data set. OpenRocket does not export data.
- **Generate** is to employ an algorithm, method, or set of rules normally used by a computer program to generate the end product.
- **Convert** is changing a file into another type. For example, taking a file used in Microsoft Word (.docx) and converting it to the Corel WordPerfect (.wdp) file format.
- **Encode** is a hardware device or software program used to convert, or encode data from one format into another. For example, a software MP3 encoder can take an audio track from a CD and encode it into MP3 format. Where the term "convert" is being used in connection with design applications, "encode" is more distinguishing term to use with 3D printing.

JoePfeiffer:
Wow, this is a great piece of work!  The most important thing is you've gotten away from "everything is export" so we can really think about what to call stuff (something I confessed to being stuck on a little earlier).

I would suggest "Convert to..." should come right after "Save as...", and "Import" should be changed to "Convert from..." and placed up with the "Opens"  so they are complementary. I do wonder, though, if it would be better to just overload "Save as..." and "Open..." to save/open the desired file format (when I use the Gimp, I find "Save as..." only saving .xcf files and having to go to "Export..." to save a .png file to be jarring).

Rather than "Generate specs", use "Generate..." since it opens a dialog letting the user select what to generate.

But these are all quibbles.  I really like the direction you're going.

hcraigmiller:
Good news, there is a consensus on the icons, no one objected to any of them.

And, I stand corrected on the “Quit” button, as @neilweinstock  stated, it does quit all open instances of OpenRocket. I’m not certain why I thought otherwise.

Implemented the changes suggested by @neilweinstock. . . 

File-Menu

Unless there are further suggestions for improvement, I will commit the changes.

neilweinstock:
Because I am a pain in the butt, I will add a few more "final" comments.

1) "Print specifications" should have a trailing ellipsis (Joe mentioned this as well).  
2) I prefer the word "documentation" to "specifications" I think, but that's not a strong opinion
3) Another thing that Joe alluded to: I think it makes sense for the section with the "Save" items should be functions of the design file.  Therefore, Export should go under "Save As...", and the other two (decal and specs) should go into the separated section below. 
4) If we go along with the reshuffle suggested in (3), then "Export design as" can become simply "Export as", because that whole section consistently applies to the design file, and you don't need to call it out on that one item.
5) You can either remove the "Import..." item as part of this PR or I will submit a separate issue for it. :)

With these changes I would be very happy with this, a big improvement of where it started.

hcraigmiller:
Just remember @neilweinstock, you said your most comments were your "'final' comments."

1. Added trailing ellipsis to "Publish specs. . ." (the text needed the room, neither documentation nor specifications fit well).
2. See, 1 above.
3. "Export as. . ." moved to save section.
4. "Export design as. . ." changed to "Export as. . ."
5. "Import" removed

Be happy. . . I like the term publish better that print, the feature does more than just spit our what's shown on the screen.

If all is well, I'll commit the changes.

neilweinstock:
Looks good to me. Huge improvement over what was there before. Thanks for taking this on, and suffering my input.

hcraigmiller:
> Looks good to me. Huge improvement over what was there before. Thanks for taking this on, and suffering my input.

Some of the best things in life come from candid discussions, improvements in this case that benefit all. Not a problem.

SiboVG:
Hm, now I'll be the one that's a bit naggy, but I'm not all to sure about the term 'publish', maybe it's my bias as a software developer, but 'publish' feels more like uploading it somewhere. It's also a term you don't often see in other programs, so it may confuse new users.

neilweinstock:
> Hm, now I'll be the one that's a bit naggy, but I'm not all too sure about the term 'publish', maybe it's my bias as a software developer, but 'publish' feels more like uploading it somewhere. It's also a term you don't often see in other programs, so it may confuse new users.

Agree 100%.  Publish has a specific meaning, and this isn't it.

hcraigmiller:
So, which terms should be used, "Print," "Publish," "Produce," or "Process"?

File-Menu Rev_03

neilweinstock:
Interesting, what program is that?

I would rank Print first, then Publish, and neither of the others third.

BTW one other idea popped into my head, as alternative to "specs":  "design info" (i.e. "Print design info").  That phrase conveys most accurately to *me* what the actual output of that function is.

hcraigmiller:
> Interesting, what program is that?

Corel WordPerfect. And, here's Microsoft Word:

File-Menu Rev_04

That's why I was leaning toward the use of "Publish."

neilweinstock:
The MSWord use of Publish is exactly what Sibo was describing: putting the file onto a server of some kind for public consumption.  That is *not* what our menu item does.

The CorelDraw example seems, from your screenshot, to be closer to an export.

hcraigmiller:
> The MSWord use of Publish is exactly what Sibo was describing: putting the file onto a server of some kind for public consumption. That is _not_ what our menu item does.
> 
> The CorelDraw example seems, from your screenshot, to be closer to an export.

Yes, and it does appear that the term "Publish" means that it is being sent somewhere else, not used locally. "Print" looks like the best choice.

neilweinstock:
Yeah, should be unicode ellipsis... https://www.fileformat.info/info/unicode/char/2026/index.htm

hcraigmiller:
Changed ". . ." to "...", good to go.

SiboVG:
@hcraigmiller it appears the changes did not come through correctly, I've pushed a commit to fix it

neilweinstock:
Good stuff.

JoePfeiffer:
Let's do this.

SiboVG:
> BTW one other idea popped into my head, as alternative to "specs": "design info" (i.e. "Print design info"). That phrase conveys most accurately to _me_ what the actual output of that function is.

I actually quite liked this idea, anyone else in favor?

JoePfeiffer:
> > BTW one other idea popped into my head, as alternative to "specs": "design info" (i.e. "Print design info"). That phrase conveys most accurately to _me_ what the actual output of that function is.
> 
> I actually quite liked this idea, anyone else in favor?

I didn't notice it as it flew by -- yes, I would prefer that too.

SiboVG:
Changed to 'design info' in PR #1331

PR #1333, opened 2022-05-05
Refine tube drag

JoePfeiffer:
With additional data from a larger tube fin rocket provided by Rocketry Forum user Neutron95 (see https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-2-is-now-available.171295/post-2272694), I was able to use a more detailed model of tube fin drag to improve results.  With this PR and the flights I have, results are now as follows:
<table>
<tr><th>Motor</th><th>Actual</th><th>Simulated</th><th>Error (%)</th></tr>
<tr><td>B6</td><td>352</td><td>349</td>-<td>-0.85</td></tr>
<tr><td>C6</td><td>693</td><td>690</td><td>-0.43</td></tr>
<tr><td>J360</td><td>4065</td><td>4209</td><td>3.54</td></tr>
<tr><td>K1103</td><td>5009</td><td>4994</td><td>-0.30</td></tr>
<tr><td>K2050</td><td>4263</td><td>4307</td><td>1.03</td></tr>
</table>
The B and C flights were with my _Totally Tubular_  (and are the average of two flights each).  The J and K flights were with Neutron95's _Blue Tuber v3_. Data for another J flight was discarded, as it was on a windy day.


SiboVG:
Out of interest: don't the empirical variables depend on the build material and shape of the tube fins? Is there also no theoretical calculation method for the pressure drop?

neilweinstock:
Those results are very good.  If they hold generally we'll be in great shape.

This is what public betas are for. :)

JoePfeiffer:
Build material, no -- the equations I'm using assume they're "smooth" but I don't really know how smooth it has to be.  Apparently these parameters are smooth enough!  On shape, round is the only shape we support.  If we start getting into rectangular tubes (and other shape) tubes, the area of the tube is used to calculate an equivalent diameter.  I expect at some point a really wide rectangular tube would start giving bad results.  The equation I'm using for the actual pressure drop calculation is theoretically derived, but with several parameters (most importantly the volume flow rate) measured.  I am actually pretty happy that all worked out well so I don't have to try to justify adjusting it.

Frankly, both the first iteration (which only had the one rocket) and this one (which adds in a much faster rocket) surprise me with how nicely they work out.  Note that I did have to adjust the stagnation pressure to make it fit (but if you go back and look at the complete lack of justification for the numbers up in BarrowmanCalculator I don't feel too badly about it).

We'll see if it holds up...

One thing I'd like to do is create an archive of designs and experimental results, so we can really see how accurate we are for a variety of rockets.

SiboVG:
Thanks for the clarification! The initial results indeed look promising, we'll just have to wait for more reference data to call it a victory, but in any case good job @JoePfeiffer!

PR #1346, opened 2022-05-10
Minor change in parachute packed length and diameter code

hcraigmiller:
Simplified code by replace new if statements with else statements. No change in functionality.


SiboVG:
Please remove the comments, they are not needed as you can tell from the if-else conditions what the code does, they just clutter the code otherwise

PR #1367, opened 2022-05-21
Small UI Fix To Update ComboBox To Display Selection

RemingtonHolder:
Small UI fix to update the ComboBox to display the selected item in: Preferences --> Simulation --> Simulation Options --> Geodetic Calculations

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/preferences/SimulationPreferencesPanel.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/preferences/SimulationPreferencesPanel.java
@@ -116,7 +116,7 @@ public class SimulationPreferencesPanel extends PreferencesPanel {
 				GeodeticComputationStrategy gcs = (GeodeticComputationStrategy) gcsCombo
 						.getSelectedItem();
 				gcsCombo.setToolTipText(gcs.getDescription());
-				gcsCombo.repaint() // On some machines, the combobox did not visually update to the selected item
+				gcsCombo.repaint(); // On some machines, the combobox did not visually update to the selected item
 			}
 		};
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
I know it's a one-liner, but I'm having a little trouble understanding -- unless I'm misreading, this is setting the selected item to the already-selected item, so it should already be displaying it.  Are there conditions under which it isn't?

RemingtonHolder:
When I was using the newest OR release for some reason it was not visually updating whenever you changed the value, it may just be a one off bug on my machine but this one liner fixed it.

SiboVG:
I'm also not experiencing this issue, so must be a one off.

I would however suggest trying `gcsCombo.repaint()`, does that also solve the issue? And at a comment next to it stating why that line is added, e.g. `gcsCombo.repaint()    // On some machines, the combobox did not visually update to the selected item`

swing/src/net/sf/openrocket/gui/dialogs/preferences/SimulationPreferencesPanel.java
```diff
@@ -116,6 +116,7 @@ public void actionPerformed(ActionEvent e) {
 				GeodeticComputationStrategy gcs = (GeodeticComputationStrategy) gcsCombo
 						.getSelectedItem();
 				gcsCombo.setToolTipText(gcs.getDescription());
+				gcsCombo.setSelectedItem(gcsCombo.getSelectedItem());
```
SiboVG, line 119: Replace by `gcsCombo.repaint() // On some machines, the combobox did not visually update to the selected item`

SiboVG (changes requested):

PR #1371, opened 2022-05-23
Improve parachute preset code/gui

hcraigmiller:
Corrected drag coefficient return from “DEFAULT_CD” to cd.” Condensed code related to setting parachute preset values. Added code to set parachute characteristics to default values when a particular characteristic data field either does not exist, is empty, or (when numerical) has a value that is not greater than zero.

Parachutes Elements n_02

Here's a [jar file](https://www.dropbox.com/s/ajxks4e430eh44j/OpenRocket.1371.jar?dl=0) for testing 

Added removal of preset status upon changes to parachute Cd, line length, and line material (jar file updated).

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Parachute.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Parachute.java
@@ -9,22 +9,23 @@ import net.sf.openrocket.util.MathUtil;
 
 public class Parachute extends RecoveryDevice {
 	private static final Translator trans = Application.getTranslator();
-	private final double defaultDiameter = 0.3;
+	private final double DEFAULT_DIAMETER = 0.3;
 	private double diameter;
 	public static double DEFAULT_CD = 0.8;
-	private final Material defaultLineMaterial;
+	private final Material DEFAULT_LINE_MATERIAL;
 	private Material lineMaterial;
-	private final int defaultLineCount = 6;
-	private int lineCount = defaultLineCount;
-	private final double defaultLineLength = 0.3;
-	private double lineLength = defaultLineLength;
+	private final int DEFAULT_LINE_COUNT = 6;
+	private int lineCount;
+	private final double DEFAULT_LINE_LENGTH = 0.3;
+	private double lineLength = DEFAULT_LINE_LENGTH;
 	private final double InitialPackedLength = this.length;
 	private final double InitialPackedRadius = this.radius;
 
 	public Parachute() {
-		this.diameter = defaultDiameter;
+		this.diameter = DEFAULT_DIAMETER;
+		lineCount = DEFAULT_LINE_COUNT;
 		this.lineMaterial = Application.getPreferences().getDefaultComponentMaterial(Parachute.class, Material.Type.LINE);
-		defaultLineMaterial= lineMaterial;
+		DEFAULT_LINE_MATERIAL = lineMaterial;
 		super.displayOrder_side = 11;		// Order for displaying the component in the 2D side view
 		super.displayOrder_back = 9;		// Order for displaying the component in the 2D back view
 	}
@@ -173,7 +174,7 @@ public class Parachute extends RecoveryDevice {
 		if ((preset.has(ComponentPreset.DIAMETER)) && preset.get(ComponentPreset.DIAMETER) > 0) {
 			this.diameter = preset.get(ComponentPreset.DIAMETER);
 		} else {
-			this.diameter = defaultDiameter;
+			this.diameter = DEFAULT_DIAMETER;
 		}
 		//	//	Set preset parachute drag coefficient
 		 if ((preset.has(ComponentPreset.PARACHUTE_CD)) && preset.get(ComponentPreset.PARACHUTE_CD) > 0){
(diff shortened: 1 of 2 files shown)


core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
 
 public class Parachute extends RecoveryDevice {
 	private static final Translator trans = Application.getTranslator();
-	
-	public static double DEFAULT_CD = 0.8;
-	
+	private final double defaultDiameter = 0.3;
```
SiboVG, line 12: I would change this to DEFAULT_DIAMETER, since it's a constant, not a real variable

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
...
 	private double diameter;
+	public static double DEFAULT_CD = 0.8;
+	private final Material defaultLineMaterial;
```
SiboVG, line 15: I would change this to DEFAULT_LINE_MATERIAL, since it's a constant, not a real variable

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
...
+	private Material lineMaterial;
+	private final int defaultLineCount = 6;
```
SiboVG, line 17: I would change this to DEFAULT_LINE_COUNT, since it's a constant, not a real variable

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
...
+	private int lineCount = defaultLineCount;
```
SiboVG, line 18: I would declare this in the Parachute() constructor, so leave this as `private int linecount;` and in `public Parachute()` after `this.diameter = DEFAULT_DIAMTER` add `lineCount = DEFAULT_LINE_COUNT`. It's just a better Java practice

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
...
+	private final double defaultLineLength = 0.3;
```
SiboVG, line 19: You know the drill --> DEFAULT_LINE_LENGTH

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
```diff
@@ -9,24 +9,31 @@
...
+	private double lineLength = defaultLineLength;
```
SiboVG, line 20: Also declare this in the constructor

SiboVG:
Added some final remarks. May I also ask to not play around too much with switching the location of methods, makes code reviewing a bit exhausting to know what is new code and what's not. For instance now you moved `public String getComponentName()` to the front instead of where it was before, but for that I have to scroll through the whole 200 code edits to see what actually changed.

SiboVG:
> Added some final remarks. May I also ask to not play around too much with switching the location of methods, makes code reviewing a bit exhausting to know what is new code and what's not. For instance now you moved `public String getComponentName()` to the front instead of where it was before, but for that I have to scroll through the whole 200 code edits to see what actually changed.

That's a good way to reduce the WTFs/minute
image

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

core/src/net/sf/openrocket/rocketcomponent/Parachute.java
hcraigmiller: Changed.

hcraigmiller:
> > Added some final remarks. May I also ask to not play around too much with switching the location of methods, makes code reviewing a bit exhausting to know what is new code and what's not. For instance now you moved `public String getComponentName()` to the front instead of where it was before, but for that I have to scroll through the whole 200 code edits to see what actually changed.

Understood... I will refrain from doing that again. Your comments are very much appreciated.

PR #1380, opened 2022-05-26
Component Icon updates for launch lug and tube fins

neilweinstock:
Before:
image

After:
image


SiboVG:
Thanks for the update!

Issue #1168, opened 2022-02-26
[New feature request] Option to turn CG and CP markings on and off in the 3D views


hcraigmiller:
In the current unstable, in the _3D views_, the visible CG and CP markings interfere with the user's ability to take screen shots for esthetic purposes, particularly in the _3D Finished_ view. TRF user K'Tesh has asked about an option to allow these markings to be turned on and off.

SiboVG:
Would also be a good way to mask the CG/CP issues in the 3D view on Mac's 🙃

RemingtonHolder:
Just tinkering around, this is a surprisingly frustrating thing to do. The code for CG and CP points seems to be pretty deeply integrated to just always exist. 🤷

SiboVG:
> Just tinkering around, this is a surprisingly frustrating thing to do. The code for CG and CP points seems to be pretty deeply integrated to just always exist/display. 🤷

It's not that hard to do in fact, if you comment out `drawCarets(gl, glu);` on `line 327` in `swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java`, the markings are gone:
<img width="1840" alt="image" src="https://user-images.githubusercontent.com/11031519/160042296-4082e2bd-c650-4506-9639-f8dc4fa846bc.png">

So the only thing that would need to be added for this PR is create a checkbox in the preferences dialog, store that checkbox value in preferences and create a simple if-condition on `line 327` that if the checkbox is checked, that you should call `drawCarets(gl, glu);`.

RemingtonHolder:
Seems as if I were looking at the wrong things. 😂

RemingtonHolder:
@SiboVG Do you think a preference for this would best fit in the Design or Graphics preference tab.

JoePfeiffer:
My first thought is that it could be a button in the rocket figure toolbar -- there's plenty of room for it there, and a user could turn it on/off without having to go to another tab or something.  We could disable it for the non-3D views (much like the scale buttons are disabled for 3D views now -- though I think there's a consensus we'd like that to change that some day).  For that matter, there's no compelling reason to not let users turn them off in the non-3D views as well.

hcraigmiller:
Fix complete by way of PR #1393.

PR #1381, opened 2022-05-26
[fixes #898 & #1168] Improve rocket view ribbon for smaller windows + add CG/CP show/hide checkbox

### LINKED ISSUE #898: Enable window to be narrower
There are many occasions when I want the OR window to be fairly small, so I can have it open side-by-side with either another design(s) or an image editor while i'm working on decals.  Right now, the narrowest the layout can get before it starts clipping is shown here:
image

When there are multiple stages its worse, because the stage buttons are laid out horizontally and make the minimum window width even larger.

It looks to me like the limit it being set by the display area.   Within that area, there are two bits of GUI that are consuming excessive horizontal space:  the text "View type" and "Flight configuration".  Arguably "View type" is redundant, since the selected option describes itself as a view.  I'm not sure about "Flight configuration".  Probably need some sort of label for that.

If we can't get rid of the text, then maybe when the window gets sufficiently narrow the text could stack on top of the selector or something?

If I can think of a more coherent suggestion of what I would like to see done, I'll add it here; in the meantime I just wanted to log the issue for future consideration.

SiboVG:
This PR fixes #898 and #1168 by rearranging the rocket view ribbon to be better suited for smaller windows + it adds a checkbox to hide the CG/CP markings in the rocket view. I also added a tooltip text for the rotation slider on the left, since I found it very confusing when I first opened OR what that slider actually did.

Demo:

https://user-images.githubusercontent.com/11031519/170521933-8dddacc5-478f-4f63-b725-0e797a8dabd9.mov

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6673568362/artifacts/252777629) for testing.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -56,6 +56,7 @@ RocketPanel.lbl.ViewType          = View Type:
 RocketPanel.lbl.Zoom = Zoom:
 RocketPanel.checkbox.ShowCGCP = Show CG/CP
 RocketPanel.lbl.Stages = Stages:
+RocketPanel.ttip.Rotation = Change the rocket's roll rotation (only affects the rocket view)
 
 ! BasicFrame
 BasicFrame.tab.Rocketdesign = Rocket design
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -53,6 +53,9 @@ RocketPanel.FigTypeAct.Unfinished = 3D Unfinished
 RocketPanel.lbl.Flightcfg = Flight configuration:
 RocketPanel.lbl.infoMessage = <html>Click to select &nbsp;&nbsp; Shift+click to select other &nbsp;&nbsp; Double-click to edit &nbsp;&nbsp; Click+drag to move
 RocketPanel.lbl.ViewType          = View Type:
+RocketPanel.lbl.Zoom = Zoom:
+RocketPanel.checkbox.ShowCGCP = Show CG/CP
+RocketPanel.lbl.Stages = Stages:
 
 ! BasicFrame
 BasicFrame.tab.Rocketdesign = Rocket design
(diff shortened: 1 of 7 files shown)


SiboVG:
@hcraigmiller @JoePfeiffer could you give a last check as to whether the UI looks good on Windows/Linux?

hcraigmiller:

From a cosmetic standpoint, two comments, flush left the zoom label to match the other labels, and decrease the spacing between the labels and the boxes a bit

CP-CG

Works very well, K'tesh will be ecstatic.

JoePfeiffer:
Looks good.

I agree on left-justify (though I've got a hunch it'll look better starting over the dropdown rather than over the zoom-out button).  I don't know if the spacing can be reduced when the zoom and stage buttons are larger than the dropdowns.  I haven't looked at the zoom buttons so don't know whether they could reasonably be made smaller.

I still think Stability: as a heading over CG/CP would be better, but recognize I'm outvoted.

SiboVG:
Latest commits align the zoom label to the zoom combobox and fix the vertical gap issue on Windows (even fixed it for macOS and Windows without the negative gap hack) + I also removed the gap between the zoom in and zoom out buttons and the zoom combobox (bothered me that they were far apart).

And to make @JoePfeiffer sleep well at night, I have these comparisons between including and excluding the stability label:
Comparisons

Please tell me which one you prefer, with or without a label.

hcraigmiller:
The inclusion of the label looks best to me.

SiboVG:
I must say, seeing it in on screen, with the label also looks better to me.

JoePfeiffer:
lgtm

JoePfeiffer:
Ah, oops, I just realized I wasn't sure whether the current state of the PR includes the label!  I committed under the assumption it did....

SiboVG:
> Ah, oops, I just realized I wasn't sure whether the current state of the PR includes the label! I committed under the assumption it did....

No problem, I'll quickly push a fix for it

PR #1390, opened 2022-05-29
Uniformity of mass component dialogs

hcraigmiller:
Unified the look of the dialogs for parachutes, streamers, shock cords, and mass objects.


SiboVG:
Looks good, thanks for the changes! Just noticed a missing colon after the mass type label in the mass component configuration, fixed that in the last commit.

Issue #1286, opened 2022-04-08
Minor: first click selects configuration but "Run simulations" disabled


jimmiedave:
Steps to repro:

1. Cold-start OpenRocket 22b2
2. Open a model
3. Go to Motors & Configuration tab
4. Add a "New Configuration" and choose a motor (can click on filters, select with double-click or button, whatever)
5. Go to "Flight simulations" tab. As expected, no config is selected. Scroll **_with scroll wheel, not by clicking_** if needed to make new config visible.
6. Click on new config.

Result: new config is selected (and should be runnable) but the "Run simulations" button and all buttons in that row except "New Simulation" are still disabled. A second click, or a click on a different config will un-stick the "Run simulations" button and make it operate as expected.

There appears to be a race condition component: if you pretty quickly run through steps 5 and 6 - not a breakneck speed, but before a couple seconds pass, normal enough speed that you'd do it without thinking about how fast, the bug repros. If you count off maybe 5-10 seconds before step 6, the bug will not repro.

Have run into this a number of times, and the experience is puzzling, though easily resolvable. Not a huge problem, but leaves the impression that the software is buggy.

Testing on Mac OS 10.15.7, installed from beta 2 installer.

jimmiedave:
Just noted that I can repro this issue in 15.03 as well.

PR #1395, opened 2022-06-01
Implement right-click popup menus (+ others...)

### LINKED ISSUE #1123: [Feature request] Right-click actions in component tree
It would be useful (and in line with how a lot of other programs work) to be able to right-click a component in the component tree and to have a context menu that has options such as to edit the component, delete it or copy it.

SiboVG:
This PR fixes #1123.... among with some other stuff (I let myself go a bit)

This PR now implements appropriate right-click popup windows in the componentTree, rocket view, motor configuration table, recovery configuration table, stage configuration table, and simulation table.

Demo:

https://user-images.githubusercontent.com/11031519/171317352-9b115eab-606f-4b2a-8e22-13c44a601b67.mp4

Whoops, noticed I went over the motor popup window a bit too quickly, here's the motor popup more in detail (also showcases that when you right-click on the configuration column, the other motor-related options are not shown):

https://user-images.githubusercontent.com/11031519/171318710-88d36fbc-d48c-4271-b9d2-770ad486b53a.mp4

Besides those changes the following other changes were made:

1. Added a 'Duplicate' button in the rocket design window and in the application ribbon (with shortcut ctrl/cmd + D; changed the debug log shortcut to ctr/cmd + shift + D instead of the previous ctrl/cmd + D)
2. Added an 'Edit' action in the application ribbon (applies to editing components and simulations)
3. The scale dialog now automatically opens in the 'scale selected component' or 'scale component and subcomponents' option when a component is selected, if not, the 'entire rocket' option is selected. Previously, the 'entire rocket' option was selected by default
4. Changed the reset button text of the recovery and stage tab from 'Reset to default' to 'Reset deployment' and 'Reset separation' respectively
5. Fixed a selection issue in the motor, recovery and stage table, similar to #1286 
6. Changed the copy icon (it now looks less like the scale icon and more like a generic copy icon)
7. Changed the 'Copy configuration' button in the 'Motors & Configuration' tab to 'Duplicate configuration'
8. Added a small gap between the motor buttons and ignition buttons in the motor configuration tab

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6738746297/artifacts/257187605) for testing.


hcraigmiller:
I was unable to get "the entire rocket" to ever select, except manually. That aside, personally, I thick most users scale the entire rocket, not a portion of the rocket, and that not having that setting as the default will create confusion with new users and frustration with old (having to do something that the user did not previously have to do).

Otherwise, all changes function as expected with no anomalies found; Issue #121 may still exist, but was not evident on a single  scale (took at least two at 1001.33).

Scale

SiboVG:
> I was unable to get "the entire rocket" to ever select, except manually. That aside, personally, I thick most users scale the entire rocket, not a portion of the rocket, and that not having that setting as the default will create confusion with new users and frustration with old (having to do something that the user did not previously have to do).

Having no component selected brings up the 'the entire rocket' option, so clicking on a blank space in the rocket view deselects all components and you will then get the right option. My mentality was 'If a user actively selects a component and then goes for the scale option, it is because he wants to actively only change that component'.

hcraigmiller:
I deselected all of the components by clicking a blank spot in the design window, components came up on scale. I also tried highlighting the stage on the tree, components came up on scale.

As to what to scale, I only know my own experience. I rarely scale a single component, or have really paid much attention to which component was selected when I went to scale. And, that’s my concern. If a new user, or current user who forgets, just hits scale without reading what is going to happen, the user will not get the expected result. 

I could be wrong, but I would think that the frequency of scaling the entire rocket would be much greater than not. But, Neil and Joe may have different thoughts.

SiboVG:
Hm, I see, I don't mind changing it back, you all have more practical experience with the program so you'll know what is common. I'll wait for what @neilweinstock and @JoePfeiffer have to say.

JoePfeiffer:
Very nice to see context menus!  That wasn't even on my radar.

On scaling -- I very rarely scale a rocket.  I've *never* had occasion to scale just part of a rocket.

The behavior of the context menus doesn't quite match my expectation -- I'd expect that when I have the mouse over a component and right-click, it would select that component and then open the menu.  Having it perform the operation on the component that was previously selected seems odd.

JoePfeiffer:
I just passed the question by @sjh114961 since he's become a pretty prolific designer, and got the response that he scales fins a lot, and has rarely ever scaled an entire rocket. So yes, we do have users that will select a component to scale, so we should retain that possibility.

JoePfeiffer:
I've been playing with it in the tree view (I find component selection far, far easier there); selecting the entire rocket is easy in that window.

Behavior when a stage or the entire rocket is selected seems a little odd:  I expect when I've got the rocket selected, and choose to scale "selected and all subcomponents", that should end up scaling the entire rocket.  It doesn't seem to at present -- I have to choose "entire rocket" to get that effect.  Likewise, when I choose "selected and all subcomponents" with a stage selected, that entire stage should be scaled.

If it did work so "selected and all components" would scale the entire rocket when the entire rocket is selected, would we have a need for an "entire rocket" choice?

hcraigmiller:
I like the idea of implementing a scale components/subs when the stage is selected (for everything in that stage), with using the edit>scale for the entire rocket, which would only be needed for multiple stages.

neilweinstock:
I don't have time to test the whole thing right now, but from the video and the descriptions it all looks good to me.  I'm still not sure the Scale function deserves to be in the contextual menu but I don't mind leaving it there.

Overall this is a really nice "quality-of-life" improvement in the program.

JoePfeiffer:
Very, very nice piece of work. lgtm

PR #1399, opened 2022-06-01
[#1397] Fix streamer packed length updating strip width

### LINKED ISSUE #1397: Cannot change packed length of a streamer
I'm unable to change packed length of a streamer, and attempting to do so changes the strip width to either 1mm wider or 1mm narrower than the unchangeable packed length. This is easy to see with the tube fin example which for some reason has a strip width of 50mm and a packed length of 25mm. Trying to change the packed length results in a strip width of either 24mm or 26mm, but a packed length of 25mm. Strip width can be manually changed larger or smaller, but packed length can never be changed from 25mm.

I'm building from source rather than using the beta, but from testing earlier builds I've archived, I can see I have this problem going back as early as 2022-05-02 15:12 UTC.

SiboVG:
Don't ask me why, but in the streamer the setLength method (which changes the packed length) was overridden to change the strip width. This PR fixes #1397.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Streamer.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Streamer.java
@@ -63,17 +63,6 @@ public class Streamer extends RecoveryDevice {
 		fireComponentChangeEvent(ComponentChangeEvent.BOTH_CHANGE);
 	}
 	
-	@Override
-	public void setLength(double length) {
-		for (RocketComponent listener : configListeners) {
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
That's actually not unreasonable -- wrapping the streamer around the shock cord is a not-uncommon way of packing, which results in a packed length the same as the streamer width.

Though we really do want to allow users to crumple the streamer up, which results in a packed length of who-knows-what....

PR #1402, opened 2022-06-02
update iteration through rocket components to account for inactive components and instances

### LINKED ISSUE #1356: Staging can show huge negative accelerations
Some staged rockets show huge negative accelerations at staging time (see the Three-Staged Rocket example, whose first stage shows a negative acceleration of over 1000 m/s^2 at burnout).

JoePfeiffer:
Fixes #1356 

Previously, the tumble simulation aerodynamic code didn't account for either inactive components nor components with multiple instances.  The result was the every aerodynamic component in the rocket was counted exactly once, so the entire rocket was considered in looking at drag on a booster stage.  Also, if there were pods, each pod would only be counted once.

While this addresses this issue, tumble recovery still shows very high decelerations.  After comparing the code with Sampo's thesis it appears to be correct, and the final landing speeds are reasonable.  I'm guessing that the aerodynamic model being used for the tumbling stages is simply not accurate at high speeds, and those big decelerations will continue until we can get a model that's more appropriate when the booster starts tumbling.  That's not likely to happen any time soon, since all we really care about is the landing velocity.  So I'm going to call that part of it "works as intended"

### DIFF
--- a/core/src/net/sf/openrocket/simulation/BasicTumbleStatus.java
+++ b/core/src/net/sf/openrocket/simulation/BasicTumbleStatus.java
@@ -1,9 +1,13 @@
 package net.sf.openrocket.simulation;
 
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.Map;
 
 import net.sf.openrocket.rocketcomponent.FinSet;
 import net.sf.openrocket.rocketcomponent.FlightConfiguration;
+import net.sf.openrocket.rocketcomponent.InstanceContext;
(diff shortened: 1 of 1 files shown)


SiboVG:
> So I'm going to call that part of it "works as intended"

"It's not a bug, it's a feature"

neilweinstock:
This seems reasonable and a good fix.  A fully detailed model of a tumbling booster seems like it would be a non-trivial project.  File under "nice-to-have someday".

PR #1404, opened 2022-06-03
Move series selection in sim plot to the beginning

SiboVG:
This PR moves the series selection in the sim pot more to the beginning. Because of the export image button, it looked like the combobox was tied to exporting the image instead of the plot.
Screenshot 2022-06-03 at 03 59 14


neilweinstock:
Yes please.

PR #1407, opened 2022-06-04
Uniform dialog for inner tube

hcraigmiller:
Uniform dialog for inner tube configuration pane, to match mass components.


SiboVG:
Could you give a comparison of what's changed?

hcraigmiller:
Comparison

SiboVG:
Ah I see, LGTM. Btw, you should update your unstable branch on GitHub, you're a bit behind (because I saw that the component material combobox has a small gap to the left, so I first thought there was still something that needs to be fixed, but I fixed that already a couple of commits ago, but it's just not yet updated in you repo)

PR #1414, opened 2022-06-07
Found another couple of instances of fins being multiplied by fincount

JoePfeiffer:


neilweinstock:
What would be the effects of these particular bugs?  Which calculations would they affect?

JoePfeiffer:
> What would be the effects of these particular bugs? Which calculations would they affect?
> 
> Interested mainly for release notes documentation purposes. Would be good to know if there is anything for users to watch out for.

tl;dr -- no reason to mention in the release notes.

They were in the code that applies roll forcing for canted fins, and dampens roll.  A user would only notice them if they had a rocket with canted fins (so there was some roll imparted) and looked very closely at a CSV data dump.  It would start to roll a little faster than it should, but then the damping code had the same code so from then on it would act pretty much the same as they'd cancel out.

Getting rid of it was more a bit of cleanup than anything else.

PR #1415, opened 2022-06-08
[Fixes #1351] Include ymin in 3D figure bounding box calculation

### LINKED ISSUE #1351: Size 3D views based upon length, height, and width
Currently the 3D views are based upon the length and height of the model (left views) The 3D views should also be based upon the width of the model (right views).

2022-05-12 X-15 Picture_03

Reported by: Robert McNamee, Facebook OpenRocket Group

SiboVG:
This PR fixes #1351 by including the ymin coordinate into the 3D figure bounding box calculation.

It seems to work, but @hcraigmiller could you do some testing to verify?

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6842006454/artifacts/263898421) for testing.


hcraigmiller:
It worked perfect for the original design used in the issue (left side), but not so much with this user provided design (right side).

Screen Grabs

Not certain how far we can go with unusual designs without a 3D zoom feature. I'm feeling like your fix is the best option until a 3D zoom feature can be added. Anyone else's thoughts, @neilweinstock and @JoePfeiffer too.

neilweinstock:
Gliders like that are a thorny problem. If OR set the zoom size to fit those wings, then when looking at a side view it would be puzzlingly small.   This is one of those cases where I don't know what the behavior *should* be.

Certainly, a zoom function would allow the user to override whenever the program made a bad choice about fitting the design into the available space, but the zoom function is likely to be a bit of a can of worms.

My question is: exactly what is the fix doing, if it is not sizing the image to fit those wings in the picture?  Is there a limit on how much it will adjust?

hcraigmiller:
Spoke too soon on whether it functions correctly. Cutting down A simple model rocket... not so much either.

Short-Fat

[Short-Fat.ork.txt](https://github.com/openrocket/openrocket/files/8872823/Short-Fat.ork.txt)

SiboVG:
I'm really confused... Works fine for me
image

SiboVG:
Could you test again with [this jar file](https://github.com/openrocket/openrocket/suites/6867679268/artifacts/265549666) @hcraigmiller?

SiboVG:
Could you also send me the first design file that didn't work?

hcraigmiller:
No change. If you cut down A simple model rocket by changing its length to 10 cm and change the fin count to two at 90 degrees rotation, you can more easily see that in 3D the fins are beyond the bounds of the display window.

Glider
[FlatCat.ork.txt](https://github.com/openrocket/openrocket/files/8872900/FlatCat.ork.txt)

SiboVG:
Yeah it's what I thought, the resizing just doesn't happen properly if you rotate the model. If you change the radial rotation of the fins, you see that the resizing is correct.

Changed fin rotation:
image

neilweinstock:
Does it solve #1351?  I am confused what it does or does not do right now.

SiboVG:
> Does it solve #1351? I am confused what it does or does not do right now.

Okay, to see what this PR does: open 'A simple model rocket' in beta 03, and increase the height of the fins. You will see that the fins go outside of the rocket view. This PR solves that issue.

hcraigmiller:
I'm not sure I understand what you mean. Are you suggesting that as the model is rotating, the scale changes. This issue intended to have the greatest distance from the center of the rocket to the outer furthest away edge (plus a margin) as the height of the display area.

hcraigmiller:
I'm wondering if this is a screen proportion issue... a movie screen would be much different than a television screen in proportion to width and height?

SiboVG:
> I'm not sure I understand what you mean. Are you suggesting that as the model is rotating, the scale changes. This issue intended to have the greatest distance from the center of the rocket to the outer furthest away edge (plus a margin) as the height of the display area.

No. When you rotate the model in the 3D rocket view, you are simple rotating the "camera", but the zoom of the rocket view is only calculated based on the side view of the rocket. The zoom does not update when you simply rotate the model by dragging your mouse along the rocket view, because you are only changing the camera's view. The zoom does update when you change **actual components** of the rocket. E.g. if you changed the fin rotation, because in that case you are changing the rotation of the rocket in its side view, which the zoom is based on.

SiboVG:
To remove any confusion: this PR solves the zoom in the 3D rocket view if you **do not change the view rotation** in the 3D view. So if you don't touch anything in the 3D view, it works.

Now what I was proposing is to create a new issue because if you **do change the view rotation** (dus click and drag your mouse in the 3D rocket view), the scale of the window is **not updated**. It just calculates the scale ones for the rocket in its side view. It does not update when you change the view rotation.

hcraigmiller:
That's what I hoped.

But, a change in rotation would not change the zoom if the zoom is based on the greatest distance from the center of the rocket to the outer furthest away edge (plus a margin) as the height of the display area, rather than just the side view height..

SiboVG:
Okay last commit changes the view's zoom based on the maximum rocket radius.

hcraigmiller:
Looks good...

PR #1417, opened 2022-06-08
[#1231] Include non-aerodynamic components in bounding box calculations

### LINKED ISSUE #1231: "fit" doesn't always work in side view
"fit" doesn't always work in side view. The ends sometimes go off screen and you can pan over to see end. See exampl
[sjh_trident.zip](https://github.com/openrocket/openrocket/files/8200848/sjh_trident.zip)
e

SiboVG:
This PR fixes #1231 by including non-aerodynamic components in the rocket's bounding box calculations. 

I first just deleted the aerodynamic condition in the bounding box calculations, but most parts of the program (including some calculations) need the bounding box or rocket length of only aerodynamic components. Just deleting the aerodynamic condition caused changes in the calculations.

So I ended up splitting the bounding box and rocket length in two variations: only aerodynamic components and all components.

This PR not only fixes #1231, but now the rocket length displayed in the rocket view's top left corner includes all components, not only the length of the aerodynamic components.

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6851577921/artifacts/264497891) for testing.


hcraigmiller:
Functions beautifully, no anomalies found as to this issue... but see Issue #1420.

PR #1424, opened 2022-06-10
[#1423] Make sure zoom combobox can fit 'Fit (100.0%)'

### LINKED ISSUE #1423: Bug: 'Fit' text does not fit in spinner when zoom value > 100%
When the fit value in the rocket view is greater than 100%, the fit text is cut off on macOS. You can recreate this by just adding a body tube: 

image

Tested on macOS 12.4, MacBook Pro M1, latest unstable.

SiboVG:
This PR fixes #1423 by giving the zoom combobox an initial size so that it can fit the text 'Fit (100.0%)' (thus all text in the form of 'Fit (xxx.x%)'.

image

I tried making the combobox size dynamically based on its content, but was unsuccessful.


hcraigmiller:

Issue #1037, opened 2021-09-27
New Feature Request: Ability for custom number of decimals on exporting simulations


SiboVG:
This issue is related to [comments in #1036](https://github.com/openrocket/openrocket/issues/1036#issuecomment-927984493).

Right now, when exporting the simulation data to CSV (go in the simulations tab, select a simulation, click the 'Plot / export'-button and in the 'Export data'-tab exporting the simulation), the variables have a fixed number of decimals (in the case of the comments to #1036, a maximum of 3 digits). Sometimes it's necessary for users to have more precise values reported back to them. Because OR internally uses more precise values than those that are eventually exported, it is merely a matter of implementing a **new feature for users to be able to control the maximum number of decimals used upon export**.

This feature could be implemented as a JSpinner in the 'Export data'-tab in which the user can set the desired number of decimals.

Sparrowhawk120:
Solved in #1427 , can we close this?

SiboVG:
> Solved in #1427 , can we close this?

Yes

PR #1427, opened 2022-06-11
[fixes #1354 & #1037] Add support for custom decimal places and exponential notation in simulation exports

### LINKED ISSUE #1354: Export of simulated latitude and longitude needs more decimals
When doing a simulation and afterwards exporting latitude and longitude, only three decimals are exported, corresponding to a rectangular area 111 meters wide. This is a bit too coarse for my use, at least - why not just two more decimals, which would create a box just 1 meter wide.

Also, the text boxes where latitude and longitude are inputted have a strange behaviour (Ubuntu 22.04, v. 22.02.beta.01) where it's possible to paste e.g. a line break into the box.

Feature request: Export directly as KML-file for visualization of 3D flight path in Google Earth.

Feature request: Built-in 2D top-down map-view for easy viewing of flight ground track.

SiboVG:
This PR fixes #1354 and #1037. I first implemented an increased, static decimal place for the latitude and longitude (for issue #1354) - which is the first commit which you can just ignore - but then found that it wasn't much extra effort to implement #1037 and add support to enter a custom number of decimal places for the simulation export variables. So that solves #1354 automatically, because users can now just choose the decimal places for any variable.

While I was at it, I also added a checkbox to decide whether to use exponential notation in the export (e.g. '1.5e-4') or not (e.g. '0.00015').

The simulation export panel thus looks like this (see the top right, 'Format settings'):
image

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6888134264/artifacts/266851250) for testing.


hcraigmiller:

Sparrowhawk120:
Is there a way to speed up the export? The crude solution from #1037 was a bit faster at full decimal length (which was like 15 places), but it hangs for about a minute at just 6 decimal places with this pull

neilweinstock:
Do you actually desire to export 6 decimal places?  If so I would be intrigued to hear why.  It's hard for me to understand how that would be valuable.

SiboVG:
> Is there a way to speed up the export? The crude solution from #1037 was a bit faster at full decimal length (which was like 15 places), but it hangs for about a minute at just 6 decimal places with this pull

Well two things: your code from #1037 did not include the `exponentialFormat` calculation, which saves your code some while-loops + I just continued to use the `shortDecimal` method inside `decimalFormat` while you just formatted the string as a full float. I must say I'm not entirely sure why the `shortDecimal` is so elaborate and what advantage it brings to using something built-in from Java. I'll take a look at that.

But I do not see export delays longer than half a second, so I'm a bit confused why you have such huge delays? I used 'A simple model rocket' and exported all variables with 15 decimal precision and got an almost instant export.

SiboVG:
@Sparrowhawk120 I made some changes to the doubleToString code, hopefully resulting in a performance improvement for you. Could you check with [this jar file](https://github.com/openrocket/openrocket/suites/6901378165/artifacts/267833775) whether anything has changed for you?

Sparrowhawk120:
@SiboVG I timed an export of Time, Lateral Dist, Lat, & Lon, and it still took 59 seconds to do an export at 6 decimal places with comma delimiting.

Sparrowhawk120:
> Do you actually desire to export 6 decimal places? If so I would be intrigued to hear why. It's hard for me to understand how that would be valuable.

@neilweinstock I'm trying to start developing a visualizer for the 6DOF so I can use the Monte-Carlo method to create a landing splash zone. This would help for much larger-scale flights requiring FAA permits without the irritating monthly subscription of Rocksim Pro

SiboVG:
> @SiboVG I timed an export of Time, Lateral Dist, Lat, & Lon, and it still took 59 seconds to do an export at 6 decimal places with comma delimiting.

How are export times for 22.02.beta.03? And do the results differ when you check/uncheck the use exponential notation? Also: what are the export times for your code from #1037?

SiboVG:
Oh yeah and do the results change with the latest unstable whether you export with 6 decimals or 15 decimals?

Sparrowhawk120:
> How are export times for 22.02.beta.03? And do the results differ when you check/uncheck the use exponential notation? Also: what are the export times for your code from #1037?

I'm gonna preface this by saying the following results are incredibly perplexing to me because both with #1037 and the original 22.02.beta.03 I didn't previously have much issue with exporting. I'm not sure if this is a product of running the jar with Java 17, or if something is wrong with my system in some other way.

In the beta without the custom places merge (22.02.beta.03) it took 60 seconds to export.
56 seconds with exponential notation on your jar file (6 decimal).
77 seconds with 15 decimals on your jar file (no exponential).
------------------------------
Similar behavior was observed using the code in #1037 
------------------------------
Similar behavior was observed using both my .ork file and the "Simple Model Rocket" example

SiboVG:
Thanks for the test results. It's indeed very strange that you have this issue. As I've said, I don't have any issues with it at all (export times all < 1 s). It may have something to do with the fact that we don't use the OpenCSV library for exporting to CSV, which may be more efficient than the custom CSV export (I know from a prior project that OpenCSV makes huge changes in delay times for reading CSVs, so maybe the same holds for writing CSVs).

I'll file a new issue for it, thanks again for bringing it up.

PR #1431, opened 2022-06-11
[Fixes #753] Update ScaleScrollPane size when rocket changes

### LINKED ISSUE #753: Side/Back View "Fit" does not increase in size (percentage) as length is shortened
After opening a design in the side view, if you stretch the rocket (increase its  length), then the view auto sizes to fit the increased length. If you then shrink the rocket (return it to its original length), the view does not auto size to fit the shorter length. The only way to get the "fit" to work again is to close and reopen the design. The same occurs with the back view.

Open
Stretch
Shrink

The freeform fin display auto fits both directions.

SiboVG:
This PR fixes #753 where the rocket view did not resize when the rocket became shorter. The ScaleScrollPane componentResized just gets triggered whenever the rocket changes (may be not very efficient when parts of the rocket change that do not affect the rocket compound size, but shouldn't be too bad, there is a check somewhere deep that checks whether the bounds actually changed before updating the rocket view).

New behavior:

https://user-images.githubusercontent.com/11031519/173206868-33f05ccc-1a33-4594-83ea-103529209090.mov

Here is a [jar file](https://github.com/openrocket/openrocket/suites/6893520699/artifacts/267268666) for testing.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
+++ b/swing/src/net/sf/openrocket/gui/scalefigure/RocketPanel.java
@@ -219,6 +219,7 @@ public class RocketPanel extends JPanel implements TreeSelectionListener, Change
 			public void stateChanged(EventObject e) {
 				updateExtras();
 				updateFigures();
+				scrollPane.componentResized(null);	// Triggers a resize so that when the rocket becomes smaller, the scrollPane updates its size
 			}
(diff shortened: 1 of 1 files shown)


neilweinstock:
Nice

hcraigmiller:
Looking good...

PR #1432, opened 2022-06-11
Revised installer configuration to enable desktop icon and file associations

### LINKED ISSUE #1135: Get file associations working correctly on packaged installers
Right now, file association does not seem to work correctly in the packaged installers.  It is unclear what the problem is right now.

This would be a good thing to get working before the final release, if possible.

JoePfeiffer:
The major change is it turns out you need to do the file installation before trying to create the integrations.

Installers to test in https://drive.google.com/drive/folders/1xaUZXnl9q6CDG-CbkqIQqpFrBz63Gw2k?usp=sharing

Fixes #1135

### DIFF
--- a/install4j/22.xx/openrocket-22.xx.install4j
+++ b/install4j/22.xx/openrocket-22.xx.install4j
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <install4j version="9.0.6" transformSequenceNumber="9">
-  <directoryPresets config="./macOS_resources" />
+  <directoryPresets config="../../core/resources-src/pix/icon" />
   <application name="OpenRocket 22.02.beta.03" applicationId="8434-9327-1469-6373" mediaDir="media" shortName="OpenRocket" publisher="OpenRocket" publisherWeb="http://openrocket.info" version="22.02.beta.03" allPathsRelative="true" macVolumeId="5f58a2be20d8e22f" javaMinVersion="11" javaMaxVersion="11" jdkMode="jdk" jdkName="JDK 11.0">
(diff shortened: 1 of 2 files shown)


neilweinstock:
On the Mac it still does not work properly.  I believe behavior is unchanged: double-clicking an ORK file in the Finder will activate OR, but it will not open the file.

So my interpretation is that the file type is correctly associated in the OS, but the OR code is not properly handling whatever sort of message the OS is sending it to open the file.

hcraigmiller:
Worked perfect on windows, desktop icon, file associations, all good.

neilweinstock:
That said... if file associations work correctly only on Windows that would still be a useful improvement for beta 4.  Mac can be fixed later.

neilweinstock:
https://docs.oracle.com/javase/tutorial/deployment/selfContainedApps/fileassociation.html

JoePfeiffer:
They're also working on linux.  Thanks for the link; I'll see if I can use that to get things working on MacOS.  Naturally, that's the only one I can't test on!

neilweinstock:
Here for ya. :)

JoePfeiffer:
Is it successfully creating a desktop icon for MacOS, even if the file association isn't working?

Briefly looking at that web page, it looks like somebody else's conclusion that it'll take some changes in OR itself and a different main method to get associations working.

neilweinstock:
I don't see a desktop icon, but that's not a thing for the Mac so I wouldn't expect it.  I am seeing proper OR icons in the Finder for the .ork files, so that's good.  I think the only remaining issue is the need for the special code in the main method.

PR #1436, opened 2022-06-12
Feature Request 1373 - Adds a warning when there is no recovery device for a simulation

ptsongalis:


core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
```diff
@@ -547,7 +547,25 @@ public void update(){
 		updateMotors();
 		updateActiveInstances();
 	}
-	
+
+	/**
```
SiboVG, line 551: To make this more general, I would put the `hasRecoveryDevice()` method inside `core/src/net/sf/openrocket/rocketcomponent/RocketComponent.java` after line 2030. A Rocket is a RocketComponent, so the rocket would automatically have the `hasRecoveryDevice()` method as well. But having it inside RocketComponent.java gives us the opportunity to use it for other future purposes.

I would also use the rocket iterator. So put this inside `RocketComponent.java`:
```Java
/**
 * Return true if any of this component's children are a RecoveryDevice
 */
public boolean hasRecoveryDevice() {
  Iterator<RocketComponent> iterator = this.iterator();
  while (iterator.hasNext()) {
    RocketComponent child = iterator.next();
    if (child instanceof RecoveryDevice) {
      return true;
    }
  }
  return false;
}
```

And then inside `FlightConfiguration.java`:
```Java
/**
 * Return true if rocket has a RecoveryDevice
 */
public boolean hasRecoveryDevice() {
  checkState();
  if (fcid.hasError()) {
    return false;
  }

  return this.getRocket().hasRecoveryDevice();
}
```

(the checkState stuff is just some condition checking of the rocket that I copied from `hasMotors()`)

SiboVG (changes requested):

core/src/net/sf/openrocket/rocketcomponent/FlightConfiguration.java
ptsongalis: Gotcha! Looks like checkState() is not defined in FlightConfiguration, should I not be including that?

SiboVG: Yeah sorry my bad, looked at a wrong piece of code, you can just ignore the `checkstate()` in that case

SiboVG (approved):

SiboVG (approved):

SiboVG:
LGTM, thanks for the changes!

PR #1438, opened 2022-06-13
Improve double to string rounding code

SiboVG:
The previous code was just a bit odd and non-standard + variable (and probably inefficient). Decided to replace it with standard Java functions instead (+ some other custom formatting methods). This may also improve the [inefficiency noted by @Sparrowhawk120](https://github.com/openrocket/openrocket/pull/1427#issuecomment-1153324209).

What changed concretely:

- When no exponential notation is needed, this PR changes nothing, you still get the same results, albeith written more efficiently (hopefully)
- When exponential notation is needed, the outcome will be exponential if the value is < 0.001 or ≥ 10000. The previous code was quite variable in this, sometimes switching between decimal notation and exponential based on the length of the result or some other factors. I prefer going for "You know what you will get".

### DIFF
--- a/core/src/net/sf/openrocket/util/TextUtil.java
+++ b/core/src/net/sf/openrocket/util/TextUtil.java
@@ -2,10 +2,12 @@ package net.sf.openrocket.util;
 
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.text.DecimalFormat;
+import java.util.Locale;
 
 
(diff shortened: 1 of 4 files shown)


SiboVG:
Okay it turns out it only cuts away a couple of seconds from the large export times of Sparrowhawk, but still this PR offers more predictable results (in terms of how the rounding and notation happens) than the previous implementation.

PR #1444, opened 2022-06-13
[#974] Add component edit window tab remembrance

### LINKED ISSUE #974: New feature request: Moving from component to component, configuration tab selected should open to the same tab
Users commonly change a single characteristic on a number of components, especially those related to appearance, such as color and shine. Currently, as the user selects one component, then another, and so on, when each configuration dialog opens the default tab "General" is active. It would be much easier if, for example, the shine of the rocket is being changed from 0% to 50%, when the nosecone appearance tab is active and you move to the body tube that the configuration dialog opens to the appearance tab, as would transitions and fins, etc.

The trade off would be that you would have to select the "General" tab to get back to the basic data, but my experience is that doing so would be much less frequent than the clicks saved by going to the same tab between components.

New components would open the "general" tab, as would selecting a component that did not have the selected tab from the prior component.

SiboVG:
This PR fixes #974: Moving from component to component, configuration tab selected could open to the same tab.

The current behavior is this:

- If a tab is selected the component config dialog of component 1, and you edit another component, component 2 (either by clicking component 2 while the edit dialog of component 1 is still open, or by first closing component 1's edit dialog, and then editing component 2):
- 
  - If the previously selected tab of component 1 is also present for component 2 (e.g. both components have a ‘General’ tab or ‘Appearance’ tab), then that same tab is opened for component 2
  
  - If component 2 does not have the tab of component 1, the first tab of component 2 is selected (e.g. component 1 is a FinSet with the ‘Fin tabs’ tab selected, component 2 is a Body Tube, then the first tab of component 2 - ‘General’ - will be opened)

- If a new component is added, the automatic dialog pop-up will always be opened in the first tab (e.g. ‘General’)

Demo:

https://user-images.githubusercontent.com/11031519/173460087-d194d033-ce60-4345-afbb-4ae5f00a93e3.mp4


SiboVG:
I'm wondering though: should a user be able to disable this new behavior in the system preferences or do we want to make this default behavior?

neilweinstock:
I would definitely make it the default behavior... hide it in preferences and no one will ever notice it.  Those that are *bothered* by the new behavior might be motivated to go find the preference (or ask on the forum, or whatever).

I think it's worth a checkbox in preferences.  If it's easier to have the default options be *disabled* by default, or if it's too darn hard to come up with good wording for the option, then you can do it opposite: have the option be "Always open leftmost tab when opening a component edit dialog", unchecked by default.

hcraigmiller:
My preference would be to have the new behavior be the default behavior.

SiboVG:
Latest commits add option in preference panel (couldn't find a good _enabled_ wording, so I went for the _disabled_ wording proposed by Neil)
image

neilweinstock:
Typo in tooltip:  almost -> always

hcraigmiller:
I would change "leftmost" to "General".

SiboVG:
> Typo in tooltip: almost -> always

My bad, it was too late when I made that change

SiboVG:
> I would change "leftmost" to "General", ditto on @neilweinstock.

I would not do this because it's not always General, for stages and boosters it's 'override' and 'separation' respectively.

PR #1452, opened 2022-06-16
[fixes #1100] Implement macOS idle application mode after all windows are closed

### LINKED ISSUE #1100: Feature request: on Mac, implement standard multi-document behavior
This request applies to Mac only, because it's a much clearer situation than on Windows (I'm unsure about Linux).  Frankly, I would want this behavior no matter what platform I was working on, but only the Mac really standardizes this behavior.

On the Mac, an app that supports multiple open documents (such as OR) is expected to behave as follows:

1) Remain running even when all document windows are closed

2) If all windows are closed, and the app is activated from background -> foreground (e.g. by clicking the icon in the dock), there are two options: 
 a) Open the file chooser
 b) Open a new blank document

Implementing this behavior in OR would make it much friendlier on the Mac.

__Discussion__
Regarding number 1:
Current behavior is quite frustrating.  Usually, I'm only working on a single rocket at at time.  Therefore, I want to close one rocket before opening the next.  Right now, that closes OR, and I swear at it every single time, because then I have to re-open OR (which takes a while using the packaged version), then open the file I want to view, then close the window that was open previously.  Over and over this repeats.  By letting the app stay open after the last window is closed (again, standard Mac behavior), I can do what makes most sense: close one app, then open another.

To save my sanity, what I try to do nowadays is keep a garbage window always open and made as small as possible and tucked into a corner of the screen, so the program won't close.  This is kind of ridiculous.

Regarding number 2:
Either one of these is acceptable I think, but in my own usage I am much more likely to want to open an existing file than create a new one.  So if it were up to me, I would choose to see the file open dialog whenever either (a) the app is first opened and there isn't a last-open file to open, or (b) app is brought to foreground and there are no windows open.  If what you actually want is a new rocket, then you dismiss the file dialog (app stays running!) and then File -> New.

The best modification of all is if you can also have a "Create new rocket" button added to the File Open dialog, so right there you have the choice of how to proceed.  This behavior is, regrettably, much less common, despite being ideal.  I don't know how difficult it is to implement, though.  It is _nice_ but not required.

SiboVG:
This PR fixes #1100 by making OR go into idle mode when all the design windows are closed, instead of quitting the application. In idle mode, an alternative menubar is present (that does not contain the edit tab etc.).

Demo:

https://user-images.githubusercontent.com/11031519/174087249-536fd7f6-b924-4899-b953-7e36d16ac627.mp4

Small sidenote: it seems like even in idle mode OR still uses a significant amount or memory, but that's just due to the horrible memory management of OR in general...

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
+++ b/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
@@ -34,6 +34,7 @@ import javax.swing.tree.TreeSelectionModel;
 import net.miginfocom.swing.MigLayout;
 import net.sf.openrocket.aerodynamics.WarningSet;
 import net.sf.openrocket.appearance.DecalImage;
+import net.sf.openrocket.arch.SystemInfo;
 import net.sf.openrocket.document.OpenRocketDocument;
 import net.sf.openrocket.document.OpenRocketDocumentFactory;
 import net.sf.openrocket.document.StorageOptions;
@@ -1687,7 +1688,8 @@ public class BasicFrame extends JFrame {
 		ComponentAnalysisDialog.hideDialog();
 
 		frames.remove(this);
-		if (frames.isEmpty()) {
+		// Don't quit the application on macOS
+		if (frames.isEmpty() && (SystemInfo.getPlatform() != SystemInfo.Platform.MAC_OS)) {
 			log.info("Last frame closed, exiting");
 			System.exit(0);
 		}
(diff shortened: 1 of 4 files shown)


hcraigmiller:
Uncaught Exception

[Windows 11 Pro; Version 21H2; Build 22000.675; Feature Experience Pack 1000.22000.675.0]
[Java version "11.0.15" 2022-04-19 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.15+8-LTS-149)]

SiboVG:
Whoops my mistake, forgot to add a clause to only add the OpenFileHandler when you're on a mac, good thing you tested it @hcraigmiller 😳

hcraigmiller:
I'm assuming that this is solely an Apple related issue. If so, no anomalies found. If not, what should I be looking for in Windows?

[Windows 11 Pro; Version 21H2; Build 22000.675; Feature Experience Pack 1000.22000.675.0]
[Java version "11.0.15" 2022-04-19 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.15+8-LTS-149)]

SiboVG:
Yes, this is only for macOS.

PR #1462, opened 2022-06-18
Update Three-stage rocket example

hcraigmiller:
Changed sustainer motor delays to avoid recovery device deployment at high speed warnings.


SiboVG:
@neilweinstock you know more about rocket design, care to give this PR a check?

neilweinstock:
Will do so as soon as I get a chance.  I don't foresee any problems; Craig knows this stuff as well (probably better) as I do.

PR #1463, opened 2022-06-18
Update Apocalypse example

hcraigmiller:
Changed nose cone and transition shoulder diameters to be consistent with body tube inner diameter; changed parachute characteristics to reduce decent to appropriate rate; moved forward and middle centering rings to ends of fins; changed fin tabs to be consistent with 15.03; rotated launch lugs so that rod is not going through fins.


SiboVG:
@neilweinstock you know more about rocket design, care to give this PR a check?

PR #1466, opened 2022-06-19
[#1465] Include xmin in sideview canvas origin calculation

### LINKED ISSUE #1465: [Bug] Side view centering not correct when child component goes outside of parent
Fit in side view still broken. See ork and image

[Mercury.zip](https://github.com/openrocket/openrocket/files/8935898/Mercury.zip)
MercuryFit

SiboVG:
This PR fixes #1465 by including the rocket's xmin position into its sideview origin calculation. For normal rockets, xmin should equal 0, thus nothing should have changed for them. For rockets where some of its child components reach further to the left than their parent component, xmin becomes smaller than zero, which should then also be included in the origin calculation. The design file from #1465 was an example of the child components reaching further to the left. A body tube contained a pod set that was positioned to the left of its parent body tube. Another way of recreating the issue from #1465 is by opening 'A simple model rocket' and moving the launch lug all the way to the left.

image


hcraigmiller:
Works well, no anomalies found,

PR #1483, opened 2022-06-25
Dependency Updates

thzero:
Refreshed the test and the non-jakarta based (XML) dependencies.   Also removed the Yasson instead of updating it, as it was not used anywhere in the code base.


hcraigmiller:
I'm not quite sure what functionality needs to be tested, but a general test of all components showed no anomalies.

thzero:
The two user space ones are
commonmark - this is used for converting markdown document syntax into html
rsyntaxtextarea - text areas

Used in the logging..
logback-classic
logback-core

The test dependencies are only used when running tests during build, etc.

PR #1493, opened 2022-06-26
XMLReaderFactory.createXMLReader Deprecation

thzero:
The XMLReaderFactory.createXMLReader() factory has been marked as deprecated.   Noticed this as part of starting to clean up the save/importt/export functionality to hopefully remove the jakarata (Java EE) dependencies from the code base.

So this PR is just to update the createXmlReader to using the SAXParserFactory to create an instance of XmlReader.

It runs through all the tests on Windows 10 and Linux (Ubuntu) fine; and any testing of opening/saving both OR and Rocksim seems to work fine through the tool.

Probably some additional testing is warranted.


SiboVG:
Seems to work well, thanks for the change!

PR #1495, opened 2022-06-27
Update Three-stage rocket example

hcraigmiller:
Corrected design error related to staging couplers.

Relocated centering rings at each stage to be placed at the top/bottom edge of the staging couplers (no overlap).


SiboVG:
Could it be that you didn't save the file with all simulation data saved? (because OR re-ran the simulation when I went to the sim tab)

hcraigmiller:
Yes, I’ll redo do it later tonight.

PR #1500, opened 2022-06-27
[#1487] Fix configuration referencing after loadFrom rocket

### LINKED ISSUE #1487: Bug: Rocket does not remember stage activeness after deleting a stage and undoing that deletion
Steps to reproduce:
1. Open 'Three-staged rocket' example design
2. Disable the final booster stage by clicking on its button in the stage selector panel (above the rocket view)
3. The configuration in the combobox on the top right above the rocket view has changed and if you click on that combobox, you can still see the configuration change is applied
4. Delete the last booster stage
5. Undo that deletion
6. If you now disable any of the stages by clicking the stage selector buttons, you will see that on the top right, the configuration is again changed in the combobox, as you would expect, but if you click on the combobox, the original configuration is displayed (C6-7;C6-0;C6-0) instead of (C6-7;C6-0). It's not just a display issue, because if you would now switch to another configuration and then switch back to the previous configuration, you will notice that the last booster stage is not disabled anymore.

Demo:

https://user-images.githubusercontent.com/11031519/175786557-e73d8b07-9960-48bd-b8e7-0a9de96cd7dc.mp4

Edit: this behavior is not only related to deleting a stage; deleting any component and then undoing that deleting brings up this bug.

SiboVG:
This PR fixes #1487 by updating the selected configuration of a rocket in the loadFrom method (called after an undo action). There was then still an issue that when updating the figures, a NullPointerException was thrown because RocketInfo still used an old configuration, which is solved by updating stabilityUnits in RocketInfo when the configuration is changed.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
@@ -390,10 +390,12 @@ public class Rocket extends ComponentAssembly {
 		this.stageMap = r.stageMap;		
 
 		// these flight configurations need to reference the _this_ Rocket:
+		this.configSet.reset();
 		this.configSet.setDefault(new FlightConfiguration(this));
 		for (FlightConfigurationId key : r.configSet.map.keySet()) {
 			this.configSet.set(key, new FlightConfiguration(this, key));
 		}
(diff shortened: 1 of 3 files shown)


hcraigmiller:

Build 767

PR #1478, opened 2022-06-24
[#1477 & #1460] Don't include inactive stages in calculations

SiboVG:
This PR fixes #1477 by not including inactive stages in calculateForceAnalysis.

Note: unintentionally this is already the start of fixing #1460, so I will leave this PR as a draft and add more commits as I go.

Issue #1477, opened 2022-06-23
[Bug] NullPointerException in the component analysis window when disabling a stage 


SiboVG:
I get a NullPointerException when disabling the stage in the component analysis window in 'A simple model rocket'.

Steps to reproduce:
1. Open 'A simple model rocket'
2. Go to the Component Analysis window
3. Click on the 'Sustainer' button, *bam* you get the error

<details>
<summary>Bug Report (click on me to open)</summary>

---------- Bug report ----------

Please include a description about what actions you were performing when the exception occurred:
(You can edit text directly in this window)

1. Open 'A simple model rocket'
2. Go to the Component Analysis window
3. Click on the 'Sustainer' button, *bam* you get this
Include your email address (optional; it helps if we can contact you in case we need additional information):

(Do not modify anything below this line.)
---------- Exception stack trace ----------
java.lang.NullPointerException
at net.sf.openrocket.aerodynamics.BarrowmanCalculator.calculateComponentNonAxialForces(BarrowmanCalculator.java:207)
at net.sf.openrocket.aerodynamics.BarrowmanCalculator.calculateForceAnalysis(BarrowmanCalculator.java:148)
at net.sf.openrocket.aerodynamics.BarrowmanCalculator.calculateForceAnalysis(BarrowmanCalculator.java:156)
at net.sf.openrocket.aerodynamics.BarrowmanCalculator.calculateForceAnalysis(BarrowmanCalculator.java:156)
at net.sf.openrocket.aerodynamics.BarrowmanCalculator.getForceAnalysis(BarrowmanCalculator.java:83)
at net.sf.openrocket.gui.dialogs.ComponentAnalysisDialog.stateChanged(ComponentAnalysisDialog.java:535)
at net.sf.openrocket.gui.adaptors.DoubleModel.fireStateChanged(DoubleModel.java:985)
at net.sf.openrocket.gui.adaptors.DoubleModel.setValue(DoubleModel.java:791)
at net.sf.openrocket.gui.dialogs.ComponentAnalysisDialog.stateChanged(ComponentAnalysisDialog.java:526)
at net.sf.openrocket.util.ComponentChangeAdapter.componentChanged(ComponentChangeAdapter.java:20)
at net.sf.openrocket.rocketcomponent.Rocket.notifyAllListeners(Rocket.java:545)
at net.sf.openrocket.rocketcomponent.Rocket.fireComponentChangeEvent(Rocket.java:491)
[...]

PR #1504, opened 2022-06-30
[#1440] Add all context drags to component drag

### LINKED ISSUE #1440: Component Analysis window isn't properly reporting multiple instances of components
Fin set friction drag seems to be bugged after the refactor #1280. When changing fin count, the total Fin Set friction drag does not change.

```java
// iterate across component instances
			final ArrayList<InstanceContext> contextList = entry.getValue();
			for(InstanceContext context: contextList ) {
				double componentFrictionCD = calcMap.get(c).calculateFrictionCD(conditions, componentCf, warningSet);

				if (c instanceof SymmetricComponent) {
					SymmetricComponent s = (SymmetricComponent) c;
					bodyFrictionCD += componentFrictionCD;

					final double componentMinX = context.getLocation().x;
					minX = Math.min(minX, componentMinX);

					final double componentMaxX = componentMinX + c.getLength();
					maxX = Math.max(maxX, componentMaxX);

					final double componentMaxR = Math.max(s.getForeRadius(), s.getAftRadius());
					maxR = Math.max(maxR, componentMaxR);
					
				} else {
					otherFrictionCD += componentFrictionCD;
				}

				if (map != null) {
					map.get(c).setFrictionCD(componentFrictionCD);
				}
			}
```
I'm not sure if OR treats each fin as a component, but if it treats them as a FinSet I think this is iterating only once per fin set.
In FinSetCalc, a single fin planform area is being used:

```java
public double calculateFrictionCD(FlightConditions conditions, double componentCf, WarningSet warnings) {
		double cd = componentCf * (1 + 2 * thickness / macLength) * 2 * finArea / conditions.getRefArea();
		return cd;
	}
```

SiboVG:
This PR fixes #1440 by including all context drags to the component drag (for friction and pressure CD) instead of simply setting the component drag to the last context's drag.

FinSets should now display the correct CD:
image

### DIFF
--- a/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/net/sf/openrocket/aerodynamics/BarrowmanCalculator.java
@@ -390,6 +390,7 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 			}
 			
 			// iterate across component instances
+			double componentFrictionCDTotal = 0;		// Total friction drag of the component
 			final ArrayList<InstanceContext> contextList = entry.getValue();
 			for(InstanceContext context: contextList ) {
 				double componentFrictionCD = calcMap.get(c).calculateFrictionCD(conditions, componentCf, warningSet);
@@ -411,9 +412,11 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 					otherFrictionCD += componentFrictionCD;
 				}
 
-				if (map != null) {
-					map.get(c).setFrictionCD(componentFrictionCD);
-				}
+				componentFrictionCDTotal += componentFrictionCD;
+			}
+
+			if (map != null) {
+				map.get(c).setFrictionCD(componentFrictionCDTotal);
 			}
 		}
 		
@@ -590,17 +593,17 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 				continue;
 
 			// iterate across component instances
+			double pressureCDTotal = 0;		// Total pressure drag for this component
 			final ArrayList<InstanceContext> contextList = entry.getValue();
 			for(InstanceContext context: contextList ) {
 
 				// Pressure drag
-				double cd = calcMap.get(c).calculatePressureCD(conditions, stagnation, base,
+				double pressureCD = calcMap.get(c).calculatePressureCD(conditions, stagnation, base,
 															   warningSet);
-				total += cd;
-				
-				if (forceMap != null) {
-					forceMap.get(c).setPressureCD(cd);
-				}
+				pressureCDTotal += pressureCD;
+
+				// Total pressure drag of the rocket
+				total += pressureCD;
 				
 				if(c.isCDOverridden())
 					continue;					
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Single stage rockets: 
Without confirming the accuracy of the values (just comparing to pre-PR designs), this PR appears to function as expected without any anomalous behavior; single fin rockets do not always have matching Cd values.

Multi-stage rockets:
Changing fin count with open analysis window or turning a stage on/off.

Multi-Stage

SiboVG:
> or turning a stage on/off.

That issue was already addressed in #1477 and fixed in #1478. The exception when changing the fin count is however new, I'll look into it.

SiboVG:
@hcraigmiller could you elaborate on how I can recreate the NullPointerException when changing the fin count?

My steps:
1. Open 'Three-staged rocket' example file
2. Open Component Analysis window
3. Move to 'Drag characteristics' tab
4. Open the fins from the sustainer stage
5. Change the fin count
No error... Is it possible you first turned of the stage, turned it back on and then changed the fin count? In that case, I get the exception, but that has nothing to do with this PR, and is because of #1477

hcraigmiller:
After further testing, simplified steps (using Build 770):

1. Open _Three-staged rocket_ example (closing _Rocket configuration_ pane),
2. Open _Tools_ -> _Component analysis_ pane, 
3. Left-click _Drag characteristics_, then
4. On that pane (or the display stages pane), turn off any stage in any order and the error occurs.

The error only occurs when turning a stage off; when turned back on no error occurs.

I concur that the error has nothing to do with the fin count, it was just coincidence. I will be testing further.

SiboVG:
> 4\. On that pane (or the display stages pane), turn off any stage in any order and the error occurs.

Yes, so that's #1477, which you can just ignore. So for your further testing please do not touch the stage selector buttons :)

hcraigmiller:
Functions as expected, not so much of slight anomaly as a question found.

It appears that the pre-total Cd results are not being mathematically rounded at .001 before the percentage of the total is calculated. Because of this, the total Cd percentage can be a full percentage point higher or lower than if the pre-total results were rounded before the percentage of the total is calculated (the results still add up to 100%). This is apparent when compared to a spreadsheet, but I don't know whether this will be a user issue or not.

SiboVG:
> I would suggest a small change that would make it more in line with the treatment in the "Stability" tab: show pressure, base, friction, and total Cd per instance, and add an aggregate total Cd column (this last would have the same content as the current "Total Cd" column

Interesting that you mention this, because I just noticed that the instance mass only seems to work for pods . For instance for a fin set, the instance mass and aggregate mass are the same, but for a nose cone inside a pod set, the aggregate mass is the nose cone mass times the number of pod instances. Is this how it should work? And should the same behavior be applied for the drag tab (the instance CD only differs from the aggregate CD for podsets, not for individual fins of a finset)?

SiboVG:
> Interesting that you mention this, because I just noticed that the instance mass only seems to work for pods . For instance for a fin set, the instance mass and aggregate mass are the same, but for a nose cone inside a pod set, the aggregate mass is the nose cone mass times the number of pod instances. Is this how it should work? And should the same behavior be applied for the drag tab (the instance CD only differs from the aggregate CD for podsets, not for individual fins of a finset)?

Ping @JoePfeiffer

JoePfeiffer:
That is how it's intended to work, but now that you raise the question it isn't clear it should be.

The program regards a fin set as a single entity -- it is a single fin set, not three individual fins.  So what the component analysis says is correct; the instance mass is the mass of the whole fin set.

But a user is likely going to be expecting the instance mass to be the mass of a single fin, while the aggregate is the mass of all three.  That might well be a better way to express it to the user.

JoePfeiffer:
Since it's now correctly reporting how the simulation is calculating finset instances, let's merge it.  We may want to revisit whether we want the component analysis to present an instance mass of a single fin to the user, as discussed above.

Issue #1506, opened 2022-07-01
After a single click in an editable field, start by selecting the entire value 


neilweinstock:
Right now if I click my mouse in an editable field, the first action is for the cursor to appear in front of the existing value:
![image](https://user-images.githubusercontent.com/243872/176961570-be7ffda1-1b59-47e2-a4e4-b83591e9c3ff.png)

This is the wrong behavior most of the time.

Instead, the most useful action upon first click into the field is to select the entire field.
![image](https://user-images.githubusercontent.com/243872/176961789-914de7f7-3aec-4ffe-bc17-9b8a2b4299ab.png)

That way, if the user's desire is just replace the old value with a new one, then can just start typing, without having to execute extra clicks to select the whole value.

If their desire is to edit at one point in the field, then they can just use the arrow keys to locate the correct spot.  They would have had to do that anyway in the existing behavior.

An alternative (inferior, in my opinion, but still better than what's there today) would be to put the cursor in the location where the user clicks, rather than always going to the beginning of the field, which is often the least useful thing to do.

Some of this behavior might be constrained by the Swing, I don't know.  But it sure is annoying the way it is now.

SiboVG:
(Assigning myself because I'm working on #1488 and need to do something similar)

neilweinstock:
Thank you kindly :)

hcraigmiller:
The behavior that most of the programs I work with that use "cell" type information is this...

1. A left-click anywhere in the cell selects all of the cell content, a change in that data of any type erases the existing data and substitutes what is being entered.
2. A second left-click in the cell positions the cursor where the pointer is when the click occurs, and allows data to be added or deleted at that insertion/deletion point. For example, if the field contained 2345, the first left-click selects "2345" and the second left-click, with the pointer between the "3" and the "4", deselects "2345" and puts the cursor between "23" and "45".

As @neilweinstock related, this would likely be the best desired behavior.

PR #1508, opened 2022-07-02
[#1488 & #1506] Focus most commonly edited parameter after preset selection + Highlight spinner text after focus gained

### LINKED ISSUE #1488: After selecting a preset component, put focus on most commonly edited value
This is primarily focused on body tubes, but could apply to other components as well (centering rings, especially)

Tubes of all kinds (body tubes, inner tubes, launch lugs) are usually cut-to-length... or at least, edited-to-length in OR.  Therefore, after selecting a preset body tube, it would a small additional convenience to automatically put focus on the "length" field and select its contents, so you can immediately type in the length, rather than needing to mouse over to the field and double-click.

To me this is a no-brainer for tubes.  For other body components I'm not sure there's a good analog.

For internal components, it's "position" that is going to be edited most of the time, and that's the field that I'd pre-select.

I haven't gone through all the others to make this determination yet.

Basically, this is another way to reduce needless redundant pointing and clicking, to make it quicker and easier to get an initial design into OR.

SiboVG:
This PR fixes #1488 and #1506. If setting focus on any spinner text field (either by clicking on it when it is not yet in focus, or by pressing the 'Tab' key until that object is in focus), all the text in the text field is highlighted/selected (fixing #1506). When you click again in the text field, e.g. the text '123' is highlighted and you click between the 1 and 2, then the highlight will stop and the cursor will be placed between 1 and 2, so '1|23'.

For the following components, after selecting a preset, the length spinner text field is highlighted:
- body tube
- tube fin
- launch lug
- inner tube
- tube coupler
- engine block

For the following components, the axial position spinner is highlighted:
- rail buttons (even though they don't have any presets...)
- streamer
- parachute
- centering ring
- bulkhead

I also took the liberty of changing the text of the Coupler component button to 'Tube Coupler', since that's also the title of the component edit dialog after adding the coupler.

Demo:

https://user-images.githubusercontent.com/11031519/177004270-f412621f-00ce-458b-a007-237ed2f91661.mp4


hcraigmiller:

PR #1507, opened 2022-07-01
Resolving issue #1270, unexpected exception with simulation scripting

thzero:
This PR fixes #1270, fixes #308, fixes #1108 and fixes #826.

As the Nashorn scripting engine is deprecated, as part of the resolution this replaces Nashorn with the Graal.Js scripting engine.  This includes the addition of the following dependencies

graal-sdk-22.1.0.1.jar
icu4j-71.1.jar
js-22.1.0.1.jar
js-scriptengine-22.1.0.1.jar
truffle-api-22.1.0.1.jar

PR #1514, opened 2022-07-04
Cleanup of scripting examples and removal of deprecated methods

thzero:
Partially tied to [Issue #1507](https://github.com/openrocket/openrocket/pull/1507).   

During testing for 1507, it was noticed that there was a depecated use of the getActiveInstances in two of the scripts (one for Java and one for Javascript) - removed use of the deprecated getActiveInsances method.   There were some performance tweaks (Java and Javascript) to not lookup the CUSTOM fin each step; did not show too much of improvement from testing, however it was left in order to help resolve reduce the change of ConcurrentModification exceptions happening.

Others were cleaned up to conform more to the standard Java code formatting.


SiboVG:
There are a couple of instances where you refactor the code to bring variables declared inside of a for-loop outside of that for-loop. You should always declare variables in the smallest scope possible (which is inside the for-loop). Changing the variables inside or outside the loop does nothing for performance or anything; the byte code generated by the Java compiler is exactly the same.

Also: is the change to the scripting example still needed; you made those changes in PR #1507 right?

thzero:
> Also: is the change to the scripting example still needed; you made those changes in PR #1507 right?

Yup. I just downloaded from thzero:scripting and checked there and yeah, its been changed.

SiboVG:
Since the scripting example is already updated in #1507 and the code refactoring isn't really correct, I'm gonna close this PR.

PR #1515, opened 2022-07-04
[#1513] Combine 'Add component' and 'Modify new component' undo action

### LINKED ISSUE #1513: Feature request: combine undo action of 'new component' and 'modify (new) component'
When you add a new component and edit a property of that component, if you then want to undo adding the component, you always have to first undo modifying the component and then undo adding the component. So you need two undo actions. It would be better to just combine them into one 'Add new component' undo action since it's no use to first be able to undo the rocket modifications, because you're then just left with a "virgin" component.

SiboVG:
This PR fixes #1513 by combining the 'Add new component' and the 'Modify component' of that newly added component into a single 'Add new component' undo action. So if you now add a new component, and modify that new component in the edit dialog that popped up when the new component was added, you do not have perform two undo actions to undo adding that new component. (I also wanted this PR up because it will ease the development for issue #960).

Demo (includes first the prior behavior in OR 22.02.beta.04 and then the new behavior from this PR):

https://user-images.githubusercontent.com/11031519/177221365-c8e581c7-5195-4176-8118-fc51cdee1923.mp4

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/ComponentConfigDialog.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/ComponentConfigDialog.java
@@ -230,8 +230,9 @@ public class ComponentConfigDialog extends JDialog implements ComponentChangeLis
 	 * @param document		the document to configure.
 	 * @param component		the component to configure.
 	 * @param rememberPreviousTab if true, the previous tab will be remembered and used for the new dialog
+	 * @param includeUndoModify	if true, include a 'Modify component' undo action
 	 */
-	public static void showDialog(Window parent, OpenRocketDocument document, RocketComponent component, boolean rememberPreviousTab) {
+	public static void showDialog(Window parent, OpenRocketDocument document, RocketComponent component, boolean rememberPreviousTab, boolean includeUndoModify) {
 		if (dialog != null) {
 			// Don't remember the previous tab for rockets or stages, because this will leave you in the override tab for
 			// the next component, which is generally not what you want.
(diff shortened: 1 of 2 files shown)


hcraigmiller:
I think this is a nice improvement. Two questions. 
1. When the addition of a component is undone, should the parent of that component be selected by default? For example, with a new design, I add a nose cone and close, then undo adding the nose cone. Should the sustainer be selected so that I don't have to select the sustainer before adding a new/different component at that level? Or, with a new design, I add a body tube and close, then add an inner tube and close, then undo the inner tube. Should the body tube be selected so that I don't have to select the body tube before adding a new/different component at that level?
2. Is this sufficiently related to Issue #960 such that the two should be combined (the undo function here has the same function as the cancel button in Issue #960)? 

Build 787

neilweinstock:
I like this change.

SiboVG:
> 1. When the addition of a component is undone, should the parent of that component be selected by default? For example, with a new design, I add a nose cone and close, then undo adding the nose cone. Should the sustainer be selected so that I don't have to select the sustainer before adding a new/different component at that level? Or, with a new design, I add a body tube and close, then add an inner tube and close, then undo the inner tube. Should the body tube be selected so that I don't have to select the body tube before adding a new/different component at that level?

Could be a good new feature request (so will not be solved in this PR).


> 2\. Is this sufficiently related to Issue 960 such that the two should be combined (the undo function here would have the same function as the cancel button in that issue

Not really a solution for #960, I just meant that with this functionality, it will be easier to implement a fix for #960.

Issue #1018, opened 2021-09-13
Update fileformat.txt for custom inner appearance + PhotoStudio saving


SiboVG:
This is a reminder to update core/fileformat.txt (and the GitHub Wiki) to a new file format version that includes the custom inner appearance saving and Photo Studio settings saving (and other changes that are also new).

Issue #874, opened 2021-01-02
Photo Studio smoke effect does not function


hcraigmiller:
Windows 10 OS, Photo Studio, Effects tab. Checking the Smoke box has no visual change, neither does increasing or decreasing the Smoke Opacity, nor changing the smoke color.

neilweinstock:
I don't believe it worked in 15.03 either.

neilweinstock:
Yup, now working for me on MacOS 11.4 as well.

@hcraigmiller still a problem for you?

hcraigmiller:
Using A simple model rocket, with the C6-5 motor... smoke effect is non-functional. All other effects work.

OS Windows 10 with Java version 11.0.9 2020-10-20 LTS.

SiboVG:
Tested it on Windows 10 Home, Java Version 8 (1.8.0_291) and it works. Pretty strange...

hcraigmiller:
It may be the version of Java. The current unstable is intended for use with Java 11 LTS, and will likely be packaged with that Java version. It is my understanding that the current unstable is not fully compatible with Java 8.

SiboVG:
Yeah, but wouldn't the Java 8 version be less compatible then and also have the smoke issue? I don't know, might have a look at it later

SiboVG:
I've [read](https://gamedev.stackexchange.com/questions/147218/glsl-shader-error-on-different-computers) about texture2D in swing/resources/datafiles/flame/smokeShader.glsl being deprecated for some GLSL versions. It should be replaced by texture. I've tried replacing this, but a simple replacement isn't enough; other things need to be edited for it to work. If someone wants to work on this, this might be a way to solve it.

The problem might be that hcraigmiller has a different GLSL version (I have no idea, don't know anything about shaders yet, so this is a shot in the dark)

PR #1528, opened 2022-07-12
[#1524] Add sliders to PhotoStudio

### LINKED ISSUE #1524: Provide sliders in Photo Studio settings dialog
This is the only place in the UI where there aren't sliders, and this is probably the place where they would add the most value.

SiboVG:
This PR fixes #1524 by adding sliders to PhotoStudio.

I also noticed that the smoke opacity spinner had no effect on the smoke's opacity, which is fixed in this PR.
I also changed the default spark concentration to 20% instead of 0, so that when you enable sparks, you immediately see the sparks (instead of first having to change the concentration value to something non-zero).

In addition to adding sliders, I also added units for some parameters, such as a percentage unit for the smoke opacity.

Demo:

https://user-images.githubusercontent.com/11031519/178478007-c9e542b6-777c-4c0d-90c0-46636ef18436.mp4

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/Rocket.java
@@ -287,6 +287,11 @@ public class Rocket extends ComponentAssembly {
 		refType = type;
 		fireComponentChangeEvent(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);
 	}
+
+	@Override
+	public double getLength() {
+		return selectedConfiguration.getLength();
+	}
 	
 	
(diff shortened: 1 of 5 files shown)


hcraigmiller:
These sliders are going to be my new best friends.

**Functions as expected, no anomalies found,** except as to Issue #874 (Photo Studio smoke effect does not function) still exists.

Question: Does this fix, or has Issue #1018 been otherwise fixed?

Build 796

SiboVG:
> Question: Does this fix, or has Issue #1018 been otherwise fixed?

No, still present.

Issue #1533, opened 2022-07-14
Sluggish UI in  V22.02


neilweinstock:
There are a number of areas where the v22 UI is noticeably sluggish compared to v15.03.  Two obvious examples:

- [x] Dragging around a vertex in the finset editor
- [ ] Adding a new component and waiting for the config window to open

I don't know if this is related to Java 11 or if something changed in the UI implementation that is making it slower.  Accordingly, I don't know if it's readily fixable.  But it would be good to know the reason.

Tested on M1 iMac, not yet confirmed on other platforms.

hcraigmiller:
Confirmed in Windows 11.

PR #1534, opened 2022-07-14
[#1533] Fix latency of freeform finset shape editor

SiboVG:
This PR fixes one part of #1533, namely the latency in the freeform finset shape editor.

Demo (I guess I don't need to tell you which part of the video is before and after this PR...):

https://user-images.githubusercontent.com/11031519/179102597-63843fc6-b8d7-43c7-a738-27d097bdcb39.mp4

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/FreeformFinSetConfig.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/FreeformFinSetConfig.java
@@ -81,6 +81,7 @@ public class FreeformFinSetConfig extends FinSetConfig {
 	private Point dragPoint = null;
 
 	private FinPointFigure figure = null;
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Well, this interesting. What I thought was sluggish in Windows is nothing like Mac users were experiencing; I had no idea what they were putting up with. To be more specific, there is no discernible different between the Beta 4 and Build 801 using Windows 11.

But, when I compare Windows 11 to the screen recording, Windows 11 performs the same as the Mac in Build 801. So, it looks like both platforms are now performing the same. Although there is a little lag if you move the cursor quickly (such as moving in a circle, I have never found the short lag time to be so slow that it was an annoyance.

Subject to the foregoing, functions as expected, no anomalies found.

Build 801

neilweinstock:
Much much better, looks good.

PR #1538, opened 2022-07-18
[#1537] Add additional configuration parameters for rail buttons

### LINKED ISSUE #1537: Feature request: Add more shape configuration parameters for rail button
Currently, the only parameters to change the shape of a rail button is the outer diameter and total height. You should also be able to change the rail button's inside diameter, slot width/diameter, slot height, and inner and outer height.

SiboVG:
This PR fixes #1537 by adding the following configuration parameters for a rail button:
- Inner diameter (controls the diameter of the middle part of the rail button)
- Base height (controls the height of the cylinder closest to its parent component)
- Flange height (controls the height of the cylinder furthest away from its parent component)

Demo:

https://user-images.githubusercontent.com/11031519/179631165-a1f6ed63-e0ae-4549-bb4c-79dcbbdc168c.mp4

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1052,7 +1052,10 @@ LaunchLugCfg.tab.Generalprop = General properties
 
 ! RailButtonConfig
 RailBtnCfg.lbl.OuterDiam = Outer Diameter:
-RailBtnCfg.lbl.TotalHeight = Total Height
+RailBtnCfg.lbl.InnerDiam = Inner Diameter:
+RailBtnCfg.lbl.TotalHeight = Total Height:
+RailBtnCfg.lbl.BaseHeight = Base Height:
+RailBtnCfg.lbl.FlangeHeight = Flange Height:
 RailBtnCfg.lbl.Angle = Rotation:
 RailBtnCfg.lbl.PosRelativeTo = Position relative to:
 RailBtnCfg.lbl.Plus = plus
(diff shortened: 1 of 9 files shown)


SiboVG:
> 2\. Where there are three basic profiles for rail buttons...
> 
> Rail Button Component Rev_03
> 
> I suggest adding a _ScrewHeight_ field as part of this PR so that all three profiles will be addressed. Functionally, the display height of the rail button would equal _Height_ + _Screw Height_, and the displayed flange height would equal _Flange Height_ + _ScrewHeight_.
> 
> Otherwise, an issue to add the _ScrewHeight_ field can be opened.

Hm, I'm a bit opposed to adding configuration options for something that the user will not get to see. I'm happy to add the screw height parameter, but only if there is some sort of dropdown menu added to the rail button configuration panel that lets you switch between one of those 3 rail button types. In that case, the rocket view would/should change the rail button's display to match your drawing, and the user would actually see what changing the screw height is doing. For that, I would create a new issue.

hcraigmiller:
> I'm a bit opposed to adding configuration options for something that the user will not get to see.

I agree, the above was the math behind it. From a display standpoint, this is what I am proposing (side view).

Rail Button Component Rev_04

The user would be able to set the screw height, but the diameter of the screw head would be _OuterDiameter_. There would be no need for the complexity of a drop down to implement this.

SiboVG:
> The user would be able to set the screw height, but the diameter of the screw head would be _OuterDiameter_. There would be no need for the complexity of a drop down to implement this, just the _ScrewHeight_ field.

Fair enough. But I won't be including the screwHeight as a parameter for this PR, since that would involve adding visuals for the screw etc. @hcraigmiller can you make a feature request for that?

SiboVG:
@neilweinstock @JoePfeiffer any remarks on this implementation?

neilweinstock:
I’m not too knowledgeable about the ins and outs of rail buttons. 

I feel like all that button configuration doesn’t really accomplish that much until aerodynamics are implemented.

JoePfeiffer:
Very nice. Like Craig, I'd also like to see screws added, but this is plenty for one PR.

PR #1544, opened 2022-07-20
[#1539] Apply preset after double-click

SiboVG:
This PR fixes #1539: double clicking a row in the preset selection dialog will now apply that preset to the component and close the preset window. Double clicking the first column is not included in this behavior (since this is a checkbox).

I also did some cleanup of translation keys on the way.

PR #1546, opened 2022-07-20
[#1516] Keep preset highlight after favorites checkbox action

### LINKED ISSUE #1516: Bug: Keep checked preset row highlighted after checked
When checking a preset in the preset dialog, the table row first gets highlighted, but then loses its highlight. I would keep the highlight so that it is also more clear for users that highlighting (and checking) a preset applies that preset to the component. 

(if you think that it would also be beneficial to add an additional info text stating that checking + highlighting a preset applies it to the component, then do reply. I'm not yet sure if this is needed, although I do find it confusing now to know the difference between checking a preset and simply selecting a preset in the preset table)

SiboVG:
This PR fixes #1516 and keeps a preset row highlighted in the preset selection window after the favorites checkbox has been clicked.

Additionally, there is also a clarifying text about directly applying a preset, below the table:
Screenshot 2022-07-21 at 01 22 02

Note: I already mention applying a preset by double-clicking it, this is from PR #1544.


hcraigmiller:

Build 824

Issue #1480, opened 2022-06-25
Automatically pre-check "Match aft diameter" and "Match fore diameter" options in preset selection window


neilweinstock:
I believe that the *vast* majority of the time, users will want to check the "Match aft diameter" and/or "Match fore diameter" options in the preset selection window, to narrow down the (vast) selection of parts.  We should pre-check these to save the user a step.

hcraigmiller:
Forgive me if I am mistaken, but this is the current behavior when adding non-preset components. And, when adding preset components, the match boxes should not be set because the part characteristics are specifically defined.

neilweinstock:
I think you're talking about the "automatic" checkbox under OD.  I'm talking about this:
![image](https://user-images.githubusercontent.com/243872/175752756-8127cb2d-0a0a-47eb-8604-d30a9a36daa4.png)

hcraigmiller:
That is correct...

As to the "Match... diameter" boxes shown your screen grab, I have always held a certain level of disdain for this feature because it does not function as the user (meaning me in this case) expects. If I add an Estes PNC-50Y nose cone with an outside diameter of .976 in., then choose preset from database, check the "Match fore diameter" box, and then sort by outer diameter, the results range from .941 to 1.014 in. I think everyone can agree that anything other than .976 **is not a "match"** even though one **might** be able to modify the parts to get the body tube to _work_.

If this feature is changed to show only **true matches**, I agree that checking the "Match... diameter" box makes sense. Otherwise, all you are doing is creating a smaller set of unmatched parts to look through, although I guess only having to look through a smaller set may have some benefit.

neilweinstock:
The amount of "slop" allowed by the "Match diameter" option is certainly something that should be looked at; it seems too large right now (when trying to match BT50, you should not see BT51 and BT52 on the list).  However  I do not believe that it needs to be exactly the same down to three decimal places... there are times when slightly mismatched parts might be used.  For instance, if using a heavy-walled BT (BT-50H), then you'll usually need to use a mismatched nose cone because virtually no commercially available cone is designed to match BT-50H.  But that's a much narrower slop than the current filter is allowing.

I don't know how the slop is currently determined, whether it's by percentage or by a fixed amount.  It would probably require some study to determine what a better value might be.

And I would argue that even with the slop, *reducing* the size of the list is still valuable.  It is certainly more useful to have it checked than not, isn't it?

hcraigmiller:
[...]

PR #1547, opened 2022-07-21
[#1480] Pre-check diameter filters in component preset chooser

JonathanJagt23:
This PR solves #1480 in which it was requested to automatically check the "Match fore diameter" and "Match aft diameter" options in the component preset chooser window. This was done by adding a line for each option which selects the JCheckbox after checking if the option is applicable but before adding the checkbox to the pane.

### DIFF
--- a/core/src/net/sf/openrocket/startup/Preferences.java
+++ b/core/src/net/sf/openrocket/startup/Preferences.java
@@ -63,6 +63,9 @@ public abstract class Preferences implements ChangeSource {
 	public static final String MOTOR_DIAMETER_FILTER = "MotorDiameterMatch";
 	public static final String MOTOR_HIDE_SIMILAR = "MotorHideSimilar";
 	public static final String MOTOR_HIDE_UNAVAILABLE = "MotorHideUnavailable";
+
+	public static final String MATCH_FORE_DIAMETER = "MatchForeDiameter";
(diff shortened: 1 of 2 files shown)


SiboVG:
I would vouch to not always have the checkboxes set to true (which is the case for this PR), but to cache the checkbox state and initially check it as true. In that case, users who do not want the checkboxes checked by default can uncheck the checkboxes and OR will remember the checkbox state. Dunno if that's the behavior that @hcraigmiller and @neilweinstock agree upon.

neilweinstock:
Hadn't thought about it from that aspect... I agree with @SiboVG 's proposed operation.  Pre-check, then subsequently cache.

hcraigmiller:
I, too, agree with the behavior proposed by @SiboVG.

SiboVG:
@JonathanJagt23 I'd suggest you take a look at `core/src/net/sf/openrocket/startup/Preferences.java`, for example the SHOW_MARKERS is an example of a boolean preference. You should implement a similar preference for MATCH_FORE_DIAMETER and MATCH_AFT_DIAMETER.

JoePfeiffer:
Also hadn't thought about that, but also agree the suggested caching behavior would be a further improvement.

SiboVG:
Code LGTM and functions as expected, good work 👍

hcraigmiller:
Functions as expected, no anomalies found. Well done.

Build 830

PR #1550, opened 2022-07-23
[#1549] Ctrl/Cmd + A select all shortcut in simulation table and component tree

### LINKED ISSUE #1549: Feature Request: Select all objects in simulation table with Ctrl/Cmd + A shortcut
As requested in [this TRF post](https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-4-is-now-available.171295/post-2301466), it would be useful to add a Ctrl/Cmd + A shortcut for selecting all objects in the simulation table. IMO this should also be extended to the component tree (where Ctrl/Cmd + A would select all objects except the rocket instance - probably won't be a much used feature there, but it helps for OR's consistency).

Shortcut for:

- [ ] Simulation table
- [ ] Component tree

SiboVG:
This PR fixes #1549 and allows all elements of the simulation table to be selected with the ctrl/cmd + A command. The same shortcut also works in the component tree, where all components are selected, except for the rocket.

I also noticed a bug in the simulation table where if you select e.g. row 1, then row 2, the plot button was disabled, as should be, but if you would then deselect row 1, then the button's status did not re-enable. This is fixed now in this PR. I also disabled the 'Edit simulation' when multiple simulations are selected. The 'plot' and 'edit' context menu items are now also disabled when multiple simulations are selected.

While working on the previous bug, I also noticed that the 'Edit', 'Delete' and 'Duplicate' button in the home screen (next to the component tree) had a smaller text than the move up/down buttons above. I think this was only macOS specific, but in any case this PR fixes that.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
+++ b/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
@@ -418,18 +418,18 @@ public class BasicFrame extends JFrame {
 		button = new SelectColorButton(actions.getMoveDownAction());
 		panel.add(button, "sizegroup buttons, aligny 0%");
 
-		button = new SelectColorButton(actions.getEditAction());
-		button.setIcon(null);
+		button = new SelectColorButton();
+		RocketActions.tieActionToButtonNoIcon(button, actions.getEditAction());
 		button.setMnemonic(0);
 		panel.add(button, "sizegroup buttons, gaptop 20%");
 
-		button = new SelectColorButton(actions.getDuplicateAction());
-		button.setIcon(null);
+		button = new SelectColorButton();
+		RocketActions.tieActionToButtonNoIcon(button, actions.getDuplicateAction());
 		button.setMnemonic(0);
 		panel.add(button, "sizegroup buttons");
 
-		button = new SelectColorButton(actions.getDeleteAction());
-		button.setIcon(null);
+		button = new SelectColorButton();
+		RocketActions.tieActionToButtonNoIcon(button, actions.getDeleteAction());
 		button.setMnemonic(0);
 		panel.add(button, "sizegroup buttons");
 
--- a/swing/src/net/sf/openrocket/gui/main/RocketActions.java
+++ b/swing/src/net/sf/openrocket/gui/main/RocketActions.java
@@ -12,6 +12,7 @@ import java.util.List;
 
 import javax.swing.AbstractAction;
 import javax.swing.Action;
+import javax.swing.JButton;
 import javax.swing.JCheckBox;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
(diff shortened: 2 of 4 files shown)


neilweinstock:
When I just tried this, it only selected them all after I had selected one.  If I started with none selected, cmd-A didn't do anything.

SiboVG:
> When I just tried this, it only selected them all after I had selected one. If I started with none selected, cmd-A didn't do anything.

You first need to put focus on the simulation table. If you simply switch to the simulation tab, the simulation table is not focused (you can actually see the blue outline around the 'New simulation' button, so that button has focus).

neilweinstock:
How do I give it focus without selecting an existing sim?  Clicking in the blank part of the list area doesn't do it.

hcraigmiller:
Testing results, using _A simple model rocket_ example. There are two behaviors, the first of which is expected, the second of which is not.

Action One (Expected Behavior)
1. Open the example and go straight to the _Flight simulations_ tab.
2. **Press Ctrl+A results in no selection; expected behavior.**
3. Left-click on _Simulation 1_ (highlighting that simulation).
4. **Press Ctrl+A results in selection of all simulations**.

Action Two (Unexpected Behavior)
1. Open the example and go straight to the _Flight simulations_ tab.
2. Left-click on _Simulation 1_ (highlighting that simulation), and left-click _Run simulations_ to re-run that simulation (the simulation remains highlighted after being re-run and may be re-run any number of times without re-selecting the simulation).
3. **Press Ctrl+A results in no selection; unexpected behavior**, _the expected bahavior would be to select all simulations_.
4. Left-click on any simulation, including _Simulation 1_.
5. **Press Ctrl+A results in selection of all simulations**.

Build 826

SiboVG:
> How do I give it focus without selecting an existing sim? Clicking in the blank part of the list area doesn't do it.

You can't, you can only do so by selecting a sim (the same is true for the motors, recovery and stage configuration table).

SiboVG:
> Action Two (Unexpected Behavior)
> 
> 1. Open the example and go straight to the _Flight simulations_ tab.
> 2. Left-click on _Simulation 1_ (highlighting that simulation), and left-click _Run simulations_ to re-run that simulation (the simulation remains highlighted after being re-run and may be re-run any number of times without re-selecting the simulation).
> 3. **Press Ctrl+A results in no selection; unexpected behavior**, _the expected bahavior would be to select all simulations_.
> 4. Left-click on any simulation, including _Simulation 1_.
> 5. **Press Ctrl+A results in selection of all simulations**.

Hm well this issue isn't really inherent to the implementation of this PR... The same behavior holds in the motors, recovery and stage config tables. E.g. the motors config: select a config, hit ctrl + A, selects all configs, select an individual config again and click 'reset deployment', and then hit ctrl + A again, nothing happens. 

On macOS you can also see that Java Swing make a visual distinction between a row being selected and in focus and a row being selected, but not in focus (because the focus was lost after clicking 'reset deployment'). See video below.

https://user-images.githubusercontent.com/11031519/180621350-597ad77f-554e-41d4-acf3-4990051e510d.mov


So I would create a new issue for this to restore the focus to the motors, recovery, stage and simulation table after a certain action has been performed. That will fix the issue you described.

hcraigmiller:
As described by @SiboVG above, functions as expected with no anomalies found.

Build 826

neilweinstock:
> On macOS you can also see that Java Swing make a visual distinction between a row being selected and in focus and a row being selected, but not in focus

That is hilariously subtle.

PR #1556, opened 2022-07-25
[#1552] Use tab and arrow keys to traverse sim table

### LINKED ISSUE #1552: Feature request: Parse through simulation table using tab key and up/down arrow key
In the motors, recovery and stage configuration table, you can parse through different table cells by using the tab key or up/down arrow key. The same should also be possible in the simulation table.

SiboVG:
This PR fixes #1552 and allows to traverse the simulation table by pressing the tab, right arrow or down arrow key to go to the next row in the sim table, and shift-tab, left arrow or up arrow to go to the previous row in the sim table.


hcraigmiller:
In the motors, recovery and stage configuration tables, when you use the **tab** key to parse through the fields, pressing the tab key when in the last field takes you to the first field. Likewise, when using **shift+tab**  to reverse the direction, shift+tab in the first field takes you to tha last field.

The simulation table does not function this way. A shift+tab in the first field leaves you in that field, and a tab in the last field leaves you in that field.

Should the former behavior be consistent across all of the tables, including the simulation table?

SiboVG:
> In the motors, recovery and stage configuration tables, when you use the **tab** key to parse through the fields, pressing the tab key when in the last field takes you to the first field. Likewise, when using **shift+tab** to reverse the direction, shift+tab in the first field takes you to tha last field.
> 
> The simulation table does not function this way. A shift+tab in the first field leaves you in that field, and a tab in the last field leaves you in that field.
> 
> Should the former behavior be consistent across all of the tables, including the simulation table?

Good point, wasn't aware that the tab key had that effect in the other tables. I would keep it consistent. I'll push a fix for it soon.

SiboVG:
> I'll push a fix for it soon.

'Soon' has passed.

PR #1557, opened 2022-07-26
Mark the snap grade as stable

### LINKED ISSUE #1384: Latest openrocket beta not available in snap
Could you please provide the latest openrocket beta on snapcraft.io?
Thank you very much for your work!

wolsen:
The grade of the snap is marked as devel, which is great while the
app is being developed but means that the snap is not eligible to be
published to candidate or stable release channels. This change marks
the snap grade as stable to allow it to be released.

Additionally, the automatic snap publishing that happens for each
commit pushed to the unstable branch has been failing to release
the snap revisions due to a symlink introduced in the openjdk package
for which leaves a dangling symlink certs. The blacklisted.certs
should be now contained in blocked.certs so we add that to the list
of files to remove when being primed (final preparation for snap
packaging).

Fixes #1384

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: openrocket
 adopt-info: openrocket
-grade: devel
+grade: stable
 summary: A free, fully featured model rocket simulator.
 description: |
   OpenRocket is a free, fully featured model rocket simulator that allows you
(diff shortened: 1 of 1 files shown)


wolsen:
This is needed to resolve publishing in the Ubuntu snap store.

SiboVG (approved):

SiboVG:
Thanks for the change!

Issue #1171, opened 2022-02-26
Stage drag coefficient override has no effect on simulation results


neilweinstock:
Tested on macOS 12.

SiboVG:
I think the slider is just messed up. If you type in Cd values in the typebox, then it works fine, same goes for when you use extreme values in the slider. But In the first half of the slider, the sims don't get affected. It's just a weird slider, also looks like it is a logarithmic slider instead of a linear one.

https://user-images.githubusercontent.com/11031519/155882061-2628bb1a-8595-47d5-9b3f-85bf595daeb5.mov

Do we want to have this logarithmic behavior? I must say I was also confused the first time I used it; thought the slider just wasn't working at all until I got to the end of the slider.

neilweinstock:
I agree that the exponential slides is bad, and should be changed.  However, I cannot replicate your results where changing the value actually changes the simulation.  On my rocket, changing the setting from 0-1 makes no difference in the sim results.  That includes the simple model rocket example that you're showing above.

SiboVG:
That's odd, I can clearly see the values changing in the simulations tab...

neilweinstock:
I'm trying to think of what could possibly be different between our setups, and coming up empty so far.

SiboVG:
Could you maybe send a quick demo of what you're seeing in 'A simple model rocket'? (you can make a screen recording on Mac using 'Cmd + Shift + 5' and use 'Record Selected Portion' to only record OR's screen are. When you're done recording just go to hit the stop-button in the menu bar in the top right)

neilweinstock:
https://user-images.githubusercontent.com/243872/155891060-962be4b7-2006-4789-8a32-5fa2e256c664.mov

SiboVG:
Aha, I see what you did, you are overriding the Cd of the stage, not of an individual component like the nose cone. So I guess the problem is that Cd overriding for stages doesn't work.

neilweinstock:
Yes, that is what I assumed was the most likely use for the override.

SiboVG:
> Yes, that is what I assumed was the most likely use for the override.

Yes, most likely, just wasn't on the same page as you :)

PR #1573, opened 2022-08-03
Multiply component drag by number of instances

JoePfeiffer:
This borders on embarrassing:  while working on #1171 today, it finally dawned on me that when we iterate through all the instances of a component when calculating drag we're wasting time: all we need to do is multiply the component's drag by the number of instances.

This is separate enough from the issue of getting drag coefficient overrides to work I wanted to make it its own PR.


SiboVG:
Oof, good catch.

PR #1577, opened 2022-08-09
RK4SimulationStepper: populate accelerationData

rckTom:
Currently simulation listener callbacks like `postStep` or `postAccelerationCalculation` do not get acceleration data from the RK4SimulationStepper. 

This sets the acceleration for the `SimulationStatus` object and initializes the `accelerationData` member of the `DataStore` object

### DIFF
--- a/core/src/net/sf/openrocket/simulation/RK4SimulationStepper.java
+++ b/core/src/net/sf/openrocket/simulation/RK4SimulationStepper.java
@@ -249,6 +249,7 @@ public class RK4SimulationStepper extends AbstractSimulationStepper {
 		deltaO = k2.rv.add(k3.rv).multiply(2).add(k1.rv).add(k4.rv).multiply(store.timestep / 6);
 		
 
+		status.setRocketAcceleration(store.linearAcceleration);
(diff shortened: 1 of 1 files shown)


SiboVG:
@JoePfeiffer care to code-check this?

JoePfeiffer:
If I'm reading this correctly, you're ending up with the acceleration data in more than one place in the store -- 
linearAccelerationWC is a copy of store.linearAcceleration
rotationalAccelerationWC is a copy of store.angularAcceleration
(you're also saving the acceleration in rocket coordinates, of course, which isn't in the store).

Would it make sense to get rid of store.linearAcceleration and store.angularAcceleration, and only use the members of store.accelerationData?  This would require changes to all the simulation steppers, of course.

SiboVG:
@rckTom?

JoePfeiffer:
There's a bunch of stuff that needs to be straightened out between the simulation status and the data store (for instance, the data store claims to have both an AccelerationData object and contains linear and angular rotation separately; the AccelerationData object doesn't seem to ever get updated in the stepper!  I think we're better off putting effort into that, and not adding a band-aid.

PR #1583, opened 2022-08-13
[#1305] Bigger default preset window size + remember column width

### LINKED ISSUE #1305: Remember column widths and window size in parts database dialogs (and/or provide better defaults)
Upon first opening a parts database dialog, all columns are equal-width.  This is rarely a good choice, but it is an easy default.  Here's a freshly-opened parachute database:
image
Obviously, the Diameter, Sides, Line Count, and LineLength columns could all be narrower, while most of the others could be wider.  If you go and adjust the column widths to be better, your changes are *not* remembered when you close the window, and so then when you reopen the window it'll be back to equal-width columns.

Furthermore, if I enlarge the window to actually fit all the columns, that too is not remembered.  Each time the dialog is opened it is back to default size.

Both of these should be remembered, and saved as part of app memory, not per-file, because these settings are not specific to a rocket design, but to the user's preferences, and it should be (I think) saved separately for each component type, because each one has a different set of data columns.

Having a bunch of good default column widths and window sizes would also be good, and would possibly negate the need to save the user's changes.

SiboVG:
This PR fixes #1305 by adding the following changes to the preset dialog UI:
- By default larger window size (based on the size of BasicFrame - the main OR window)
- Rememberance of window size and location
     - The window size and location is the same for the preset dialogs accross all component types. If you change the dialog size in the body tube preset dialog, then that same window size will be used for the nose cone preset dialog. Seems like the most reasonable approach to me.
- Rememberance of column widths
     - This is custom for each different preset dialog. Changing e.g. the width of the second column in the body tube dialog does not alter the second column width in the nose cone preset dialog.

Demo:

https://user-images.githubusercontent.com/11031519/184496917-b7e5f9e1-91c6-465c-a3a9-adc6f6eebb78.mp4

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/EditDecalDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/EditDecalDialog.java
@@ -163,9 +163,11 @@ public class EditDecalDialog extends JDialog {
 		panel.add(cancelButton, "tag cancel");
 		
 		this.add(panel);
-		
-		GUIUtil.rememberWindowSize(this);
+
 		GUIUtil.setDisposableDialogOptions(this, okButton);
+		GUIUtil.rememberWindowSize(this);
+		this.setLocationByPlatform(true);
+		GUIUtil.rememberWindowPosition(this);
 		
 	}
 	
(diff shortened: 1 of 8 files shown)


neilweinstock:
You're spoiling me.

Are these settings remember for each rocket design or globally for the program?  (Globally would be the best choice here).

hcraigmiller:
Functions as described, no anomalies found. A really nice improvement.

Build 884

SiboVG:
> You're spoiling me.
> 
> Are these settings remember for each rocket design or globally for the program?  (Globally would be the best choice here).

The settings are global.

Issue #1574, opened 2022-08-07
Photo Studio: no motor=no flame, add motor: no flame


jimmiedave:
Bring a model into Photo Studio without a motor. Turn on the Flame effect. You'll see no flame, although the non-existent flame will illuminate the rear of the rocket when turned on.

Without closing Photo Studio, select a motor for the rocket, and you'll still see the same effects (not) displayed.

Close Photo Studio and re-open it without changing anything and the flame appears (provided you had a motor and the flame effect selected before you closed it).

Tested Mac OS Catalina 10.15.7, OpenRocket v22.02.beta.04, Intel Mac: MacBook Pro 16in 2019, onboard Radeon Pro 5300M 4GB (and Intel UHD graphics 630 1536MB).

Tested first with my own X-15 model, and with A Simple Model Rocket.

Edit: Same effect with/without a background picture.

**Flame Off**
<img width="1133" alt="Screen Shot 2022-08-07 at 4 14 47 PM" src="https://user-images.githubusercontent.com/35353530/183309900-a7f42ca4-8ed5-4143-ab47-a476e4e8c283.png">

**Flame On, showing illumination of the rocket by the invisible flame.**
<img width="1133" alt="Screen Shot 2022-08-07 at 4 14 58 PM" src="https://user-images.githubusercontent.com/35353530/183309894-3246c78a-17ba-412b-b493-37d5acb79e19.png">

hcraigmiller:
> Bring a model into Photo Studio without a motor. Turn on the Flame effect. You'll see no flame, although the non-existent flame will illuminate the rear of the rocket when turned on.

Behavior confirmed.

> Without closing Photo Studio, select a motor for the rocket, and you'll still see the same effects (not) displayed.

Behavior confirmed.

> Close Photo Studio and re-open it without changing anything and the flame appears (provided you had a motor and the flame effect selected before you closed it).

Behavior confirmed.

See also Issues #1122 (flame), #874 (smoke).

[Windows 11 Pro; Version 21H2; OS Build 22000.739; Windows Feature Experience Pack 1000.22000.739.0]
[Java "11.0.15" 2022-04-19 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.15+8-LTS-149)]

neilweinstock:
I believe this is one of many (!) known 3D breakages, all of which will have to wait until the whole 3D library is replaced.

SiboVG:
This issue is twofold:

[...]

PR #1586, opened 2022-08-14
[#1574] Fix motor flame lighting when no motor

SiboVG:
This PR fixes part of #1574, namely now there will no more flame lighting when the rocket has no motors.

The other part of #1574 that needs to be fixed is while PhotoStudio is active, and you add a new configuration with a motor, that motor is not drawn in PhotoStudio. Only if you close PhotoStudio and reopen it, the motor becomes visible. I will not handle a fix for that in this PR.


hcraigmiller:
> This PR fixes part of https://github.com/openrocket/openrocket/issues/1574, namely now there will no more flame lighting when the rocket has no motors.

Build 885

PR #1588, opened 2022-08-14
[#1523] Fix copy order

### LINKED ISSUE #1523: [Bug] Copy/paste and cut/paste reverses order of subcomponents
Copy/paste and cut/paste, either within a design or from an existing design to a new design results in the following:

1. The order of the components attached to a body tube reversed; main airframe compnents did not (all individual component characteristics preserved).
01 Body Tube Order

2. The order of the components attached to a pod reversed.
02 Pod Order

Restoring the correct order, restores the design. 

03 Restore Order

SiboVG:
This PR fixes #1523.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/RocketActions.java
+++ b/swing/src/net/sf/openrocket/gui/main/RocketActions.java
@@ -433,12 +433,13 @@ public class RocketActions {
 			return new Pair<RocketComponent, Integer>(destComponent, destComponent.getChildCount());
 
 		RocketComponent parent = destComponent.getParent();
-		return getPastePositionFromParent(srcComponent, parent);
+		return getPastePositionFromParent(srcComponent, destComponent, parent);
 	}
 
-	private Pair<RocketComponent, Integer> getPastePositionFromParent(RocketComponent component, RocketComponent parent) {
-		if (parent != null && parent.isCompatible(component)) {
-			int index = parent.getChildPosition(parent) + 1;
+	private Pair<RocketComponent, Integer> getPastePositionFromParent(RocketComponent srcComponent, RocketComponent destComponent,
+																	  RocketComponent parent) {
+		if (parent != null && parent.isCompatible(srcComponent)) {
+			int index = parent.getChildPosition(destComponent) + 1;
 			return new Pair<>(parent, index);
 		}
 
--- a/swing/src/net/sf/openrocket/gui/main/RocketActions.java
+++ b/swing/src/net/sf/openrocket/gui/main/RocketActions.java
@@ -851,10 +851,18 @@ public class RocketActions {
 		@Override
 		public void actionPerformed(ActionEvent e) {
 			List<RocketComponent> components = selectionModel.getSelectedComponents();
+			List<RocketComponent> topComponents = new LinkedList<>();		// Components without a parent component in <components>
 			if (components != null) {
 				components.sort(Comparator.comparing(c -> c.getParent() != null ? c.getParent().getChildPosition(c) : 0));
 				components = new ArrayList<>(components);
 				fillInMissingSelections(components);
+			} else {
+				return;
+			}
+			for (RocketComponent c: components) {
+				if (!RocketComponent.listContainsParent(components, c)) {
+					topComponents.add(c);
+				}
 			}
 			Simulation[] sims = selectionModel.getSelectedSimulations();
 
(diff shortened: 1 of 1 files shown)


hcraigmiller:
So long as the entire level is being copied/pasted, it functions without error. But if part of the copy is a higher level than the rest, then the paste reorders the highest level. In all other respects, functions as expected.

Example, copying upper body tube sub components (without the body tube) and the lower body tube with sub components, when pasted, reorders body tubes.

03 Phantom Tube

[Estes 24mm x 95mm BT60 Motor mount.txt](https://github.com/openrocket/openrocket/files/9333951/Estes.24mm.x.95mm.BT60.Motor.mount.txt)
[BT-60 Body Tube.txt](https://github.com/openrocket/openrocket/files/9333955/BT-60.Body.Tube.txt)

SiboVG:
> So long as the entire level is being copied/pasted, it functions without error. But if part of the copy is a higher level than the rest, then the paste reorders the highest level. In all other respects, functions as expected.
> 
> Example, copying upper body tube sub components (without the body tube) and the lower body tube with sub components, when pasted, reorders body tubes.
> 
> 03 Phantom Tube

I'm afraid that's just not fixeable... (or at least it will be a big mess trying to fix it). If the parent component is not copied, there is no real info to know where the paste position should be. So in your example OR has no way of knowing if the second body tube was above or below the first phantom body tube, because the first body tube is not copied.

So I'm inclined to mark that as "works as intended" 🙃 I may revise this at one point, but not now.

hcraigmiller:
Strange, because it works correctly in Beta 4.

SiboVG:
> Strange, because it works correctly in Beta 4.

There were some rewrites in the copy/duplication code because of other bugs that were present in the copy/duplication system in beta 4 (don't exactly recall which ones, but also component ordering being incorrect sometimes). So in totality, there should be less copy/duplication bugs in unstable.

If you're lucky, I will find the energy to revise your mentioned issue in a couple of hours.

hcraigmiller:
You have no idea of how much I appreciate your hard work.

SiboVG:
FYI: the behavior is also broken in beta 4, but just broken in a different way. If you insert a nose cone on top, and do the same selection and copy procedure as described earlier, but now with the top nose cone included, you will see that beta 04 moves all the components down, whereas this PR moves everything up. So in beta 04 the nose cone is in the wrong place, in this PR the body tube is in the wrong place.
Screenshot 2022-08-14 at 22 56 10

hcraigmiller:
It looks like user instructions will be needed. I think that the behavior needed is that in my comment containing the motor mount, with instructions that _omitting components between the copy start and end point_ may result in unexpected placement as to the order of the components after the paste.  Your thoughts?

SiboVG:
> It looks like user instructions will be needed. I think that the behavior needed is that in my comment containing the motor mount, with instructions that _omitting components between the copy start and end point_ may result in unexpected placement as to the order of the components after the paste. Your thoughts?

Well I don't see a straightforward way of fixing it code-wise. So I'm okay with your proposition of just documenting the "bug".

hcraigmiller:
If you belive it can be done, I think being able to do the motor mount copy/paste would be highly desireable. I'll start on the user instructions.

SiboVG:
> If you belive it can be done, I think being able to do the motor mount copy/paste would be highly desireable.

Hm, that's tricky, but doable. You can create a new ticket for it.

hcraigmiller:
Done. . . Issue https://github.com/openrocket/openrocket/issues/1591#issue-1338382688

SiboVG:
> FYI: the behavior is also broken in beta 4, but just broken in a different way. If you insert a nose cone on top, and do the same selection and copy procedure as described earlier, but now with the top nose cone included, you will see that beta 04 moves all the components down, whereas this PR moves everything up. So in beta 04 the nose cone is in the wrong place, in this PR the body tube is in the wrong place. Screenshot 2022-08-14 at 22 56 10

Latest commits fixed a small issue in the code, which restores this beta 4 paste behavior. So incompatible components are pasted after the selected component, instead of before it (also makes more sense to me).

hcraigmiller:
This feature will require Wiki instructions/warnings for certain (to the effect that skipping a higher level component while copying its subcomponents may have unexpected paste results), but it functions as expected with regard to the motor mount copy/paste functionality.

One error though. An attempt to paste on the stage results in an error.

Sustainer

Build 919

SiboVG:
> One error though. Using your copy from original design above, an attempt to paste on the stage results in an error.

Fixed.

hcraigmiller:
Functions as expected with no anomalies found. I will start working up the Wiki instructions/warnings.

Build 921

PR #1589, opened 2022-08-14
[#411] Improvements to scale dialog

### LINKED ISSUE #411: Improve the Scale Dialog
Based on an email from Joe Pfeiffer:

- [ ] 1. the "Scaling to apply" window is only three digits, allowing only a factor of 999% to be displayed.  Putting in a wider number is possible, but requires scrolling back and forth in the window to see.  
- [ ] 2. The Scale from and to windows are only one digit wide (though after I've scaled a rocket, they get wider?).
- [x] 3. When I apply a scale factor, it doesn't recenter the rocket in the window, and doesn't bring up scrollbars.  The result is that the rocket disappears on a large scale.
- [ ] 4. The dialog presents scaling in terms of %, but the error dialog presents it in terms of scale factor.  I actually like scale factor better, but they both need to be one or the other

SiboVG:
This PR fixes #411 by displaying the scale dialog error message in terms of the currently selected unit + the spinner widgets now have a larger width.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/ScaleDialog.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/ScaleDialog.java
@@ -193,6 +193,7 @@ public class ScaleDialog extends JDialog {
 	
 	
 	private final DoubleModel multiplier = new DoubleModel(1.0, UnitGroup.UNITS_RELATIVE, SCALE_MIN, SCALE_MAX);
+	private UnitSelector multiplierUnit;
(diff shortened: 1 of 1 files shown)


hcraigmiller:

1. Scaling to apply window > 3 digits: Functions as expected
2. Scale from and to windows > 1 digit: Functions as expected
3. Recenter the rocket in the window: Functions as expected
4. Unable to trigger scale factor error.

Build 888

PR #1590, opened 2022-08-14
[#1555] Don't open config dialog for duplicate config

### LINKED ISSUE #1555: Don't automatically open first stage motor selection when duplicating motor configurations
See: https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-4-is-now-available.171295/post-2302738

Right now OR automatically opens the first stage motor selection when creating a new motor configuration *or* duplicating an existing configuration.

Most likely this behavior is still appropriate and useful for new motor configs, but maybe not for existing ones.  I'm not sure if the best behavior would be to just don't open anything, or "take the hint" and open the selector depending on which cell was selected prior to the duplication.  My proposal right now is just to not open anything automatically when duplicating.

SiboVG:
This PR fixes #1555.


hcraigmiller:
Assuming that the intent is to implement Option 1, functions and expected with no anomalies found.

Build 890

SiboVG:
> Assuming that the intent is to implement Option 1, functions and expected with no anomalies found.

Yes, sorry for not clarifying, option 1 made the most sense to me as a solution, as it is conceptually the easiest for users.

PR #1595, opened 2022-08-16
Add icons to buttons

SiboVG:
This PR adds icons to the main buttons in the rocket design, motors & configuration, and flight simulations tab. The intent is to visually aid users, so they don't have to read every button text to know what it does + it helps to more quickly find the button you're looking for.

A counterargument could be that the icons clutter the UI, so I'll leave that to you to decide.
buttons

I also noticed that when you create a new design and go straight to the 'Motors & Configuration' tab, that every button except for the 'New configuration' button was still enabled. This PR fixes that issue and disables the 'Rename Configuration', 'Remove Configuration' and 'Duplicate Configuration' button when no motor mount is defined and no configuration is selected.


neilweinstock:
I'm sort of mixed on this one.

If those are actual screenshots, then my two specific comments are:
1) I don't like the box around the "delete" icon.  Should be just the red X
2) The icons all seem a bit too large to me, with not enough padding between the icons and the button edges.  I *don't* want to make the buttons bigger, therefore I would propose making the icons smaller.

SiboVG:
> * I don't like the box around the "delete" icon.  Should be just the red X
> * The icons all seem a bit too large to me, with not enough padding between the icons and the button edges.  I _don't_ want to make the buttons bigger, therefore I would propose making the icons smaller.

Done. Icons are now 90% their original size (making them smaller than that made them hard to read).
buttons

neilweinstock:
Better. I still think the green triangle could be smaller(like, same size as the red X).

hcraigmiller:
Looks nice to me as is.
unctions as expected, no anomalies found.

Build 914

SiboVG:
> Better. I still think the green triangle could be smaller(like, same size as the red X).

Updated version:
image

neilweinstock:
Better. Looks good to me.

hcraigmiller:

Build 915

SiboVG:
@JoePfeiffer you want to give this a quick visual check of whether you like this change?

neilweinstock:
Very, very few apps use an icon for the "rename" function, probably because
they know that no one really has any intuitive feel for what it should look
like.

From my googling, it seems the one Sibo used here is as close to standard
as it gets.  If it wouldn't look so weird to have rename be the only button
without an icon, I'd say just remove it entirely, but might as well put the
icons on all the buttons.

SiboVG:
> If it wouldn't look so weird to have rename be the only button without an icon, I'd say just remove it entirely, but might as well put the icons on all the buttons.

I first tested the UI without the rename icon, because indeed a rename action often doesn't have an icon, but it looked a bit off compared to the other buttons. So now all the buttons have their icon, and no one gets discriminated :)

PR #1598, opened 2022-08-17
[#1227] Fix fin root edge display in 3D view and fin template

### LINKED ISSUE #1227: When exporting templates, draw fin roots correctly for fins-on-nose cones and fins-on-transitions
As reported by @MadRocketer on TRF: https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-is-underway-download-it-now.171295/post-2250080

When a fin is placed on a transition or nose cone, the exported template still shows a straight edge.
image
Exports like this:
image

Obviously it would be preferable for the exported template to show the correct root edge contour.

SiboVG:
This PR fixes #1227. The root edge of a fin set now correctly outlines a transition's shape both in the 3D view and in the fin template export. Besides the fins, the fin tabs also weren't drawn correctly (neither in the 2D view, nor in the 3D view - the fin tab root edge was always just a straight line). This PR solves that issue as well.

Demo (with a transparent transition):

https://user-images.githubusercontent.com/11031519/185255413-d52b3dcc-c1fa-4c30-9025-a2d433d4c4f4.mp4

Fin template export also works with fin tabs:
Screenshot 2022-08-18 at 00 27 28

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -1244,7 +1244,7 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 		// for anything more complicated, increase the count: 
 		if ((body instanceof Transition) && (((Transition)body).getType() != Shape.CONICAL)) {
 			// the maximum precision to enforce when calculating the areas of fins (especially on curved parent bodies)
-			final double xWidth = 0.005; // width (in meters) of each individual iteration
+			final double xWidth = 0.0025; // width (in meters) of each individual iteration
 			divisionCount = (int) Math.ceil(intervalLength / xWidth);
 			
 			// When creating body curves, don't create more than this many divisions. -- only relevant on very large components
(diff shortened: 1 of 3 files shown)


neilweinstock:
LOL, that was quick, gonna need to come up with something more challenging for you. ;)

If I may nitpick: in the last screenshot there, you can see the line segments in the curved root line.  Is it possible to use smaller steps so it looks smoother?

hcraigmiller:
The height of the fin tab does not auto calculate correctly.

1. Create transition with inner tube and centering rings.
2. Add freeform fin (default).
3. On _Fin tabs_ tab, left-click _Calculate autimatically_.

Results in fin tab height to centerline of transition.

SiboVG:
> If I may nitpick: in the last screenshot there, you can see the line segments in the curved root line. Is it possible to use smaller steps so it looks smoother?

Done, export now looks like this:
image

SiboVG:
> The height of the fin tab does not auto calculate correctly.
> 
> 1. Create transition with inner tube and centering rings.
> 2. Add freeform fin (default).
> 3. On _Fin tabs_ tab, left-click _Calculate autimatically_.
> 
> Results is fin tab height that goes to centerline of transition.

That's not a result of this PR, so there should be a new issue created for it. Also a side note: if the transition's fore diameter is smaller than the aft diameter, the calculations is done correctly. It's only when the aft diameter is smaller than the fore diameter that the tab height is calculated incorrectly.

hcraigmiller:
Actually, the cause was user error on my part. . . I failed to check the inner tube "motor tube" checkbox.

**Functions as expected, no anomalies found.**

Build 918

Issue #1107, opened 2022-02-06
[Feature request] Ability to designate top or bottom centering ring for automatically calculated tabs


hcraigmiller:
Many high-power rockets use 3 or more rings to center the motor mount tube. Currently, automatically calculated fin tabs reference only the uppermost centering ring and create a tab between that centering ring and the next ring down. However, most high-power rockets need fin tabs between the lowermost centering ring and the next ring up.

An option to select either the uppermost or lowermost centering ring as the reference for automatically calculated fin tabs is desirable.

PR #1601, opened 2022-08-18
[#1600] Calculate auto tab for all inner tubes

### LINKED ISSUE #1600: [Improvement] Auto fin tab height calculation for all inner tubes
Currently, the auto fin tab height only functions correctly if the inner tube is also a motor mount tube. If the inner tube is not a motor mount tube, the auto fin height function makes the fin tab height the distance from the outside of the component to the internal centerline of the component.

SiboVG:
This PR fixes #1600.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/FinSetConfig.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/FinSetConfig.java
@@ -233,61 +233,62 @@ public abstract class FinSetConfig extends RocketComponentConfig {
 		autoCalc.addActionListener(new ActionListener() {
 			@Override
 			public void actionPerformed(ActionEvent e) {
-				log.info(Markers.USER_MARKER, "Computing " + component.getComponentName() + " tab height.");
-
-				double inRad = 0.0;
-				RocketComponent parent = component.getParent();
-				if (parent instanceof SymmetricComponent){
-					try {
-						document.startUndo("Compute fin tabs");
-						
-						List<CenteringRing> rings = new ArrayList<>();
-                        // Do deep recursive iteration to find centering rings and determine
-						// radius of inner tube
-                        Iterator<RocketComponent> iter = parent.iterator(false);
-                        while (iter.hasNext()) {
-                            RocketComponent rocketComponent =  iter.next();
-							if (rocketComponent instanceof InnerTube) {
-								InnerTube it = (InnerTube) rocketComponent;
-								if (it.isMotorMount()) {
-									inRad = it.getOuterRadius();
-								}
-							} else if (rocketComponent instanceof CenteringRing) {
-								rings.add((CenteringRing) rocketComponent);
-							}
-						}
-						
-						//Figure out position and length of the fin tab
-						if (!rings.isEmpty()) {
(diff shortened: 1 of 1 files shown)


hcraigmiller:
Nested inner tubes result in error. The fin tab height attaches to the smallest diameter tube, rather than the first tube that the fin tab comes in contact with (delete the nested inner tube and the height calculates correctly). The next error is in the current unstable and PR # 1598 as well. The nested centering rings causes the fin tab length to be too short (delete the nested centering rings and the tab length calculates correctly). 

See the following:

Fin Tab Auto

[Inner Tube.txt](https://github.com/openrocket/openrocket/files/9378377/Inner.Tube.txt)
[Inner Tube.Nested.txt](https://github.com/openrocket/openrocket/files/9378379/Inner.Tube.Nested.txt)

SiboVG:
> Should this be an instructional issue for the Wiki, or a code change?

The calculate auto tab code was just really janky. I rewrote the whole thing and it now alway looks for the biggest radius CenterRing or InnerTube, no matter where they are located in the component tree. The tab length is set the same length as the max radius component. If there are multiple components with equal maximum size, e.g. 2 centering rings, the fin tab will span both ring lengths and the distance between the two rings.

hcraigmiller:
The fin tab auto calculate is making the fin tab length the distance between the outside (not inside) of the cetering rings, and the height is from the outer diameter to the inner diameter of the component the fins are attached to (not to the outer diameter of the inner tube).

Fin Tab Auto n02

SiboVG:
> The fin tab auto calculate is making the fin tab length the distance between the outside (not inside) of the cetering rings, and the height is from the outer diameter to the inner diameter of the component the fins are attached to (not to the outer diameter of the inner tube).

Ah, I wasn't aware it was supposed to work like this...

hcraigmiller:
It's not. . . 

Fin Tab Auto n03 02

[Tab Height.Short.txt](https://github.com/openrocket/openrocket/files/9383364/Tab.Height.Short.txt)

SiboVG:
> Like this. . .

Yes, I was aware of how it's supposed to work after your prior post, my apologies.

SiboVG:
Okay, latest commit should have everything up and running.

hcraigmiller:
**Functions as expected, no anomalies found.** Works great.

FYI: Issue #1107 is an extension of this work.

Build 926

hcraigmiller:
Should this PR be working with nose cones, body tubes, and transitions? Currently works with body tubes and transitions that have an aft diameter that is equal to or less than the fore diameter.

SiboVG:
> Is this PR intended to be working with nose cones, body tubes, and transitions? Currently works only with body tubes and transitions that have an aft diameter that is equal to the fore diameter.

This PR isn't supposed to solve that, but there is a bug that the tab height is limited when the fore diameter is smaller than the aft diameter. If the fore diameter is zero, you cannot increase the tab height; it's locked to 0. I would file this as a new issue.

hcraigmiller:
> This PR isn't supposed to solve that, but there is a bug that the tab height is limited when the fore diameter is smaller than the aft diameter. If the fore diameter is zero, you cannot increase the tab height; it's locked to 0. I would file this as a new issue.

Issue #1603 created.

Build 926

PR #1604, opened 2022-08-22
[#1315] Update auto length in MassObject

### LINKED ISSUE #1315: [Bug] Changes to parachute automatic resizing feature related to parachute length
The automatic parachute sizing feature is a wonderful improvement, and there are three changes that will make it even better.

Currently, the parachute automatic resizing has three problem areas:
1. With the automatic box checked, when a user moves a parachute from a tube having a diameter of A to a tube having a diameter of B (or changes a tube's inner diameter), _the length of the parachute is not automatically resized_; and
2. With the automatic box checked, when a user unchecks the automatic box, _the length of the parachute is not automatically resized to its preset length_. For example, if a preset parachute with a diameter of 2 and length of 3 is placed inside of a tube with an inner diameter of 1, then the parachute will be given an automatic diameter of 1 and length of 12. If the user then unchecks the automatic box, the parachute diameter returns to the preset of 2, but the length stays at 12, rather than also returning to the preset of 3.

Those two changes would be very beneficial. It's already a thing of beauty

SiboVG:
This PR fixes #1315. New behavior for all mass objects:

- If the 'Auto' checkbox at (packed) diameter is unchecked:
     - No special behavior; you can just edit the diameter and length as you please
- If you then enable the 'Auto' checkbox:
    - The diameter will change to fit the parent component
    - The length will change in such a way that the (packed) volume after enabling the 'Auto' checkbox is the same as the (packed) volume before enabling the checkbox. E.g. if the diameter is 5 cm before, length 5 cm (volume of 98.17 cm^3), enabling the 'Auto' checkbox changes the diameter to 10 cm, and the length to 1.25 cm (volume still 98.17 cm^3)
    - You can still change the length in the auto mode, but this will now change the volume of the (packed) mass object. So let's say the diameter in auto mode was 10 cm and the length 1.25 cm, if you now set the length to 2.5 cm, the volume changes to 196.35 cm^3. If you now disable the auto setting, again, the volume between auto-mode and non auto-mode will be preserved. So after disabling the 'Auto' checkbox, the diameter falls back to 5 cm, and to keep the volume at 196.35 cm^3, the length will be set to 10 cm.
    - When changing the parent diameter, the volume of the mass object will also be preserverd, so the length of the mass object dynamically changes with the (auto) diameter

In short: there is a preservation of the mass object's volume when converting from auto mode to non auto mode and vice versa. There is a preservation of volume when in auto mode and the parent's diameter is altered.

Demo:

https://user-images.githubusercontent.com/11031519/185869289-ef9d5bdf-090a-4593-9176-3b71b1d6a7c4.mp4

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/savers/MassObjectSaver.java
+++ b/core/src/net/sf/openrocket/file/openrocket/savers/MassObjectSaver.java
@@ -13,7 +13,7 @@ public class MassObjectSaver extends InternalComponentSaver {
 
 		MassObject mass = (MassObject) c;
 
-		elements.add("<packedlength>" + mass.getLength() + "</packedlength>");
+		elements.add("<packedlength>" + mass.getLengthNoAuto() + "</packedlength>");
 		if (mass.isRadiusAutomatic()) {
 			elements.add("<packedradius>auto " + mass.getRadiusNoAuto() + "</packedradius>");
 		} else {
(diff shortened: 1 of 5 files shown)


hcraigmiller:

Build 931

PR #1608, opened 2022-08-23
Increase test error margin

SiboVG:
Still sometimes a failed test due to this

### DIFF
--- a/core/test/net/sf/openrocket/simulation/DisableStageTest.java
+++ b/core/test/net/sf/openrocket/simulation/DisableStageTest.java
@@ -19,7 +19,7 @@ import org.junit.Test;
  * @author Sibo Van Gool <sibo.vangool@hotmail.com>
  */
 public class DisableStageTest extends BaseTestCase {
-    private final double delta = 0.08;  // 8 % error margin (simulations are not exact)
+    private final double delta = 0.1;  // 10 % error margin (simulations are not exact)
 
     /**
      * Tests that the simulation results are correct when a single stage is deactivated and re-activated.


JoePfeiffer:
Have all of the failures been with max acceleration on multistage rockets (that's what I hit this morning that had me looking into this)?  I would suggest that we'd be better off not testing max accel, and then we can tighten the acceptable error quite a bit on the other parameters.

Right after a stage separation the thrust from a motor changes so quickly that individual acceleration numbers can vary wildly from run to run even though the final result of the simulation is extremely close.  For instance, I just ran the "three stage" example rocket several times.  One of the runs shows a max acceleration of approximately 311 ft/sec^2 after both stage separations; the other shows 341 after the first but only 311 after the second.  In spite of this spike, max altitude for the first is approx 1410 ft while the second is 1427.

I really think the difference is where exactly on the thrust curve the thrust is sampled, and over the course of a motor burn the variations cancel out.  All the same, we do get a much larger variation in max acceleration than we do the other simulation results, to the point that I don't think comparing max accel is useful in locating possible bugs.

I'm attaching the CSV files for two runs; first stage separation is at 1.86 seconds.
[lomax.csv](https://github.com/openrocket/openrocket/files/9409078/lomax.csv)
[himax.csv](https://github.com/openrocket/openrocket/files/9409079/himax.csv)

SiboVG:
> Have all of the failures been with max acceleration on multistage rockets (that's what I hit this morning that had me looking into this)? I would suggest that we'd be better off not testing max accel, and then we can tighten the acceptable error quite a bit on the other parameters.

I recall it being present in other parameters than the max acceleration as well, but I'll run some quick tests.

SiboVG:
Turns out that it's max acceleration and ground hit velocity that's the culprits. Removing those from the unit tests gave me 15 successive runs without a failure, at an error margin of 2.5%. I'll push a PR that removes those + with more strict error margin.

PR #1612, opened 2022-08-25
[#1603] Fix fin set tab height validation + dynamic max slider value

### LINKED ISSUE #1603: [Bug] Auto calculate fin tab limited to 0 height when transition fore diameter < aft diameter
@SiboVG wrote:
> "[T]here is a bug that the tab height is limited when the fore diameter is smaller than the aft diameter. If the fore diameter is zero, you cannot increase the tab height; it's locked to 0.

See discussion PR #1601.

SiboVG:
This PR fixes #1603 by correcting the fin set tab height validation (this also fixes an issue where if the aft diameter was smaller than the fore diameter, the tabs were able to grow over the rocket's centerline and overlap with each other). I also made it so that the slider of the fin tab height now dynamically updates its maximum value (in beta 4 you could still slide the slider even though the height value got clipped e.g. halfway).

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -309,7 +309,8 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 		}
 		
 		tabHeight = newTabHeight;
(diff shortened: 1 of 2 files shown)


hcraigmiller:

Build 951

PR #1615, opened 2022-08-25
[#236] Limit simulation time step to 0.01 s

### LINKED ISSUE #236: Simulation time step display bug
Simulation time step dialog rounds display value without rounding internal value, so what is displayed to the user is not what is used for simulation. This occurs only after the cursor has left the time step text box (eg changing to different setup tab) For example:

image

Yet a simulation timestep of 0.005
image

SiboVG:
This PR fixes #236 and limits the simulation time step setting to 0.01 s.

Although this is a simple PR, I did implement something that **needs testing** of whether it doesn't brake other stuff in the program. I added a check to `setValue` in `DoubleModel.java` that clips the set value to the model's maximum and minimum value. I must say I was surprised that this wasn't implemented already. The effect of this operation should be that when you type a numerical value in a spinner box and hit enter, that if that value was larger than the DoubleModel's maxValue, that the set value will be clipped to maxValue. Same goes for minValue in the other direction.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/adaptors/DoubleModel.java
+++ b/swing/src/net/sf/openrocket/gui/adaptors/DoubleModel.java
@@ -773,6 +773,14 @@ public class DoubleModel implements StateChangeListener, ChangeSource, Invalidat
 	 */
 	public void setValue(double v) {
 		checkState(true);
+
+		if (v > maxValue) {
+			log.debug("Clipping value " + v + " to maximum " + maxValue + " for " + this);
+			v = maxValue;
+		} else if (v < minValue) {
+			log.debug("Clipping value " + v + " to minimum " + minValue + " for " + this);
+			v = minValue;
+		}
 		
 		log.debug("Setting value " + v + " for " + this);
 		if (setMethod == null) {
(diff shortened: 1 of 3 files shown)


hcraigmiller:
My experience in this area is limited. That said, Functions as expected, no anomalies found.

Build 955

swing/src/net/sf/openrocket/gui/adaptors/DoubleModel.java
```diff
@@ -773,6 +773,14 @@ public double getValue() {
...
 	public void setValue(double v) {
 		checkState(true);
+
+		if (v > maxValue) {
+			log.debug("Clipping value " + v + " to maximum " + maxValue + " for " + this);
+			v = maxValue;
+		} else if (v < minValue) {
+			log.debug("Clipping value " + v + " to minimum " + minValue + " for " + this);
+			v = minValue;
+		}
```
wolsen, line 783: Note, there is already a `MathUtil.clamp(double x, double min, double max)` method that does this (minus the logging - though that could be added with a check on whether the value has changed.

wolsen: This change seems extremely reasonable (and correct to me - to clamp the value). It would be *incorrect* to be out of the bounds of the model itself. Naturally, @SiboVG concern is around ensuring that there is proper testing as other components need to be validated that the bounds are specified correctly - which will require a fair amount more effort on the review than what I have this evening (though I am doing this).

SiboVG: > Note, there is already a `MathUtil.clamp(double x, double min, double max)` method that does this (minus the logging - though that could be added with a check on whether the value has changed.

Yes, I wrote it explicitly to have the logging message, but I can rewrite it.

PR #1627, opened 2022-08-30
Rewrite CD overrides so it works for component assemblies

### LINKED ISSUE #1171: Stage drag coefficient override has no effect on simulation results

JoePfeiffer:
Fixes #1171


hcraigmiller:
Preliminary observation, is this the intended behavior and, if so, why?

- Stage: The "Override mass, CG, and CD of all subcomponents is checked by default and will not uncheck.
- Boosters: The "Override mass, CG, and CD of all subcomponents is checked by default and will not uncheck.
- Pods: The "Override mass, CG, and CD of all subcomponents is checked by default and will not uncheck.

This behavior is different than other components and may cause user confusion.

JoePfeiffer:
> Initial observation, is this the intended behavior and, if so, why?
> 
>     * Stage: The "Override mass, CG, and CD of all subcomponents" is checked by default and will not uncheck.
> 
>     * Boosters: The "Override mass, CG, and CD of all subcomponents" is checked by default and will not uncheck.
> 
>     * Pods: The "Override mass, CG, and CD of all subcomponents" is checked by default and will not uncheck.
> 
> 
> This behavior is different than other otherwise identical component override panes and may cause user confusion.

It is the intended behavior and is also the current behavior.  Component assemblies have no physical reality except as "containers" for other components, so setting their CD isn't meaningful except as an override for their subcomponents.

Though, now that you bring it up, we could enable that, and then setting a CD override on a stage (for instance) without overriding subcomponents would act as a "bias" on the calculated CD of the subcomponents.

hcraigmiller:
< Though, now that you bring it up, we could enable that, and then setting a CD override on a stage (for instance) without overriding subcomponents would act as a "bias" on the calculated CD of the subcomponents.

That would be a nice additional feature of the override.

JoePfeiffer:
The more I thought about it the more I like the idea.  By the time I was done with this PR I wasn't convinced allowing CD overrides was even a good idea, since an overridden CD can't change in the course of  a flight like a calculated CD can.  Allowing the override on a component assembly to act as an adjustment on its calculated CD does a better job letting users use a test flight to calibrate CD for better predictions of behavior with a different (likely larger) motor.

JoePfeiffer:
Thinking about it, a wiki page about how to use overrides would not be a bad idea...

hcraigmiller:
> Thinking about it, a wiki page about how to use overrides would not be a bad idea...

Working on it. . . will likely need input and review (on Slack) before posting.

SiboVG:
Btw good work on this @JoePfeiffer 👍 I know a couple of users who will really appreciate this.

PR #1630, opened 2022-08-31
Add icon for Cd override

JoePfeiffer:
The component tree view currently has icons marking when CG or mass has been overridden.  This adds a similar icon for drag coefficient overrides.  I couldn't think of an icon that was better than CD with the D somewhat smaller than the C in lieu of a real subscript

This is sufficiently independent from #1627 I made it a separate PR

cd-override-example


JoePfeiffer:
> Good to add this, but the D should be lower case.

There was a discussion on Slack last May (in a different context) as to whether it should be C<sub>D</sub> or C<sub>d</sub>, with the last few comments seeming to indicate C<sub>D</sub> was preferred, so I came as close to that as I could. I'd be happy to change it to Cd (with, presumably, the d a bit smaller than the C, as the D is now) if people want.

[https://app.slack.com/client/T5WV162BG/C02BHPA26JK/files/F03J4DBNLHW](https://app.slack.com/client/T5WV162BG/C02BHPA26JK/files/F03J4DBNLHWl)

SiboVG:
I think there should be 1 pixel of empty space between the 'C' and the 'D', to differentiate them better.

Also something that I now notice: the mass override icon should not have the 'kg' text, analogous to the small icon on a mass component in the component tree.

JoePfeiffer:
cd-override-d

JoePfeiffer:
> As to the icon itself, on my display it looks somewhat fuzzy. I'd like to see something with crisper edges.

Does it seem fuzzier than the CG icon?

hcraigmiller:
> Does it seem fuzzier than the CG icon?

Yes.

JoePfeiffer:
Ah, I see what's happening...

JoePfeiffer:
cd-override-example

SiboVG:
Why are we switching back to Cd instead of CD all of a sudden?

JoePfeiffer:
> Why are we switching back to Cd instead of CD all of a sudden?

It was Neil's suggestion and got no comments, so I thought I'd go with it for a while and see if anyone said anything -- and now someone has! I prefer CD myself, but it's a really weak preference.

neilweinstock:
CD looked really ugly there.  Maybe it could have been better with a smaller "D".

Didn't we actually settle on C_D, which was Sampo's original usage?  That would look even worse as an icon, but... whatever.

JoePfeiffer:
> CD looked really ugly there. Maybe it could have been better with a smaller "D".
> 
> Didn't we actually settle on C_D, which was Sampo's original usage? That would look even worse as an icon, but... whatever.

My assumption during that conversation was that really meant C<sub>D</sub>, but without being able to go below the baseline it looked *really* bad.  CD with D smaller than C was an approximation to that.

JoePfeiffer:
I'm finding Cd easier to recognize than CD in this context
override-example-d
override-example-D

JoePfeiffer:
Yes, need consistency, and we're using CD elsewhere.  One last try; this time I went all in on the idea that there won't be a real subscript.  I think I like this one better than any of the previous options, so I'm pushing this one to the PR.
cd-override-example

hcraigmiller:

Build 976

neilweinstock:
I think this is a good approach.  The one final tweak I would make is to make the D smaller than the C, e.g.:
image

Subtle difference but IMHO better.  That is the last comment I'll make on this. :)

SiboVG:
> I think this is a good approach. The one final tweak I would make is to make the D smaller than the C, e.g.: image
> 
> Subtle difference but IMHO better. That is the last comment I'll make on this. :)

Agreed.

JoePfeiffer:
Took the CD from Neil's suggestion

hcraigmiller:
Functions as expected, no anomalies found. A little difficult to read, the top right of the "C" seems a little odd.

00 Cd

Build 981

neilweinstock:
The rendering of all the icons in your screenshot is very strange, as if they’re all being slightly scaled.

I am wondering if that’sa serif font that Is not scaling well to that very small size. I did my mock-up in Ariel, for what it’s worth.

hcraigmiller:
I don't know. . . the application is running full screen. It's not enough to bother me (and it can be tweaked later), just FYI if it's still being worked on.

PR #1642, opened 2022-09-04
[#1638] Use RocketInfo mach for figure extra

### LINKED ISSUE #1638: Rocket design window doesn't change displayed Mach number
In the upper right corner of the rocket design window, where the stability, CG, and CP are displayed, the displayed Mach number remains fixed even though the CP moves as the user changes the speed in the component analysis dialog.

SiboVG:
This PR fixes #1638.

https://user-images.githubusercontent.com/11031519/188316614-52e00682-aece-4ee1-bf3a-81926f8cdf19.mov

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figureelements/RocketInfo.java
+++ b/swing/src/net/sf/openrocket/gui/figureelements/RocketInfo.java
@@ -182,7 +182,7 @@ public class RocketInfo implements FigureElement {
 	private void drawStabilityInfo() {
 		String at;
 		//// at M=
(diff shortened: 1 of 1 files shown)


SiboVG:
One remark: exiting the component analysis window resets the mach, angle of attack, and wind direction settings. Is this something we want? Or do we want to keep those changes?

hcraigmiller:

My presference would be the current behavior, exiting resets.

Build 986

JoePfeiffer:
My first thought would be to leave them alone on exit.  But I don't have a really strong sense of which is better...

PR #1647, opened 2022-09-06
[#1646] Increase fin root resolution

### LINKED ISSUE #1646: Fin 3D appearance goes transparent on non-conical transitions

SiboVG:
This PR fixes #1646. Turns out the fin root points resolution of 2.5 mm was too small. Increasing it to 5 mm appears to have solved the issue.

So fin template export will be **slightly** less smooth (see discussion [here](https://github.com/openrocket/openrocket/pull/1598#issuecomment-1218776893)).

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -33,7 +33,7 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 	 * Maximum number of root points in the root geometry.
 	 */
 	private static final int MAX_ROOT_DIVISIONS = 100;
-	private static final int MAX_ROOT_DIVISIONS_LOW_RES = 20;
+	private static final int MAX_ROOT_DIVISIONS_LOW_RES = 15;
 
     public void setOverrideMass() {
     }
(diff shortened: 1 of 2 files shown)


SiboVG:
[See testing of @hcraigmiller in #1646.](https://github.com/openrocket/openrocket/issues/1646#issuecomment-1238853053)

If you scale the rocket 200% (fin tabs and no fin tabs), this is what happens with the parachute _**and fins**_:

Tab 200
Tab 200 Parachute

As to the parachute, current unstable does the same thing, I'll create a new parachute issue?

Build 992

_Originally posted by @hcraigmiller in https://github.com/openrocket/openrocket/issues/1646#issuecomment-1238853053_

SiboVG:
Okay, turns out that if the root has too many points, the 3D geometry can't handle it. So now, the number of root points is limited to 20 points instead of 100, but the default resolution is set to 2.5 mm (as was before this PR). So for small fins, you should still get a smooth fin geometry (2.5 mm resolution, if the root length is 5 cm or smaller). If the root length is larger than 5 cm, then there will only be 20 points for the root geometry.

SiboVG:
Basically the 3D rendering of the fin should be replaced with drawing the fin using triangles. Unless that is fixed, I'll have to limit the root points...

SiboVG:
Okay, new approach: I'm now using a special low-res root geometry only for the 3D rendering, so the rest of the program is not affected by the limitations in the 3D rendering.

hcraigmiller:
Still occurs with a variety of back-to-back scaling, but the one that works every time is (1) scale 50% the (2) scale 1000%.

hcraigmiller:

Build 998

PR #1654, opened 2022-09-08
Finpoint snap

### LINKED ISSUE #427: FinSetConfig Improvements (5/5)
Feature Config:

"I would be super-nice if there were a way to automatically align a point vertically or horizontally with adjacent points (e.g., holding down a modified while dragging or something like that.)"

Option 1:  Snap to nearest point / vertical / horizontal (when within 5? x? pixels? mm?)
Option 2: Snap to grid (with tolerance: 5? mm?)
Option 3: Manual Edit via point table

neilweinstock:
This PR enables snap-to-cardinal (vertical/horizontal) or snap-to-ordinal (45 degree or 135 degree diagonal) when the shift key is held down while dragging a point in the finset editor.  Alignment is with respect to the previous point, so there is no effect when dragging point 0 (front of the root).

[edit] Fixes #427.

Known issues: 
1) The layout of the explanatory text under the editor is not good, I don't now how to do that yet
2) Slightly weird behavior when dragging the last point (aft root) when mounted on a nose cone or transition.  I'm inclined to leave that as it is, but see what you think.


hcraigmiller:
Is this associated with Issue #427?

neilweinstock:
Yes

neilweinstock:
It is absolutely not always true that the one you want to snap to is the nearest.  Keeping behavior consistent and predictable is (IMHO) the only way to maintain sanity here, no unpredictable guessing by the app.

An alternative would be to use shift+ctrl+click+drag to align relative to the following point, but that just seems to be too much complication for the user.  It's not at all hard to implement that way, but it would require still more explanatory text under the editor, and it would consume the ctrl-shift-click combination, in case we ever wanted it to be used for something else in the future.  But I could go along with it if folks prefer it.  What do you think?

Either way it might be worth adding to the existing text that it operates relative to the previous point.

SiboVG:
Well the behavior I talked about is just the one I know from other software, such as in Adobe Illustrator (which I'm used to, that's why the current behavior felt off to me). It's just more intuitive to me that way, but no idea what others are used to or prefer.

I wouldn't add additional shortcuts, that would overcomplicate things.

SiboVG:
But I'm not sold on only snapping to the previous point, that would only give you 50% of the functionality that you could get.

neilweinstock:
Well, I can add it to another key modifier, but IMNSHO it must be under user control.

There are other ways to make it user controllable but all the ways I can think of end up being more complex.  A second key modifier is simple, although (again) it requires more explanatory text.

I'll create another test build with it implemented that way and we can see how we like it.

neilweinstock:
Ctrl-Shift-Click-drag function implemented in latest check-in.  Still layout problems with explanatory text below.

SiboVG:
I'm not a fan of the extra modifier, but we'll see what others say...

neilweinstock:
I'm open to any reasonable alternative that keeps it under user control.  I haven't thought of any good ones yet.

neilweinstock:
One alternative to the extra modifier is to use a momentary press of another key (perhaps space bar), while dragging, to toggle which point is used for reference.  In other words: click and drag to start moving the point.  Hold down shift to snap: initially it will be referencing the previous point.  While holding down shift, momentarily press <space> to toggle to using the next point for reference.  Keep pressing <space> to toggle back and forth, while shift key is held.  It will remember the setting as long as the drag operation is in process.  Once you release the mouse button, it will revert to the default for the next drag operation.

I don't know if this is any better, but it is different.  If it sounds more appealing, I can try to figure out how to implement it (it's somewhat more complex than what's there now).

BTW I already plan an update where it shows the segment that is being aligned in a different color during the operation, so you can easily see how it's working in real time.

PR #1656, opened 2022-09-08
[#1651] Always do super component scales first

### LINKED ISSUE #1651: [Bug] Fin tabs do not scale "down" correctly
Using the Issue #1646 .ork file with fin tabs, scale the rocket 400% and the fin tabs scale correctly. Scale down to 25% and the fin tab proportionate length is lost (it is shorter from fin tip backwards), as it is if the design is reopened and scaled to 25%.

SiboVG:
This PR fixes #1651. The problem was the order in which scaling was done. For fin sets, first the fin points were scaled down, which would clip the tab length because the tab length before scaling down would now be larger than the fin length, and then the clipped tab length was further scaled down. So now, the scaling order is changed to first perform scaling according to the super class (for fins just FinSet), and afterwards the component-class (e.g. FreeFormFinSet).


hcraigmiller:

Build 1000

PR #1657, opened 2022-09-08
[#1653] Scale RadiusRingComponent according to scale direction

### LINKED ISSUE #1653: [Bug] Centering ring inner diameter does not scale "down" correctly
Using the Issue #1646 .ork file with fin tabs, scale the rocket 50% and the centering ring inner diameter is too small (it scale _up_ correctly).

Scale CR-Inner

SiboVG:
This PR fixes #1653. The problem was the order of scaling. The previous behavior was that first the outer radius of the ring was scaled, and then the inner ring. This works for scaling up, but if scaled down, the outer radius would first scale down, would then become smaller than the inner radius, so the inner radius got clipped, and then the clipped inner radius got scaled down again. This PR changes the scaling order based on when you're scaling up or down. If scaled up, first the outer radius is scaled and then the inner radius, when scaling down, first the inner radius is scaled, and then the outer radius.


hcraigmiller:

Build 1001

PR #1674, opened 2022-09-13
[#1662] Fix scaling of pod axial offset + fix ThicknessRingComponent scaling down

### LINKED ISSUE #1662: [Bug] Pod placement does not scale correctly when positioned relative to middle/bottom of parent
The placement of a pod relative to the parent scales correctly when positioned relative to the tip of the rocket or the top of the parent component. The placement of a pod relative to the parent DOES NOT scale correctly when positioned relative to the middle or bottom of the parent component. Nesting pods positioned relative to the middle or bottom of the parent component multiplies the error.

Image

[Anti-Virus.Scale.Rev_02 - Copy.txt](https://github.com/openrocket/openrocket/files/9541629/Anti-Virus.Scale.Rev_02.-.Copy.txt)

SiboVG:
This PR fixes #1662. The issue was in the scaling order, whereby the parent component of the pod was scaled first, and then the pod (and its offset). This caused the parent reference length for the axial offset to change first, and the scaled axial offset to be calculated incorrectly. This PR first does all the component offset scalings, and only then the component dimension scalings.

I also noticed the same issue as #1653, but for ThicknessRingComponents (e.g. TubeCoupler), which this PR fixes.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/PodSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/PodSet.java
@@ -1,15 +1,11 @@
 package net.sf.openrocket.rocketcomponent;
 
-import java.util.ArrayList;
-import java.util.Collection;
 
-import jdk.jshell.spi.ExecutionControl;
 import net.sf.openrocket.l10n.Translator;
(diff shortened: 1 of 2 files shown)


hcraigmiller:

Build 1024

PR #1677, opened 2022-09-15
[#1676] Treat zero-length SymmetricComponent as tube/disk

### LINKED ISSUE #1676: [Bug] Zero length transitions cause uncaught exception
Put a zero length transition on the back end of the _A simple model rocket_ example, and click the simulation tab. Zero length transition causes uncaught exception; change length to .001, and error goes away; body tubes allow zero lengths without uncaught exception.

SiboVG:
This PR fixes #1676 by setting the longitudinal and rotational inertia of a SymmetricComponent (transition/nose cone) to zero if the component length is zero. To then have the same behavior as a phantom body tube (the component acts as a disk and can still cause drag), I match the fore and aft radius to the max value of the two when the component length is zero. This is because a zero-length transition is just a disk, and a disk is just a zero-length tube. So matching the fore and aft radius will let the transition behave like a zero-length tube.


hcraigmiller:

This should definitely reduce the number of discontinuity errors related to Estes engine hook hacks.

MMT

Build 1031

Issue #1668, opened 2022-09-12
Centering thickness is incorrect.


sjh114961:
I reported to Dave cook, "Thickness of CR5060-W should be 0.125 NOT 0.05. Balsa Machining part". His response was "Sounds like a plausible error but where are you seeing this? I've not included a BMS file in my database (yet) due to a bunch of missing manufacturer specs for various balsa parts. If you're looking in the old 15.03 parts source code, none of that code has been carried forward to the new release directly." Is this true or do you guys need to fix this? Dave's github issue https://github.com/dbcook/openrocket-database/issues/13

JoePfeiffer:
@dbcook was incorrect on that -- the old database is still available, since it contains parts Daves's doesn't (like BMS!).  You can control whether or not use see legacy parts by opening the Parts Library dialog and clicking the "Show Legacy Database" checkbox.

sjh114961:
According to OR this is NOT Legacy  
![centeringRings1](https://user-images.githubusercontent.com/53155232/189723751-6e4bd2b6-e26b-43e9-9a3e-3185fc8e8ff5.png)

sjh114961:
So something is wrong

dbcook:
> @dbcook was incorrect on that -- the old database is still available, since it contains parts Daves's doesn't (like BMS!). You can control whether or not you see legacy parts by opening the Parts Library dialog and clicking the "Show Legacy Database" checkbox.

I stand corrected on that, though I strongly recommend against keeping or using the legacy database due to the massive number of errors it contains (several hundred before I stopped counting).  I would like to enable the total removal of the old database because of the low fidelity, age (the data set is a decade old) and lack of substantiation.

It seems like there's enough interest that I'll start on a BMS file with whatever parts can be substantiated - CRs, tubes and some balsa parts.  I can't say how long that will take since the primary dev computer I use has died with hardware issues and I'm not sure it will come back to me with the hard drive intact with my "special" installations of OR, vscode XML tools, etc.  It will be a week and maybe two before I can do anything that requires serious testing.  In the meantime I can make simple textual fixes only.

@JoePfeiffer , the non-legacy designation for the BMS parts looks wrong.  Not sure exactly what the bug is since I'm not the maintainer of the logic that decides what's legacy or not.  I just confirmed that there are no parts in any file in my DB repo with mfr containing "BalsaMachining", so I've not accidentally included the bms file in my own repo.  It has to be either a packaging problem or a logic flaw upstream in OR.

[...]

PR #1678, opened 2022-09-16
Tidy up directories a little and better identify legacy components

JoePfeiffer:
This tidies up the component source directories, so they are now named components-openrocket and components-dbcook.  The intent is that component files that are added directly by the OpenRocket team will be in components-openrocket, while the components-dbcook directory will contain Dave's database as a subproject (of course).

All of the component files that came in from 15.03 are now marked as Legacy, whether we have a better source for the data (like Estes) or not (like BMS).

Finishes fixing #1668


hcraigmiller:

Build 1032

PR #1709, opened 2022-10-02
Fixed error on using DampingMoment simulation listener

raynichc:
When attempting to run a simulation with the DampingMoment simulation listener added an exception is thrown and the variable is not computed.

It seems that AerodynamicCalculator.getForceAnalysis was modified to take in a WarningSet argument, however, for the DampingMoment listener, this was given as null, causing the error.


JoePfeiffer:
This cures the null pointer exception, but passing in an empty WarningSet that then gets discarded makes me nervous.  Should we be passing our existing WarningSet into firePostFlightConditions(), and then passing that to the listeners?

raynichc:
I've found that I can forward the SimulationStatus warnings and it will add any new warning appropriately.

PR #1711, opened 2022-10-02
Fixed flight event labels getting cut off

### LINKED ISSUE #1622: [Bug] Flight event labels can get cut off
Flight event labels on the far right of the sim plot can get cut off.
image

I also think at one point there should be a much better way of displaying the flight event labels, as they often overlap each other, and clutter the sim plot...

raynichc:
Fixes #1622, though I'm not a fan on the magic number, but I couldn't figure out how to determine the text's height as I can't get the font metrics. Any suggestions?

Before Fix:
not_fixed

After Fix:
fixed


SiboVG:
If you can somehow transform the font size of 13 (on line`m.setLabelFont(new Font("Dialog", Font.PLAIN, 13));`) to replace the magic number, that would be great. I'm not very familiar as to how to do that.

SiboVG:
After giving it some thought, this is good enough for now.

PR #1717, opened 2022-10-03
Fixed photo studio save indents

### LINKED ISSUE #1714: Add tabs in PhotoStudio .ork file settings
Very minor personal annoyance, but there should be tabs before the photostudio fields in the .ork file:
image

raynichc:
Fixes #1714, this approach felt cleaner than keeping the start and end tags in the list and handling those as special cases.


SiboVG:
Thanks for the PR! Functions as expected. Now I go back to reading .ork files without frustrations :)

PR #1720, opened 2022-10-03
Fix 1712

JoePfeiffer:
Correctly simulate case where a core section with side boosters separates from the upper stages before the side boosters separate from the core, but warn the user about it.


JoePfeiffer:
This is ready to go, but is likely to interfere with the work @SiboVG is doing assigning motor ignition events to correct stages. We can merge that in when it's ready and I can rework this to fit, or we can merge this now and work around it in the motor event work.

SiboVG:
> This is ready to go, but is likely to interfere with the work @SiboVG is doing assigning motor ignition events to correct stages. We can merge that in when it's ready and I can rework this to fit, or we can merge this now and work around it in the motor event work.

Don't worry, haven't had the time to properly work on that PR 🙃 so I'll just give this a code-check and we're good to go.

core/src/net/sf/openrocket/simulation/BasicEventSimulationEngine.java
```diff
@@ -436,19 +438,24 @@ else if (event.getType() == FlightEvent.Type.EJECTION_CHARGE)
...
+					// Create a new simulation branch for the booster
 					SimulationStatus boosterStatus = new SimulationStatus(currentStatus);
+					
+					// Prepare the new simulation branch
 					boosterStatus.setFlightData(new FlightDataBranch(boosterStage.getName(), FlightDataType.TYPE_TIME));
-					// Mark the booster status as only having the booster.
-					boosterStatus.getConfiguration().setOnlyStage(stageNumber);
+
+					// Mark the current status as having dropped the current stage and all stages below it
+					currentStatus.getConfiguration().clearStagesBelow( stageNumber);
```
SiboVG, line 466: Hm, what if there are two booster stages in the same BodyTube, booster 1 and booster 2. Booster 1 is a lower stage than booster 2. Both are ignited at the same time, but booster 1 burns longer than booster 2. Will the separation event of booster 2 wrongfully set booster 1 inactive? (I may be talking nonsense, please correct me if I do)

SiboVG (changes requested):

core/src/net/sf/openrocket/simulation/BasicEventSimulationEngine.java
JoePfeiffer: No, because a new simulation branch is created. When booster 2 burns out, both boosters separate from the sustainer. So, the existing simulation branch marks both of them as inactive. A new branch is created for the boosters; this branch marks the sustainer as inactive but both booster stages are active. Since this is an absurd separation order, the warning gets set. Later, when booster 2 burns out, boosters 1 and 2 separate.

In the existing code, when the misleadingly named clearStage() is called, it actually clears all the axial stages (but not parallel stages) below the stage being cleared. Then, in the new simulation branch for the boosters, it marks the topmost stage in the branch as the only active stage. So in this case, any axial stages below the separation point get lost. Any parallel stages are simulated as if they were still attached to the sustainer. But now, when they try to separate from the sustainer, it is observed that the stage above them is already inactive so the separation doesn't get performed.

SiboVG (approved):

PR #1721, opened 2022-10-04
Don't explicitly save smokeAlpha in photostudio .ork

SiboVG:
This PR fixes the bug reported on [TRF](https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-5-is-now-available.171295/post-2328000) where opening PhotoStudio could throw an exception. The issue was that smokeOpacity in the .ork file was saved as 2.0, which was the converted to an alpha value of 510. Don't ask me why that happened, but it was either way redundant for a smokeOpacity field in the .ork file, since the alpha value was already stored in smokeColor. So this PR simply removes the smokeOpacity field in the .ork file.


hcraigmiller:

Build 1074

Issue #1729, opened 2022-10-10
Adding more motors to the database (feature request)


Nighten7:
I'd like to see some more motors in the database. I was working on one of my rockets the other day and noticed that the Aerotech H219T and I175W were not present. if you could add some more high power motors (particularly L1 in my case, but others would be nice as well) that would be great. maybe you could use thrustcurve.org to get the information necessary to run simulations with a motor?

SiboVG:
We use thrustcurve.org for our motor database. Both the H219T and I175W are present, but you need to search them using their common name from thrustcurve.org (H219 and I175). I thought the database search box also worked for searching based on the common name, but apparently not, pinging @JoePfeiffer for that.

JoePfeiffer:
Both of those motors are in our database.  Some things to look at:
1) Don't include the T in H219T or the W in I175W.  For consistency across manufacturers we aren't currently showing the propellant type code; it's been requested that we add that back in, and we will (soon...). I hadn't realized until now that including it in the search would make the search fail -- though while you're typing the motor name in the search box you should see the motor until you reach the propellant designation.
2) Make sure that if you're planning on the motor extending out the front of your motor mount you don't have the "Limit motor length to mount length" box checked.
3) What version of OpenRocket are you running? The H219 wasn't in 15.03 (though the I175 was, so this is unlikely to be it).

PR #1738, opened 2022-10-12
Add motor designation to motor selection table

SiboVG:
See the discussion in #1729. Users sometimes search motors based on their designation, but the search only works for name-based searches. This PR adds a designation column to the motor selection table (apparently this was also requested by some users).

Screenshot 2022-10-12 at 17 31 18


neilweinstock:
Where do those designations come from?  Some in the screenshot above I find to be curious.

For example, the 13mm Quest is A3T, but Estes is just A3... I would normally refer to the Estes also as A3T.

SiboVG:
> Where do those designations come from? Some in the screenshot above I find to be curious.
> 
> For example, the 13mm Quest is A3T, but Estes is just A3... I would normally refer to the Estes also as A3T.

Don't know their backstory, but they are the same ones as shown on thrustcurves.org ([Estes A3](https://www.thrustcurve.org/motors/Estes/A3/), [Quest A3](https://www.thrustcurve.org/motors/Quest/A3T/))

neilweinstock:
OK, guess I can't argue with thrustcurve.

JoePfeiffer:
Yes, there's a consistency problem with the designations.  For instance, the "real" designations of the Estes C6 motors are C6-0, C6-3, and C6-5.  thrustcurve lists all these as C6, with several delays.  The Quest A8-3, meanwihle, has a designation of A8-3 (I'm not sure why a search on thrustcurve for C6 turns the Quest A8-3 up, incidentally).

I don't think we really need a designation column in the table; what we need is (1) more robust searching (there was a recent issue that turned out to be caused by searching failing if the user had the propellant designation in the search term) and (2) a checkbox for displaying propellant codes for motors that have them.

SiboVG:
> (2) a checkbox for displaying propellant codes for motors that have them.

Where would the propellant code be shown?

JoePfeiffer:
The manufacturers put it right after average thrust.  So Aerotech calls the G64 a G64W (W for White Lightning). That seems to be consistent for manufacturers that use a propellant code; note that not all of them use one -- I don't think I've ever seen a propellant code on a black powder motor, for instance.  The "T" in A3T above is a case size code that Estes used at one time for their 13mm motors, and apparently Quest still does.

neilweinstock:
Estes still uses the "T" for 13mm motors.

The weird thing here is that in the motor names themselves, the extra letter is added after the *delay*, not after the average thrust.  So you would have an A10-3T or an E20-7W.  But when the delay is not included (e.g. for thrustcurve data) they do just squash the suffix next to the average thrust.

I will further note that on Rocket Reviews, the Estes A3 *is* listed as A3T: https://www.rocketreviews.com/estes-a3t.html.  I would have thought they are getting their data from thrustcurve.org but maybe not.

JoePfeiffer:
Ouch, I stand corrected on both counts.  However, I still believe we should just display <motor class><avg thrust><optional propellant code>.  For the 13mm black powder motors, it wouldn't be hard to display the "T" if the propellant code box is checked if we feel that's valuable.

neilweinstock:
Without any consideration towards implementation, what I would want as a *user* is:
1) Always show me Designation.  I don't care about name.  (Possible exception: micro-maxx, not sure what to do there)
2) Allow me to search without the extra suffix.  E20 should find me E20W and E20 and E20X or whatever.

I can't think of an occasion where it is useful to me to see the names *without* the extra suffix.  And I don't really want to see both of them next to each other since they're almost the same thing, and it looks redundant.

hcraigmiller:

I agree with @neilweinstock's view on two points: (1) most users are more interested in the designation than the name and (2) displaying both the name and designation is redundant and wastes precious space within the motor pane.

If both are deemed necessary, perhaps a happy medium is to have a checkbox that switches between the two so that the user can choose which to display.

Build 1091

JoePfeiffer:
I have to say I'm surprised to see the strong preferences you have for the designation -- between its inconsistency between manufacturers, and truly bizarre designations like CTI uses, I'd expect users to want the common name (like I do). Please, let's have a checkbox selecting between them.

neilweinstock:
I think an option to select is fine.  Either way, though, when using designations, search should be able to handle queries where just the name is used (that is, searching for E20 should match E20W).

SiboVG:
Any suggestions on what the checkbox should look like? To me, a checkbox if for enabling/disabling one specific feature, not for switching between options (i.e. replacing the "Name" column with the "Designation" column). So in that regard, a checkbox isn't very intuitive to me. But if you can find the right wording for the checkbox, or a better UI widget, I'm all ears.

hcraigmiller:
Checking one, unchecks the other?

Motor Designation

neilweinstock:
> Checking one, unchecks the other?

In that case ==> radio buttons.  Could work.

SiboVG:
> In that case ==> radio buttons. Could work.

Radio buttons could work. So just a "Name" column, and  radio buttons with something like "Motor name display:", "Common name", and "Designation" (or "Manufacturer designation"?)

neilweinstock:
That's the general idea.  I think I'd say "Motor name column:" and then "Use common name" and "Use manufacturer designation" as the two buttons.  If there's no room for "manufacturer" then leave it out, but I'd include it if there's room.

This is pretty subjective, others may have different preferences here.

hcraigmiller:
I think I would use ThustCurve’s nomenclature, “Common Name” and “Designation.”

SiboVG:
> I think I would use ThustCurve’s nomenclature, “Common Name” (or just “Name”) and “Designation.”

Yes, but I find ThrustCurve's "Designation" label a bit confusing, adding the "Manufacturer" term clears the concept IMO. And the term "designation" is still present, so people who are used to ThrustCurve's nomenclature should still understand it.

JoePfeiffer:
Yes, "Manufacturer's Designation"

SiboVG:
New functionality:

https://user-images.githubusercontent.com/11031519/195698114-e812b682-c249-4738-b5d6-9b6ac85b31a5.mp4

neilweinstock:
That looks pretty good to me.

hcraigmiller:
Functions as expected, no anomalies found. Very nice additional feature.

Build 1096

PR #1741, opened 2022-10-14
[#1167] Remove modality for thrust curve import error dialog

### LINKED ISSUE #1167: Hanging on thrustcurve import
see https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-is-underway-download-it-now.171295/post-2246069

SiboVG:
This PR fixes #1167. I'm not a 100% pleased with this solution, but it's good enough for now. This PR just removes the modality from the thrust curve import error dialog, so instead of the BasicFrame waiting for the error dialog to be closed, the BasicFrame already loads while the error dialog is displaying. You will notice a small delay where the error dialog is blank, while BasicFrame loads, but after that delay (a second or two), both dialogs show up.
Screenshot 2022-10-14 at 15 43 54
Screenshot 2022-10-14 at 15 43 20

Here is a faulty test thrust curve, to fire the import error.
[Quest_A3T.eng.zip](https://github.com/openrocket/openrocket/files/9786887/Quest_A3T.eng.zip)

### DIFF
--- a/swing/src/net/sf/openrocket/database/MotorDatabaseLoader.java
+++ b/swing/src/net/sf/openrocket/database/MotorDatabaseLoader.java
@@ -153,7 +153,8 @@ public class MotorDatabaseLoader extends AsynchronousDatabaseLoader {
 						"<br>" + trans.get("MotorDbLoaderDlg.message2") + "</p></body></html>";
 				JOptionPane pane = new JOptionPane(message, JOptionPane.WARNING_MESSAGE);
 				JDialog dialog = pane.createDialog(null, trans.get("MotorDbLoaderDlg.title"));
-				dialog.setModalityType(Dialog.ModalityType.DOCUMENT_MODAL);
+				dialog.setModalityType(Dialog.ModalityType.MODELESS);
+				dialog.setAlwaysOnTop(true);
 				dialog.setVisible(true);
 			}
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
Is it intended that it also seems to not show the "loading" dialog? Regardless, I got as far yesterday as establishing I couldn't reproduce the bug under either Linux or Windows so seeing you've gotten it now looks good.

SiboVG:
> Is it intended that it also seems to not show the "loading" dialog? Regardless, I got as far yesterday as establishing I couldn't reproduce the bug under either Linux or Windows so seeing you've gotten it now looks good.

The loading dialog never shows on my machine, so I don't know what's different. You mean the "Loading motors..." dialog?

JoePfeiffer:
> > Is it intended that it also seems to not show the "loading" dialog? Regardless, I got as far yesterday as establishing I couldn't reproduce the bug under either Linux or Windows so seeing you've gotten it now looks good.
> 
> The loading dialog never shows on my machine, so I don't know what's different. You mean the "Loading motors..." dialog?

Yes -- you never see it, with or without an error in the motor loader?  I've always seen it until now.  Loading the motor database is fast enough that I'm not really sure what purpose it serves anyway.

PR #1752, opened 2022-10-17
[#1746] Fix "Seek value of" unit selector in rocket optimizer

### LINKED ISSUE #1746: [Bug] Unit selector in rocket optimization "seek value of" does not work
The unit selector for the "seek value of" spinner is just blank and unresponsive:
image

In addition, I'd also like to increase the spinner width to at least 3 digits, right now it's just one '0'.

SiboVG:
This PR fixes #1746.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/adaptors/DoubleModel.java
+++ b/swing/src/net/sf/openrocket/gui/adaptors/DoubleModel.java
@@ -580,7 +580,7 @@ public class DoubleModel implements StateChangeListener, ChangeSource, Invalidat
 	
 	private final ArrayList<EventListener> listeners = new ArrayList<EventListener>();
 	
-	private final UnitGroup units;
+	private UnitGroup units;
 	private Unit currentUnit;
 	
 	private final double minValue;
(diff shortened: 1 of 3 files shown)


hcraigmiller:
The spinner itself functions as expected with no anomalies found. Not so much for the results.

Parachute (ft/sec):
Optimize Parachute

This appears to be a situtation where each parameter will need to be tested and reported if in error. For example, body tube length optimization **functions perfectly when seeking stability by calibers**, but not so much for stability by inches:

Optimize Stability Inches

Build 1107

hcraigmiller:
After further testing, it appears that the "Seek value of" function is non operable across all attributes.

SiboVG:
Should hopefully be solved now (I didn't encounter the bug because my unit system was set to metric instead of imperial).

hcraigmiller:
Seek value works well; tested about 100 variations on different components. Functions as expected, no anomalies found.

Build 1117

SiboVG:
> tested about 100 variations on different components

Wow, kudos to you!

PR #1765, opened 2022-10-24
Improve sampling for SimulationRunDialog

### LINKED ISSUE #1728: [Bug] Uncaught exception when adding/running new simulation
With run outdated simulations on:

1. Open .ork
2. Run simulation with existing configuration.
3. Add second configutation (same one); runs very slow.
4. Go to flight simulations tab.
5. Add third configutation (same one)
4. Go to flight simulations tab (uncaught exception).

Crash
[Bug Report.txt](https://github.com/openrocket/openrocket/files/9742769/Bug.Report.txt)
[Sea Killer.ork.txt](https://github.com/openrocket/openrocket/files/9742770/Sea.Killer.ork.txt)

JoePfeiffer:
It turned out there were two (more) issues:

1. The simulation ran so fast the rocket was on the ground before the first progress update happened.  Since the SimulationStatus was simply being published, this meant that all of the entries in the publish queue were references to the same status, and all the samples before it were effectively lost. Publishing a copy of the SimulationStatus fixes this.

2. Since vertical velocity is slightly less than 0 when apogee is detected, the max altitude code wasn't picking up on the apogee event. Looking for either positive z velocity or actual increase in altitude fixes this (we still need to also update if z velocity is positive so we reset after "false apogees" caused by things like calculating optimal recovery delay).

It also turned out the apogeeAltitude variable wasn't quite redundant; it included a call to MathUtil.max() to make sure the recorded apogee couldn't be 0 (though, as the original bug report shows, there were pathological conditions under which that didn't work).  So that variable is back in "just in case".

Fixes #1728 (for real this time, I hope!)

### DIFF
--- a/swing/src/net/sf/openrocket/gui/simulation/SimulationRunDialog.java
+++ b/swing/src/net/sf/openrocket/gui/simulation/SimulationRunDialog.java
@@ -270,6 +270,7 @@ public class SimulationRunDialog extends JDialog {
 		private final int index;
 		private final double burnoutTimeEstimate;
 		private volatile double burnoutVelocity;
+		private volatile double apogeeAltitude;
 
 		private final CustomExpressionSimulationListener exprListener;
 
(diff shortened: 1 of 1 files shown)


SiboVG:
Functions as expected, no anomalies found (sorry for loaning your catchphrase @hcraigmiller 🙃).

PR #1773, opened 2022-10-27
[#1771, #939, #1191] FIX OVERLAY ISSUES IN 3D VIEW!!!!

### LINKED ISSUE #939: Numerous problems in 3D view
These problems are observed on a Mac with M1 processor on Big Sur 11.4.  As far as I can tell, they were all introduced in a recent PR.

1) Load the "Simple model rocket" example, and enable 3D Finished view.   Adjust the position of the rocket with the mouse.  Observed frame rate is now much lower than 15.03 (it is noticeably jumpy).

2) Rotate it so that the nose cone is pointing to the right (180 degrees from default).  On my machine it now looks like this:
image.  There are two problems here:
a) the original CG and CP markings on the right never moved.
b) new CG and CP markings on the left, in the correct location, are only drawn half.
c) there are some other weird aspects to the CG and CP markings, but hopefully it's all the same problem so I won't try to describe them all.

3) Add a mass component to the nose cone, set it to 4 oz. by using the "up" arrow on the Mass field.  Now looks like this:
image
a) The new CG marking has now disappeared.
b) The CG location in the upper right hasn't changed.  It will change to the correct value if you switch to the side view.
c) On my machine, it takes about a second to respond to each "up" click, vs. 15.03 which is nearly instant.

4) After switching back and forth between 3D view and Side view a few times, I was able to get the app to permanently beachball, and I needed to force quit.  I haven't been able to reproduce this yet, but it definitely happened once and should never happen.

### LINKED ISSUE #1191: [Bug] CP duplication in 3D views
TRF user K'Tesh found this bug: https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-public-beta-is-underway-download-it-now.171295/page-5#post-2247144.

K'Tesh provided the .ork file and I deconstructed the components to identify the cause. It appears to be related to _Fin 3_ on the sustainer. The attached screenshots show the reconstruction of _Fin 3_, in steps, and the effect of those steps as the bug manifests itself.

01 Reconstruction Fin_3

The .ork files are attached.  And, the file names are labeled on each screenshot.

In _Revision 4_ , _Fin 3_ has been reconstructed, except for the final point; its position is Base 0. This is the first that you get a hint that there's a problem when a second partial CP point appears.

In _Revision 5 , _Fin 3_ has been fully reconstructed; its position is Base 0.  Again, there is a hint that there is a problem when the CP disappears and reappears as the rocket is rotated horizontally.

In _Revision 6_ , _Fin 3_ has been fully reconstructed, ; its position is Base -1.125, the same as the other two fins. A second full CP point appears.

This particular design has a booster, and the Fin 3 on the booster causes the same issue.

Renamed .ork to .txt.

Original file:
[Estes Magnum Payloader (2032)(Clear Testbed).txt](https://github.com/openrocket/openrocket/files/8153272/Estes.Magnum.Payloader.2032.Clear.Testbed.txt)

Reconstructed _Fin 3_
[00.Build.Partial.01.04.05.Body_Tube.Fin-3.Booster_False.txt](https://github.com/openrocket/openrocket/files/8153212/00.Build.Partial.01.04.05.Body_Tube.Fin-3.Booster_False.txt)
[00.Build.Partial.01.04.06.Body_Tube.Fin-3.Booster_False.txt](https://github.com/openrocket/openrocket/files/8153213/00.Build.Partial.01.04.06.Body_Tube.Fin-3.Booster_False.txt)
[00.Build.Partial.01.04.04.Body_Tube.Fin-3.Booster_False.txt](https://github.com/openrocket/openrocket/files/8153214/00.Build.Partial.01.04.04.Body_Tube.Fin-3.Booster_False.txt)

### LINKED ISSUE #1771: [Bug] Stability info, warnings, and flight info are not updated in 3D view
When you are in the 3D view and update the rocket, only the main info (top-left) of the rocket is updated. The stability info (top-right), warnings (bottom-right), and flight info (bottom-left) are not updated. Only when you switch back to the 2D view, and then go back to the 3D view, the infos are updated again (well, that is, they just updated in the 2D view).

SiboVG:
This PR fixes #1771, fixes #939, and fixes #1191. No more duplicated CG/CP markings!!!

The issue was that the carets (and extra text, such as stability info - see issue #1771) are drawn on the JOGL Overlay class. The overlay class has a function to mark areas as "dirty", which means they should be repainted. The dirty area apparently wasn't sized properly, which this PR solves.

https://user-images.githubusercontent.com/11031519/198417059-70aaa186-4d41-4315-9d24-3c4438739b72.mp4

### DIFF
--- a/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
+++ b/swing/src/net/sf/openrocket/gui/figure3d/RocketFigure3d.java
@@ -324,9 +324,9 @@ public class RocketFigure3d extends JPanel implements GLEventListener {
 		}
 		rr.render(drawable, configuration, selection);
 		
-		drawExtras(gl, glu);
+		drawExtras(drawable, gl, glu);
 		if (drawCarets) {
-			drawCarets(gl, glu);
+			drawCarets(drawable, gl, glu);
 		}
 		
 		// GLJPanel with GLSL Flipper relies on this:
@@ -353,12 +353,12 @@ public class RocketFigure3d extends JPanel implements GLEventListener {
 		return og2d;
 	}
 	
-	private void drawCarets(final GL2 gl, final GLU glu) {
+	private void drawCarets(final GLAutoDrawable drawable, final GL2 gl, final GLU glu) {
 		final Graphics2D og2d = createOverlayGraphics(caretOverlay);
 		
 		og2d.setBackground(new Color(0, 0, 0, 0));
-		og2d.clearRect(0, 0, canvas.getWidth(), canvas.getHeight());
-		caretOverlay.markDirty(0, 0, canvas.getWidth(), canvas.getHeight());
+		og2d.clearRect(0, 0, drawable.getSurfaceWidth(), drawable.getSurfaceHeight());
+		caretOverlay.markDirty(0, 0, drawable.getSurfaceWidth(), drawable.getSurfaceHeight());
 		
 		// The existing relative Extras don't really work right for 3d.
 		Coordinate pCP = project(cp, gl, glu);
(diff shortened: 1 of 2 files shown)


neilweinstock:
You are _the man._  :)

hcraigmiller:
I have been unable to replicate any of the issues referenced in Issues 939, 1191, and 1771.

Build 1134

PR #1789, opened 2022-11-04
[#106] Support SubAssembly RockSim import

### LINKED ISSUE #106: Rocksim Subassemblies are not supported in OpenRocket, No warning on import
Rocksim Subassemblies are not supported in OpenRocket, with no warning given when a rocksim file containing those is opened in OpenRocket - those features will be silently dropped from the model.

Rocksim file showing problem (bottom half of rocket is lost on import) http://www.rocketreviews.com/wildman-rocketry-wildman-rocket-dual-deploy-7143.html

Possible fix: either warn on open, or add support for subassemblies.

Original Report:
http://www.rocketryforum.com/showthread.php?55541-TWO-OpenRocket-Questions

SiboVG:
This PR fixes #106 and adds support for RockSim SubAssembly importing. 

Comparison after import from RockSim:
image

Note that 'Body tube 1' and 'Transition 2' are ignored because they cannot be added to respectively a nose cone and transition. Attempting to import it, you get the following warning:
image


hcraigmiller:
Perfection. Functions as expected, no anomalies found.

FYI, the mass and CG overrides in RockSim are connected; when you override the mass or CG, it overrides both. So, when imported into OpenRocket, both are overridden, and _this is the correct behavior_. The result is that is you save an .ork as an. rkt. then open it again in OpenRocket, all components with a mass override (but not a CG override) will have both the mass and CG overridden, and _this is the correct behavior_. 
 
Build 1151

PR #1795, opened 2022-11-06
[#1348] Add support for RockSim importing and exporting of Pods

### LINKED ISSUE #1348: Does not import RKT pods
I have not tried many file formats but it seems to be well-known that OR does not import pods from RKT files. It would be incredibly valuable if it could parse the data and at least get whatever comparable elements of the pods imported that it can.

For example, this X-15 RKT files does not import: https://www.apogeerockets.com/downloads/rocksim_files/ApogeeX-15.rkt

SiboVG:
This PR adds the support for importing and exporting pods from and to RockSim. This fixes #1348, although the pods from the design file of #1348 do not get imported, because those use pods on a nose cone, whereas OpenRocket only supports pods on body tubes. Maybe make a new feature request for pods on nose cones and transitions?

"Demo":
OpenRocket Pods example (simplified to use only one pod - as RockSim does, and an offset angle)
Screenshot 2022-11-06 at 17 04 20

Imported into RockSim:
Screenshot 2022-11-06 at 17 04 31

RockSim file imported back into OR:
image


hcraigmiller:

Build 1158

neilweinstock:
This is good.

I suspect that the real useful feature will be when we support pods on fins.  That's the most natural way to do things like fins on fins and designs like the sample used in the PR.

If I recall correctly, Rocksim supports pods on just about everything, and you don't need to have phantom body tubes (you can just attach fins directly to pods).

Pods on nose cones and transitions are probably not frequently used.

PR #1797, opened 2022-11-06
[#1794] Clamp IntegerModel to min and max on set

### LINKED ISSUE #1794: [Bug] Pods "Number of copies" spinner can have 0 and negative value
You can type 0, and negative number in the pods configuration "Number of copies" spinner, which should not be possible. Typing e.g. '0' does not have an effect on the actual pod instance count though.

SiboVG:
This PR fixes #1794 by clamping IntegerModel values to min and max, similar to what was done in PR #1615.


hcraigmiller:

Build 1160

Issue #1790, opened 2022-11-04
[Bug] Wrong simulation results from TARC Payloader example rocket


SiboVG:
Bug report from Chris Flanigan on the SourceForge bug mailing list:

---------- Bug report ----------

*Include detailed steps on how to trigger the bug:*

*(You can edit text directly in this window)*

1. start OpenRocket

2.  "File, Open Example, TARC Payloader"

3.  Select "Flight simulations" tab

4.  "Plot/Export, Plot" (to display the saved simulation results)

5.  Close the plot windows

6.  "Run Simulations" (to recalculate the simulation results)

7.  "Plot/Export, Plot" (to display the newly calculated simulation results)

The new results do not match the saved results.  It appears that the
parachute for the "Payload" never gets activated.
*What does the software do and what in your opinion should it do in the
case described above:*

The TARC Payloader example indicates that something about upper stage
parachutes isn't working.  The results plotted after doing "Run
simulations" doesn't match the original results.  The results also don't
match the OpenRocket v15 results for the TARC Payloader example.

If you designate the "Body tube" of the "Payload" stage as a motor mount,
add a tiny motor with a short delay (such as 1/4A3 with 0.1 second delay),
and change the "Payload" parachute to "First ejection charge of this
stage", the results still don't look correct.  It looks like the "Payload"
stage is using the delay time from the booster motor.

Include your email address (optional; it helps if we can contact you in
case we need additional information):

chris...@gm...

(Do not modify anything below this line.)

---------- System information ----------

OpenRocket version: 22.02.beta.05

OpenRocket source: default

OpenRocket location:
C:\Users\cflanigan\AppData\Local\OpenRocket\.\OpenRocket.jar

JOGL version: 2.4.0-rc-20210111

Current default locale: en_US

System properties:

awt.toolkit=sun.awt.windows.WToolkit

exe4j.consoleCodepage=cp0

exe4j.isInstall4j=true

exe4j.launchName=C:\Users\cflanigan\AppData\Local\OpenRocket\OpenRocket.exe

exe4j.moduleName=C:\Users\cflanigan\AppData\Local\OpenRocket\OpenRocket.exe

exe4j.semaphoreName=Local\c:_users_cflanigan_appdata_local_openrocket_openrocket.exe0

exe4j.tempDir=

exe4j.unextractedPosition=0

file.encoding=Cp1252

file.separator=\

install4j.appDir=C:\Users\cflanigan\AppData\Local\OpenRocket\

[...]

PR #1799, opened 2022-11-07
Wait to do TUMBLE processing until actual TUMBLE event is obtained from queue

JoePfeiffer:
Inhibit tumbling if recovery device has been deployed.

Also, run TARC payloader simulation and save new simulation data.  I compared our simulation data with an on-line descent rate calculator at https://descentratecalculator.onlinetesting.net/ and saw that the results are well within 1 fps of each other.

Fixes Issue #1790.

The tumble condition was being triggered on the same step as the recovery system deployment.  Consequently, the recovery system was being ignored and the rocket was being simulated as tumbling to the ground.  It turns out this issue also existed under 15.03, but a much slower descent rate was being calculated for the tumbling rocket because 15.03 wasn't considering the motor mass when the rocket was tumbling.


hcraigmiller:

Build 1163

PR #1800, opened 2022-11-07
Fix 1702

JoePfeiffer:
Calculates CP and CNa for entire component tree when performing component stability analysis so results can be shown for entire rocket.

Replaces name of Rocket component with Total, for consistency with drag analysis

Roll dynamics should probably also have a Total row, but that should be a separate issue.

I was thinking about showing the stability impact of all component assemblies, but concluded that it added more confusion than clarity, especially since we can look at the individual stages by disabling the ones we aren't interested in.


hcraigmiller:

Build 1168

PR #1805, opened 2022-11-09
[#1792] Don't clamp SymmetricComponent thickness when importing

### LINKED ISSUE #1792: [Bug] Transition wall thickness is lost when saved and reopened
Issue Centering Ring

[Terrier.Rev_02.Stripped.ork.txt](https://github.com/openrocket/openrocket/files/9944776/Terrier.Rev_02.Stripped.ork.txt)

SiboVG:
This PR fixes #1792. The issue was that when a transition, nose cone, or body tube got imported, first the thickness of the component was set and then the fore and aft radius. However, the thickness setter method clamped the thickness if it was greater than the aft and fore radius. Since the correct aft and fore radius was not yet set at that point, the thickness setter took the default fore and aft radius, which is 2.5 cm. So: if you saved the thickness with a value greater than 2.5 cm, that value would get clamped when re-opening the design.

This PR ignores the clamping of the thickness during import.

I also took the liberty of fixing the maximum value of the thickness slider, which was fixed at 1 cm. Now it updates dynamically with the fore and aft radius.

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
+++ b/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
@@ -171,9 +171,10 @@ class DocumentConfig {
        
 		// SymmetricComponent
 		setters.put("SymmetricComponent:thickness", new DoubleSetter(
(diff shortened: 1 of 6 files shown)


hcraigmiller:

Build 1181

PR #1811, opened 2022-11-10
[#1479] Add option to automatically open preset dialog when creating a new component

### LINKED ISSUE #1479: Automatically open Preset selection window when creating a new component (for those types that support it)
I propose that whenever creating a new component, the preset selection window should be opened automatically.

This is not a no-brainer, and I'm not sure I've completely convinced myself of this.  But it's worth consideration as a quality-of-life improvement.  Along with it, it might be necessary to provide a way to override it, perhaps a checkbox in the preset dialog that says "Always open this dialog when creating a new <component type>".  This preference should be separate for each component type, rather than global, and therefore I think it would have to be here rather than in a big mess in Preferences.

To some extent, it undercuts the ability to add a part to the drop-down menu.  But I don't know how many people use that, *and* the way it works isn't super useful (that will be a subject of another feature request).

Anyway, something for consideration.  I'm trying to think of ways to simplify the initial entry of a rocket, and this would be one.

SiboVG:
This PR fixes #1479. There is now an option in the component preset dialog for opening the dialog when creating a new component. The dialog will not automatically open when you have created a component, close the config dialog, and edit the component again. It only works for newly created components. This feature is enabled by default. If you want it disabled by default, let me know.

Demo:

https://user-images.githubusercontent.com/11031519/201204718-232c73f5-e9f0-4736-a122-58e0dc4e6151.mp4

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -2127,6 +2127,7 @@ ComponentPresetChooserDialog.menu.sortDesc = Sort Descending
 ComponentPresetChooserDialog.menu.units = Units
 ComponentPresetChooserDialog.checkbox.showLegacyCheckBox = Show Legacy Database
 ComponentPresetChooserDialog.lbl.favorites = Check to add preset to the preset drop-down menu in the component edit dialog<br>Directly apply a preset by double-clicking it or by selecting it and closing this window.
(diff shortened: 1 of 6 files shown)


hcraigmiller:

I like the ability to turn individual component preset dialogs on and off (I think this is needed). It would just be nice to have a "master" reset on way or the other for all of the component preset dialogs (to be able to have either all on or all off as a starting point). Just a thought, but works well as is too.

Build 1186

PR #1822, opened 2022-11-13
Provide user with warning when a simulation branch contains no data

JoePfeiffer:
Also, add a method to WarningSet making it easy to add a "discriminator" to warnings, informing the user what in a design or simulation is causing the warning.

The attached .ork results in an empty data branch for Stage 1
[empty-booster.ork.txt](https://github.com/openrocket/openrocket/files/9998418/empty-booster.ork.txt)


JoePfeiffer:
> Running simulation 3 in empty-booster.ork does not throw a warning, shouldn't it?
Yes... it does when I run it...
empty-branch-warning

PR #1834, opened 2022-11-15
[#1831] Don't highlight sustainer stage at startup

### LINKED ISSUE #1831: [Feature request] Don't highlight rocket on start-up
OpenRocket automatically selects the sustainer stage when a design is opened, which will also highlight all that stage's children in the rocket view. This is 99% of the case not what users want (at least I catch myself deselecting the components every time I open a design). So I propose to not highlight the stage's children at start-up. However, if the sustainer stage is selected on start-up, with no children highlighted, clicking the sustainer again should highlight the children, instead of first having to select a different component and then the sustainer to highlight the sustainer components.

SiboVG:
This PR fixes #1831.

https://user-images.githubusercontent.com/11031519/202036630-730980cc-94ae-4ae0-bc6f-3fe572114cd4.mov


neilweinstock:
Wait, so the word "Sustainer" is highlighted, but then you click on it again and Sustainer stays highlighted but then that toggles the bolding of the lines?

That is a bit strange, although probably better than what's there now.

SiboVG:
This was the only "logical" solution IMO. You want the sustainer highlighted so you can add components to it straight away. It's just simply disabling the highlight behavior at start-up and if the user desires it, he/she can click the sustainer again. This is the solution with the least amount of extra clicks.


hcraigmiller:

I like this behavior much better than it was before.

Build 1205

PR #1835, opened 2022-11-16
Tell user where discontinuities in body tube diameter are

JoePfeiffer:
Now that warnings have an easy way to add additional information, let's make use of it to tell users *where* discontinuities exist in body tubes.

A simple .ork showing the changed warning is attached.

Fixes 241

[discontinuous.ork.txt](https://github.com/openrocket/openrocket/files/10024071/discontinuous.ork.txt)

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1829,7 +1829,7 @@ PlotConfiguration.Groundtrack = Ground track
 ! Warning
 Warning.LargeAOA.str1 = Large angle of attack encountered.
 Warning.LargeAOA.str2 = Large angle of attack encountered (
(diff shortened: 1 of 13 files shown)


neilweinstock:
What is the change in the warning system?  I must have missed that when it flew by.

JoePfeiffer:
Tiny change as part of #1822 -- we can now pass in an optional string which is appended to the warning message, so we can give the user details on what caused a warning.

neilweinstock:
Ah.  That's a good incremental improvement (and this PR is good) but a deeper solution in the longer term will be to provide warnings with much more than text... like a link to a list of components which led to the warning.  Then links could be provided to take the user directly to those components, or the components could be highlighted when the warning is clicked or moused over or something like that.

Issue #1672, opened 2022-09-13
[Feature request] Additional surface roughness selections


hcraigmiller:
If the intention is to be able to export to a RASAero design file, the component finish, surface roughness selections needs to be expanded and/or allow user input values.

At the present time, OpenRocket (with 5) and RASAero (with 8) vary so greatly on this factor that it will be difficult to have the export be accepted by RASAero without the user having to input the surface roughness. This is where the two stand:

![Surface](https://user-images.githubusercontent.com/68821492/189810257-733960a0-c2df-442e-8bde-419669084da6.png)

 Only two have the same name, but are materially different in value.  I believe that OpenRocket should at least have six options, with 0.00 being the addition, if that is the minimum expansion that is done.

JoePfeiffer:
To make it even better, RASAero and OR use different definitions of roughness.  RASAero uses equivalent sand roughness while OR uses measured roughness height.  I've found papers trying to develop algorithms to convert between them, but haven't pursued them.

What I'd really like to see would be a roughness slider running from 0 to .02 (which is basically the full range from the existing smooth to rough), with a way to "snap" to the existing values in the current dropdown.  Unfortunately, the friction drag code heavily relies on caching values in a table indexed by the materials; I want to change that, but haven't really thought about it.  I figure it's a topic for "after this release".

PR #1852, opened 2022-11-25
[#1672] Add additional surface roughness selections

hcraigmiller:
Adds additional surface roughness (finish) selections that conform with RASAero II.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1505,11 +1505,14 @@ material.tubular_nylon_14_mm_9_16_in = Tubular nylon (14 mm, 9/16 in)
 material.tubular_nylon_25_mm_1_in = Tubular nylon (25 mm, 1 in)
 
 ! ExternalComponent
-ExternalComponent.Rough = Rough
+ExternalComponent.Sawnsurface = Sawn Surface (Very Rough)
+ExternalComponent.Roughsand = Rough Sand
 ExternalComponent.Unfinished = Unfinished
-ExternalComponent.Regularpaint = Regular paint
-ExternalComponent.Smoothpaint = Smooth paint
+ExternalComponent.Flatpaint = Flat Paint
+ExternalComponent.Glosspaint = Gloss Paint
+ExternalComponent.Highglosspaint = High Gloss Paint
 ExternalComponent.Polished = Polished
+ExternalComponent.Smooth = Smooth (Zero Roughness)
 
 ! LineStyle
 LineStyle.Solid = Solid
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1506,10 +1506,10 @@ material.tubular_nylon_25_mm_1_in = Tubular nylon (25 mm, 1 in)
 
 ! ExternalComponent
 ExternalComponent.Sawnsurface = Sawn Surface (Very Rough)
-ExternalComponent.Roughsand = Rough Sand
+ExternalComponent.Rough = Rough Sand
 ExternalComponent.Unfinished = Unfinished
-ExternalComponent.Flatpaint = Flat Paint
-ExternalComponent.Glosspaint = Gloss Paint
+ExternalComponent.Regularpaint = Flat Paint
+ExternalComponent.Smoothpaint = Gloss Paint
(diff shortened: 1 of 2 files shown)


JoePfeiffer:
Looking at the old values, the values from Hoerner (which are where Sampo got the old values), RASAero's values, and the values in the PR it seems like it would be a good idea to use the same terms for the roughnesses as the original source (I'm disappointed to see OR doesn't do that at present).

We also need to retain the old finish names since those are stored in the .ork files.

SiboVG:
We indeed need to find a way to maintain compatibility with the old roughness values.

Maybe add an option for custom roughness values, similar to component materials, and set the name the something like "Legacy Regular Paint (60 µm)"?

SiboVG:
> Maybe add an option for custom roughness values, similar to component materials, and set the name the something like "Legacy Regular Paint (60 µm)"?

Actually, we could probably even keep the current names, instead of adding the "Legacy" text. Will save us some complexity for backward compatibility.

hcraigmiller:
I changed the 5 original names back, just changing the text description and value (slightly), and then added 3 new ones.

JoePfeiffer:
Something I'd like to do (which I've been postponing until after final release) is allowing the user to directly enter a surface roughness, rather than having it tied to a name.  Backward compatibility of this will be easy; if the  stored finish is text it needs to be a name; if it's a number it's the actual roughness.

JoePfeiffer:
I still have two questions -- why change the values (if the answer is you've found a more accurate source they should definitely be changed), and how are the new descriptions clearer than the old ones? I don't know how to think about the roughness of a sawn surface.

neilweinstock:
I'm unclear what the plan is here, can't test it because build is failing.

"Flat paint", "gloss paint", and "high gloss paint" do not IMHO convey the surface smoothness.  Are those the terms that RASAero uses?

It's not the type of paint so much as the quality of the application.  I routinely use gloss paint, and after applying a coat or two of Future I suppose it might even qualify as "high gloss".... but it's nowhere near as smooth as the super-polished paint jobs I've seen pictures of on TRF.  Likewise, I can easily imagine that a very well-applied flat paint job would have less surface drag than a crappily-applied gloss paint.

The existing terms do a good job of conveying the _quality_ of the paint job, regardless of the paint type.

hcraigmiller:
I am in the process of reviewing the "Thesis" and other sources so that I may put forth a reasoned proposal. At least one value in the current unstable is significantly different than the thesis (the thesis polished value is .5 microns and OpenRocket is using 2 microns for the polished value). I hope to have the balance of the data no later than 2022-11-29.

JoePfeiffer:
The thesis values are a subset of the values from Hoerner's book; as long as there are no name conflicts I wouldn't object at all to merging the names and values from Hoerner and RASAero.

hcraigmiller:
I have compiled the surface roughness data available from the Thesis, Hoerner Fluid Dynamics, RASAero II, Rocksim, and sources related to sandpaper surface roughness. I have created two suggested proposals in which eight surface finishes are proposed in dark shaded red (current finishes, if not reused, are shown in light shaded red).

[Proposal 1.pdf](https://github.com/openrocket/openrocket/files/10109129/Proposal.1.pdf)
[Proposal 2.pdf](https://github.com/openrocket/openrocket/files/10109130/Proposal.2.pdf)

The names are sourced from the Thesis, Hoerner Fluid Dynamics, OpenRocket, RASAero II, Rocksim, or a combination of these. However, some names may need to be changed.

Eight surface finishes are being suggested so that the future implementation of an export to RASAero II format will be to the same number of finishes, although this is not absolutely necessary.

hcraigmiller:
I received a response from Tim at Apogee regarding what Rocksim uses for surface roughness values. He wrote: "Unfortunately, I don’t have any data on the finish of the rocket that you want. I’m sure it is buried in the code, but even I can’t see that. Nor would I know where to look, since I’m not a programmer."

hcraigmiller:
PR #1863 created to resolve error in polished surface roughness value. In light of desire to complete the version release before Christmas, and this PR being in need of further research and discussion, especially where RASAero II compatibility will not be complete for the version release and the Rocksim values are still unknown.

PR #1858, opened 2022-11-30
[#1768] Add welcome dialog

### LINKED ISSUE #1768: Add welcome dialog on first start-up of new OR version
Whenever a user downloads a new OR version, and opens it for the first time, it would be nice to have a welcome dialog pop up stating something like "Thank you for downloading OpenRocket version xx.xx", and the release notes for that release. I would also add a checkbox stating "Don't show this dialog again". If unchecked, the welcome dialog will pop up every time a user opens OR. If checked, the dialog will be ignored.

For the official 22.02 release, we should also include the known issues in the GitHub release comments, so that it is clear what users should be aware of **ahum** 3D rendering **ahum**.

SiboVG:
This PR fixes #1768 and adds a welcome dialog when opening a new version of OR for the first time:
Screenshot 2022-11-29 at 23 53 04

The release notes are retrieved from the `ReleaseNotes.md` file, which I've had to refactor to use html-tags in order to easily fetch the release notes from that file. Note: the markdown layout (what GitHub previews when you open that file) did not change because of this.

There is an option to don't show the welcome dialog again. This setting-saving is version-specific. So if you ignore the dialog in OR 22.02.beta.04, the welcome dialog will still show up in 22.02.beta.05.


hcraigmiller:
Very nicely done. Functions as expected, no anomalies found.

Build 1236

PR #1861, opened 2022-12-01
Remove bounds on minimum and maximum zoom sizes

### LINKED ISSUE #1857: (beta 5): Opening file causes stack overflow error
Opening this file in beta 5 causes a crash.
image

[Saturn V Full Scale.ork.zip](https://github.com/openrocket/openrocket/files/10109313/Saturn.V.Full.Scale.ork.zip)

JoePfeiffer:
Trying to set a zoom smaller than the minimum size was provoking an infinite recursion.  It's not clear to me why we need min and max in the first place (the user can always use "Fit" to get back to something reasonable) so just get rid of them.

Fixes #1857


hcraigmiller:

Build 1240

PR #1863, opened 2022-12-01
Correct "polished" surface roughness value 

hcraigmiller:
Change incorrect value for polished surface roughness from 2 to .5 microns per Thesis and Hoerner Fluid Dynamics.

01 01 02 Thesis Roughness

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/ExternalComponent.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/ExternalComponent.java
@@ -29,7 +29,7 @@ public abstract class ExternalComponent extends RocketComponent {
 		//// Smooth paint
 		SMOOTH("ExternalComponent.Smoothpaint", 20e-6),
 		//// Polished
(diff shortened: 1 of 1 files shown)


SiboVG:
Hm, there's still some compatibility issues. If a user has a design saved with the old finished and polished surface value and opens that design in OR 22.02 (with the adjusted value), then the simulation results will change because suddenly you drop from 0.5 microns to 0.02 microns.

I really admire your efforts on fixing the surface roughnesses in OR @hcraigmiller, but IMO this is something we should fix after the final release, because it's really easy to break compatibility and change users' simulations unwantedly.

neilweinstock:
I agree with @SiboVG.  This is a good thing to clean up, but needs to be done carefully and it is not urgent to squeeze into this release.

hcraigmiller:
Just trying not to have future designs perpetuate a known error.

hcraigmiller:
Closed, see PR #1875.

Issue #1548, opened 2022-07-21
Support bilateral symmetry in instanced components (fins, pods, boosters)


neilweinstock:
Could have sworn I submitted a request for this in the past, but I don't see it anywhere so here it is.

Instanced components should have an option for bilateral symmetry, in which case:
1) The number of instances is 2
2) The "angle" parameter specifies offset of each instance from vertical
3) In the case of pods, the angle offset for child components is opposite on each side (whereas today they all go in the same direction)
4) from a CP standpoint, each set of two would need to be treated as two single instances, since there's likely no shortcut to creating an aggregate effect of the set.

I envision a checkbox for "bilateral symmetry" that, when checked, forces the number of instances to 2 and disables that field.

This would be a real time- and effort-saver for any airplane-like design (of which there are *many*).

hcraigmiller:
Would you describe/give examples of how OR functions now and how it would function with the suggested changed?

Though only one component bears the nomenclature, there are currently 7 components in 3 groups that have instances:
- Booster Set (Number of copies); Pod Set (Number of copies);
- Trapezoidal Fin Set (Number of fins); Elliptical Fin Set (Number of fins); Freeform Fin Set (Number of fins); Tube Fin Set (Number of fins); and
- Rail Button (Instance Count).

I am assuming that where you reference "airplane-like" designs that, of the foregoing components, Rail Buttons and Fin Sets are not within what you are calling "instances components." In other words, are you referring only to Booster and Pod sets when you say "instanced components"?

Is this an example of what you are trying to achieve, but with only 1 pod set, by using "bylateral symmetry"?

![Winglet Annotated](https://user-images.githubusercontent.com/68821492/180341499-faa507ee-a381-4309-ae69-c0f379366133.png)

neilweinstock:
Rail buttons should be excluded, but fin sets are most definitely included. Consider if the wings in your image above have a dihedral or anhedral.  Consider all the fins on the Estes Interceptor.

In the present implementation, all those fins must be entered as two single fins, rather than as sets of two as would be more appropriate.
________________________________

hcraigmiller:
So, is this closer to descibing what you are trying to achieve?

![Rotation Rev_01](https://user-images.githubusercontent.com/68821492/180354164-56636d9a-00a0-4b01-8ace-4272170c4358.png)

neilweinstock:
Exactly.
________________________________

neilweinstock:
[...]

PR #1866, opened 2022-12-01
Support multi-comp fin set shape editing

SiboVG:
This PR support multi-component fin set shape editing.

The basic principle is:
1. Editing the first fin point of the main fin set will edit the first fin point of the listener fin set
2. Editing the last fin point of the main fin set will edit the last fin point of the listener fin set
3. Editing intermediate fin points of the main fin set will edit the correspondingly indexed fin points of the listener fin set
    1. If the main fin set has more points than the listener fin set, then ignore the out-of-bound points for the listener fin set
    2. If the main fin set has more points than the listener fin set, e.g. main set has 5 points, listener set 4 points, if you edit point 4 of the main fin set (second-to-last point if the main set, last point of the listener set), then point 4 of the listener fin set will **not** be edited. You can only edit the last point of the listener set by editing the last point of the main fin set.

Basically a lot of rules that try to say: "Just use multi-comp fin set shape editing for fins that have the same shape".

https://user-images.githubusercontent.com/11031519/205123751-037f18f7-aa4f-4dbb-9cdb-cb41008396bd.mp4

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FreeformFinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FreeformFinSet.java
@@ -103,6 +103,43 @@ public class FreeformFinSet extends FinSet {
 		}
 		return freeform;
 	}
+
+	/**
+	 * Converts a point of this fin set to edit into a point for a config listener to edit.
+	 *
+	 * The editing is as follows:
+	 * 		1) Editing the first point of this fin set will always edit the first point of the listener set
+	 * 		2) Editing the last point of this fin set will always edit the last point of the listener set
+	 * 		3) Editing any other point of this fin set will edit the corresponding point of the listener set, except
+	 * 		   for when the current point is not the last of this set, but it is the last of the listener set. In that
+	 * 		   case, no listener point will be edited.
+	 *
+	 * @param listener the listener which point needs to be edited
+	 * @param index the point index of this fin set that is being edited
+	 * @return the point index of the listener fin set that needs to be edited. Returns -1 if the listener's point should not be edited.
+	 */
+	private int getConfigListenerPointIdx(FreeformFinSet listener, int index) {
+		/*
+				The editing is as follows:
+					1) Editing the first point of this fin set will always edit the first point of the listener set
+					2) Editing the last point of this fin set will always edit the last point of the listener set
+					3) Editing any other point of this fin set will edit the corresponding point of the listener set, except
+					   for when the current point is not the last of this set, but it is the last of the listener set. In that
+					   case, no listener point will be edited.
+				 */
+		if (index == this.points.size() - 1) {
+			// If editing the last point, also edit the last point of the listener
+			return listener.getPointCount() - 1;
+		} else {
+			if (index == listener.getPointCount() - 1) {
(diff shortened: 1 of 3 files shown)


SiboVG:
@neilweinstock made the remark that this behavior may be too complex for users to phantom and we should protect them from messing things up by hiding the "Shape" panel when multi-editing freeform fin sets. I'll leave that here as a discussion topic.

hcraigmiller:
Functions as described; no anomalies found as to the fin-point change capabilities described. I presume that this multi-component editing PR is intended only to change the fin shape, not to increase the fin count, split fins, or change fin types (such as changing trapezoidal to freeform) this PR does none of these.

I disagree with @neilweinstock regarding hiding the "Shape" panel when multi-editing freeform fin sets. First, I'm not sure that an objective in developing OpenRocket should be to dumb-down its capabilities to "protect" users from "messing things up." I have always believed, and still do believe that an important element of OpenRocket is to educate, to improve a user's knowledge and abilities, not intentionally suppress that opportunity.

Second, this feature would have tremendous benefits to those who regularly use freeform fin sets. Again, I see no need to blunt the capabilities of OpenRocket for one user because another may not understand how to correctly use it. Look at the overrides, talk about an area that is rarely fully understood and has the potential for huge mistakes by less knowledgeable users.

What's the saying, we learn from our mistakes. My thoughts.

Build 1247

neilweinstock:
> I disagree with @neilweinstock regarding hiding the "Shape" panel when multi-editing freeform fin sets. First, I'm not sure that an objective in developing OpenRocket should be to dumb-down its capabilities to "protect" users from "messing things up." I have always believed, and still do believe that an important element of OpenRocket is to educate, to improve a user's knowledge and abilities, not intentionally suppress that opportunity.
> 
> Second, this feature would have tremendous benefits for those who regularly use freeform fin sets. Again, I see no need to blunt a capability of OpenRocket for one user because another may not understand how to correctly use it. Look at the overrides, talk about an area that is rarely fully understood and has the potential for huge (possibly catastrophic) mistakes by less knowledgeable users.

I'm not convinced it's so useful, especially when the fins are not the exact same shape to begin with.

Also, in general, I would not equate "dumbing down" with "removing features that are pointlessly dangerous and/or confusing, with low benefit".  Learning how to navigate software pitfalls and/or poorly implemented features is not the sort of "improving a user's knowledge" that I consider to be worthwhile.

However, in this particular case, although the proposed functionality is weird, I don't think that it's likely to be activated by accident and cause unintended problems for a user, and therefore I'm OK with it.  And I understand the reason for the behavior, although it's more of a developer reason than a user-focused reason IMHO.

The main use I foresee for it is in managing bilateral symmetric fin sets (which of necessity must be implemented as two separate single fins, see #1548)....  In that case the different fins will always be identical.  I can't think of an instance where editing two different-shaped fins at the same time makes any sense.

SiboVG:
I'll ask @JoePfeiffer to make the final decision on this feature.

neilweinstock:
In case it was lost in my previous screeds, I'm OK with this feature as it currently stands.  I don't think there's any reason to hold it back.

JoePfeiffer:
Let's put it in and see what the users say.

PR #1873, opened 2022-12-07
Convert user simulation listeners to extensions

### LINKED ISSUE #1710: Simulation Listener Flight Data Types not added to Custom Expression variables
### Description
When using a Simulation Listener such as `net.sf.openrocket.simulation.listeners.example.DampingMoment`, the FlightDataType defined by the listener (in this case the Damping moment Coefficient, Cdm) is not added to the Custom Expression's variable list. The data tpye will show up in the simulation results though.

### Instructions for replication
- Create/open rocket and edit/add simulation.
- Add `net.sf.openrocket.simulation.listeners.example.DampingMoment` as Simulation Extension with User Code > Java code.
- Close simulation (running simulation should not be required as per wiki, however, when attempting to run this listener, another bug is found see #1709).
- Open Custom Expressions and add new expression
- Insert variable
- Notice "Damping moment coefficient" with symbol "Cdm" is not present
- Attempting to manually input Cdm will show the expression as invaild.

### Apparent Cause
`OpenRocketDocument.getFlighDataTypes` is responsible for acquiring all flight data types and is used by Custom Expressions for validation and display. This function does loop through all simulations and in each simulation gets the flight data types from each simulation's extensions, however, as simulation listeners are wrapped in the `JavaCode` simulation extension, the flight data types defined in a simulation listener are not properly forwarded to the simulation extension and thus are not included by `OpenRocketDocument.getFlighDataTypes`.

I attempted to fix this myself, however, am not yet familiar or confident enough with OpenRocket's code to make the modifications appropriately. I would be interested in contributing to this fix, but would need a few pointers/suggestions.

Thanks,
Ray

JoePfeiffer:
Investigating #1710 led to me learning that there is a much nicer infrastructure for adding extensions to OR than the listeners documented in http://wiki.openrocket.info/Simulation_Listeners

Adding extensions using the extension infrastructure puts them in the menu structure opened by clicking "Add extension" in the "Simulation options" pane of the "Edit simulation" dialog, and also makes it possible to configure them with a crude GUI.

It looks like somebody created this extension infrastructure, ported one of the existing simulation listeners to it, and stopped. I don't know whether the listeners were ever able to write values to the flight status, but they certainly don't now -- this was the actual #1710.  Finishing the switch to the more general extension architecture seems like a better use of time than fixing the actual reported issue.

This PR converts all of the listeners to extensions, and creates GUIs where appropriate.  It leaves the old Java listener in place for any simulations out there that may use it, and also leaves the AirStart listener both so we can test and so the relationship between it and the AirStart extension is clearer to programmers coming behind us.

Testing:
Open a .ork, go to the "Flight simulation" pane, open the "Edit simulation" dialog.
Go to the "Simulation options" pane, and click "Add extension" to open the simulation extension menu.
The individual extensions are:

- Launch conditions->Air-start
Opens a GUI permitting you to select an altitude and velocity for the first iteration of a simulation
- Reports->CSV Save
Saves some Flight status values to a CSV file after every simulation step
- Reports->Print Simulation
Dumps a summary of simulation progress to standard out after every simulation step
- Post-step flight conditions->Damping moment Coefficient (Cdm)
Computes the damping moment coefficient and stores it as a flight variable after every simulation step
(the old listener not doing this was the actual issue reported in #1710)
- Scripts->Java listeners
Permits the user to enter a fully qualified class name to be used as a simulation listener.  This is the old listener interface; entering net.sf.openrocket.simulation.listeners.example.AirStart in the box will begin a simulation at an altitude of 1000m.  My next task is to write the wiki page referenced in the info box for this item
- Scripts->Javascript
Opens a window in which you can enter a JavaScript script.  To see this in use, open the example .ork file "Simulation extensions and scripting".  There will be a warning while opening the file; click OK.  Editing the "Active roll control" simulation and then the "Javascript script" configurator will show the code
- Control Enhancements->Roll control
Opens a GUI to configure a PID controller providing active control of a canard stopping roll.  To see this one in use, use the "Simulation Extension" example .ork
- Simulation Conditions->Stop Simulation
Opens a GUI allowing the user to stop a simulation after a given amount of simulation time or number of steps (whichever comes first)
- Copy extension->(several)
Allows you to copy an extension from another simulation in the same .ork file

Fixes #1710


hcraigmiller:
Functions as described when following instructions above.

Build 1254

hcraigmiller:
Functions as described when following instructions above.

Build 1257

PR #1874, opened 2022-12-08
[#1021] Add extra root point(s) if fin is outside parent's bounds

### LINKED ISSUE #1021: Bug: Fin root edge not drawn correctly on transitions.
There is a anomaly when fins are attached to transitions. Rather than referencing the end of the body tube as the set point when drawing the root edge beyond the body tube, the first fin point is used. So, you get A instead of B, as shown below.

0000 Fins

I am assuming that the same technique is being used when drawing the root edge beyond the body tube, it just doesn't create a visual anomaly because the root edge is flat instead of angled.

SiboVG:
This PR fixes #1021. There is now an extra root point added at the front of the parent when the first fin point goes out of the parent's bounds. Same effect for the last point. When both the first and last fin point are outside the bounds (either to the left or right), no extra root point is added.

https://user-images.githubusercontent.com/11031519/206334470-56ddebac-726c-4f22-b6a6-e6a4d8cac56c.mov

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -1036,14 +1036,83 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 	 * @return points representing the fin-root points, relative to ( x: fin-front, y: centerline ) i.e. relto: fin Component reference point
 	 */
 	public Coordinate[] getRootPoints(){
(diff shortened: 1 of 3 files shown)


hcraigmiller:
Very nice. Functions as expected, no anomalies found.

Build 1255

JoePfeiffer:
Merged it and the github action build failed (ummm, it was working before I commited?), so reverted.

PR #1875, opened 2022-12-08
Corrects polished value and adds 4 additional surfaces

hcraigmiller:
Corrects polished value and adds 4 additional surfaces.


JoePfeiffer:
I like this approach better than my earlier suggestion on Slack -- just report POLISHED to the user as sheet metal.  I'd suggest adding a comment so people will understand the reason when looking at the code later.

JoePfeiffer:
Exactly

PR #766, opened 2020-08-30
Feature Implements Aerodynamics on Railbutton components

teyrana:
```
public class RailButtonCalc extends RocketComponentCalc {

       private double CDmul;
       private double refArea;

       public RailButtonCalc(RocketComponent component) {
               super(component);

               final RailButton button = (RailButton) component;

               final double outerArea = button.getTotalHeight() / button.getOuterDiameter();
               final double notchArea = 2*( (button.getOuterDiameter() - button.getInnerDiameter()) * button.getInnerHeight());

               CDmul = 1.0;
               refArea = outerArea - notchArea;
       }

       @Override
       public void calculateNonaxialForces(FlightConditions conditions, Transformation transform,
                       AerodynamicForces forces, WarningSet warnings) {
               // Nothing to be done
       }

       @Override
       public double calculatePressureDragForce(FlightConditions conditions,
                       double stagnationCD, double baseCD, WarningSet warnings) {

               return CDmul*stagnationCD * refArea / conditions.getRefArea();
       }
```

PR #1876, opened 2022-12-11
[#1567] Implement rail button screw

### LINKED ISSUE #1567: Feature request: Extend configuration parameters for rail buttons to include exposed screws
Extend the current parameters to include exposed screws. The user would be able to set the screw height, but the diameter of the screw head would be OuterDiameter. There would be no need for the complexity of a drop down to implement this, just the ScrewHeight field.

Rail Button Component Rev_04

The screw head height is already a field within the rail button database (see PR #1540).

See discussion in PR #1538.

SiboVG:
This PR fixes #1567 and implements a rail button screw.

https://user-images.githubusercontent.com/11031519/206908350-c272c66b-a97c-4229-92e7-b0debfe0edf4.mp4

I also noticed that the height sliders in the rail button config panel were not scaled linearly and that the minimum and maximum value of the sliders did not adjust dynamically. This PR fixes that as well.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -1077,9 +1077,10 @@ LaunchLugCfg.tab.Generalprop = General properties
 ! RailButtonConfig
 RailBtnCfg.lbl.OuterDiam = Outer Diameter:
 RailBtnCfg.lbl.InnerDiam = Inner Diameter:
-RailBtnCfg.lbl.TotalHeight = Total Height:
+RailBtnCfg.lbl.TotalHeight = Total Height (excl. screw):
 RailBtnCfg.lbl.BaseHeight = Base Height:
 RailBtnCfg.lbl.FlangeHeight = Flange Height:
+RailBtnCfg.lbl.ScrewHeight = Screw Height:
 RailBtnCfg.lbl.Angle = Rotation:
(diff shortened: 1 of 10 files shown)


neilweinstock:
Very nice.  Are the buttons and/or screws factored into aerodynamic calcs yet?  I can't remember.

SiboVG:
> Very nice. Are the buttons and/or screws factored into aerodynamic calcs yet? I can't remember.

Nope, not yet. PR #766 started it, but it was never finished.

neilweinstock:
In that case, since it looks like that will not get done for this release, I think we should have a warning in the config dialog to that effect.  Rail buttons are the only component in the "body components and finsets" group that are not factored in, and so users might not realize their special status.

neilweinstock:
As a further thought on this, I think the ability to fine-tune the screw dimensions will even further lead folks to think that it will have an effect on aerodynamics, otherwise why bother?

Definitely should have a visible warning there.

hcraigmiller:
Beautiful. Functions as expected, no anomalies found.

Build 1258

SiboVG:
> In that case, since it looks like that will not get done for this release, I think we should have a warning in the config dialog to that effect. Rail buttons are the only component in the "body components and finsets" group that are not factored in, and so users might not realize their special status.

Filed as #1878.

PR #1879, opened 2022-12-12
Surface roughness comments

hcraigmiller:
Sorry @JoePfeiffer, I hadn't pushed the comments yet, I wanted to make sure that the comments were okay with you before doing so.


JoePfeiffer:
Oops!  Sorry for jumping the gun.

Addendum to #1875

PR #1881, opened 2022-12-12
Increase default PS settings window size

SiboVG:
If you opened PhotoStudio for the first time, the settings window was way too small:
Screenshot 2022-12-12 at 21 05 40

This PR fixes it:
image

You can now also close the settings dialog using the escape key, and the settings dialog now gains focus when starting up PhotoStudio, instead of the PhotoFrame having focus.


neilweinstock:
Nice.

PR #1884, opened 2022-12-13
Update unit tests for fin point exception

SiboVG:
This PR fixes the build exception on PR #1874, by including the fin point exception in the unit test.


SiboVG:
Okay, you know what, this is gonna be too complicated. I hoped I could undo the revert commit using this PR, but I'll just push it directly on the unstable branch and ignore this PR.

PR #1887, opened 2022-12-14
This is a rework of the railbutton aerodynamics part of PR 766.

JoePfeiffer:
Drag calculation should be reasonably accurate up to transonic speeds.  Hoerner summarizes several sources demonstrating Cd of a cylinder is roughly 1.2 from Reynolds numbers 10^4 to something over 2*10^5, which is something a little over supersonic (page 3-9, figure 12).  I expect we'll get into compressibility effects before that, so this should be reasonably good up until transonic range.


SiboVG:
It looks like you didn't update your branch with the PR that implements the rail button screw? (because you've deleted some of the code that came from PR #1876).

neilweinstock:
If this is good until transonic range, what happens above that?  Can you characterize the way in which it might be inaccurate at that point?

JoePfeiffer:
> 
I'm not sure -- the figure from Hoerner shows Cd remaining pretty constant until a Reynolds number of a little over 2*10^5, and then a quick drop until it's at about .2 at something like 3*10^5 (for a standard 1010 railbutton, the Reynolds number is about 2x10^5 at 300m/s, which is just under Mach 1).

I'm guessing that compressibility will raise the Cd before that, and then the shock wave will mask the railbutton above it.  So "good until transonic" is all I really want to commit to.

neilweinstock:
In that case do we still need to attach a warning to it?  One unobtrusive way to do that would be to put a warning on any *simulation* that includes railbuttons and exceeds mach1.

wolsen (approved):
LGTM

JoePfeiffer:
> LGTM
Ah, let's not (yet).  I came across some very good information on fixing the supersonic part of it.

wolsen:
> Ah, let's not (yet).  I came across some very good information on fixing the supersonic part of it.
> 

Ah yeah, was just marking my approval. As it's in draft, I wasn't willing to merge.

Good to hear on the supersonic part.

hcraigmiller:

Note of comparison with RASAero. Using the _A simple model rocket_ example, removing launch lug and matching simulated altitude (Sustainer Cd set to 0.009 [visually rounds to 0.01]), using Estes C6-7, yields altitude of 1201ft in both programs. Add mini rail buttons and RASAero drops to 936 ft with OR dropping to 1129. In order to match the RASAero altitude projection, the OR rail button Cd needs to be set to .160 (.157 if using 2 separate instances).

Build 1292

JoePfeiffer:
Hmmm, please post the final .rkt and .ork files.

hcraigmiller:
Here are the OpenRocket and RASAero II design files for comparison; Rocksim does not support rail buttons.

[A simple Model Rocket.Rail Buttons.zip](https://github.com/openrocket/openrocket/files/10254824/A.simple.Model.Rocket.Rail.Buttons.zip)

JoePfeiffer:
I'm seeing the same thing...

I wish there was documentation from Rogers regarding just how he simulates rail guides (I don't find "The rail guide drag model for model and high power rockets is particularly accurate" to be helpful).  If I go in to my code and set the boundary layer depth to 0 and the referance area to simple height times width (ignoring the notch), my results get closer to RASAero's (I matched the rocket mass of 1.75oz dry weight, then had to override Cd by -.015 to match altitude) -- with those changes, I get a closer result of 963 feet.

hcraigmiller:
No argument here, my note was kind of an FYI to be aware of and check in case it was something simple.

wolsen (approved):

SiboVG:
Thanks @JoePfeiffer! 💪

PR #1908, opened 2022-12-19
Update A simple model rocket.ork

hcraigmiller:
Change root and tip chords from 50mm to 2in and sweep length from 25mm to 1in to avoid unknown conversion error.

Change_Chord_Dimensions


SiboVG:
Is the conversion error caused by RockSim, or by the OpenRocket conversion? From what I can tell, OR does its job, but RockSim has an anomaly.

hcraigmiller:
It has something to do with the way Rocksim is dealing with the RootChord. Follow these steps:

- Open the _A simple model rocket_ example and export it to Rocksim.
- Open the export file in Rocksim and immediately save it. Call it "A simple model rocket.Save_00.Unopened.rkt".
- Open the export file in Rocksim again, then open the configuration for the fins and immediately click "okay" to close the window. Save this file as "A simple model rocket.Save_01.Opened.rkt".

When these files are reopened in Rocksim. unopened looks like the left picture above and opened looks like the right picture.

If you run a file comparison between the two saved files, this is what you get (blue is what Rocksim changed, red is what Rocksim changed it to):
[Compare.pdf](https://github.com/openrocket/openrocket/files/10282585/Compare.pdf)

Although there are several changes that Rocksim has made to the original dimensions, the only one that matters is the length of the RootChord value. When the fin configuration window is opened, then closed, the RootChord dimension changes from "50." to "49.9999". If you open "A simple model rocket.Save_00.Unopened.rkt" in a text editor and change the value of the RootChord from "50." to "49.9999", then save and reopen the file in Rocksim, it looks as it should, like the right picture.

The value of 49.9999 is what you get if you convert 50mm to inches (rounding to 5 significant digits), then convert it back to inches with 4 significant digits. This seems to be what is being done with most values, convert from mm to inches, round, reconvert to mm. I don't know why the RootChord is the only fin dimension that, when changed, makes it so the fin displays correctly

[A simple model rocket.Save_00.Unopened.rkt.txt](https://github.com/openrocket/openrocket/files/10282716/A.simple.model.rocket.Save_00.Unopened.rkt.txt)
[A simple model rocket.Save_01.Opened.rkt.txt](https://github.com/openrocket/openrocket/files/10282720/A.simple.model.rocket.Save_01.Opened.rkt.txt)
[A simple model rocket.Save_00.Unopened.RootChordChange.rkt.txt](https://github.com/openrocket/openrocket/files/10282722/A.simple.model.rocket.Save_00.Unopened.RootChordChange.rkt.txt)

SiboVG:
Alright, thanks for the detailed analysis!

PR #1909, opened 2022-12-19
Require designation match

JoePfeiffer:
Require motor designations to match before motors can be said to match.  I *think* this is really leftover unfinished business from my PR #1032; that PR stopped trying to parse motor designations to create a "simplified designation" instead using common names for most purposes that previously used simplified designations.

The main reason for this is to get rid of the annoyance of MICRO_MAXX and MICRO_MAXX_II motors being confounded with each other in the motor selection dialog.

In general, I can't think of  a reason we'd want to be able to have multiple designations turn up as a single motor.

One possible concern with this is we want to make sure we can still find motors saved in 15.03 designs when loading from the new version.  That was only really a problem with CTI motors; I tried creating a couple of designs with CTI motors under 15.03 and then opening the file; at least for the ones I tried this found them.


hcraigmiller:
Looks good to me. Functions as expected, no anomalies found.

Build 1308

Issue #1898, opened 2022-12-17
[Bug] Simulation fails


hcraigmiller:
Open the attached design file and go to the Flight simulations tab. Add a simulation using the Estes E12-0 motor and close the pane without running the simulation. Then run the simulation from the Fight simulation tab.

![Time](https://user-images.githubusercontent.com/68821492/208264328-85eee803-b928-4ff0-b738-586e91bbac25.png)

[01.03.01.Motor Mount.24mm.Estes 95mm Engine Hook.ork.txt](https://github.com/openrocket/openrocket/files/10252661/01.03.01.Motor.Mount.24mm.Estes.95mm.Engine.Hook.ork.txt)

[Windows 11 Pro; Version 22H2; OS Build 22621.521; Windows Feature Experience Pack 1000.22634.1000.0]
[Java "11.0.15" 2022-04-19 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.15+8-LTS-149)]

JoePfeiffer:
I'm not sure this is even a bug...  I'll see about either putting a clamp on the simulation step and/or putting up an exception for a "rocket" that's got so little aerodynamic surfaces it can't be simulated.

hcraigmiller:
Fixed by way of PR #1914]

JoePfeiffer:
No, #1914 is just helping me find what's going on with this.  The simulation is still failing on an exception.

Trying to solve this one is an interesting exercise, but in spite of the labels I'm not sure it's really any kind of blocker.  Basically a totally ridiculous simulation fails.  My current plans on solving it are --

1. see if I can find a way to limit the sim step size to keep crazy aerodynamics from provoking a numerical instability.  This shouldn't be difficult, but I need to look at it seriously in the morning.
2. if not, see if I can characterize "crazy" in a way that can bring up a popup refusing to even try.  This one has a zero-length total external symmetric components so it would be easy to identify if the first idea is too hard; is there a more general characterization?
3. punt.  Anybody who tries to sim a naked motor mount really deserves what they get...

PR #1914, opened 2022-12-21
Attach flight data to simulation exceptions so it can be preserved to help user track down the problem

JoePfeiffer:
So for instance, in the flying motor mount in Issue #1898, the simulation can proceed through many steps before failing with a SimulationException.  This lets a user look at the data generated before the exception.

### DIFF
--- a/core/src/net/sf/openrocket/document/Simulation.java
+++ b/core/src/net/sf/openrocket/document/Simulation.java
@@ -384,15 +384,19 @@ public class Simulation implements ChangeSource, Cloneable {
 			simulatedData = simulator.simulate(simulationConditions);
 			t2 = System.currentTimeMillis();
 			log.debug("Simulation: returning from simulator, simulation took " + (t2 - t1) + "ms");
-			
-			// Set simulated info after simulation, will not be set in case of exception
+
+		} catch (SimulationException e) {
+			simulatedData = e.getFlightData();
+			throw e;
+		} finally {
+			// Set simulated info after simulation
 			simulatedConditions = options.clone();
 			simulatedConfigurationDescription = descriptor.format( this.rocket, getId());
 			simulatedRocketID = rocket.getFunctionalModID();
 			
 			status = Status.UPTODATE;
 			fireChangeEvent();
-		} finally {
+
 			mutex.unlock("simulate");
 		}
 	}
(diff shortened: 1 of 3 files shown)


hcraigmiller:
I know this is going to be a dumb question, but how do you look at the data?

JoePfeiffer:
> I know this is going to be a dumb question, but how do you look at the data?

Not dumb at all, I should have thought to include some screenshots.

If you follow the instructions in issue #1898 with current unstable, after running the simulation the data table looks like this:
no-data
There's no simulation summary data stored, and the Plot/Export button attempts to rerun the simulation.  As long as it continues to get the exception, it doesn't save any data.

With Build 1314, following the same instructions and then closing the exception popup window, we get the following data table:
with-data
Yes, the data is ridiculous -- but that's what #1898 is reporting.  It's now possible to plot the data or export it as a csv to try to see where the sim went wrong.

hcraigmiller:

Question. The plot of the simulation ends with "Exception", but the Warning pop-up displays the standard warnings (no recovery device, large angle of attack, etc.). If the simulation fails, would it be more visible/recognizable to the user to replace the standard warnings with something like "Exception--Simulation failed" or something akin to that. 

Build 1314

wolsen (approved):
It would be nice to see some unit tests with this... however, this is quite a straightforward change.

FWIW, I think there are pros and cons to reveal the flight data to the user this way. The biggest pro is that its already there and safe for release inclusion. The biggest con is that it relies on the user being able to interpret the data, which may not be the case for a novice. However, there's already a lot of cases that does require the user to be able to interpret the data. This at least makes the data available so its better in that sense. In the future, we should probably explore what the best way to present scenarios like this to help the user.

wolsen:
> Functions as described; no anomalies found.
> 
> Question. The plot of the simulation ends with "Exception", but the Warning pop-up displays the standard warnings (no recovery device, large angle of attack, etc.). If the simulation fails, would it be more visible/recognizable to the user to replace the standard warnings with something like "Exception--Simulation failed" or something akin to that.

Heh, I merged it at the same time that @hcraigmiller made this comment. I think this is a good suggestion, but I'd like to suggest to hold off. This close to release, the minimal number of changes should be considered unless something *really* warrants more than a basic fix.

JoePfeiffer:
This is all data that's already available for any sim that doesn't have an exception, so any issues with the user being able to interpret the results are already present. I actually did this so I could look at what's going on with #1898!

To me it seems like a popup saying the sim data is out of bounds is about as clear a statement that something has gone *really* wrong as we can get. We still want those warnings, since those are things that (most of the time) happened earlier in the simulation and hopefully will give users clues to where to look for their issues.

Issue #1893, opened 2022-12-16
[Feature request] Add "Top view" to View Type


hcraigmiller:
After reviewing nearly four dozen designs created by users, it has become painfully clear that users are building their designs using the side view as if it were the top view. Rather than perpetuate this technical error, a "top view" should be added to the View Type options. Doing so will allow users build in the top view (many currently use the side view as if it were the top view) and to more easily switch between top, side, and back views, as if looking at a blueprint.

PR #1916, opened 2022-12-21
[#1893] Add top view option to 2D display

hcraigmiller:
Adds a 2D top view option.


hcraigmiller:
Something got lost in translation. . . with the exception of the message and the label, it works as expected (I like the suggestion of @SiboVG to keep the side view as the primary view).

Untitled

SiboVG:
> Something got lost in translation. . . with the exception of the message and the label

Was this produced with [openrocket_build_1319](https://github.com/openrocket/openrocket/suites/10001235923/artifacts/485585890)? I can't reproduce it. There was an issue with the translation file, but that was fixed in the last commit.

hcraigmiller:
Perfection again. I wish I had your skill set @SiboVG. Functions as expected, no anomalies found.

Build 1322

SiboVG:
> Perfection again. I wish I had your skill set @SiboVG.

I don't know, being skilled in Java is both a blessing and a curse.

hcraigmiller:
It's been a great blessing for the development of OpenRocket. Your efforts are much appreciated.

neilweinstock:
How is a top view different from what’s there now?  I won’t ba able to try it until tomorrow at least.

wolsen:
> How is a top view different from what’s there now? I won’t ba able to try it until tomorrow at least.

This was my thought - isn't this the side view with a 90 degree rotation?

Additionally, I think this should be considered post release? I mean, is it strictly necessary?

hcraigmiller:
It's like a 90-degree snap of the 2D view. You can easily switch between the views without trying to get the exact angle on the slider and then have to put it back.

Side-Top

SiboVG:
I had the same initial reaction as @neilweinstock and @wolsen, but have since "seen the light". You can indeed achieve the same effect with using the rotation slider, but the slider is **very often** overlooked (or users don't understand what it does - as was the same for me), and the extra view also has the extra benefit that you can quickly switch between different views, without having to always pin-point the rotation slider to 90°.

SiboVG:
To more easily differentiate between the 2D and 3D views, I added a horizontal separator between the items
image

hcraigmiller:
Nice touch. I think that will be helpful to users.  Functions as expected, no anomalies found.

Build 1326

JoePfeiffer:
Would it be reasonable to move the rotation slider to 90 when switching to top view? I think of the switch as being a 90 rotation; it would also remind users of the slider's existence.

hcraigmiller:
This actually an approach decision that needs to be made. Coming from a background in drafting, the menu order should be Top View, Side View, Back View. That same background suggests that _the default view should be the side view_. So, if the slider is to snap 90 degrees when changing between the top and side views, to which view should the back view be tied. From a drafting standpoint, that would be the side view.

Against that backdrop, there are two approaches, either have the slider stay the same when changing between the top and side view or snap 90 degrees when this change is made. As to the first approach, @SiboVG (correct me if I misunderstood) took the approach that the slider is an X degree rotation from the Y view, so the slider does not change when views are changed.

The second approach would be that the top and side views are 90 degrees in rotation from one another, so the slider should reflect that change. This was my initial thought, but after considering the interplay with the back view, the back view is neither a 90-degree rotation from the top nor side view.

Although snapping the rotation 90 degrees would make the use of the slider more evident, as @JoePfeiffer noted, I wonder if it would also bring with it some level of confusion for the user. My vote would be for not snaping the slider 90 degrees when changing between the top and side view. The slider would merely be an X degree rotation from the Y view.

SiboVG:
I dislike the idea of the 90° value for the top view, because it changes how the rotation slider is supposed to work. It's supposed to give a rotation offset to the current view type. Setting the slider to 90° in the top view makes it feel like the top view is not a dedicated view, but rather a 90° rotated side view. (which it is, but it feels wrong with my - sorry for the subjective argument)

It does bring attention to the rotation slider, but I'd rather stick with the equal rotation value across all view types.

neilweinstock:
I don't think there's a strictly "right" answer here, but I think I agree with @SiboVG, let's leave it at 0 by default.

JoePfeiffer:
To me, it seems like one view is a rotation of the other, so the slider should move.  But I'll bow to you guys.

SiboVG:
Last commit sets the top view as the first view type (makes a bit more sense for the top view to be actually on top). The default view type is still the side view.

image

hcraigmiller:
Looks like this one is done. Nicely done. Functions as expected, no anomalies found.

Build 1347

Issue #1919, opened 2022-12-24
[Bug] OR 22.02.beta.05 freezes on malformed thrust curve import, throws exception in 15.03


SiboVG:
From [this thread on TRF](https://www.rocketryforum.com/threads/openrocket-wont-load.176381/):

A user was reporting he could no longer open OpenRocket. After further investigation, OpenRocket 15.03 threw a `java.lang.IllegalArgumentException: MALFORMED` exception:

> java.lang.IllegalArgumentException: MALFORMED
	at java.util.zip.ZipCoder.toString(ZipCoder.java:58)
	at java.util.zip.ZipInputStream.readLOC(ZipInputStream.java:300)
	at java.util.zip.ZipInputStream.getNextEntry(ZipInputStream.java:122)
	at net.sf.openrocket.file.motor.ZipFileMotorLoader.load(ZipFileMotorLoader.java:56)
	at net.sf.openrocket.file.motor.GeneralMotorLoader.load(GeneralMotorLoader.java:36)
	at net.sf.openrocket.database.MotorDatabaseLoader.loadDirectory(MotorDatabaseLoader.java:124)
	at net.sf.openrocket.database.MotorDatabaseLoader.loadDatabase(MotorDatabaseLoader.java:69)
	at net.sf.openrocket.database.AsynchronousDatabaseLoader.doLoad(AsynchronousDatabaseLoader.java:103)
	at net.sf.openrocket.database.AsynchronousDatabaseLoader.access$100(AsynchronousDatabaseLoader.java:8)
	at net.sf.openrocket.database.AsynchronousDatabaseLoader$LoadingThread.run(AsynchronousDatabaseLoader.java:126)

However, OpenRocket 22.02.beta.05 just freezes, with no exception popping up.

This issue is not about solving the MALFORMED exception, but about figuring out why the beta does not throw the exception. The user should always get feedback on why something fails, instead of the program just freezing... 

The fix for the exception itself will be dealt with on the TRF, or if necessary, a new issue will be created for it.

SiboVG:
Update: the user reported that he re-installed the beta, and now it opens just fine... So I'm just gonna close this issue, because it seems like it'll be difficult to replicate the issue now. Also, PR #1920 should reduce the likelihood of this happening again.

PR #1920, opened 2022-12-24
Generalize error-catching of motor file loading

SiboVG:
In light of PR #1919, I think it is reasonable to generalize the exception-catching for motor file loading. So, instead of only catching IOExceptions during motor file loading, now every exception is caught. The upside is that however messed up the files and directories in the user's thrust curve directory is, it will (hopefully) not crash the entire program - as was the case for the TRF user from PR #1919. The downside is that - assuming that the exception dialogs work (PR #1919 shows that this is not always the case) - users will not have a direct feedback that an import failed due to a file/directory issue. You can still view the import error in the debug log though. 

FYI: with "file/directory issue", I mean an issue caused by the raw format of the file/directory, e.g. when you have a corrupted file. I'm not talking about the normal thrust curve import issues like "Thrust curve xx has two different thrust values for the same time value". Those issues will still show up fine. This PR only affects raw file import errors.

### DIFF
--- a/core/src/net/sf/openrocket/file/motor/ZipFileMotorLoader.java
+++ b/core/src/net/sf/openrocket/file/motor/ZipFileMotorLoader.java
@@ -44,7 +44,7 @@ public class ZipFileMotorLoader implements MotorLoader {
 	
 	
 	@Override
-	public List<ThrustCurveMotor.Builder> load(InputStream stream, String filename) throws IOException {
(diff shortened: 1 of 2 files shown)


hcraigmiller:
I couldn't get it to fail or act abnormally, but this is not an area that I'm very familiar with.

Build 1331

SiboVG:
> I couldn't get it to fail or act abnormally, but this is not an area that I'm very familiar with.

This is a hard thing to test, this PR is more about code-checking. I also couldn't produce a scenario where I got anything else from an IOException.

PR #1924, opened 2022-12-24
Throw exception at simulation startup if CP can't be calculated.

### LINKED ISSUE #1898: [Bug] Simulation fails

JoePfeiffer:
Exception says "Can't calculate rocket stability."

The basic problem with #1898 is it's impossible to calculate the flying motor mount "rocket"'s center of pressure except when it's flying backwards (at nearly 180 degrees).  This results in ever-faster gyrations until one parameter or another exceeds OR's limits.  I spent a little while seeing if I could adjust the simulation time step in response to the speed of the gyrations; the end of result of these experiments was I could calculate with ever-tinier timesteps which made the simulation speed drop to a crawl before finally throwing the exception.

This PR observes that the CP of the rocket assembly has a 0 or negative weight at launch and throws a more meaningful exception at that point.

Fixes #1898

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -562,6 +562,7 @@ SimuRunDlg.msg.Unabletosim = Unable to simulate:
 SimuRunDlg.msg.errorOccurred = An error occurred during the simulation:
 
 BasicEventSimulationEngine.error.noMotorsDefined = No motors defined in the simulation.
+BasicEventSimulationEngine.error.cantCalculateStability = Can't calculate rocket stability.
 BasicEventSimulationEngine.error.earlyMotorBurnout = Motor burnout without liftoff.
 BasicEventSimulationEngine.error.noConfiguredIgnition = No motors configured to ignite at liftoff
 BasicEventSimulationEngine.error.noIgnition = No motors ignited.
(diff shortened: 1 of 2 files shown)


hcraigmiller:

Build 1340

SiboVG:
Doesn't the exception need extra information? As a user I would be very confused as to which part of my rocket to adjust.

JoePfeiffer:
> Doesn't the exception need extra information? As a user I would be very confused as to which part of my rocket to adjust.

Yes and no.  At the point the exception is thrown, there's really no way (that I can think of anyway) to pinpoint what components are wrong; it just all adds up to a calculation that can't be right.

What I do want to do (which is separate from this PR) is add some geometry warnings similar to the existing jagged fin warning -- two that immediately come to mind is that there are internal components that extend forward of the rocket airframe, and the front of the airframe is open.  I also hope to implement Craig's suggestion that we should flag pods that are in line with a parent component.  Finally, at least on my machine, those geometry warnings are very light in the rocket panel; they need to be bold and a darker color so they stand out more.

SiboVG:
> Finally, at least on my machine, those geometry warnings are very light in the rocket panel; they need to be bold and a darker color so they stand out more.

I'll open a PR for it.

PR #1927, opened 2022-12-25
Display nan as n/a

JoePfeiffer:
This is related to PR 1924, but separate enough I wanted it to be its own PR.  Either of these two can be merged without the other.

This one makes two small changes: 
Iinstead of displaying Not a Number values as NaN, Unit.toString() now displays them as N/A.  This brings it into agreement with Unit.toStringUnit(), which already did that.  While a developer knows what NaN means, someone without a software background will probably find N/A more understandable.  You can see a place where toStringUnit() was using N/A in the RocketPanel display from Issue #1898, where it is used to identify the CP (which has a negative weight) as N/A

In the ComponentAnalysisDialog, if the rocket CP has a negative weight, this now displays it as N/A.  Previously, it would go ahead and display the calculated value, while the RocketPanel would display N/A.  The dialog will now display the same value, including N/A where appropriate, as the RocketPanel.

### DIFF
--- a/core/src/net/sf/openrocket/unit/Unit.java
+++ b/core/src/net/sf/openrocket/unit/Unit.java
@@ -90,8 +90,11 @@ public abstract class Unit {
 	 * @return       A string representation of the number in these units.
 	 */
 	public String toString(double value) {
-		double val = toUnit(value);
+		if (Double.isNaN(value))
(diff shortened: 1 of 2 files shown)


hcraigmiller:

Build 1345

PR #1928, opened 2022-12-26
New app icons *DEMO*

neilweinstock:
This PR is intended exclusively for previewing a new proposed icon design.  @SiboVG and I have been working on this for a while.  Updates include all app icons, About logo, and splash screen.

The quality here is not perfect, again it is just for a preview.  Grab the JAR and try it out and provide feedback.

Although I *really* think the new version should have new icons, I'm pretty much exhausted by this effort, so if folks don't like it I'll drop it.


neilweinstock:
Hmm, build is failing but all I changed were PNG files.  This effort is doomed. :(

neilweinstock:
Working now.  ¯\_(ツ)_/¯ 

Here's a shortcut to the JAR file: https://github.com/openrocket/openrocket/suites/10053241838/artifacts/489225878

hcraigmiller:
I think the artwork is great, but the task bar icon in Widows should be changed to have a transparent background.

Task Bar

neilweinstock:
Sibo and I have gone back and forth on this quite a bit. 

Would like to hear from @JoePfeiffer and @wolsen before I continue fiddling.

hcraigmiller:
My preference is "D" below.

Task Bar Icons

JoePfeiffer:
To me it looks great.  Clearly in the same lineage as the old, but an unmistakable update.  I agree with Craig that a transparent background would be desirable. I don't fully understand the difference between icons C and D above; the main difference I see is that the icon content is slightly larger?  Neither of them have a transparent background.

hcraigmiller:
Icons B through D all have transparent backgrounds (any color is that of my task bar). The difference between C and D is that the content of D is larger because the border space was reduced to 1 pixel before it was resized to 32 pixels.

JoePfeiffer:
Ah, OK -- the others had no apparent background, so I assumed the background color I was seeing was "really" there.  In that case I agree that D is the best of the examples; is there a reason the border can't be eliminated completely (reduced to 0 pixels) to fill the space as completely as possible?  Visually, it appears to have more than one pixel.

hcraigmiller:
The reduction of the border to 1 pixel out of 100 pixels, prior to reducing the icon to 32 pixels, resulted in a 32 by 32 with no border with minimal effect on the rocket itself.

neilweinstock:
I will optimize the images to fill the available area, fear not.  If there’s no background then there will be no border.

My main qualm about the plain digital rocket is that there are many logos (a surprising number) that look like that, up to snd including the rocket 🚀 emoji.

I wish there were something too we could add just to give it a bit of identity.

hcraigmiller:
I think the earth in the icon had that effect, but the pods cannot really be seen in the version with the earth. I think most rely on the desktop icon to be distinctive enough to be identifiable, and I believe the pods alone version does that.

Here is a comparison between the simulators (I didn't have a higher resolution of the new "earth" one):
DeskTop

JoePfeiffer:
Could the pods be emphasized either by making them larger or darker?  It does seem like the earth adds a lot of identity.

hcraigmiller:
Task bar icon comparison (with new updated OR icon):

Task Bar 2022-12-27 01

neilweinstock:
The orientation of the rocket has changed a few times along the way.  Without the background, it may be better to move away from the (almost) pure side view.  It's a bit more distinctive (maybe?) this way.

Here's a version with crappy flames added back, since they now fit.
image
image
image
image

wolsen:
> Could the pods be emphasized either by making them larger or darker? It does seem like the earth adds a lot of identity.

Overall the new artwork looks great! I agree that the Earth adds a lot of identity here. It is included in some of the icons, but not all of them. Either way, this is a great new set of logos with the transparent background and such.

Also, can you please update the `snap/gui/openrocket.png` file as well? It is used for the snap's desktop icons.

hcraigmiller:
In windows:

DeskTop-Task Bar Rev_02

PR #1931, opened 2022-12-29
Add open airframe warning and use text compare to find airframe discontinuities

### LINKED ISSUE #999: Unit rounding creating discontinuity issues.
See prior discussion thread, https://github.com/openrocket/openrocket/issues/871#issuecomment-899110349

JoePfeiffer:
Adds a warning if the forend of an airframe is open (ie there isn't a nose cone or equivalent at the start of an airframe or pod)

Changes the test for diameter discontinuity to simply be whether the two diameters end up with the same display string.  This will cause some surprises if a user goes back and forth between two units, but that seems less problematic than the current situation where a discontinuity error can be set when the two diameters are displayed as being equal.

Finally, it turned out that the code looking for the preceding body tube section was broken at axial stage boundaries, so it would never set the warning in that case. This has been fixed.

Fixes #999


hcraigmiller:
The rounding error (that was invisible) is gone, but two other errors were found.

First, open OR and add a default body tube, then add a default nose cone.

Exception

Second, the discontinuity check is crossing over between pods.

Discontinuity

[Discontinuity Error.ork.txt](https://github.com/openrocket/openrocket/files/10317285/Discontinuity.Error.ork.txt)

JoePfeiffer:
Thanks.  Hmm, that second one looks more like this turned up a bug in the pod handling....

JoePfeiffer:
OK, let's try this again....

hcraigmiller:
Nice. . . Functions as expected with no anomalies found.

Build 1374

broofa:
Hey, 'just wanted to drop you a note here saying, _thank you_!

I've been running the `beta 5` release and ran into the mysterious "discontinuity in body diameter" issue with no explanation, and eventually tracked it down to the nose cone diameter ("4.16cm") not being the same as the body diameter (also "4.16cm").  Landed here after googling around for issues and discovered that this had been fixed (here) in the latest release.

As a long-time user of OpenRocket, I'd grown accustomed to looong delays between releases.  I can't tell you how refreshing it is to see bugs getting fixed and released.

As a fellow opensource-er I know how thankless this can be sometimes.  So, again, thank you.  You guys rock!

SiboVG:
> Hey, 'just wanted to drop you a note here saying, _thank you_!
> 
> I've been running the `beta 5` release and ran into the mysterious "discontinuity in body diameter" issue with no explanation, and eventually tracked it down to the nose cone diameter ("4.16cm") not being the same as the body diameter (also "4.16cm"). Landed here after googling around for issues and discovered that this had been fixed (here) in the latest release.
> 
> As a long-time user of OpenRocket, I'd grown accustomed to looong delays between releases. I can't tell you how refreshing it is to see bugs getting fixed and released.
> 
> As a fellow opensource-er I know how thankless this can be sometimes. So, again, thank you. You guys rock!

You're spot on with your last paragraph, so thanks for your kind words, appreciate it :).

Issue #1937, opened 2022-12-31
Make example rockets' design file title the same as its file name


SiboVG:
It sometimes confuses me that the names listed in the rocket example menu do not have that name when you open the design file. For example if you open the example file "Parallel Staging Example", the actual design file is named "Booster Example" (because that's the name of the rocket in the design file).

<img width="1564" alt="image" src="https://user-images.githubusercontent.com/11031519/210121898-dbbe4b4f-f4bc-4d05-9287-3bce07231298.png">

It would make me happy if all the rockets in the example files were named after their file name (the name that shows up in the examples menu).

SiboVG:
Fixed by PR #1940.

PR #1939, opened 2022-12-31
Rename and re-order examples

hcraigmiller:
Rename and re-order examples (see Issue #1937).

Because of the way that OpenRocket processes the example file names to the menu order, the names of some of the files had to be changed in order to achieve the desired order in the menu.

Menu Order Rev_02

For each example, the file name, menu name, and design name are now the same.


neilweinstock:
Artful naming. :)

PR #1940, opened 2022-12-31
Rename and re-order example designs

hcraigmiller:
Rename and re-order examples (see Issue https://github.com/openrocket/openrocket/issues/1937).

Because of the way that OpenRocket processes the example file names to the menu order, the names of some of the files had to be changed in order to achieve the desired order in the menu.

Menu Order Rev_02

For each example, the file name, menu name, and design name are now the same.


SiboVG:
@hcraigmiller I actually liked your initial naming conventions, so I just pushed a commit that allows you to set a custom menu item ordering, by editing `exampleFileOrder` inside `swing/src/net/sf/openrocket/gui/main/ExampleDesignFileAction.java`. I also added the different horizontal separators.

image

Could you change the names back to your initial suggestions?

Small note: The "Simulation extensions .ork" has an extra space at the end. That should be fixed (also don't forget to edit that file's String entry in `exampleFileOrder`!).

hcraigmiller:
Nicely done @SiboVG.

PR #1944, opened 2023-01-04
New app icons

neilweinstock:
Here's the fully-implemented new app icon, with all comments (I think) incoporated.


SiboVG:
Some of the new icons look very blurry compared to the previous ones.
image

neilweinstock:
Hmm, it is a bit soft, although it looks better in real life than in the slightly enlarged version shown above.
image

neilweinstock:
I'm not sure why that happened.... will fix.

neilweinstock:
OK, check again.

SiboVG:
Yup, it's sharp now, thanks 👍

Some notes though:
1. I'm not too fond of the small black outline on the rocket in the splash screen
Screenshot 2023-01-04 at 22 57 49

3. I would add back a bit of the outside blue haze on the Earth. The Earth almost looks almost too sharp now and takes away some of the attention from the rocket (which should have the main focus)
E.g. if you compare the about dialogs:
image

You can also see this in the welcome dialog, where everything looks too sharp actually (some aliasing issue?).
image

4. I would make the Earth slightly smaller; I liked your original mock-up the best, where part of the right pod is outside the Earth:
image

neilweinstock:
I'll fix most of what you pointed to.  The about box, though, appears to be bad scaling inside the Java app... my supplied image does not look all jaggy like that.

neilweinstock:
All requested changes have been made.

SiboVG:
Thanks! Didn't do a proper test yet, but is it possible that you didn't update the about icon? It still has its right pod inside the Earth.

neilweinstock:
There was an old version in the source directory; wasn't supposed to be there at all.  The correct one is up to date (in resources).

SiboVG:
Sorry @neilweinstock, I wasn't very clear. What I meant was that the outline haze of the Earth should be more defined, not that the blur over the Earth should be more. The new Earth should have the same sharpness and outline as the old icon. What I mean by outline is this:
image

SiboVG:
Also: I think the aliasing issue is because you used "GIMP built-in sRGB" color profile instead of "sRGB IEC61966-2.1". I already tried fiddling with the code to render it with anti-aliasing, but that didn't fix it, so could you try exporting with the "sRGB IEC61966-2.1" color profile?

neilweinstock:
Like this?  The haze is not as extensive as before because it was Too Much.
image

SiboVG:
Yeah that looks better!

SiboVG:
I do hope you're not pulling the old "I'm just gonna send the same image twice and hope the customer magically thinks this is better" graphics designer trick on me.

neilweinstock:
LOL.

The difference is subtle, to be sure.  I'll check in just the new source (XCF file), won't try doing any of the exporting until you take a whack at the color profile stuff.

SiboVG:
Something minor: the 256 icon has a small artefact on top:
image
image

Sounds small, but it is noticeable in the program.

neilweinstock:
Argh, will fix

SiboVG:
> Argh, will fix

Graphic design is fun :)

hcraigmiller:
As to function only, functions as expected; no anomalies found.

Overall, the icons look good to me, and the new taskbar icon is even clearly distinguishable from the old one.

Build 1396

neilweinstock:
Anyone else want to review this before we merge?

JoePfeiffer:
Looks excellent to me.

PR #1945, opened 2023-01-04
Inhibit reignition of already ignited motor

### LINKED ISSUE #1938: [Bug] Plot of flight events is not correct when powered pods have recovery devices
Flight Plot Icons No_03

See also comments of @SiboVG on PR #1935.

JoePfeiffer:
The problem in #1938 turned out not to be the recovery devices in the pods, but rather that the sustainer motor was ignited again when each of the motors in the first stage burned out.

If a later stage motor's ignition event is triggered by burnout of earlier stage motors, an ignition event gets queued for every burnout in the earlier stage. This checks to see if a motor is already burning when an ignition event is processed for it, and ignores the event if so.

Fixes #1938


hcraigmiller:
Functions as expected; no anomalies found.

However, I discovered that the powered pod with recovery example has stored simulations that contain and display the duplication. After this PR is merged, I will re-run all of the simulations in that example and upload a new example file.

Build 1376

PR #1953, opened 2023-01-06
[#1942] Fix text component selection key behavior

### LINKED ISSUE #1942: [Improvement] Hitting left-arrow key when all entry text is selected should place cursor to the front
If all the text of a text entry field is selected, it would make sense that when you hit the left-arrow key that the text cursor jumps to the front of the text, and the right-arrow key to the back of the text. However, now the cursor is by default placed at the end of the text.

So now it looks like this when I hit the left-arrow key:

https://user-images.githubusercontent.com/11031519/210173514-397b28e1-e042-46a7-b2a1-c1edded3a374.mov

The text cursor should be at the front.

SiboVG:
This PR fixes #1942:

https://user-images.githubusercontent.com/11031519/211022585-567b42e9-32c4-4134-bdd3-8f1fb1292f15.mp4

Note: this fix is now only applied to spinner editors (practically all number-input fields in OR), and the component name and comment text fields. In the future, I may post a general fix for every text field etc., which would just require writing a new JTextField class and applying the fixed key listener to it. But for now, this PR should be good enough.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/SpinnerEditor.java
+++ b/swing/src/net/sf/openrocket/gui/SpinnerEditor.java
@@ -1,5 +1,7 @@
 package net.sf.openrocket.gui;
 
+import net.sf.openrocket.gui.adaptors.TextComponentSelectionKeyListener;
+
 import javax.swing.JSpinner;
 import javax.swing.JTextField;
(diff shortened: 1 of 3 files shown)


hcraigmiller:

Build 1402

Issue #1923, opened 2022-12-24
[Bug] Simulation extension error on load of examples


hcraigmiller:
The following examples are throwing simulation extension errors when opening the example file:

- Simulation Extensions
- Simulation extensions and scripting

![Simulation Extension](https://user-images.githubusercontent.com/68821492/209439792-f54821bb-a528-4aac-ba03-a5906e04e3c1.png)
The above is the Simulation Extensions example.

SiboVG:
This is the result of PR #1873.

I would also suggest to merge the "Simulation extension" and "Simulation extension and scripting" example, since they're basically the same.

JoePfeiffer:
Rather than combine them, maybe we should separate them?  Having the Java extension do the roll control and air start in the extension example, and having the JavaScript script do both in the scripting example (currently a script is used for roll control, and an extension for air start)?

JoePfeiffer:
@hcraigmiller could you check to see if this is still happening for you?  It isn't for me.  I do still want to better separate the examples, as I suggested above, before we close this.

hcraigmiller:
Using the current unstable, both appear fully functional, not only when loaded, but when all of the simulations are run as well.

PR #1954, opened 2023-01-06
Separate extension and scripting examples.

JoePfeiffer:
The "extension" example now uses the extension mechanism to add both listeners
The "scripting" example now uses JavaScript scripts to perform the same tasks as in the extension example

Finishes fixing #1923


hcraigmiller:
I guess this is an area that I'm going to have to educate myself about. So, rather than commenting on whether the behavior is correct, I'll stick to describing the behavior.

First, when opening the example menu, there are now only two options, instead of three.

Menu

I re-ran the simulation for the _Presets_ example from the Flight simulations tab, and it completed without error. To me it appeared to be functioning properly.

I re-ran each of the simulations for the _Simulations extensions_ example from the Flight simulations tab, individually, and each completed without error. To me, each, individually, appeared to be functioning properly. When re-ran 1 and 2 simultaneously, and 1 and 3 simultaneously, each completed without error. But, if 2 and 3 are re-run simultaneously, an error appears.

Error

An error appears if all three simulations are run simultaneously too.

If there is any other testing procedure that should be being used for this PR, please let me know.

JoePfeiffer:
Hmmm, this is unexpected.  My local build is showing both "Simulation Extensions" and "Simulation scripting", so something seems to have gone wrong with my commit.

Errors when running two sims at once is really weird.  I suspect some sort of crosstalk between them...  which will be unpleasant to find.  Darn.

Incidentally, thanks for doing these PR checks.  There is no more thankless role than QC, and I for one really appreciate you doing it.  So at least one person here thanks you!

SiboVG:
> Incidentally, thanks for doing these PR checks. There is no more thankless role than QC, and I for one really appreciate you doing it. So at least one person here thanks you!

Make that two!

hcraigmiller:
I've been looking more closely at these simulations, and I would recommend resizing the parachutes. Changing the drogue to 55 cm will reduce the descent to just under 100 fps and changing the main to either the FRS-16-96 or the IFC-72-N will reduce the descent rate to below 18 fps (either should open safely up to 120 fps), or, if space is a premium, the IFC-72-S. To me, 32+ fps is too fast on the ground hit.

JoePfeiffer:
> I've been looking more closely at these simulations, and I would recommend resizing the parachutes. Changing the drogue to 55 cm will reduce the descent rate to just under 100 fps and changing the main to either the FRS-16-96 or the IFC-72-N will reduce the descent rate to below 18 fps (either should open safely up to 120 fps), or, if space or mass is at a premium, the IFC-72-S. To me, 32+ fps is too fast on the ground hit.

Good idea (I was just looking at the extensions and scripting, not whether the rest of the example made sense!).  I enlarged the drogue to get a main deploy speed under 70fps (24in drogue) and the main to get a landing speed just a little over 20fps (60in main).

JoePfeiffer:
This has proved to be much knottier than I expected -- it's enough of an edge case that I'm happy for it to simply be a known issue when RC1 goes out.

JoePfeiffer:
Long time to make a small change in the end.

15.03 made a copy of the Rocket when executing a simulation; at some point early in the changes leading to 22.02 that was dropped.  Trying to have both simulation branches changing settings in a single rocket caused a conflict and crash.

This now also makes a copy of the Rocket, as 15.03 used to do.

hcraigmiller:
Functions as expected; no anomalies found.

Will _Simulation scripting_ be the new name going forward, instead of _Simulation extensions and scripting_?

Build 1453

JoePfeiffer:
> Looks good to go then. I'm going to have to spend some time learning how to use the extensions and scripting features.

Coincidentally, I'm documenting extensions now on the wiki while it's still fresh in my mind.  Do you know if one user can see a draft on the wiki in another user's namespace?  If so, here's the current status:
https://wiki.openrocket.info/User:JoePfeiffer/Extensions_and_Listeners

SiboVG:
> Coincidentally, I'm documenting extensions now on the wiki while it's still fresh in my mind. Do you know if one user can see a draft on the wiki in another user's namespace? If so, here's the current status:
> https://wiki.openrocket.info/User:JoePfeiffer/Extensions_and_Listeners

Yes we can. Small tip: if you ever want to know if someone else can see a webpage you can see, you can open an incognito window on Google Chrome and open the page there. Incognito tabs don't use your stored cookies/caches. Of course that is assuming you use Chrome.

PR #1955, opened 2023-01-07
Update parachutes of 4 example rockets to reduce descent fps

hcraigmiller:
Adjusted example parachutes to reduce descent fps on mains to about or less than 18 fps and on drogues to about or less than 90 fps. The affected examples are: Two-stage rocket, TARC payload rocket, Airstart rocket, and Dual parachute deployment rocket.

These examples had unsafe descent rates.


hcraigmiller:
@neilweinstock, I believe that your _Pods--airframes and winglets_ example is fixed. check it out and let me know.

SiboVG:
> @neilweinstock, I believe that your _Pods--airframes and winglets_ example is fixed. check it out and let me know.

Did you ever find out why the value changed? Neil correctly expressed the concern that this may be a bug.

hcraigmiller:
I was unable to replicate the issue. I tried everything I could think of to try to break it, but it worked fine.

neilweinstock:
I guess we’ll just have to leave it for now and keep an eye on it in case it comes back.

hcraigmiller:
I agree, merge the correction and hope it was something someone did inadvertently, as opposed to a bug.

PR #1971, opened 2023-01-15
Update config dialogs

SiboVG:
What started out as a small fix for the inner tube cluster rotation spinner/slider (the slider would jump to the start position when you changed the spinner value from 0° to 1°) quickly turned into something bigger.

This PR does a full update of all the component configuration dialogs to make them:

- more **homogenous**: there were **a lot** of differences between the layout of different
  - this PR tries to have ± the same input field ordering. For example attribute settings are located in the panel's left column, and material settings and placement settings in the right column. As an illustration, here is an overview of how the material and placement settings were distributed amongst the different config dialogs:
  
image

You can see it's all over the place...
  - all the tabs should have their widgets starting in the top left corner. Sometimes there were tabs that had centered widgets, sometimes in the top left corner etc. This should be fixed
  - the fin set config dialogs where the only one with a vertical separator. I experimented both with and without a separator in all the dialogs, but eventually settled on no separator, but with just a bit of blank space between the left and right tab column.
- Better **separation** between different-typed input fields
    - Placement input fields (e.g. "Position relative to:", "Fin rotation") are now grouped in a Placement border. Material settings ('Component material', 'Component finish') are grouped with a material border, unless there is only the 'Component material' setting. In that case, the border is emitted because it made the UI too crowded.

Here's a demo of the new and old config dialogs (it's a long video):

https://user-images.githubusercontent.com/11031519/212562342-6bd8c533-65d2-4f3a-a7a7-baa3e749ec41.mp4


neilweinstock:
Generally very good.  Only anomaly I noticed is that for parachutes and streamers, the Placement box is upper right instead of lower right.  Was there a reason for this?

SiboVG:
> Generally very good. Only anomaly I noticed is that for parachutes and streamers, the Placement box is upper right instead of lower right. Was there a reason for this?

I think it makes more sense for the deployment box to be the last box (lower right). I tried to think chronologically: first you set general attributes, then material, then placement, then deployment, or something else specific. But I don't have a strong opinion on this.

neilweinstock:
I don't feel very strongly about it, but I liked seeing Placement in the same location in all different component types.  I don't actually think lower right is necessarily the best place for it, but I don't have a better proposal for now.

hcraigmiller:
Two possible inconsistencies:

1. The nose cone shoulder data is not in a box, while the transition shoulder data is.
2. Inner component configuration (all), Component material is not in a Material box while all other components with a material choice are.

My preferences on placement would be to have the same information in the same place, if possible, from the top of the rocket down. For example, all components (except mass components) have a material, so material should have a high placement priority. And, all components (except nose cones, transitions, and body tubes) have a placement, so placement should have a high placement priority.

Applying this to the right-side elements, to the extent that a component has the elements, I would suggest the following order:
1. Materials,
2. Placement,
3. Root fillets,
4. Deployment.

With all of these elements in boxes.

neilweinstock:
I think Placement should go above Materials... basically, you must *set* the placement on every component, whereas in many cases you can leave the material untouched, either because it's already set properly from a preset or because you're just roughing out a design and don't know or care about materials yet.

IMHO Placement is the single most important set of parameters that are common to all internal components.

hcraigmiller:
Makes sense.

SiboVG:
> 1. The nose cone shoulder data is not in a box, while the transition shoulder data is.

That was intentional, the box is not needed for the nose cone, since there is only one shoulder. You just have duplicate information otherwise (a "Shoulder" tab and then again a "Nose Cone Shoulder" box).

SiboVG:
Okay, I'll change the ordering:
1. Placement
2. Materials
3. Root fillets / deployment

SiboVG:
Samples of new layouts:

image
image
image
image
image
image

SiboVG:
Okay I changed my mind, I'm gonna use a material border even for single-combo material boxes. It looks "naked" without a border.

SiboVG:
Samples of the new layouts.

image
image
image
image
image
image

SiboVG:
Oh and thanks for the feedback. I've been staring at so many different layout configurations that I've almost lost my objectivity (I think @neilweinstock can relate to this with the logo design).

hcraigmiller:
Very nice. Functions as expected; no anomalies found.

Build 1450

SiboVG:
@JoePfeiffer do you have any remarks on the new look?

JoePfeiffer:
> @JoePfeiffer do you have any remarks on the new look?

Not really.  It looks really nice, and you guys have been nailing the details down really well without inserting another cook in the kitchen.

neilweinstock:
I like it, definitely an improvement.  More consistency in this area is a good thing, the lack thereof has been an irritation to me for a long time.

With Craig's previous set of changes and now this, we're in a ***much*** better place than we were in 15.03.

PR #1974, opened 2023-01-15
Issue - 1970 - [Feature request] Change Inner Tube cluster tube separation from unitless to distance unit

### LINKED ISSUE #1970: [Feature request] Change Inner Tube cluster tube separation from unitless to distance unit
In the Cluster tab of an inner tube, the tube separation is set as a unitless value, where 1 is tubes touching, 2 is 2 tubes distance I guess etc. Wouldn't it make more sense to have this be an actual distance instead of the ambiguous mess that it is now.

jura0207:
This fixes #1970 - [Feature request] Change Inner Tube cluster tube separation from unitless to distance unit

The problem was that the distance between multiple clusters inside the tube were measured by scaling the outer diameter of the tubes in the cluster. The more precise solution is to make the distance a length unit.

This was implemented by changing the unit from scaling to length. Frontend was also adjusted to implement this with adding a measurement unit and adjusted slider. Saving and loading .ork now have both clusterScale and clusterSeparation variables so that the older supported versions would also work. Language tooltip was removed since there is no more scaling in the menu.

[jar file](https://drive.google.com/file/d/1J5hyOR1HzLuSce3_aVcMddxDtp1zQBjG/view?usp=sharing)


core/src/net/sf/openrocket/rocketcomponent/InnerTube.java
```diff
@@ -177,35 +178,51 @@ public void setInstanceCount( final int newCount ){
...
 	public boolean isAfter(){ 
 		return false;
 	}
-	
+
 	/**
-	 * Set the cluster scaling.
-	 * @see #getClusterScale()
+	 * Return the distance between the closest two cluster inner tube center points.
+	 * This is equivalent to the cluster scale multiplied by the tube diameter.
```
SiboVG, line 199: I would remove this line, since we're gonna remove all references to the cluster scale after we remove backward compatibility with OR 15.03.

core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
```diff
@@ -359,6 +359,8 @@ class DocumentConfig {
 		setters.put("InnerTube:clusterconfiguration", new ClusterConfigurationSetter());
 		setters.put("InnerTube:clusterscale", new DoubleSetter(
```
SiboVG, line 360: Can you add a comment above this line: `// TODO: delete clusterscale when compatibility with OpenRocket 15.03 is not needed anymore`

core/src/net/sf/openrocket/file/openrocket/savers/InnerTubeSaver.java
```diff
@@ -28,6 +28,7 @@ protected void addParams(net.sf.openrocket.rocketcomponent.RocketComponent c, Li
 
 		elements.add("<clusterconfiguration>" + tube.getClusterConfiguration().getXMLName()
 				+ "</clusterconfiguration>");
+		elements.add("<clusterseparation>" + tube.getClusterSeparation() + "</clusterseparation>");
 		elements.add("<clusterscale>" + tube.getClusterScale() + "</clusterscale>");
```
SiboVG, line 32: Can you add a comment above this line: `// TODO: delete clusterscale when compatibility with OpenRocket 15.03 is not needed anymore`

SiboVG (changes requested):

SiboVG:
Thank you for your changes @jura0207! Looks very good overall, just a couple of small notes (see my review).

However, I may have been too quick with posting issue #1970. I know see some logic as to why cluster scale was used before, instead of just cluster separation (in distance). If you change the diameter of the inner tube, and want the tubes to be touching, then always you have to update the cluster separation value...

https://user-images.githubusercontent.com/11031519/212574077-e00e3e97-ce16-44fa-ab6c-fe45bb080ba4.mov

Not entirely sure on what the best behavior would be (maybe add a radio button to switch between cluster scale/cluster separation?), I'll ping @hcraigmiller, @neilweinstock and @JoePfeiffer.

SiboVG:
And by the way, RockSim functionality functions as expected.

core/src/net/sf/openrocket/file/openrocket/savers/InnerTubeSaver.java
jura0207: Added

core/src/net/sf/openrocket/rocketcomponent/InnerTube.java
jura0207: Removed

jura0207:
Changes were implemented.

> Thank you for your changes @jura0207! Looks very good overall, just a couple of small notes (see my review).
> 
> However, I may have been too quick with posting issue #1970. I know see some logic as to why cluster scale was used before, instead of just cluster separation (in distance). If you change the diameter of the inner tube, and want the tubes to be touching, then always you have to update the cluster separation value...
> 
>  Screen.Recording.2023-01-16.at.00.45.00.mov 
> Not entirely sure on what the best behavior would be (maybe add a radio button to switch between cluster scale/cluster separation?), I'll ping @hcraigmiller, @neilweinstock and @JoePfeiffer.

You're welcome! I'm glad I could help a little. :D

Would adding something like:
`clusterSeparation = clusterSeparation * (r / outerRadius);` where r is new radius
in the ThicknessRingComponent.java in setOuterRadius fix this?
Or maybe better add a function that does that in case the cluster exists.
I'm not sure this is the right approach since this function is used in a lot of other places and both the scope and risk would be too big.

JoePfeiffer:
I'm thinking of a number of use cases for which either the old or the new could make sense.  I think we do want that checkbox.

I do wonder what the semantics of "separation" should be -- to me, that should be the distance between motor tubes, not the distance between their centers (so a separation of 0 would have touching motor mounts).  That would also eliminate the issue of needing to change separation with changes in motor tube diameter.  Would that be reasonable?

SiboVG:
> I do wonder what the semantics of "separation" should be -- to me, that should be the distance between motor tubes, not the distance between their centers (so a separation of 0 would have touching motor mounts). That would also eliminate the issue of needing to change separation with changes in motor tube diameter. Would that be reasonable?

Sounds like a good idea. I would suggest adding radio buttons to switch between absolute and relative separation. For relative separation, I would keep the "1 == tubes touching", since that's the standard that OR uses right now. For absolute separation, 0 cm == tubes touching.
Screenshot 2023-01-18 at 17 24 1@jura0207

@jura0207 do you feel like implementing this? If this PR became more than you bargained for, that's also fine, then we can pitch in and finish it.

SiboVG:
Superseded by #2269.

PR #1977, opened 2023-01-15
Add JAR files to proper work of project load

janforys:
**W**hen added those files - all errors has gone


SiboVG:
Thanks for the changes!

PR #1978, opened 2023-01-16
[1972] Update example rockets to include simulation data

hcraigmiller:
All designs checked/updated simulations.


hcraigmiller:
The only simulation that does not have all green bubbles is the Pods--powered with recovery deployment because the last two of the four are the sustainer only and the booster only. Is there a preference for including or removing these last two simulations?

SiboVG:
> The only simulation that does not have all green bubbles is the _Pods--powered with recovery deployment_ because the last two of the four simulations are the sustainer only and the booster only simulations. Is there a preference for including or removing these last two simulations?

Why exactly can't you save the simulation data of the second-to-last one? I don't have an issue doing that on my machine.

hcraigmiller:
> Why exactly can't you save the simulation data of the second-to-last one? I don't have an issue doing that on my machine.

I can run and save all four simulations with green bubbles. But, my understanding of what to expect may be lacking.

This is what I am observing. Simulation 1 and 2 were created with both the sustainer and booster stages active. Simulation 3 was created with only the sustainer stage active, and Simulation 4 was created with only the booster stage active.

If I open the example and go to the Flight simulations tab, all four simulations have green bubbles. If I plot the simulations, Simulation 1 and 2 correctly plot both stages (one with and one without pod recovery). If I plot Simulation 3, the sustainer only, it also correctly plots. But, if I plot Simulation 4, instead of plotting just the booster, it incorrectly plots both the sustainer and booster stages (not just the booster). Moreover, on the simulation results, the booster is shown without a deployment event, and the plot shows the sustainer descending without a parachute (presumably because it has no motor/ejection charge to deploy the parachute).

Booster Only Not Recognized

It appears that Simulation 3 is correctly recognizing that it was created without the booster, but Simulation 4 is not recognizing that it was created without the sustainer.

In other words, it appears that I am observing a new issue, not an issue related to saving the simulation data. If you agree, I will create a new issue.

hcraigmiller:
Hold up. I think I found figured out why.

hcraigmiller:
Okay, it was an incorrect setting that was overlooked. All fixed, will update.

PR #1979, opened 2023-01-16
Warn for phantom tubes earlier on

### LINKED ISSUE #1894: [Bug] Warning that inline pods "may not simulate accurately"
An increasing number of users are positioning pods so that a component (I'll use a transition) can be placed inline as if the transition is the next airframe component after the body tube. @JoePfeiffer confirmed that from an aerodynamic standpoint, this technique may not be (is not) being simulated accurately.

Transition Pod_Transition

With the increasing number of users implementing this technique, if there is a way to **_fairly easily_** determine whether a pod component is positioned in line with the airframe to which it is attached (_only warn if **the pod position is relative to center of the parent component AND the radial distance of zero**_), a warning should be given that "Pods with inline components may not simulate accurately."**_ In any event, a description of this issue and warning should be included in the Wiki._**

I am not suggesting aerodynamic correction. If I understood @JoePfeiffer, he does not believe that aerodynamic correction is feasible, even if possible, because of too many nuances in design variations and should not be undertaken.

[Test.Boosted Dart Example.Transition.zip](https://github.com/openrocket/openrocket/files/10243488/Test.Boosted.Dart.Example.Transition.zip)

SiboVG:
This PR fixes #1894 and adds a warning when a pod set is in line with the rocket's center line. Question: should the same warning also be generated for boosters?

I also noticed that the zero-volume warning for phantom tubes was only added when fins were added to the phantom tube. I thought it would be better to generalize this error and check for zero-volume tube warnings even when there are no fins attached to the component.

I also added the phantom tube component name to the zero-volume warning.

https://user-images.githubusercontent.com/11031519/212781178-7e7e6363-2141-434e-87f5-5b53f1eba50a.mp4


JoePfeiffer:
It turns out that the inline pod warning is only needed if the pod is both in line, and there is a gap between it and the previous component, and the better thing to warn about is a gap in the external airframe, not the inline pods (something I realized when fixing the previous/next component code).

SiboVG:
> It turns out that the inline pod warning is only needed if the pod is both in line, and there is a gap between it and the previous component, and the better thing to warn about is a gap in the external airframe, not the inline pods (something I realized when fixing the previous/next component code).

Ah, right, I'll adjust it.

SiboVG:
I could also apply the gap warning to designs with a middle stage disabled, e.g. here:
image

JoePfeiffer:
Yes, please do.

neilweinstock:
Wait, wasn't there going to be an inline pod warning when the pod is overlapping the airframe?

SiboVG:
> Wait, wasn't there going to be an inline pod warning when the pod is overlapping the airframe?

I never saw that mentioned, did I misread #1894?

JoePfeiffer:
> Wait, wasn't there going to be an inline pod warning when the pod is overlapping the airframe?

I remember that being mentioned along the way; I don't remember us deciding to do it. Though replacing the in-line check with an overlapping check doesn't sound hard.

neilweinstock:
> > Wait, wasn't there going to be an inline pod warning when the pod is overlapping the airframe?
> 
> I never saw that mentioned, did I misread #1894?

It probably was not mentioned explicitly... but I considered it to be implied.  Any overlap between the pod and the airframe will screw up simulation.  Of course that also applies to pods that are *not* inline, so.... ?

The idea that there's no warning when there's no gap is also a bit suspect to me.  If I put no nose cone on the pod, and set the gap to be zero, I'm pretty sure that OR will not sim it as a single contiguous airframe.  You'll get a warning about open-ended pod, but you could fix that by putting a zero-length nose cone at the front.

All in all, I am concerned that we're putting too much untested heuristics into this warning, too close to release.

JoePfeiffer:
> The idea that there's no warning when there's no gap is also a bit suspect to me. If I put no nose cone on the pod, and set the gap to be zero, I'm pretty sure that OR will not sim it as a single contiguous airframe. You'll get a warning about open-ended pod, but you could fix that by putting a zero-length nose cone at the front.

The previous/next component code is smart enough that they actually should simulate the same (uh, oh -- I hadn't noticed until now that an inline pod with no gap actually does  put up the open airframe warning and, while the CP and CG calculations are the same, the drag is not.  Hmmmm)

SiboVG:
I agree with @hcraigmiller that warning about overlaps is a bit too much for now. We can think about it when we want to improve the warning set system, but I wouldn't include it in 22.02.

> The previous/next component code is smart enough that they actually should simulate the same (uh, oh -- I hadn't noticed until now that an inline pod with no gap actually does put up the open airframe warning and, while the CP and CG calculations are the same, the drag is not. Hmmmm)

So for zero-gap, inline pods there needs to be an additional warning, i.e. the initial warning from this PR? Then once there is a gap, the inline pods warning should be replaced with the gap warning?

Finally, the open airframe warning should be adjusted to detect inline pods?

JoePfeiffer:
> So for zero-gap, inline pods there needs to be an additional warning, i.e. the initial warning from this PR? Then once there is a gap, the inline pods warning should be replaced with the gap warning?
> 
> Finally, the open airframe warning should be adjusted to detect inline pods?

There shouldn't be any warning if there is no gap, and drag should be calculated correctly in that case (basically, since it can get it right for axial stages it should be able to get it right for pods.  I thought that it would when I fixed the previous/next component code, but clearly I didn't do enough testing).

SiboVG:
> There shouldn't be any warning if there is no gap, and drag should be calculated correctly in that case (basically, since it can get it right for axial stages it should be able to get it right for pods. I thought that it would when I fixed the previous/next component code, but clearly I didn't do enough testing).

Are you able to fix this for the 22.02 release? If not, we should add the warning.

JoePfeiffer:
Ah, I apologize -- I didn't see that you'd updated this one to account for gaps.  Not sure which one we should merge now...

SiboVG:
> Ah, I apologize -- I didn't see that you'd updated this one to account for gaps. Not sure which one we should merge now...

Your PR looks more elaborate, but there's still some issues to fix in yours. I'll adjust this PR to remove the gap warnings, but I also did some refactoring of zero-volume bodies, so I'll include that.

JoePfeiffer:
Running yours, I'm getting an "open front airframe" warning from my inline pods test.  Is that intended?

SiboVG:
> Running yours, I'm getting an "open front airframe" warning from my inline pods test. Is that intended?

Well not intended, I just didn't think of fixing it.

SiboVG:
Changed the PR to only refactor the phantom tube warning.

PR #1983, opened 2023-01-17
[#1980] Render booster markers even if no children

SiboVG:
This PR fixes #1980.

PR #1984, opened 2023-01-18
[#1981] Update pod/booster marker position based on axial method & Correct for pod/booster marker radius offset

### LINKED ISSUE #1981: [Bug] Pod/booster markers do not account for position method
Pod/booster markers are always plotted on the fore side of the pod/booster. However, if the "Position relative to" method is set to "Bottom of the parent component", the marker should move to the aft side of the set since that now became the reference point of the pod/booster.

image

I also have an extra possible bug: in the previous image, the marker is plotted on the tip of the nose cone. However, the pod's radial distance is set the 0 relative to the surface of the body tube. So the pod's **actual reference point** is: 
image

Should the marker be moved there instead?

SiboVG:
This PR fixes #1981. The pod/booster marker is now at the reference point of the pod/booster, which is a combination of the axial and radial offset.

https://user-images.githubusercontent.com/11031519/213063905-3f5d7c17-8f39-4d27-ae23-8508e3918e2b.mov

Note: this PR is a sub-branch of PR #1983, so best to merge that first.


neilweinstock:
I like it.

PR #1985, opened 2023-01-18
Check for gaps and overlaps in airframe

### LINKED ISSUE #1894: [Bug] Warning that inline pods "may not simulate accurately"

JoePfeiffer:
Since users are using pods with a radial offset of 0 to create sci fi rockets with exposed inner tubes and stuff, warn them that they aren't likely to simulate accurately.

Also, if a user sets up a rocket using a pod that doesn't have a gap in the airframe, we are able to simulate that so we should get rid of the warning in that case.

First screenshot:  a rocket with a single long body tube.
long-tube

Second screenshot: rocket with the same geometry but two body tubes, using a PodSet to position the second one. Note CP and performance summary are roughly the same (yeah, it needs a bigger motor)
inline-pod

ork files:
[long tube.ork.txt](https://github.com/openrocket/openrocket/files/10449130/long.tube.ork.txt)
[inline pod.ork.txt](https://github.com/openrocket/openrocket/files/10449135/inline.pod.ork.txt)

Fixes #1894


SiboVG:
It doesn't appear to work when there is a component before or after the parent. I fixed this in my PR by searching for the most aft and most fore symmetric component.

image
image

Also: why is there a discontinuity error?

SiboVG:
The overlap error only shows the nose cone:
image

Maybe just replace "Nose Cone" with "Pod Set"? If the pod has multiple components that overlap, the warning message may become long.

And this should be an overlap warning, not a gap warning: (but I think I explained that in my previous post)
image

SiboVG:
Contrary to what it looks like, I'm not deliberately tracking down mistakes because you murdered my PR @JoePfeiffer, it's just issues that I encountered myself when coding my PR.

SiboVG:
And there should be a gap warning for disabled stages (see my PR).
image

JoePfeiffer:
OK, let's give this another try...  I'm attaching another pair of example rockets.

The inline pod rocket has the same stability and simulation results as the non-pod rocket, with no warnings.

If the pod is moved aft, the gap warning is triggered, specifying the upper body tube and middle body tube.

If it is moved forward so the first tube in the pod overlaps the parent body tube, the overlap warning is triggered, specifying the upper body tube and middle body tube.

If it is moved farther forward so both the middle and lower body tube overlap the upper body tube, it triggers an overlap warning saying only that the in-line podset overlaps the parent airframe component.

If it is moved far enough forward that the pod set is entirely forward of the airframe, a warning is triggered specifying that.
[nopod.ork.txt](https://github.com/openrocket/openrocket/files/10470062/nopod.ork.txt)
[inline pod.ork.txt](https://github.com/openrocket/openrocket/files/10470063/inline.pod.ork.txt)

hcraigmiller:
Unable to create a booster or pod.

1. Start OpenRocket.
2. Add default nose cone.
3. Add default body tube.
4. Attempt to add either booster or pod.

2023-01-20 n_01

hcraigmiller:
Two observations:

1. If you have an inline pod body tube (without a gap warning) and change the reference point from center to surface, it triggers the forward open warning even though the body tube has not moved. _I actually believe that this is appropriate_; so this is an FYI.

2023-01-20 n_03 Ref_Surface

2. When the inline pod body tube is gapped (and the pod body tube outer diameter is set to automatic), the pod body tube outer diameter defaults until its configuration window is opened.

2023-01-20 n_04 Ref_Surface
 
Overall, everything else seems to be working.

JoePfeiffer:
On the array out of bounds exceptions -- how did *that* happen?  I'm not sure whether I'm more troubled by the fact that I somehow missed that, by the fact that I had to add a pod to create the example rocket and somehow didn't set it off (and the bug in the code was definitely a "this will definitely, reliably fail"), or the fact that the unit tests didn't catch it. Should be fixed now.

On the open airframe when the reference is changed -- that's intended behavior.  We're only trying to catch a user making a mistake, not one working at misleading the program.  Now that I think about it there's probably a more robust colinearity check that wouldn't be hard to implement....  but not right now.

On the diameter going to default when we change the radial distance -- what exactly was the sequence of steps that created that?  I can see that you started with the example .ork, but you've added a stage and body tube. I *think* that's correct behavior:  if the pod isn't exactly lined up it isn't collinear. so the automatic diameter doesn't work.

hcraigmiller:
> On the diameter going to default when we change the radial distance -- what exactly was the sequence of steps that created that? I can see that you started with the example .ork, but you've added a stage and body tube. I _think_ that's correct behavior: if the pod isn't exactly lined up it isn't collinear. so the automatic diameter doesn't work.
>

Using your inline pod example, open the pod and change the radial distance to something other than zero. The behavior that I would expect is that it would grab the reference diameter, remove the automatic diameter, and use set the outer diameter to the grabbed reference diameter.

hcraigmiller:
Everything appears to be working in Build 1468. But I'll retest everything more in-depth when you say it's ready.

JoePfeiffer:
> Using your inline pod example, open the pod and change the radial distance to something other than zero. The behavior that I would expect is that it would grab the reference diameter, remove the automatic diameter, and set the outer diameter to the grabbed reference diameter.

OK, I'm seeing it now.  That's not the only place where a component with an automatic diameter reverts to its default diameter when it no longer has a reference -- for instance, create a nose cone and body tube; set a diameter on the nose cone and make the body tube diameter automatic.  Now if you delete the nose cone the body tube reverts to the default diameter.  Thinking of what is most reasonable to do in these cases is a good question, but outside the scope of this PR.  Everything ought to be working now.

hcraigmiller:
New build not available.

JoePfeiffer:
OK, merged unstable (including @SiboVG 's (phantom body check) in and all should be good.

SiboVG:
I still notice some issues. The gap check does not work correctly when there is a component aft to the body tube. This should be an inline overlap warning between the transition and pod set.
image

Maybe it's better to also include the nose cone in the in-line pod set overlap warning?
image

No gap detected when the pod set if before the nose cone:
image


There is a consistent discontinuity warning between the body tube and nose cone, that is activated once the pod set is in-line.
image

neilweinstock:
This is a heck of a feature to jam in at the last second.

JoePfeiffer:
Remember, the goal here is just to warn people about their sci-fi rockets having a gap in the airframe, and also warning about overlaps since it's easy.  We've extended it a bit into looking at more general structure relating to in-line pods, but we really don't want to keep pursuing really oddball cases with a goal of pointing specifically to the errors forever.

SiboVG:
> Remember, the goal here is just to warn people about their sci-fi rockets having a gap in the airframe, and also warning about overlaps since it's easy. We've extended it a bit into looking at more general structure relating to in-line pods, but we really don't want to keep pursuing really oddball cases with a goal of pointing specifically to the errors forever.

I understand, but some of my mentioned anomalies are either not super uncommon (gap warning when the pod is before the parent component) or not very hard to fix (check for gap warnings before the nose cone or after the transition by checking for the furthest previous/next symmetric component). I probably mentioned too many issues, but I still think there needs to be some improvements.

neilweinstock:
I am sympathetic to these desires but we need to get this release out the door.

JoePfeiffer:
It fixes the previous/next problems that were the original motivation, and does report the actual issues in the component tree.  Further improvements in the error reporting can wait.

PR #1993, opened 2023-01-21
Grey out texture edit button if using default material

SiboVG:
When the default material is used, the edit button should be greyed out.

image

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/AppearancePanel.java
@@ -576,13 +576,19 @@ public class AppearancePanel extends JPanel {
 		order.add(textureDropDown);
 		JButton editBtn = new SelectColorButton(
 				trans.get("AppearanceCfg.but.edit"));
-		editBtn.setEnabled(builder.getImage() != null);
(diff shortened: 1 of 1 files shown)


neilweinstock:
Also if there is no texture selected for the component?

SiboVG:
> Also if there is no texture selected for the component?

That's already implemented, no?

SiboVG:
From 22.02.beta.05:

https://user-images.githubusercontent.com/11031519/213885732-c425ab72-e0c4-4b0d-8db5-54be06024d30.mov

neilweinstock:
Yes, but in my experience it's also greyed out when default appearance is used... at least *some* of the time.

hcraigmiller:

Build 1472
Java Version "11.0.17" 2022-10-18 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.17+10-LTS-269)
Microsoft Windows 11 Pro; Version 10.0.22621 Build 22621.1105; Windows Feature Experience Pack 1000.22638.1000.0

SiboVG:
> Yes, but in my experience it's also greyed out when default appearance is used... at least _some_ of the time.

Hm never noticed that. If you can replicate it, please do tell. In the meantime I'll merge this PR already.

PR #1995, opened 2023-01-24
Enable decal editing in Linux

JoePfeiffer:
There appear to have been a total of three typos preventing decal editing in Linux
1) The test for snap confinement ended up not allowing any UNIX-like OSes to set the editor in the preferences
2) When the tests for desktop support failed, the text entry field for the command name and the command chooser were not enabled.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/dialogs/preferences/GraphicsPreferencesPanel.java
+++ b/swing/src/net/sf/openrocket/gui/dialogs/preferences/GraphicsPreferencesPanel.java
@@ -145,7 +145,7 @@ public class GraphicsPreferencesPanel extends PreferencesPanel {
 		 * we will rely on using the xdg-open command which allows the user to pick
 		 * their preferred application.
 		 */
-		if (SystemInfo.getPlatform() != Platform.UNIX || !SystemInfo.isConfined()) {
+		if ((SystemInfo.getPlatform() != Platform.UNIX) || !SystemInfo.isConfined()) {
 			this.add(editorPrefPanel, "growx, span");
 		}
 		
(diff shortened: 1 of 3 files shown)


swing/src/net/sf/openrocket/gui/dialogs/preferences/GraphicsPreferencesPanel.java
```diff
@@ -145,7 +145,7 @@ public void stateChanged(ChangeEvent e) {
 		 * we will rely on using the xdg-open command which allows the user to pick
 		 * their preferred application.
 		 */
-		if (SystemInfo.getPlatform() != Platform.UNIX && !SystemInfo.isConfined()) {
+		if (SystemInfo.getPlatform() != Platform.UNIX || !SystemInfo.isConfined()) {
```
wolsen, line 148: Just looking at this and reading the comment defining the intention, I feel that this should actually be:

```java
if (SystemInfo.getPlaform() == Platform.UNIX && !SystemInfo.IsConfined()) {
...
```

wolsen: Scratch that - the logic is correct, it just took me a moment.

Essentially, we want to run it we're not on Linux and its not a confined state. Your's does the trick

wolsen:
This works correctly in the snap packaging. There is another issue where the gimp snap gets rejected access to the decal png - but this bug is unrelated to this change, previously worked, and is likely an issue with the snapd or gimp snap. I'll follow up on those.

JoePfeiffer:
At least under Linux, it isn't necessary for isSupported(Desktop.Action.EDIT)  to be true in order to launch an editor program -- as I understand it, it can't be done from inside a snap, but isn't a problem otherwise.  If the EDIT action is always defined under MacOS and Windows, this should be ready to go.  If not, we'll need another round.

Is there documentation anywhere listing Java's supported desktops and actions, on a per-platform basis?

Issue #1392, opened 2022-05-30
22.02.beta.03, MacOS 12.2.1 : "Can't be opened because Apple cannot check it for malicious software"


broofa:
I just installed the latest beta, `22.02.beta.03` and upon opening it I get the following error:

![CleanShot 2022-05-30 at 16 21 45@2x](https://user-images.githubusercontent.com/164050/171067567-0f5c2db4-0135-4b76-831e-715fa3ca9691.png)

Note: [There is a workaround](https://support.apple.com/guide/mac-help/apple-cant-check-app-for-malicious-software-mchleab3a043/mac), but it looks like this is an issue that should be addressed here, hence filing this issue.

JoePfeiffer:
The workaround is in the Mac instructions on the download page.  The "real" issue appears as issue #1144

hcraigmiller:
Have you tries following the instructions here: https://openrocket.info/downloads.html?vers=22.02.beta.03

"Show macOS installation instructions"

SiboVG:
Also want to point out that we are working on getting OpenRocket code signed by Apple so that it is included as a trusted software on macOS.

broofa:
@JoePfeiffer, @hcraigmiller :  FWIW, I ran into this issue after launching `beta.02` and being instructed to "Install Update" (below).  Clicking that button triggers a direct browser download of the `.dmg` image.  There's no warning about this issue in that workflow.

![CleanShot 2022-05-30 at 19 14 09@2x](https://user-images.githubusercontent.com/164050/171079572-e7de5436-6e4b-4674-86fc-d169c52debff.png)

SiboVG:
> @JoePfeiffer, @hcraigmiller : FWIW, I ran into this issue after launching `beta.02` and being instructed to "Install Update" (below). Clicking that button triggers a direct browser download of the `.dmg` image. There's no warning about this issue in that workflow.

The direct download in the software updater is meant to ease the pain of installing a new update. But again: this issue is only temporarily, once we get the code signing to work, the warnings will be gone. Thanks for bringing it up though.

PR #2004, opened 2023-01-28
[#2002] Restore plugged

### LINKED ISSUE #2002: "None" or "Plugged" not listed in pull-down for every motor
This was discussed in the past but apparently never implemented.

We had decided that we would put "None" (or maybe "Plugged") into the list of options in the pull-down for motor delay, and remove the terrible mechanism where you have to type the word "none" into the combo box to enable that.

SiboVG:
This PR fixes #2002 and restores #1921.

It's actually funny to see how different commits undid and then redid the changes of #1392 

https://user-images.githubusercontent.com/11031519/215229959-6e1416ba-68a2-4494-b2ec-af75b0a967c9.mov


JoePfeiffer:
The ease with which you can accidentally write over a commit is downright startling, and very unfortunate.  Regressions are all too easy.

PR #2006, opened 2023-01-31
[#1999] Limit geometry warnings + add option to hide warnings & show all warnings

### LINKED ISSUE #1999: Too many warnings displayed in rocket figure
I believe this image speaks for itself:
image

We should do something about this before release.  Since there's obviously not time to do any sort of major rethinking of the warning system (we can save that for later), is there perhaps a way to put the warnings in a box and make it expand/collapse, so you can get it all out of the way?  I'll try to come up with a detailed proposal.

SiboVG:
This PR fixes #1999. Geometry warnings are now limited to 3 lines max. If there are more than 3 warnings, only 2 warnings will be displayed and the last warning line will display "...X warning(s) remaining (click "Show all warnings" below)...", with X being the number of undisplayed warnings.

There is now also a checkbox to show/hide warnings. If the warnings are hidden, there will still be a message in the bottom-right corner indicating how many warnings there are in the design, e.g. "4 warning(s)".

Lastly, there is a button to show all warnings. This will open a new dialog with a table containing all the warnings and their corresponding warning source. A warning source is the rocket component that caused the warning, e.g. if you have a phantom body tube called "Phantom Tube" with a zero-volume warning, the source is "Phantom Tube".

As an extra bonus (I had some time left on my flight), I added the feature that selecting a row in the warning dialog will select the corresponding warning sources in the component tree. E.g. if there is a discontinuity warning between a nose cone and body tube, then selecting that warning in the warning table will also select that nose cone and body tube in the component tree. This makes it easier for users to pinpoint where the issue is.

The warning dialog updates dynamically to changes in the warnings.

Demo:

https://user-images.githubusercontent.com/11031519/215725518-9bcfdd3f-d427-4a5a-803a-554383ee21fb.mp4

A couple of weeks ago I noticed that if you have two components that have the same name, and those components have the same geometry warning, then only one warning was displayed. E.g. if you have two body tubes named "Body Tube", with a zero-volume warning, then you would only get one warning "Zero-volume bodies may not simulate correctly: "Body Tube"". So there was no feedback that there were actually two warnings of the same kind. This PR fixes that and displays both warnings (yes, they're twice the same message).

Lastly, the "Show all warnings" button is only activated when there are warnings:

https://user-images.githubusercontent.com/11031519/215723171-1520e950-aab4-4521-836c-25cbfbb3911b.mov

----

I actually like the warning dialog, it could be our gateway to improving the OR warning system. You could for instance add an extra column in the warning table to indicate the warning severity, or add a text box on the bottom of the warning dialog that displays extra information on a selected warning, or even methods to resolve it.


neilweinstock:
Hoo boy, this is an awful lot of new UI to jam in at the last second.

neilweinstock:
6. Not all warnings are have the clickability implemented (Jagged edge fins, Too many fins).  I don't recall all the different warnings to check.  Note it is not necessarily critical to implement all of these right now, just pointing it out.  For the above two warnings the "Sources" just says "N/A".

neilweinstock:
OK, here's my first proposal for what to do with this PR.  I may be out of step with what others think (we'll see when they chime in) but I can only give my opinion.  My recommendation would have been very different two or three weeks ago.

1) Get rid of "Show all Warnings" and the associated dialog.
2) Keep "Display Warnings" as-is (although move to lower-right corner)
3) Either show all warnings or none in lower right (according to "Display Warnings" checkbox), not the current "show two plus message to show all".

That will satisfy the needs of this PR right now in a simple and direct way.  

We can then plan for a mid-year release to get this in plus whatever else we get to.  It can wait 6 months for this to be done right, with more test and feedback and thought.

JoePfeiffer:
>     1. Get rid of "Show all Warnings" and the associated dialog.
Agree (well, don't put it in on this release.  Think about it early on in the next one)

>     2. Keep "Display Warnings" as-is (although move to lower-right corner)
Agree

>     3. Either show all warnings or none in lower right (according to "Display Warnings" checkbox), not the current "show two plus message to show all".

I like the "show 2 or all" that's here.  For a reasonable design with only a few small problems, it points to the issue so the user can easily see what's going on.  It's only when the design has a *lot* of issues that we start hiding things. to keep them from covering up the user's view of the rocket.

neilweinstock:
> > 3. Either show all warnings or none in lower right (according to "Display Warnings" checkbox), not the current "show two plus message to show all".
> > ```
> 
> I like the "show 2 or all" that's here. For a reasonable design with only a few small problems, it points to the issue so the user can easily see what's going on. It's only when the design has a _lot_ of issues that we start hiding things. to keep them from covering up the user's view of the rocket.

In that case other things would have to change.  The checkbox would probably then be called "Show All Warnings", which would be confusing if there were <3 warnings and we were already showing all warnings.  Checking and unchecking that box would have no effect.  I suppose it would have to be disabled when there are fewer than 3 warnings.  And to be honest a checkbox would not really be the correct control for that scenario.

I'm not sure exactly how it should all look, need to think about it.  Would prefer all or nothing for now, but if necessary we could work out the details.

neilweinstock:
OK, I've thought about it some more, and will explain why I think the show/hide warnings should be all or nothing.

1. There is no utility to only showing 2 out of _n_ warnings.  I'm still going to need to see the rest of them.  There is nothing special about "the first two" that makes them worth displaying on their own.
2. Show/hide is only a single mouse click, and it is a persistent setting.  If you have a design with a small number of warnings, you can leave them showing all the time (in fact that is what I would expect to be most common usage).  Only in the relatively rare case where there are a *lot* of them will it be worth hiding.  And then I'm going to want to make them all go away, not just show a randomly selected 2 of them.
3. We'll have a clearly visible message in lower right of how many warnings there are, when hidden, and right below that will be a button to "show warnings".  I don't see how that could be clearer.

And so, overall, I think show/hide, all or nothing, is clearer and cleaner for this design pass.  For next release (or whenever we can look into a more sophisticated implementation, but any additional display or control complexity should come with clear benefits.

Oh, and I would label the checkbox "Show warnings" (rather than "Display warnings", although that's a minor point.

hcraigmiller:
Insofar as function is concerned, the display warnings checkbox functions as expected with no anomalous behavior found. The same is true of a warning count greater than three changing the display to show on the third line that additional warnings exist. The show all warnings button also pops up the warning table as expected, and a single left-click on an entry identifies the elements in the component tree (except jagged fins). But, if a component dialog is open when the show all warnings button is used to pop up the warning table, the component configuration dialog goes into multi-editing mode without the ability to see what is causing the error; in such an instance, **the component configuration dialog should be closed before the table is opened**.

My thoughts.
1. I generally prefer "Show" warnings instead of "Display" warnings, but question that when such similar dialogs are next to each other. 
2. I understand the reasoning behind having the show warnings checkbox below the warnings, but there is also weight to be given to putting like things together. I believe that the "show" CP/CG and warnings are both primarily visual in effect and should be together in the upper bar (as shown in the issue itself).
3. The show all warnings button is intended to display additional and should be located below the warnings being displayed. Insofar as whether to use a list of warnings in the display pane or a pop-up table of warnings, both have their pros and cons. But I guess what I prefer is to have the number of warnings stated at the bottom with the show all table as a pop-up for the list.  **I really like being able to click on a selection and have the component tree highlight he affected parts.**
4. If my 3 is implemented, I question whether there is a need for both the show warnings checkbox and the show all warnings button, although users such as K'tesh would probably want both.

My struggle with the foregoing has been whether to require the use of the show all warnings button to see a single warning. On balance, I think the best approach is to (1) put the show warnings checkbox up top with the CP/CG check box; (2) show up to three warnings below before it just puts **a warning count with no other warning**, and have the table show all of the warnings (closing an open configuration dialog before it does).

neilweinstock:
> My thoughts.
> 2. I understand the reasoning behind having the show warnings checkbox below the warnings, but there is also weight to be given to putting like things together. I believe that the "show" CP/CG and warnings are both primarily visual in effect and should be together in the upper bar (as shown in the issue itself).

I think in this case the proximity makes it more useful in the lower right, and that space is unused otherwise.  I'm not convinced there's any usability advantage to having it with "Show CG/CP".  It seems like a consistency for the sake of consistency.

I don't think it would necessarily be *bad* to have it up there, but I think it's clearer and easier in the lower right.

> **I really like being able to click on a selection and have the component tree highlight he affected parts.**
I do too.  I have been asking for it for a while.  But it feels half-baked right now and we don't have time to bake it.

> My struggle with the foregoing has been whether to require the use of the show all warnings button to see a single warning. On balance, I think the best approach is to (1) put the show warnings checkbox up top with the CP/CG check box; (2) show up to three warnings below before it just puts **a warning count with no other warning**, and have the table show all of the warnings (closing an open configuration dialog before it does).

I continue to struggle to see any advantage of showing a fixed number and then a "show more".  I mean, I literally can't think of a single instance where that is useful.  Maybe I'm just missing it.

> Insofar as to what to include in the next release, my philosophy is that you don't waste (or delay) something good. If it functions reliably, the best way to obtain information to tweak the final version of the feature is with user feedback. If this feature is put out there in whatever final form, users will express their dissatisfaction if they don't like the way it looks or functions.

I can understand that thought process, but there's a reason where there's such a thing as a feature freeze.  We should have been frozen for weeks or months already, but we keep piling in more stuff. To add a complex feature like this after the Release Candidate, when we're trying to get the damned thing out the door already, is just not the right way to do things.  I consider the current implementation to be a prototype of something that will be great, but not ready yet, and we will need to debate it and possibly go through a few iterations to get it right.  Now is not the time for that.  Simple now, fancy later.

Just MHO.

SiboVG:
Can someone please provide me with a clear and final needed implementation? I really don't have the mental energy to read through all these messages or come up with something myself.

neilweinstock:
Sure, as soon as we agree on something. :)

JoePfeiffer:
I see the "show at least a couple regardless" as a nag to remind people there is a problem.

I could be OK with hiding them all if (1) there was text in that area saying "(n) warnings are being hidden", and (2) the "show all" was always enabled when the user first loaded a rocket or started editing a new rocket, so they had to explicitly turn it off.

neilweinstock:
Maybe? :)

I am proposing exactly this, for above stated reasons:
image

JoePfeiffer:
Looks OK to me.

PR #2007, opened 2023-02-01
Base Drag Hack Example

hcraigmiller:
Adds the "Base drag hack (short-wide)" example.


SiboVG:
Is it possible that by default, it has its flight configuration set to "No motors"?

hcraigmiller:
That is correct.

Issue #2034, opened 2023-02-11
Add title border around instance settings of rail button


SiboVG:
It would be better to add a a title border "Instances" around the instance settings of a rail button, to distantiate it from the other settings. 
<img width="1026" alt="image" src="https://user-images.githubusercontent.com/11031519/218279957-a3efa736-040f-42f9-8e77-fb96a45f1234.png">

PR #2041, opened 2023-02-12
[#2035] Add instances setting in launch lug config

### LINKED ISSUE #2035: [Feature Request] Add instance settings for launch lugs
Rail buttons have instance settings to duplicate the rail button instance. For consistency, it would make sense to also have this option for launch lugs.

SiboVG:
This PR fixes #2035 and adds instances options to launch lugs. Apparently all the code for launch lug instances was already programmed, but the instances settings were just not present in the launch lug config dialog. (Or maybe it was once, but got deleted 👀).

Note: this PR is a subbranch of #2034, so merge that first before merging this PR.


hcraigmiller:
No anomalous behavior found.

Upon increasing the instance count, I expected to see two separate launch lugs, but only saw one (the first completely overlapping the first). I suggest an initial instance separation value equal to that of the rail button initial instance separation value, rather than an initial instance separation value of zero.

OR Build: 1532

SiboVG:
> Upon increasing the instance count, I expected to see two separate launch lugs, but only saw one (the first completely overlapping the first). I suggest a default initial instance separation value equal to that of the rail button default initial instance separation value, rather than a default initial instance separation value of zero.

Good call, fixed now. Default separation is now 2x the launch lug length

PR #2042, opened 2023-02-12
[#2003] Increase resolution of launch temp and pressure to 2 decimals

### LINKED ISSUE #2003: Increase precision of atmospheric pressure and temperature on Launch Conditions tab to 2 digits
Looking forward, I found one of the issues in comparing OpenRocket to RASAero, it deals with air pressure. In RASAero, if you leave the pressure field blank, then RASAero calculates the air pressure internally (using inHg as the unit of measure). Regardless of the unit of measure being used, OpenRocket appears to round to the nearest integer while RASAero allows two additional digits of precision for entered data. So, the first air pressure issue is a rounding error. 101568 pa (29.99 in-Hg) to 101620 pa (30.01 in-Hg) in OpenRocket rounds to 30 in-Hg. So, there is no way to match the OpenRocket in-Hg value to the two-digit more precise value for in-Hg in RASAero. The precision in OpenRocket needs to be increased to 2 decimal points to increase comparability and to increase compatibility as we move to a RASAero export feature; the same is true for temperature.

SiboVG:
This PR fixes #2003. The temperature and pressure in the launch site settings now have a 2 decimal resolution. For the pressure, `bar` and `atm` remain the same, since that was already good at 0.001 resolution. The resolution of `Pa` is still at integers, that's accurate enough.

### DIFF
--- a/core/src/net/sf/openrocket/unit/TemperatureUnit.java
+++ b/core/src/net/sf/openrocket/unit/TemperatureUnit.java
@@ -3,9 +3,9 @@ package net.sf.openrocket.unit;
 public class TemperatureUnit extends FixedPrecisionUnit {
 
 	protected final double addition;
-	
-	public TemperatureUnit(double multiplier, double addition, String unit) {
-		super(unit, 1, multiplier);
+
+	public TemperatureUnit(double multiplier, double addition, double precision, String unit) {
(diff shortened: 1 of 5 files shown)


hcraigmiller:

OR Build: 1534

PR #2054, opened 2023-02-14
[#2051] Sanitize some XML content

### LINKED ISSUE #2051: [Bug] Quotes in part names keep .ork files from loading.
**Describe the bug**
Quotes in .ork files cause a failure to load

**To Reproduce**
Attempt to open the attached file.  An error popup appears, stating the the XML is malformed.  The issue is the 11" long tube.

**Expected behavior**
File opens

**Additional context**
Note that a workaround is to load the problematic part from the parts library, then change a parameter and then change it back so the part name isn't saved.

[ork_test.zip](https://github.com/openrocket/openrocket/files/10727839/ork_test.zip)

SiboVG:
This PR fixes #2051 by sanitizing some of the XML content when saving a .ork file.

Note: it is not possible to open files already saved with quotes, it is only possible to avoid this issue from happening in the future by sanitizing the XML content before saving it.

Steps to reproduce:
1. Add a body tube, go to its parts library
2. Search for "38mm Motor Mount tube 11" and apply it to the body tube
3. Save the file and close it
4. Open it again

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/OpenRocketSaver.java
+++ b/core/src/net/sf/openrocket/file/openrocket/OpenRocketSaver.java
@@ -627,7 +627,7 @@ public class OpenRocketSaver extends RocketSaver {
 	private void writeElement(String element, Object content) throws IOException {
 		if (content == null)
 			content = "";
-		writeln("<" + element + ">" + content + "</" + element + ">");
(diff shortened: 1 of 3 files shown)


hcraigmiller:

OR Build: 1545

PR #2057, opened 2023-02-14
[#2055] Add motor manufacturers substitutor

### LINKED ISSUE #2055: [Feature Request] Add manufacturers wildcard option to configuration naming
Request from a user on the r/rocketry Discord server. If you add `{motors}` in the configuration name field, OR automatically fills in that wildcard with the motor names of that configuration. The user is asking to have a similar effect for the manufacturer's name, so a `{manufacturers}` wildcard.

E.g. typing in `[{motors}] - [{manufacturers}]` in the configuration name field could display as `[A6, A6] - [Klima, Quest]`.

SiboVG:
This PR fixes #2055. Entering `{manufacturers}` in the configuration name field now replaces that text with the motor manufacturers.

I also noticed that the configurations in the flight configuration combo box did not use the proper RocketDescriptor substitutor. In layman terms, this means that the text in the flight configuration combo box was different from that seen in the motors configurations table. This was only noticeable when there were no motors in the configuration. In that case, the flight configuration displayed `No motors`, while the motors configuration table displayed `[No motors]` (with brackets).

This PR now ensures that the configurations in the flight configuration combo box are properly formatted. The only noticeable difference with the current behavior is that a "No motors" configuration is now displayed as `[No motors]` instead of `No motors`.

https://user-images.githubusercontent.com/11031519/218758635-7abe4bad-2b7d-4cb8-a368-ff97f8a58080.mp4

I also added an explanation of the substitution tags in the rename config dialog:
image


neilweinstock:
Cool.  Did anything like this exist before?  If so I had no idea (and, most likely, neither does anyone else).

SiboVG:
> Cool. Did anything like this exist before? If so I had no idea (and, most likely, neither does anyone else).

The {motors} tag always existed. If you edit the config name in OR 22.02, you can see the tag.

SiboVG:
> Although it would be nice if the display could be [A8-5 Estes; B6-0 Estes; B6-0 Estes] instead of [A8-5; B6-0; B6-0 Estes; Estes; Estes]

That would require a new tag IMO, like `{motors-manufacturers}` or something similar. Ideally you would be able to type e.g. `{motors manufacturers}`, which would display as `[A8-5 Estes; B6-0 Estes; B6-0 Estes]`, or `{motors | manufacturers}` which would be `[A8-5 Estes; B6-0 Estes; B6-0 Estes]`. So you could choose the separator between the motor and manufacturer. However, that would be a bit more complex to implement. I would make that a new feature request.

neilweinstock:
> Although it would be nice if the display could be [A8-5 Estes; B6-0 Estes; B6-0 Estes] instead of [A8-5; B6-0; B6-0 Estes; Estes; Estes]

I would tend to think that this is the desired behavior.  Not sure if the Discord user who made the initial request specified?

SiboVG:
> > Although it would be nice if the display could be [A8-5 Estes; B6-0 Estes; B6-0 Estes] instead of [A8-5; B6-0; B6-0 Estes; Estes; Estes]
> 
> I would tend to think that this is the desired behavior. Not sure if the Discord user who made the initial request specified?

He specified it as is implemented in this PR.

neilweinstock:
OK, fair enough.

JoePfeiffer:
> OK, fair enough, although I wonder how many other people will actually want it to work this way (doesn't make a great deal of sense to me).

Agreed.  Yes, I can see this is what was requested, but @neilweinstock 's suggestion would make more sense to me as well.

SiboVG:
Okay, I changed the substitutor to a generalized one, where you can also use a combination of motors and manufacturers, as described in #2068.

Substitution of `[{motors manufacturers}]` in "A three-staged rocket" example:
image


New rename dialog:
image

neilweinstock:
Now we're talking. :)

I would suggest that the example for the combination should be {motors manufacturers} which would yield something like "Aerotech M1350-0" which is (I think) the most likely desired output.

SiboVG:
> Now we're talking. :)
> 
> I would suggest that the example for the combination should be {motors manufacturers} which would yield something like "Aerotech M1350-0" which is (I think) the most likely desired output.

Okay, I'll change it. I was trying to demonstrate that you can choose the separator between the tags. Should I add an extra info text, e.g. "A **combination** of the above can be used. You may choose the separator between the tags."?

neilweinstock:
More show, less tell is generally preferable.  In this case, it is a feature that will be used by an incredibly small group of people; the number who care about separators is an even smaller subset. I wouldn't worry about it.

neilweinstock:
I apologize but I fumbled my original suggestion... I meant to say {manufacturers motors} which would yield "Estes D12".

Not sure why the substitutors are plural but not worth worrying about.

neilweinstock:
If we're stuck with motors then make it manufacturers for consistency.  In a perfect world it would be motor and manufacturer.  Oh well!

hcraigmiller:
Very nice.

OR Build: 1564

Issue #2065, opened 2023-02-22
[Bug] error in calculation of interstices area for tube fins


JoePfeiffer:
As discussed  in this TRF thread, https://www.rocketryforum.com/threads/base-drag-hack-why-not-always-use-it.178095/post-2395352 calculation of interstices area for tube fin drag assumes exactly six tubes and is broken besides (mea culpa).

PR #2066, opened 2023-02-22
Fix interstices area calculation and update pressure drag calculation for tube fins

JoePfeiffer:
This is a start to fixing #2065.  Fixing the interstices area calculation ends up needing to revisit the tube pressure drag calculation, this PR in its present state works for subsonic but gives a much too low drag figure in supersonic flight with the result that altitudes are too optimistic.  Putting this out as a draft so I can work on other stuff while I'm learning enough about supersonic flow in tubes to finish it up.

Note that the existing code gives very good results for the special case of exactly six fins with the same diameter as the body tube.


neilweinstock:
@JoePfeiffer, can you tell me if this is expected.  Below testing done with the jar file from this PR.

Here is a set of tube fins with 1:1 aspect ratio:
image

[As an aside, the first thing I notice is that the drag for the "Tube fin set" still only shows the drag of one fin.  It seems to me that Component analysis should show the total drag for the set, so the sum of the columns works correctly.]

Anyway: pressure drag is .056 for one of these tube fins.  At 1:1 aspect they should be reasonably transparent to the airflow.

Here is the same rocket with the length of the tube fins increased to 12", for a 12:1 aspect ratio:
image

The friction drag has scaled linearly, while the pressure drag has increased to .088, a 57% increase.

However, everything I've ever read suggests that as tube fin aspect ratio gets longer, eventually they become completely opaque to the airflow, and act like a capped tube.  I would assume that my 12:1 ratio tube fins behave like that.  If that's true, then I would have expected two changes:
1) the pressure drag would have increased more than the observed 57%
2) The friction drag would not have increased linearly, because there is no longer flow on the inside of the tubes.  Therefore, the *exterior* component of friction drag should have increased linearly, but the *interior* friction drag should have gone to zero.  Assuming interior and exterior friction drag started about equal, then I would expect a 12x increase in aspect ratio to result in (about) 6x friction drag.

The 12:1 tube fins, with no airflow through them, should also have virtually no impact on CP anymore.  That's harder for me to judge by fiddling in OR, though; I don't know exactly what I should be seeing for the different version of the tube fine.

It is (extremely) possible that everything is working correctly and my own understanding is wrong, but if so please explain why so I don't have to ask this stupid question again. :)

neilweinstock:
Wait, were you trying to address this issue or not: https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-final-is-now-available-for-download.177835/post-2391950.  I think back when you filed #2065 I was talking about the linked issue while you may or may not have been talking about the same thing.

JoePfeiffer:
That was where it started -- but those good results in the six-tube case relied on the errors in every other case.

neilweinstock:
It's definitely behaving better... 3 tube fins goes higher than 6 now. :)

JoePfeiffer:
@neilweinstock thanks -- your questions pulled me out of the rabbit hole I'd descended trying to model rounded cross sections for the tubes, at which point I finally noticed that the .ork files for both of the test rockets had the wrong tube thickness.  It's awfully hard to get pressure drag right the thickness is wrong! 

So this PR now:

1.  Corrects the interstices calculations that were causing nonsensical results with numbers of fins other than six.
2. Refines the drag calculations.
3. When a TubeFinSet is first added to a model, it copies the tube thickness from the body tube to the TubeFinSet.  It doesn't track it after that; if you change the thickness of the body tube you need to adjust the thickness of the tube fins by hand. 

Here are comparisons between the test rockets and the simulated results:

Motor | Actual | PR #1860 | This PR | Error
-- | -- | -- | --
B6 | 352 | 349 | 361| 2.6%
C6 | 693 | 690 | 729 | 5.2%
J360 | 4065 | 4209 | 3982 | -2%
K1103 | 5009 | 4994 | 5084 | 1.5%
K2050 | 4263 | 4307 | 4290 | 0.6%
I've attached the corrected .ork files that got these results.
[Blue Tuber v3-mod.zip](https://github.com/openrocket/openrocket/files/11923811/Blue.Tuber.v3-mod.zip)
[tubular.zip](https://github.com/openrocket/openrocket/files/11923814/tubular.zip)

neilweinstock:
Outstanding, a really great fix for the upcoming release.

PR #2078, opened 2023-02-26
Eliminate option to save "some" simulation data

### LINKED ISSUE #2024: Get rid of option to save subset of simulation data
This is an option that was introduced when disks were much smaller (possibly at the beginning of the project). ork files, even with many simulations, just aren't big enough to worry about any more.

There appears to have been at least one case of a user who used this option to save a subset of the data, and then thought the data hadn't been generated.

Just get rid of the "Every xx seconds" option in the file save dialog

**Additional context**

JoePfeiffer:
Save options now allow saving all sim data or only data from the sim summary table

Changes "Only primary figures" option name to "Only summary data" for clarity

Fixes #2024


hcraigmiller:
I like it.

OR Build: 1573

Issue #2077, opened 2023-02-25
[Feature Request] Add context menu ability for the simulation table to export all run simulation table rows to a CSV file


jppetrakis:
The ability to easily export the simulation data table to a csv for importation into a spreadsheet (eg. excel, google sheets, etc).  Once imported this can provide a portable, concise summary of the performance parameters of a rocket that  can be displayed on any given device (in particular a tablet at the launch site) or printed to a sheet of paper.

Being able to save the spreadsheet to google sheets and then referenced during a launch via a tablet or smartphone gives great flexibility to the user.

Considered directly printing the simulation table to a printer, pdf etc.  The approach of producing a csv file seems simpler.  I can see future expansion of this "command" to include different formats such as pdf or some other document type.

hcraigmiller:
See also, Issue #1525 "After that, I'd like to either be able to select rows and cells and copy them in a way that pastable into , or export the summary table as a delimited file."

JoePfeiffer:
#1525 probably should have been two feature requests in the first place -- the desire to export the table contents is pretty independent of the desire to configure the table.

hcraigmiller:
Resolved by way of PR #2087.

PR #2087, opened 2023-02-27
Issue #2077 [Feature Request] Add context menu ability for the simulation table to export all run simulation table rows to a CSV file

jppetrakis:
Initial attempt at providing function -

Usage:
Run one or more simulations
Right-click table area, and select "Export to csv file", file chooser will open
Select csv file location, changing file name if required (subsequent exports will default to last exported location during session)
click "Open" button

All rows with simulation data will be exported to the file.  Any row with two or more empty columns is assumed to require a simulation run and such rows will not be exported.  The current unstable version appears to fill the simulation table with data upon startup after a save that did run simulations, except for the optimum delay column, so in this case, export will output the table with no value in said column


JoePfeiffer:
Nice.  This is something several users have asked for.  

A few things --

A simulation has to be selected before the context menu can appear.

Commas aren't escaped.  If I create a sim with the name "Simulation, 6" the 6 ends up in the second column.

Now that I see it actually running, I wonder if the context menu is the right place to put it? Admittedly we aren't good at having uniformity in where our export buttons are, but I think this would be the only one that uses a context menu. Would it make more sense to have it in the row of simulation action buttons? I started to suggest making it a third pane in the Plot/Export dialog, but that button isn't active unless a simulation has been selected.

neilweinstock:
Sorry I haven't tried this out yet but I can say with some confidence that the context menu should not be the *only* way to do it.  In general, a context menu should *never* be the only way to do anything.

neilweinstock:
I also thought of the 3rd tab to the Plot/Export dialog; that at least puts the export stuff in the same place.  But you're right about it not being active always.  I've never been a fan of that button.

It could work as a File menu item: "Export sim results table".  Then you can also have it in the context menu.

jppetrakis:
@neilweinstock, @JoePfeiffer  - seems like a good idea... I originally did have another button but it made that button bar busy.  The plot/export button currently pulls up a ui that only pertains to the selected sim too.... I imagine the file or perhaps one of the other context items (I imagine that would lead only to tools if not files!)

The extra button is certainly easy enough...

neilweinstock:
My vote is the File menu, somewhere in the grouping with "Export Decal Image" and "Print Design Info".

jppetrakis:
Sounds good... btw I really like how the motor config column shows the motor description (eg. has the propellant type if available)!

jppetrakis:
I added the capability to the "file" menu and left the context menu action in place (if that's ok!).  Added csv escaping but added the apache.commons.text library because the same utility in apache.commons.lang appears to be deprecated... hope THAT is ok too!

neilweinstock:
I just tried it out.

The problem here is that the feature is only enabled when sims are selected.  This applies not only to the contextual menu item, but also to the File menu item.

Because this feature applies only to the entire table and is completely independent of the current selections, it needs to implemented that way.  That means two things:
1) The file menu item should always be enabled.
2) The contextual menu item is a bit stickier.  Technically I can think of two approaches here:
a) If we retain the current behavior that the contextual menu applies only to selected items, and the menu is only even available when there are items selected, then this feature should not be on that menu.
b) If we want to make the contextual menu always-available, then we can leave this there, but then ensure that the existing options are disabled if no sims are currently selected.

I probably lean towards #2... I like the idea that the contextual menu is always available, even if everything is disabled.  This makes it more discoverable for the users.

Ideally, this feature would be visible in the simulation tab, but there's no obvious place for it right now.  We should come up with something, might be folded into a larger rethink of the UI on this tab.  In the meantime, I think this would be satisfactory to close out this PR.

Would like some other opinions on this.

neilweinstock:
Sorry, you're correct that it should only be enabled when a document is loaded.  Or rather, I would say whenever there is at least one sim in the table.  If a rocket is loaded but has no sims, it should be disabled.

But my test was different.  Open OR, load the Simple Model Rocket example, go to Sim tab, pull up File Menu:
image

jppetrakis:
Gotcha.... your use case is easy enough to deal with.  I'm also trying to activate the menu item when a new sim has been set up eg.  not loaded.

SiboVG:
Some notes:
1. I would change the file menu item name to "Export simulation table to CSV file"
2. The sim export menu has an option to choose the field separator, I would take that value for this PR's CSV export:
image

hcraigmiller:
I would suggest the following UI changes for consistency.

Export Sim Table

SiboVG:
> I would suggest the following UI changes for consistency. In addition to relocating the menu item, I would change the file menu item name to "Export simulation table **as** CSV file"

Good point. In that case, I would name is "Export simulation table as CSV file" (see my previous post). CSV should always be capitalized and I wouldn't abbreviation "simulation" to "sim". Never over-estimate your user, some people have never heard of the sim abbreviation.

jppetrakis:
"simulation table" and "as" it is (or will be).  Since the text says "csv" I'm not certain about your #2 suggestion, @SiboVG ... 

I have been looking for an existing mechanism (although not exhaustively) for handling the activation of the entry under the File menu (before I try and create one).  Right now barring such a mechanism,  I can:
1) always have the export active, and fire a popup saying the table is empty and you should run a simulation
2) put a change listener on the table which could call back into BasicFrame and reset the enablement on the export action.

Preference??

SiboVG:
> Since the text says "csv" I'm not certain about your #2 suggestion, @SiboVG ...

It should be capitalized. CSV instead of csv.

jppetrakis:
@SiboVG - Capitalizing I understand... I mean obtaining the delimiter from the export panel value.  If someone chooses a tab, it's not a .csv file anymore, right?

SiboVG:
> @SiboVG - Capitalizing I understand... I mean obtaining the delimiter from the export panel value. If someone chooses a tab, it's not a .csv file anymore, right?

My bad. A CSV does not necessarily use a comma-separator, although the name makes it appear it does. For instance in Microsoft Excel, when you import a CSV, you can choose by which separator entries are separated. What I often see is CSV files using `;` as the separator, since numbers are sometimes displayed with commas, for instance "1,000,000.00" or if you live in France or Belgium, a comma is used as the decimal separator. Therefore, it should be possible to have a different CSV separator.

jppetrakis:
@SiboVG Now I know!  I did however use commons.text's escapeCSV at least but if someone wants tabs they should be able to get tabs!

jppetrakis:
With a few tweaks, I can (I believe safely) enable/disable the csv export action under the file menu.
Using the separator that is stored in the preferences
Capitalization of CSV, and corrected grammar of action text.

hcraigmiller:
I would change the placement of the file menu item to put like things together:

File Menu Export

Unless others disagree.

jppetrakis:
LOL could have sworn someone said to put it in the decal block... If nobody else chimes in by 7 pm Eastern, I'll move it ;-)

neilweinstock:
I prefer where it is now, for the following reason: the Save/Save as/Export as section all regards the ORK file itself.  The section below it is for auxiliary info.  So to me the CSV export fits perfectly in the bottom section.

hcraigmiller:
I guess we need @JoePfeiffer and @SiboVG to vote too.

jppetrakis:
> 

Well either way I don't think I can push this out anyway... so it'll have to be one of you guys anyway.

JoePfeiffer:
I have a very weak preference for leaving it where it is now, for exactly the same reason as @neilweinstock

hcraigmiller:
Export fails; I am also very concerned that the export attempts to default to the Windows system directory (that should never happen).

Method 1: Open A simple model rocket example, select export simulations, select file directory, attempt to save.
Method 2: Open A simple model rocket example, re-run all simulations, select export simulations, select file directory, attempt to save.

File Menu Export Failure n_01

JoePfeiffer:
Confirming the exception.  Under Linux it's bringing up the standard file save dialog, with the directory I've been saving stuff to selected.

jppetrakis:
That's due to my attempt to add apache.commons.text to the mix.  I missed something it seems.

jppetrakis:
> Confirming the exception. Under Linux it's bringing up the standard file save dialog, with the directory I've been saving stuff to selected.

RE: selected directory... 
I set that up per-session (I think I mentioned this earlier didn't I?)

JoePfeiffer:
On apache commons -- did you add the modified build.xml to the commit?

On file dialog -- it was more a response to @hcraigmiller providing you with another data point; in my case it just remembers where I last saved something.  Don't know why his Windows box is trying to put it somewhere odd.

SiboVG:
There is a Wiki item on how to properly add libraries to the OR codebase: https://github.com/openrocket/openrocket/wiki/Developer's-Guide#adding-new-libraries. It's not only the build.xml that needs editing.

jppetrakis:
I think I got it....

jppetrakis:
It's supposed to initially use this for the initial lixation: System.getProperty("user.dir")

hcraigmiller:
Unexpected results.

Method 1: Open A simple model rocket example, select export simulations, select file directory, pop-up says "Open" not "Save."
Method 2: Open A simple model rocket example, re-run all simulations, select export simulations, select file directory, pop-up says "Open" not "Save."

And it is still defaulting to the system folder.

File Menu Export Failure n_02 01

Interestingly, if you click "Open" the file is actually saved correctly.

In contrast, if you use the export feature on the export data pane, it functions correctly.

File Menu Export Failure n_03 01

jppetrakis:
@hcraigmiller - new commit improves text on file chooser for save.  This particular export function is not accessed in the Plot/Export mechanism.  It is (as you already know) accessed through the File -> Export simulations table as  and by right clicking on the simulations table itself and choosing the export action in the list.  That's why I think you're seeing diffences.  Finally I expect that how you are launching openrocket is defaulting its current working directory to system32.  That said, I changed the default open dialog location

hcraigmiller:
As to the feature generally, the save text and save location functions as expected, no anomalous behavior found.

Even so, the feature itself does not provide the spreadsheet data in a format that I expected. Ordinarily, a spreadsheet is expected to be in a format that can be sorted/organized without any significant manipulation. Here, all of the fields are text fields that cannot be sorted without removing the units (a tedious process).
File Menu Export n_01 01
I would suggest that the units be placed in Row 2 with the simulation data beginning with Row 3 (without the units) so that the data can be sorted/reorganized more easily.
File Menu Export n_01 04
OR Build: 1595

neilweinstock:
I second what @hcraigmiller said above.  Units should be in the header only to facilitate sorting.

jppetrakis:
Welllll.... sorting seems to be app dependent... f I import the current resulting csv into google sheets the columns appear to sort properly.  (I mostly use google sheets when you-know-who intentionally broke my office license with a windows update) Go figure.  That said... I'll do something about it...

JoePfeiffer:
What we're using for our detailed data exports has a header line where each column is labeled "item (units)".  That seems to work well, and there's something to be said for uniformity here.

Also, would we be better off presenting the summary table like the  above in the program, rather than having the CSV export feature edit the cells?

jppetrakis:
> What we're using for our detailed data exports has a header line where each column is labeled "item (units)". That seems to work well, and there's something to be said for uniformity here.

Sort of what I had in mind...


> 
> Also, would we be better off presenting the summary table like the above in the program, rather than having the CSV export feature edit the cells?

I suppose so, but I have it working!   Wasn't all that hard.  :-)

jppetrakis:
New commit re unit handling in the output.  It'll be easy to remove if Joe's suggestion is implemented.

hcraigmiller:
Nicely done.

OR Build: 1597

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
```diff
@@ -598,6 +624,216 @@ public void updateEnabledState() {
...
+					} catch (IOException e) {
+						String msg = e.getMessage();
+						JOptionPane.showMessageDialog(simulationTable.getParent(), msg);
+					}
+				}
+			}
+		}
+
+		private JFileChooser setUpFileChooser() {
+			JFileChooser fch = new JFileChooser();
```
SiboVG, line 685: Add `fch.setCurrentDirectory(((SwingPreferences) Application.getPreferences()).getDefaultDirectory());` to open the default saving location for files.

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
```diff
@@ -598,6 +624,216 @@ public void updateEnabledState() {
 		}
 	}
 
+	class DumpSimulationToCSVAction extends SimulationAction {
```
SiboVG, line 627: I don't really like that the code for the actual CSV writing is implemented here, doesn't make the code very modular. Instead, take a look at `core/src/net/sf/openrocket/file/CSVExport.java`. I would create a separate file in `core/src/net/sf/openrocket/file` called `SimulationTableCSVExport.java` and implement the CSV exporting functionality there. Or if you find out that `CSVExport.java` contains a lot of the same code that you implemented here, it may be better to put your code in `CSVExport.java` and then use/rewrite some of the existing code to fit your goal. (that was one of my first concerns when I watched your code, that there may be a lot of duplicate code with the code in `CSVExport.java`. I haven't dug in the code to verify that, my apologies, but be careful with this.)

SiboVG (changes requested):

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
```diff
@@ -598,6 +624,216 @@ public void updateEnabledState() {
...
+				}
+
+				@Override
+				public String getDescription() {
+					return trans.get("FileHelper.CSV_FILTER");
+				}
+				
+			});
+
+			// default output csv to same name as "the rocket".
```
SiboVG, line 709: `// Default output CSV to the rocket name`
Your comment otherwise makes it seem like you would actually name it "the rocket", instead of the rocket's name (try to reserve double-quotes in your comments only for Strings)

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
```diff
@@ -598,6 +624,216 @@ public void updateEnabledState() {
...
+
+		private JFileChooser setUpFileChooser() {
+			JFileChooser fch = new JFileChooser();
+			String saveDialogTitle = trans.get("simpanel.pop.export_to_csv.save.dialog.title");
+			String saveButtonText = trans.get("simpanel.pop.export_to_csv.save.button.text");
+			fch.setApproveButtonText(saveButtonText);
+			fch.setDialogTitle(saveDialogTitle);
+			fch.setApproveButtonToolTipText(saveDialogTitle);
+			// set up to filter for .csv's
+			fch.setFileFilter(new FileFilter() {
```
SiboVG, line 692: Use `swing/src/net/sf/openrocket/gui/util/FileHelper.java.CSV_FILTER` as the file filter, it already has the correct functionality. So:
```Java
fch.setFileFilter(FileHelper.CSV_FILTER);
```
Then you can also delete the `// set up to filter for .csv's` comment, as the code will be self-explanatory

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
SiboVG, line 685: I would add a `CsvOptionPanel` so you can choose the CSV parameters in the file chooser dialog. See the example in `swing/src/net/sf/openrocket/gui/dialogs/optimization/GeneralOptimizationDialog.java savePath()`

SiboVG (changes requested):

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
jppetrakis: Thanks!

SiboVG:
There is no way of choosing the file name on macOS:
image

Take a look at some of my change requests, especially looking at existing examples in the OR codebase that implement CSV exporting. Taking inspiration from that code will probably fix this issue.

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
jppetrakis: a matter of temperament on my part.

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
jppetrakis: This will take some time...

SiboVG: Time is all we have :) By the way, we appreciate all your hard work on this 👍 All these comments sound like nagging, so I hope you're not offended by them. I know what it feels like when you make your first contributions to an open-source project and you then get a thousand comments from seniors. It's not fun, but it improves the program 🙂.

JoePfeiffer:
> New commit re unit handling in the output. It'll be easy to remove if Joe's suggestion is implemented.

Not something I feel anywhere near strongly enough about to insist on.

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
jppetrakis: Each enough to parrot... but it adds more than the separation char, and it's unclear if the decimal precision can easily be applied to the strings extracted from the table...

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
```diff
@@ -598,6 +624,216 @@ public void updateEnabledState() {
...
+			// that this would no longer be needed because if there is no sim table yet, the context menu
+			// won't show up.   But I'm going to leave this in just in case....
+			if (modelRowCount <= 0) {
+				String msg = trans.get("simpanel.dlg.no.simulation.table.rows");
+				JOptionPane.showMessageDialog(tableParent, msg);
+				return;
+			}
+
+			JFileChooser fch = this.setUpFileChooser();
+			int selectionStatus = fch.showOpenDialog(tableParent);
```
SiboVG: Should be `showSaveDialog`. (fixes the issue for macOS not having a naming field)

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
jppetrakis: Did the initial refactor but when I tried to move it under core had some issues... so for now it's still in openrocket-swing

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
SiboVG: Well you do the String conversion yourself in the code:
```Java
String value = o.toString();
```
You can try this instead:
```Java
final String valueString;
if (o instanceof Value) {
    double value = ((Value) o).getValue();
    valueString = {implement the code to format the double using the supplied decimal precision};
} else {
    valueString = o.toString();
}
```

SiboVG (changes requested):

jppetrakis:
@SiboVG - I got the file chooser working better on Mac (luckily I happened to have one lying about...)

swing/src/net/sf/openrocket/utils/SimulationTableToCSVFileExporter.java
```diff
@@ -0,0 +1,235 @@
...
+					JOptionPane.showMessageDialog(simulationTable.getParent(), msg);
+				}
+			}
+		}
+	}
+
+	private JFileChooser setUpFileChooser() {
+		JFileChooser fch = new JFileChooser();
+		String saveDialogTitle = trans.get("simpanel.pop.export_to_csv.save.dialog.title");
+		String saveButtonText = trans.get("simpanel.pop.export_to_csv.save.button.text");
```
SiboVG, line 105: There is no real added value to defining a custom save button text (and tooltip), I would remove it.

SiboVG (changes requested):

core/resources/l10n/messages.properties
```diff
@@ -536,6 +536,10 @@ simpanel.pop.plot = Plot / Export
 simpanel.pop.run = Run
 simpanel.pop.delete = Delete
 simpanel.pop.duplicate = Duplicate
+simpanel.pop.export_to_csv = Export table as CSV file
```
SiboVG, line 539: Translation keys shouldn't be snake_cased, but camelCased.

SiboVG (changes requested):

SiboVG:
@jppetrakis there were some missing dependencies for your libraries for IntelliJ, I pushed a commit to fix that.

I would however delete the `org.apache.commons.lang3` library, we don't need it. There exists already a string joining method in `swing/src/net/sf/openrocket/utils/StringUtils.java`. So I would remove the library and use the existing method, that saves us a good .6 mB that we don't need :).

jppetrakis:
> 

should it be *completely* removed??

swing/src/net/sf/openrocket/utils/SimulationTableToCSVFileExporter.java
```diff
@@ -0,0 +1,237 @@
...
+			for (int j=1; j<modelColumnCount ; j++) { // skip first column
+				String colName = simulationTable.getColumnName(j);
+				String unitString = valueColumnToUnitString.get(colName); // unit string MAY be null!
+
+				Object o = simulationTableModel.getValueAt(i, j);
+				if (o != null) {
+					String value = null;
+					if (o instanceof Value) {
+						double dvalue = ((Value) o).getValue();
+						value = String.format("%."+precision+"f", dvalue);
```
SiboVG, line 204: Use `TextUtil.doubleToString()`

swing/src/net/sf/openrocket/utils/SimulationTableToCSVFileExporter.java
```diff
@@ -0,0 +1,237 @@
...
+			}
+
+			// piece together the column data for the ith row, skipping any rows with null counts > 0!
+			for (int j=1; j<modelColumnCount ; j++) { // skip first column
+				String colName = simulationTable.getColumnName(j);
+				String unitString = valueColumnToUnitString.get(colName); // unit string MAY be null!
+
+				Object o = simulationTableModel.getValueAt(i, j);
+				if (o != null) {
+					String value = null;
```
SiboVG, line 201: Make this a final String and remove the null initialization, it's better practice since that way you'll be sure that value will never be actually null. (take a look at my initial remark on this and the code provided)

SiboVG (changes requested):

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
SiboVG: You need to make sure that when you refactor the file to Core that there aren't any references to Java classes from Java Swing, otherwise you'll get cyclic dependency issues. So just refactor your code well enough so that all the info that the CSV export needs from the Swing-part of OR, it gets from method arguments.

SiboVG:
> should it be _completely_ removed??

Yes, we don't need it, you can accomplish the same functionality with the OpenRocket StringUtils.

swing/src/net/sf/openrocket/gui/main/SimulationPanel.java
SiboVG: But if there's no easy way to do that, it's perfectly fine to put it in `swing/src/net/sf/openrocket/file`.

jppetrakis:
Latest commit just has the code changes... I tried removing lang3 but I seem to have tripped over something... it would not initialize the StringEscapeUtils class from commons-text so I must have missed/messed something...

swing/src/net/sf/openrocket/utils/SimulationTableToCSVFileExporter.java
jppetrakis: done

SiboVG:
> Latest commit just has the code changes... I tried removing lang3 but I seem to have tripped over something... it would not initialize the StringEscapeUtils class from commons-text so I must have missed/messed something...

I'll try to take a look at it tonight.

SiboVG (approved):

SiboVG:
@jppetrakis in my latest commit, I updated the file chooser dimensions because (at least on my macOS machine) the dialog got cut off on the bottom because the CSV accessory increased the window content size. I now added some static parameters, but could you check if there may be a better way, to do it dynamically?

The same issue is present in the CSV export of the rocket optimizer ("Tools > Rocket optimization", when you click "Save path").

Thanks!

jppetrakis:
I was wondering about a possible association of the apache commons stuff....  I'll try and take a look at the chooser.  Who knew something so simple....

jppetrakis:
@SiboVG - I've been poking a bit at the chooser dialog... it's as if the Mac implementation is unaware of the fact that the "selected file" text area is there.  None of the things I've tried so far have had the desired effect...  I'm still trying to look into this at this time.

SiboVG:
> @SiboVG - I've been poking a bit at the chooser dialog... it's as if the Mac implementation is unaware of the fact that the "selected file" text area is there. None of the things I've tried so far have had the desired effect... I'm still trying to look into this at this time.

We'll just hard-code the new size in for now. I'll just push a commit that only applies the dimension upscale when the OS is macOS.

SiboVG:
Okay, code-wise, this PR should be ready. We'll wait for @hcraigmiller's official "functions as expected" approval.

hcraigmiller:
Indeed, it functions as expected, no anomalous behavior found.

OR Build: 1607

SiboVG:
Thanks for all your hard work on this @jppetrakis, we really appreciate it 👍.

jppetrakis:
Thanks everyone!  It was an eye-opener ;-) It's been a while since I did anything with swing and it's clear there is more to learn in this code base.

hcraigmiller:
Grab another issue too. . . 👍.

JoePfeiffer:
Thank you!

PR #2091, opened 2023-03-01
[#2090] Add plugged option to charge delay combobox if it wasn't present yet

### LINKED ISSUE #2090: [Feature Request] Offer "None (plugged)" option for motors that don't have an existing -P
In the motor selection dialog, the Ejection delay pulldown should offer a "None (plugged)" option at the end of the list, *unless* the motor already has a -P option in the list (in which case the new entry isn't needed).  This should eliminate the need for the user to have to type in "None" which almost no one knows to do despite the explanatory text below.  That text can be removed once this feature is implemented.

I would also like to do something different with custom delays, so the pulldown could be a pure selector and not a combo box, but I suppose that's a separate issue.

SiboVG:
This PR fixes #2090. Every motor now has the plugged option. Neil mentioned in #2090 to remove the explanatory text below the combo box, but I would prefer to leave it in.

https://user-images.githubusercontent.com/11031519/222033478-89a21949-8543-406d-aaed-56841b0487aa.mov

I also noticed an issue due to us changing the name from "None" to "Plugged (None)": selecting the plugged option in the combobox would not change the combobox text, it would still be stuck at e.g. "0". It was not only a display issue, the actual ejection delay would also not change. The issue was caused by an old translation key that still read "None" and was called when a combobox item was selected. Anyways, this PR fixes that.


neilweinstock:
In that case I would change the text to "Plugged", because that is the primary description of the option in the pull-down selector.

neilweinstock:
While we're here, should we also put in a bit of text that says you can type a custom value into the combo box as well?

JoePfeiffer:
> While we're here, should we also put in a bit of text that says you can type a custom value into the combo box as well?

Yes.  Aerotech and CTI high power motors all have a procedure to shorten the delay time, so it can be virtually anything the user wants.

SiboVG:
New texts:
image

neilweinstock:
I was just doing some experimenting with this in 22.02 and found that for motors that offer a plugged option, you can't select it!

Try selecting an Estes A10-P; it immediately reverts to A10-0.

I was also surprised to see that for the A10, the -P option was shown as "Plugged (none)".  I thought we were going to leave it at -P, since that is the way the motor is sold.

In other words, the behavior *I* expected was:
1) If the motor offered a -P option, we left the option list as it was
2) If the motor did *not* offer a -P option, we added a "plugged (none)" option to the end of the list

And of course, in case 1, it should actually allow you to select -P. :)

hcraigmiller:

OR Build: 1587

SiboVG:
> I was just doing some experimenting with this in 22.02 and found that for motors that offer a plugged option, you can't select it!

See my first post, that is fixed in this PR.

SiboVG:
> I was also surprised to see that for the A10, the -P option was shown as "Plugged (none)". I thought we were going to leave it at -P, since that is the way the motor is sold.
> 
> In other words, the behavior _I_ expected was:
> 
> 1. If the motor offered a -P option, we left the option list as it was
> 2. If the motor did _not_ offer a -P option, we added a "plugged (none)" option to the end of the list
> 
> And of course, in case 1, it should actually allow you to select -P. :)

I don't know if it's a good idea to have two separate plugged naming for the same functionality, may just confuse users. From the perspective of a user, they also don't care if the plugged option comes from the manufacturer, or from us.

What if we change the naming for all scenarios to "**-P (Plugged)**"?

neilweinstock:
> See my first post, that is fixed in this PR.

Whoops sorry, failed to read the text below the screengrab.

> I don't know if it's a good idea to have two separate plugged naming for the same functionality, may just confuse users. From the perspective of a user, they also don't care if the plugged option comes from the manufacturer, or from us.

I disagree with this; the distinction is real and relevant.  It is useful to know if a plugged version is commercially available, vs. if it's something I'll need to to myself.  To be fair this only applies to a relatively small set of motors.  

I will also note that we've had this distinction since forever, and the only problem we've had is that folks didn't realize you could type "none" to create a plugged version of any motor, and we're fixing that now.  Of course, the fact that we've been doing it this way is not necessarily a compelling argument to continue.

> What if we change the naming for all scenarios to "**-P (Plugged)**"?

My vote is to retain the existing distinction: -P when the option is in the motor file, and plugged (none) or none (plugged) or whatever when it's an option that we're adding ourselves.

I don't feel extremely strongly about it, though, and will go with consensus.  What say you, @JoePfeiffer and @hcraigmiller and @wolsen?

JoePfeiffer:
For our purposes, I think it's better to not have the distinction. In addition to whether or not a motor is sold as plugged, we've also got the ones for which plugging is such a routine, manufacturer authorized procedure that it might as well have the option of being sold that way (pretty much all medium and high power motors except the really big ones) vs. motors that users can plug in violation of the safety code (black powder with a -0 delay).  What matters to us is just what the delay is, not how it got that way.

Would the text entry prompt be clearer if it said "(or you can type in a desired delay in seconds)"?

neilweinstock:
I can go along with that, and I like that wording for the text entry prompt.

SiboVG:
New texts:
image

I decided on a shorter version of Joe's suggestion, "(or type in a desired delay in seconds)", because otherwise the width of the combobox would become too wide. I can make the combobox width independent of the info text length, but I kinda like that they're the same width, makes it more obvious that they belong together.

neilweinstock:
I think you can shorten it even a bit more to "or type in desired delay in seconds" (get rid of "a").  Shorter and also better.

hcraigmiller:

OR Build: 

SiboVG:
> I think you can shorten it even a bit more to "or type in desired delay in seconds" (get rid of "a"). Shorter and also better.

Done.

hcraigmiller:

Once this is merged, the motors on a few examples will need to be tweaked.

OR Build: 1596

JoePfeiffer:
One more thing -- the plugged ejection charge delay should not appear until a motor is selected (just as the thrust curve selector doesn't).

SiboVG:
> One more thing -- the plugged ejection charge delay should not appear until a motor is selected (just as the thrust curve selector doesn't).

Fixed now + the delay box is now also disabled, just as the curves widgets are:
image

hcraigmiller:
Unexpected behavior. It only allows changes in the default delay if there are more than one thrust curves to choose from.

Delay n_01

SiboVG:
> Unexpected behavior. It only allows changes in the default delay if there are more than one thrust curves to choose from.

Wow, sorry wasn't really paying attention, thanks for catching it, should be fixed now.

hcraigmiller:

OR Build: 1609

PR #2097, opened 2023-03-04
Fix error in sim loading, add unit test, refine loading a bit.

JoePfeiffer:
(see discussion of sim loading error at https://discord.com/channels/1073297014814691328/1078123170088878130/1080633494604107796)

The problem with nearly all the sims being marked as "out of date" was that a single loop was being used both to determine whether any of the sims had been saved with data, and to mark them as current.  I exited when a sim with data was identified; all the sims needed to be updated.  Loop isn't exiting early now.

Added a unit test to determine whether a sim was stored in a state in which it couldn't be run (I'm not sure that can happen in practice), runnable but not run, current, or out of date.  

Refined sim loading a bit to identify when loaded data was out of date when it was stored, and mark it out of date.


hcraigmiller:

OR Build: 1593

PR #2116, opened 2023-03-14
Fix simtab export

SiboVG:
This PR fixes a small bug in the sim table exporting, where an empty (e.g. new) simulation would throw a NullPointerException.

I also added the following improvements:
- Add a confirmation dialog for overwriting an existing file
- Account for exponential notation setting in CSV export file chooser

I also made the change that using the export functionality using the context menu in the sim table now only exports the currently selected simulations:
image

Not only does that make more sense to me, it also adds an extra functionality for users who only want to export a couple of sims.


hcraigmiller:

For clarification, when a partial context menu sim table is exported, where an empty (e.g., new) simulation is one of the simulations selected, the empty simulation is ignored (not included in the export).

OR Build: 1618

JoePfeiffer:
> For clarification, when a partial context menu sim table is exported, where an empty (e.g., new) simulation is one of the simulations selected, the empty simulation is ignored (not included in the export).

I understand this to be intended behavior

PR #2117, opened 2023-03-16
Set base drag of previous component before exiting loop if current component drag overridden

JoePfeiffer:
See TRF thread https://www.rocketryforum.com/threads/base-drag-hack-why-not-always-use-it.178095/post-2404817
and discord thread https://discord.com/channels/1073297014814691328/1078118846818439199/1085899733345771520


JoePfeiffer:
not the right approach.  Replaced by #2119

PR #2119, opened 2023-03-17
Fix base drag 

### LINKED ISSUE #2118: [Bug] Setting CD override on a SymmetricComponent sets base CD of preceding component to 0
Setting CD override on a SymmetricComponent sets base CD of preceding component to 0.

See https://discord.com/channels/1073297014814691328/1078118846818439199/1085899733345771520

JoePfeiffer:
The underlying problem causing issue #2118 is that the base drag calculation loop in BarrowmanCalculator compares each SymmetricComponent's forward radius to the preceding component's aft radius, and then calculates the preceding component's base drag.  In addition to requiring a special case for the final SymmetricComponent, it's also unclear which resulted in the bug.

This PR rewrites that loop, so each SymmetricComponent compares its aft radius to the forward radius of the component behind it, and sets its own base CD.  This reduces the special case handling to just noticing whether there is no next component and treating its forward radius as 0 if not, and also fixes #2118.

Also adds a unit test.


hcraigmiller:

OR Build: 1620

PR #2127, opened 2023-03-20
[#2126] Apply freeform conversion to config listeners

### LINKED ISSUE #2126: When multi-editing fins, "Convert to freeform" only operates on one
When I have selected multiple fins for multi-editing, the "Convert to freeform" button is available.  However, if I click on it, it only converts the first fin that I had selected, not all the fins that are selected.  It should operate on all of them.

SiboVG:
This PR fixes #2126.

https://user-images.githubusercontent.com/11031519/226393635-c042f27a-6d01-44d7-8289-98ec496c1692.mov

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/FinSet.java
@@ -1425,7 +1425,8 @@ public abstract class FinSet extends ExternalComponent implements AxialPositiona
 	 * Split the fin set into individual fins.
 	 * @return  A list of the new fin sets.
 	 */
-	public List<FinSet> splitFins() {
+	public List<FinSet> splitFins(boolean freezeRocket) {
+		final RocketComponent root = getRoot();
 		RocketComponent parent = getParent();
 		int index = parent.getChildPosition(this);
 		int count = getFinCount();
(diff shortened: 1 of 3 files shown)


neilweinstock:
That was quick. :)

SiboVG:
> Note: Multi-component edit correctly converts all selected to freeform, but only splits the first selected fin set. Should multi-component fin split be included in this PR or a new PR?

Implemented now.

hcraigmiller:
If you convert to freeform and then split without closing the multi-component dialog, you get an error.

Untitled

If you convert to freeform and close the multi-component dialog before reselecting both fin sets and splitting the fins, then you don't. When you select both fin sets and split the fins first, the dialog closes automatically so this isn't an issue because reselection is mandatory.

SiboVG:
> If you convert to freeform and then split without closing the multi-component dialog, you get an error.
> 
> Untitled
> 
> If you convert to freeform and close the multi-component dialog _before_ reselecting both fin sets and splitting the fins, then you don't. When you select both fin sets and split the fins first, the dialog closes automatically so this isn't an issue, reselection is mandatory.

I can't recreate this, am I missing something?

hcraigmiller:
Steps:

1. Open the _A simple model rocket example_.
2. Duplicate the fin set and set its position to -3 in; reposition to above other fin set on the component tree.
3. Select both fin sets and click edit.
4. On the _General_ tab, click _Convert to Freeform_.
5. On the _Shape_ tab, move the second point.
6. On the _Shape_ tab, click _Split fins_.

Error occurs.

SiboVG:
Should be fixed now.

hcraigmiller:

OR Build: 1639

PR #2128, opened 2023-03-20
[#2108] Don't dispose config dialog when no components selected in 3D view

### LINKED ISSUE #2108: In 3D view, clicking in blank area should not close config editor
This is a regression from 15.03.

When a component is selected, it is outlined in red in the design view.  Clicking outside the rocket in the blank area will cause the component to be unselected and the outline removed.

If the config editor is open at the same time, expected behavior is that it should stay open after the unselect operation.  This is the behavior in 15.03.

In 22.02 it works correctly in the side view, but in 3D view the unselect causes the config editor to close.  It should stay open.

SiboVG:
This PR fixes #2108.


SiboVG:
Busy day!

_Originally posted by @neilweinstock in https://github.com/openrocket/openrocket/issues/2108#issuecomment-1476504204_

Don't tell my employer 🤫.

PR #2129, opened 2023-03-20
[#2102] Add "Select components of same color" and "Select none" option

### LINKED ISSUE #2102: [Feature Request] "Select all components of same color"
Multi-edit is a great way to change the color of many components at once.  What is not great is the need to very carefully multi-select the full set of components to change.  It's slow and tedious and repetitive.

I propose a function to "select all components of the same color".  This *ignores* opacity, it just pays attention to RGB.  It should be a contextual menu item and also under the Edit menu, I think.

This way if I want to change all the white components to black, I just right-click on one of them, choose "select all components of same color", and it's done.  Then I go in and change the color.  

This would be a very nice convenience when editing appearance, and would be a perfect companion to component mult-edit.

SiboVG:
This PR fixes #2102 and adds an option to "Select components of the same color" and to "Deselect components" in the edit application ribbon and the component context menu. #2102 was a request for the same color selection, but I thought there is probably a good majority of people that don't know how to properly deselect components.

Note: if a component has the default appearance, using the "Select components of the same color" will select all components of the **same type** that have the default appearance. E.g. selecting a body tube with default appearance and using the select of same color function will only select body tubes with the default appearance.

Demo:

https://user-images.githubusercontent.com/11031519/226475908-189bb3d1-7654-40ef-ad3c-f8c53fa223fd.mp4

---

A question remains: what to do if the component has a decal? Should the "Select all components of same color" also check whether other components have the same color **and** the same decal? If the current component does not have a decal, but another component has a decal, but the same color, should that component be selected?

Another question: currently, the first selected component is used for the target color. So if you select multiple components, only the first component is considered. Should a more intelligent system be used that checks if a component has the same color as **any** of the selected components? E.g. you multi-select a red and blue component, clicking "Select all components of same color" will select all red **and** blue components.


hcraigmiller:

NOTE: The reference color is that of the highest selected component in the componenet tree, not color of the highest number of components with the same color.

OR Build: 1636

neilweinstock:
Sorry I won’t be able to test this tonight but I can comment. 

1) decals should be ignored for the purpose of this behavior. We’re only concerned about the color. 

2) having multiple items selected and *then* activating this feature is not the intended use case; in fact it doesn’t really make sense at all. I’d be happiest if it were *disabled* when multiple items were already selected, because any other behavior we choose is likely to be unintuitive and/or surprising and hence not useful.

If you wanted to get fancier and make it available only if all the currently selected components already have the same color, that would be fine because then there would be no confusion or surprise. But it’s not necessary.

SiboVG:
Okay, same color selection is now only activated when one component is selected.

neilweinstock:
Awesome.  I will try this out as soon as it finishes building.

neilweinstock:
Very nice.  Two small issues:

1) If the selected component doesn't have a color, the menu item is still available.  Example: pods, stages.  I tried it with a pod and it selected the other pod... do these components have an appearance behind the scenes?
 
2) I don't think that having the two new items in a submenu is the right approach.  Select -> none is very unintuitive and non-standard; normal menu item for that would be "unselect all" or "deselect all", which wouldn't naturally go underneath a "Select" item.

Of course, putting the two new items in the top level menu would cause it to become very wide due to "Select components of same color", and I don't have a proposal to shorten that.

And so I'm going to approve this for now, but as soon as we come up with a better menu organization we should change it.

SiboVG:
> 1. If the selected component doesn't have a color, the menu item is still available.  Example: pods, stages.  I tried it with a pod and it selected the other pod... do these components have an appearance behind the scenes?

They don't. I pushed a commit to only enable the item if the component has an appearance section.

> 2\. I don't think that having the two new items in a submenu is the right approach.  Select -> none is very unintuitive and non-standard; normal menu item for that would be "unselect all" or "deselect all", which wouldn't naturally go underneath a "Select" item.

I actually see a deselect item (if present) under a select menu quite often. For example Adobe Illustrator:
image

Blender:
image

My idea for putting it in a separate select menu was that in the future we may add other selection modes that could be put there. Just helps organize things, I can't stand it when things aren't properly grouped together *(flashbacks to horrific desktop organization I've seen with some people)*.

I agree that the "None" is not used that often, so I'll change it to "Deselect all".

hcraigmiller:
Is the color selection no longer functional?

Untitled

neilweinstock:
> I actually see a deselect item (if present) under a select menu quite often. For example Adobe Illustrator: image
> 
> Blender: image
> 
> My idea for putting it in a separate select menu was that in the future we may add other selection modes that could be put there. Just helps organize things, I can't stand it when things aren't properly grouped together _(flashbacks to horrific desktop organization I've seen with some people)_.
> 
> I agree that the "None" is not used that often, so I'll change it to "Deselect all".

Fair.  I don't think that what we have is exactly analogous to those programs, but in any case I'm OK with it for now.  Thank you for implementing this.

hcraigmiller:
Steps:

1. Start OpenRocket and open the _Three-stage rocket_ example.
2. Select all three fin sets.
3. Right-click on one of the fin sets.

Untitled n_02

neilweinstock:
That is correct behavior.  Select components of same color now only applies when a single component is selected.

hcraigmiller:
I see, I see... in that case, functions as described with no anomalous behavior found.

OR Build: 1653

PR #2147, opened 2023-03-27
[#2140] Only enable sim editing for single-selections

### LINKED ISSUE #2140: [Bug] "Edit Simulations" button disabled when multiple simulations are selected, but Edit menu option still works
If multiple simulations are selected, the "Edit Simulations" button is disabled.  However, Edit -> Edit (cntl/cmd-E) still works to multi-edit the whole group.  These should either both work or neither.

SiboVG:
This PR fixes #2140 and disables the edit action when multiple simulations are selected.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
+++ b/swing/src/net/sf/openrocket/gui/main/BasicFrame.java
@@ -1284,34 +1284,24 @@ public class BasicFrame extends JFrame {
 		////	Handle the document
 		OpenRocketDocument doc = null;
 		try {
-
 			doc = worker.get();
-
 		} catch (ExecutionException e) {
-
 			Throwable cause = e.getCause();
-
 			if (cause instanceof FileNotFoundException) {
-
 				log.warn("File not found", cause);
 				JOptionPane.showMessageDialog(parent,
 						"File not found: " + displayName,
 						"Error opening file", JOptionPane.ERROR_MESSAGE);
 				return null;
-
 			} else if (cause instanceof RocketLoadException) {
-
 				log.warn("Error loading the file", cause);
 				JOptionPane.showMessageDialog(parent,
 						"Unable to open file '" + displayName + "': "
 								+ cause.getMessage(),
 								"Error opening file", JOptionPane.ERROR_MESSAGE);
 				return null;
-
 			} else {
-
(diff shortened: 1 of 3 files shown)


JoePfeiffer:
Let's hold off on this one -- to my surprise, there's actually a reasonable use case for this (changing launch conditions for a bunch of sims).  I'm not sure whether this is the right answer, or if we should be enabling the button.

neilweinstock:
There is absolutely a good use case for multi-editing simulations, it is definitely a useful feature.  I think the problem is that the current implementation is kind of half-baked, and inconsistent.  Given that it has always been there, though, it is reasonable to leave it as-is until we can fully bake it, so we don't remove functionality that people are making use of (even if it is janky).

JoePfeiffer:
For the sake of consistency, we should be enabling the button in this case.

SiboVG:
I'm confused, in which scenarios does multi-selecting sim editing work? I did some quick testing but did not find consistent behavior. E.g. multi-editing launch site conditions does **not** work.

JoePfeiffer:
On the Simple Model Rocket example, I just selected all the simulations and used to ^E to get into the simlation editor.  I modified wind speed, latitude, launch rod length, and geodetic calculations; I also added the Damping Moment Coefficient simulation extension.  The only one from that set that didn't get reflected in all the sims was the geodetic calculations. I tried this both with 22.02 and current unstable.

I'm not sure why you're seeing something different...

SiboVG:
Using OR 22.02:

https://user-images.githubusercontent.com/11031519/228303327-50224e7e-b6f0-401f-9d52-b56ea3d59bcc.mp4

SiboVG:
Okay, I see what's going on, the multi-sim editing only works when you click the "Close" button in the edit dialog. If you close the dialog using the close button on the top-left of the dialog, or by pressing the escape key, multi-sim editing is broken.

SiboVG:
> On the Simple Model Rocket example, I just selected all the simulations and used to ^E to get into the simlation editor. I modified wind speed, latitude, launch rod length, and geodetic calculations; I also added the Damping Moment Coefficient simulation extension. The only one from that set that didn't get reflected in all the sims was the geodetic calculations. I tried this both with 22.02 and current unstable.

The checkboxes for "Always launch directly up-wind" and "Use ISA" also don't get applied.

PR #2149, opened 2023-03-28
[#2148] Apply sim changes when window closes & [#1826] Fill in gaps in multi-sim editing

### LINKED ISSUE #1826: [Feature request] Gaps in multi-sim editing functionality
It would be useful to allow multi-component editing of simulations, so you can e.g. edit the launch site parameters for all simulations simultaneously.

### LINKED ISSUE #2148: [Bug] Multi-sim editing only works when clicking the "Close" button
See discussion in #2147. Multi-sim editing only works when you click the "Close" button in the sim edit dialog. When you close the dialog using the x-button in the top-left of the window, or by pressing escape, it doesn't work.

SiboVG:
This PR fixes #2148 and fixes #1826. Now, when closing the edit sim dialog in **any way** (clicking the close button, hitting the escape key, x-button in top-left of the window) applies the simulation options to all selected simulation.

The fix for #1826 is that multi-sim editing had some gaps, namely:
- The checkbox for "Always launch directly up-wind"
- The checkbox for "Use ISA"
- Geodetic calculations

These are now fixed.

### DIFF
--- a/core/src/net/sf/openrocket/simulation/SimulationOptions.java
+++ b/core/src/net/sf/openrocket/simulation/SimulationOptions.java
@@ -442,10 +442,6 @@ public class SimulationOptions implements ChangeSource, Cloneable {
 			isChanged = true;
 			this.launchLongitude = src.launchLongitude;
 		}
-		if (this.launchPressure != src.launchPressure) {
-			isChanged = true;
-			this.launchPressure = src.launchPressure;
-		}
 		if (this.launchRodAngle != src.launchRodAngle) {
 			isChanged = true;
 			this.launchRodAngle = src.launchRodAngle;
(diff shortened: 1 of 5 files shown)


SiboVG:
Oh, and now the edit button for sims is enabled, also when multiple sims are selected (fixes #2140).

hcraigmiller:

I do have one concern. If a change is accidental, there is no way to cancel that change. I would have the esc key cancel the changes, rather than confirm the changes.

OR Build:  1673

SiboVG:
> I do have one concern. If a change is accidental, there is no way to cancel that change. I would have the esc key cancel the changes, rather than confirm the changes.

I get your point, but that would be counter-intuitive to the workflow of component editing, where pressing the escape key also applies the changes. I'm also very much used to this workflow from other software, as I don't always want to move my mouse over to the close button and click it. Pressing the escape key just accelerates my workflow + it's aligned to the component edit workflow.

neilweinstock:
Ultimately we could have a cancel/OK button like in the component editor.

hcraigmiller:
I do think something is needed to cancel an accidental change, perhaps as @neilweinstock suggested.

PR #2156, opened 2023-03-29
[#2142] Set length setting as first item

### LINKED ISSUE #2142: Body Tube and Inner Tube config editors are not consistent
image

On Body Tube, "length" appears on top (correct), while on Inner Tube, length appears at bottom (incorrect).

Also, is there a reason Inner Tube doesn't have a "filled" option for wall thickness?  It seems no more or less likely to need this than Body Tube, feels like they ought to be the same, no?

The one clear difference that seems correct is that Body Tube has an "automatic" option for diameter and Inner Tube does not, because it's not applicable.  everything else seems like it ought to be the same.

SiboVG:
This PR fixes #2142.

image

### DIFF
--- a/swing/src/net/sf/openrocket/gui/configdialog/BulkheadConfig.java
+++ b/swing/src/net/sf/openrocket/gui/configdialog/BulkheadConfig.java
@@ -21,8 +21,7 @@ public class BulkheadConfig extends RingComponentConfig {
 		
 		JPanel tab;
 		
-		tab = generalTab(trans.get("BulkheadCfg.tab.Diameter"), null, null,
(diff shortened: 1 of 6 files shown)


hcraigmiller:
This should also be done for the tube coupler and engine block.

With that caveat, functions as expected with no anomalous behavior found.

OR Build: 1676

SiboVG:
> This should also be done for the tube coupler and engine block.

Thanks for noticing. The length is now the first property for tube couplers, engine blocks, bulkheads (thickness ~ length) and centering rings (thickness ~ length).

neilweinstock:
Excellent.

hcraigmiller:

OR Build: 1678

Issue #1391, opened 2022-05-30
Feature request: ability to undo simulation deletion


SiboVG:
For a lot of important actions in OR, you can undo it afterwards (ctrl+z), but not for a simulation deletion. Right now there's simply a warning dialog stating 'Watch out, this action is irreversible'. It could save users some frustration by implementing simulation deletion undoing.

PR #2163, opened 2023-03-30
[#2158] Add ok/cancel button for simulation editing

### LINKED ISSUE #2158: [Feature request] Ok/Cancel button for simulation editing
Same as #960, but for simulation editing, see discussion in #2149.

So, instead of only a Close-button in the sim editing window, there should be an OK-button to apply the changes, and a Cancel-button to discard them.

SiboVG:
This PR is a continuation of #2149, so merge that first. It fixes #2158 by adding an ok/cancel button in the simulation edit dialog, similar to what is present for component editing.

Changes:
- The edit simulation dialog will now get an asterisk symbol once a modification has been made.
- Clicking the "Cancel" button in the edit sim dialog will pop up a confirmation dialog on whether to discard the changes. Just like with component editing, you can choose to not show that dialog again by changing a setting in the preferences.
- Clicking the "Ok" button is the usual behavior.

Demo:

https://user-images.githubusercontent.com/11031519/228711972-f438efa8-b216-45ee-a867-1ce12253a572.mp4

NOTE: this PR could be improved upon once #1391 is implemented.


hcraigmiller:
The cancel button applies the changes.

SiboVG:
> The cancel button applies the (1) change to the launch rod length; (2) always launch up or down wind check box.

Fixed now. This was actually a pre-existing bug, where setting the launch rod length or launch into wind boolean did not fire a change event.

PR #2174, opened 2023-04-03
Add preferences exporting & importing

SiboVG:
This PR adds preferences exporting & importing, to accommodate for teachers who want their students to have pre-determined settings, such as preset types (as discussed on [TRF](https://discord.com/channels/1073297014814691328/1078391456512872568/1091340170466439188)).

There is now an export and import preferences button in the general tab of the preferences dialog. The preferences are stored in an XML file. NOTE: a preference only shows up in the XML file when it has been changed by the user to a non-default value.

You can choose to enable/disable user directory exporting (for privacy reasons - is disabled by default) and to enable/disable window size/position/... data exporting (not useful when exporting for students on different machines and set-ups).

Demo:

https://user-images.githubusercontent.com/11031519/229642565-0a344224-13c8-4415-83c6-9c93edc009f1.mp4


hcraigmiller:
Functionality question: Should OR need to be closed and reopened for the imported preferences to take effect?

SiboVG:
> Functionality question: Should OR need to be closed and reopened for the imported preferences to take effect?

No, the preferences dialog is closed and then re-opened and that should normally apply the changes.

hcraigmiller:

One comment related to:
> NOTE: a preference only shows up in the XML file when it has been changed by the user to a non-default value.

After extensive experimentation with this feature, I'm not sure this is the right approach. If I want to create several export files with various settings, I have to change a specific default settings and then change it back in order to ensure that an import will set that characteristic. For example, (1) I set the launch field altitude to 1000 and export the settings, then (2) I set the altitude to 2000 and the geodetic calculation to flat Earth and export. I then import (1) and expect the geodetic calculation to be spherical (what is was when (1) was exported) but it remains flat Earth because the default (spherical) was not saved.

The same applies to wind settings and other default settings. I believe that **all of the settings need to be exported, _changed or not_**. I think doing otherwise will potentially cause a great deal of confusion and potential inadvertent user usage errors.

OR Build: 1704

SiboVG:
Well tbh this behavior isn't intentional, it's a limitation of how the preferences stuff is programmed. I'll see if I can alleviate it somewhat.

neilweinstock:
I agree with @hcraigmiller that all settings should be exported if at all possible.

SiboVG:
> I agree with @hcraigmiller that all settings should be exported if at all possible.

Saving all preferences will not be possible, but I'll see what I can do.

SiboVG:
I edited the code so that now most preferences are stored by default. I *think* there are some hidden, unimportant preferences that may not be included, since the preferences are only stored once you fire a get-preference call. This call is fired once the UI-element of that preference is opened. For most preferences that's the case, but I have a feeling there's still some small preferences that are not stored at start-up (but may be later, when you start using OR for some time).

Anyway, long story short: practically all preferences are stored now.

I also added a warning pop-up after you import the preferences stating you may need to restart OR for some of the changes to take effect. This is not the case for every preference, but for some, it's necessary.

hcraigmiller:
I think a Wiki comment may be needed. If you reset the preferences, then export with no changes, virtually no preferences are saved. But, if even one preference is changed, then all of the material preference values (as near as I can tell) are exported. Works well, and is going to make simulating multiple frequently used launch sites much, much easier.

OR Build: 1710

SiboVG:
> I think a Wiki comment may be needed. If you reset the preferences, then export with no changes, virtually no preferences are saved. But, if even one preference is changed, then all of the material preference values (as near as I can tell) are exported. Works well, and is going to make simulating multiple frequently used launch sites much, much easier.

Just pushed a commit that refreshes the preferences dialog after resetting the preferences, which solves that issue for the most part. It also forces the preference updates to be visible in the preferences dialog, which is nice. I also added the same warning message to reseting the preferences about may needing to re-open OR for the changes to take effect.

hcraigmiller:

OR Build: 1712

PR #2176, opened 2023-04-04
[#2096] Update values after auto is set

### LINKED ISSUE #2096: When unchecking "automatic" in a component editor, retain the existing value
Imagine a nose cone with base diameter set to "automatic", and the value shown is 1" because that's the size of the body tube behind it.

If I uncheck "automatic", I would like the value to stay at 1", rather than revert to the default or whatever value had been previously entered manually.  Sometimes it is useful to use the "automatic" feature as a shortcut for entering the correct value; but if you lose the value when it's unchecked then it doesn't work.  In my experience there has never been an occasion when it has been useful for the value to revert back to the previous, which is usually the default of 1.9" or whatever, a completely useless value for an LPR flier.

So, to be absolutely clear, if I uncheck "automatic" on that nose cone, I want the base diameter to stay at 1".  Of course at that point I can always change it if I like.

SiboVG:
This PR fixes #2096. After checking the "Automatic" checkbox on any rocket component setting, that setting will remain saved, even after unchecking the 'Automatic" checkbox.

### DIFF
--- a/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
+++ b/core/src/net/sf/openrocket/file/openrocket/importt/DocumentConfig.java
@@ -389,7 +389,7 @@ class DocumentConfig {
 		
 		// MassObject
 		setters.put("MassObject:packedlength", new DoubleSetter(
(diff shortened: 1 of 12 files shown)


hcraigmiller:

OR Build: 1707

PR #2189, opened 2023-04-11
[#1147] Implement RASAero exporting

### LINKED ISSUE #1147: [Feature request] Export to RASAero II file format
The RASAero II file format is very similar to that of the current unstable, though the variable names are different, and distances are measured in inches, not meters. It appears that all that is needed is a translator that (1) changes field names, (2) corrects numerical values, if needed, and (3) removes excess data fields.

See also, Issue #875, Import from RASAero II file format.

SiboVG:
This PR fixes #1147 and implements exporting OpenRocket designs to RASAero designs. Includes both rocket component exporting and simulation exporting (including motors).

Special thanks to @hcraigmiller for rigorous testing and providing feedback.


SiboVG:
A small note: because RASAero has limited design freedom, you will notice that exporting an OpenRocket design can throw a fair amount of errors (not bugs, but errors displayed in an error-warning dialog, similar to the warning dialog when importing a RockSim design). This approach forces the user to adjust the design, rather than letting OpenRocket make bold assumptions about the design which could result in discontinuities between the OpenRocket design and the RASAero design. So, we use a safer approach.

SiboVG:
A small note: because RASAero has limited design freedom, you will notice that exporting an OpenRocket design can throw a fair amount of errors (not bugs, but errors displayed in an error-warning dialog, similar to the warning dialog when importing a RockSim design). This approach forces the user to adjust the design, rather than letting OpenRocket make bold assumptions about the design which could result in discontinuities between the OpenRocket design and the RASAero design. So, we use a safer approach.

hcraigmiller:
I found one more issue. The body tube consolidation is not accounting for distance from the tip of the rocket (or the from of the consolidated airframe = from front of the parent component plus the length of all more forward components in the stage airframe).

Booster 2 Body tube consolidation

SiboVG:
> I found one more issue. The body tube consolidation is not accounting for the distance from the tip of the rocket (or from the top of the consolidated airframe = from front of the parent component plus the length of all more forward components in the stage airframe).
> 
> Booster 2 Body tube consolidation

Should be fixed now.

hcraigmiller:

OR Build: 1719

PR #2191, opened 2023-04-12
[#2083] Account for units of measure in material database loader

### LINKED ISSUE #2083: [Bug]  UnitsOfMeasure for materials in .orc files has no effect
Code analysis and functional testing shows that the UnitsOfMeasure attribute for <Material> has no effect in .orc files.

No matter what you specify, the UoM is always interpreted as the SI unit (kg/m3, kg/m2, kg/m).

Looking at some code history suggests this has been true for a very long time, perhaps since the beginning of OR.

To reproduce:  Change the UnitsOfMeasure on some <Material> entry in a .orc file in the parts database and observe that the computed mass of parts made from that material don't change.

Expected:  Units of measure will be interpreted as specified.

Note: Fixing this will break old versions of the parts database prior to 1.0.0.3

SiboVG:
This PR fixes #2083 and accounts for the units of measure in the material object in a .orc file. The density parameter in the material XML object is now relative to the specified UnitsOfMeasure attribute of the material. E.g.:
```XML
<Material UnitsOfMeasure="g/dm3">
  <Name>Test, bulk</Name>
  <Density>894.4</Density>
  <Type>BULK</Type>
</Material>
```

Here, the density is `894.4 g/dm3`. If the bulk density units are set to `g/cm3` in OR, then the material will display as `Test, bulk (0.894 g/cm3)`.

Previously, the density value always needed to be formatted using SI units, so kg/m3 for bulk materials.

I haven't found materials that use a non-SI units of measure, but at least @dbcook's database should not be affected, as we use the latest version of the component database.


SiboVG:
There is an issue in the rocketarium component database, pushed a fix for it [here](https://github.com/dbcook/openrocket-database/pull/32). But this causes this PR to throw an exception upon startup.

SiboVG:
> There is an issue in the rocketarium component database, pushed a fix for it [here](https://github.com/dbcook/openrocket-database/pull/32). But this causes this PR to throw an exception upon startup.

Fixed now.

PR #2201, opened 2023-04-24
[Enhancement] Add better warning that data will be plotted in time order

### LINKED ISSUE #2161: [Enhancement] Better warning that data will be plotted in time order, regardless of the X axis type
The plot dialog contains a warning that all data will be plotted in time order, regardless of the selected X axis type, but many people (including myself) never noticed this.
image

We should make this more clear. I propose to remove the text box stating the warning and instead when an X axis type different from time is selected, to show a dark-red warning text next to the X axis type selector (or somewhere else). Not only will the dark red color give it more attention, but the fact that it suddenly appears and only appears when the X axis type is not set to time should give users a better notion of the concept.
Screenshot 2023-03-30 at 04 01 02

MayNc10:
Fixes #2161 
The plot dialog warns that all data is plotted in time order regardless of X axis type, but this warning is easy to miss.
The message was removed by default and on changing to an X axis type other than time, a red warning will appear.

[Here is a jar file for testing](https://drive.google.com/file/d/1YJp7lEhRgSuI3k13kbFqVgx6vLZBWWz-/view?usp=share_link)

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -767,8 +767,6 @@ simplotpanel.but.NewYaxisplottype = New Y axis plot type
 simplotpanel.lbl.Axis = Axis:
 simplotpanel.but.ttip.Deletethisplot = Delete this plot
 simplotpanel.Desc = The data will be plotted in time order even if the X axis type is not time.
-! Fix this name
-simplotpanel.Warning = The data is plotted in time order even though the X axis type is not time. 
 simplotpanel.OptionPane.lbl1 = A maximum of 15 plots is allowed.
 simplotpanel.OptionPane.lbl2 = Cannot add plot
 simplotpanel.AUTO_NAME = Auto
@@ -1506,6 +1504,7 @@ TCMotorSelPan.btn.close = Close
 ! PlotDialog
 PlotDialog.CheckBox.Showdatapoints = Show data points
 PlotDialog.lbl.Chart = left click drag to zoom area. mouse wheel to zoom. ctrl-mouse wheel to zoom x axis only. ctrl-left click drag to pan.  right click drag to zoom dynamically.
+PlotDialog.lbl.timeSeriesWarning = The data is plotted in time order even though the X axis type is not time. 
 PlotDialog.btn.exportImage = Export Image
 
 ComponentTree.ttip.massoverride = mass overriden
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -767,6 +767,8 @@ simplotpanel.but.NewYaxisplottype = New Y axis plot type
 simplotpanel.lbl.Axis = Axis:
 simplotpanel.but.ttip.Deletethisplot = Delete this plot
 simplotpanel.Desc = The data will be plotted in time order even if the X axis type is not time.
+! Fix this name
+simplotpanel.Warning = The data is plotted in time order even though the X axis type is not time. 
 simplotpanel.OptionPane.lbl1 = A maximum of 15 plots is allowed.
 simplotpanel.OptionPane.lbl2 = Cannot add plot
 simplotpanel.AUTO_NAME = Auto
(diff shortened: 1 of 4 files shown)


SiboVG:
Thanks for your PR! I would make one adjustment though: use `net.sf.openrocket.util.Color.DARK_RED.toAWTColor()` for the text color of the text area. That way, it will be even more visible, especially when users re-open the plot window so that they're reminded of it.

I was also thinking to maybe add a warning text in the actual plot window, your thoughts?
Screenshot 2023-04-24 at 23 42 04

neilweinstock:
Nice change.

MayNc10:
I added the warning message to the plot, if that's not wanted I can remove that commit. 
If we decide to add that message, we have to add a new message to the translation dialogs, since the message on the plot is a bit different than the one on the setup screen.

swing/src/net/sf/openrocket/gui/components/DescriptionArea.java
```diff
@@ -202,5 +202,15 @@ public void setTextFont(Font font) {
...
 	}
+
+	public void setBackground(Color color) {
+		if (editorPane == null) return;
+		editorPane.setBackground(color);
+	}
+
+	public void setForeground(Color color) {
+		if (editorPane == null) return;
+		editorPane.setForeground(color);
```
SiboVG, line 213: That doesn't really work on my machine (maybe it does on yours)?

Change this to:
```Java
public void setForeground(Color color) {
	editorPane.setForeground(color);
	StyledDocument styledDocument = (StyledDocument) editorPane.getDocument();
	SimpleAttributeSet attributeSet = new SimpleAttributeSet();
	StyleConstants.setForeground(attributeSet, color);
	styledDocument.setCharacterAttributes(0, styledDocument.getLength(), attributeSet, false);
}
```

and change `setText` to:
```Java
public void setText(String txt) {
	editorPane.setText(txt);
	editorPane.revalidate();
	SwingUtilities.invokeLater(new Runnable() {
		
		@Override
		public void run() {
			editorPane.scrollRectToVisible(new Rectangle(0, 0, 1, 1));
		}
		
	});
	setForeground(editorPane.getForeground());
	editorPane.scrollRectToVisible(new Rectangle(0, 0, 1, 1));
}
```

Also: the `if (editorPane == null) return;` condition is unnecessary. It's good that you try to be on the safe side, but in this scenario, it's just an unnecessary calculation.

SiboVG (changes requested):

swing/src/net/sf/openrocket/gui/components/DescriptionArea.java
MayNc10: I kept the null condition check because on my machine, if I don't keep it my program crashes when I open a rocket model.

MayNc10:
I have made the changes requested in the review. However, I kept the null check, as without it the program on my machine would crash when opening a rocket model.

[Here is an updated jar file for testing](https://drive.google.com/file/d/1D6JE5sLWhO5EmQIyxpdiKFArqYrBp6F3/view?usp=sharing)

core/resources/l10n/messages.properties
```diff
@@ -767,6 +767,8 @@ simplotpanel.but.NewYaxisplottype = New Y axis plot type
 simplotpanel.lbl.Axis = Axis:
 simplotpanel.but.ttip.Deletethisplot = Delete this plot
 simplotpanel.Desc = The data will be plotted in time order even if the X axis type is not time.
+! Fix this name
+simplotpanel.Warning = The data is plotted in time order even though the X axis type is not time.
```
SiboVG, line 771: You'd have to excuse me, my initial comment to put the new translation key was not very good. It's better to look for the keys `PlotDialog` and add a new key, for instance `PlotDialog.lbl.timeSeriesWarning`

SiboVG (changes requested):

SiboVG:
Your changes look good, thanks! Just added another comment to move the new translation key, I initially gave you bad directions, my bad...

MayNc10:
Fixed!

[Here is an updated JAR file for testing](https://drive.google.com/file/d/1TmCZ3KS1GGTxKKIhAQhsefljZLoTgTeI/view?usp=sharing)

SiboVG (approved):

SiboVG:
LGTM, thanks a bunch for your changes!

PR #2208, opened 2023-04-30
Modify swing/build.xml to include jogl jars individually

JoePfeiffer:
It appears as if jpype (see https://github.com/jpype-project/jpype ) is unable to resolve libraries when an entire directory is included in a .jar.  Including the jogl libraries individually instead appears to fix the problem.

Under Linux AMD64, OR appears to run equally well with this change to the build script.  I will appreciate people checking to see that this doesn't cause regressions under other platforms and architectures. @hcraigmiller  @SiboVG @neilweinstock

### DIFF
--- a/swing/build.xml
+++ b/swing/build.xml
@@ -84,7 +84,6 @@
 			<manifest>
 				<attribute name="Main-Class" value="${main-class}" />
 				<attribute name="SplashScreen-Image" value="pix/splashscreen.png" />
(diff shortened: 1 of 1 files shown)


SiboVG:
Causes issues on macOS with Java 17, which means that it will cause issues with making an ARM build of OR for macOS.
image

JoePfeiffer:
Darn...

PR #2216, opened 2023-05-16
Downsized subcomponent override icons

### LINKED ISSUE #2188: Overriding subcomponents makes for a very ugly component tree
I appreciate the thought process that went into this last release but IMHO this is really not appealing, as it makes the whole component tree harder to read, and just generally ugly:
image

The reason I care enough to actually file an issue is that overriding mass and CG of an entire rocket is a common and in fact recommended thing to do, and so there will be a lot of finished OR files the look just like the above.  Yuck.

As is so often the case, I'm not sure how to fix it yet.  Some thoughts I had were:
1) make the grey subcomponent override icons *smaller*
2) Put the override icons in their own column at the left
3) Do something else on the left side.  I'll try to mock something up.

neilweinstock:
This PR swaps in smaller subcomponent override icons to fix #2188.


SiboVG:
Hm, seeing it live on my computer, the icons look are a bit too small. Could you try option 2, the slightly larger icons?

neilweinstock:
Screenshot?

neilweinstock:
I uploaded the middle-sized icons.  I still like the smaller ones best.

SiboVG:
Thanks for the change. The smallest icons just looked a bit illegible IMO.

PR #2053, opened 2023-02-14
[#2040] Fix CM of launch lug & rail button instances

SiboVG:
This PR fixes #2040. The CM (mass+CG) of a rail button and launch lug now accounts for multiple instances.

While working on this issue, I also noticed that the CG of the launch lug was incorrectly calculated. Here is a design with a body tube's diameter and length set to 0:
image

You can see that CG.y is not in the center of the launch lug. The same issue was present for CG.z.

Behavior from this PR:
image

PR #2219, opened 2023-05-23
[#2217] Fix mass issues for fin sets (+ CM of launch lug and rail buttons)

### LINKED ISSUE #2217: Major problem in tube fin set mass calculation
See attached ORK file:

- The body tube is 15" long.  Calculated mass is .827 oz.
- The tube fins are a set of 5, each 3" long.  Calculated mass is 4.3 oz.
- Both are based on the same library component (BMS T60-34).

There seems to be an extra multiplication by the number of fins (i.e. the mass of one fin is getting multiplied by 5 *twice*).  Pretty much confirmed this by experimenting with the number of fins in the set.

[tube_fins_mass.zip](https://github.com/openrocket/openrocket/files/11546443/tube_fins_mass.zip)

SiboVG:
This PR fixes #2217. The was caused by PR #2053 - I didn't realize that componentVolume was that of all the instances instead of a single instance.

Additionally, I noticed an issues with launch lugs and rial buttons where the y and z location of the CG wasn't right. If you created a new design with just a body tube and rail button/launch lug and overdid the mass of the body tube to 0, the CG should be located solely at the rail button/launch lug, which wasn't the case. This is now fixed.

### DIFF
--- a/core/src/net/sf/openrocket/rocketcomponent/ExternalComponent.java
+++ b/core/src/net/sf/openrocket/rocketcomponent/ExternalComponent.java
@@ -102,7 +102,7 @@ public abstract class ExternalComponent extends RocketComponent {
 	 */
 	@Override
 	public double getComponentMass() {
-		return material.getDensity() * getComponentVolume() * getInstanceCount();
+		return material.getDensity() * getComponentVolume();
 	}
 	
 	/**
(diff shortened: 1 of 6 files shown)


hcraigmiller:

OR Build: 1760

PR #2220, opened 2023-05-23
[#2079] Display secondary stability unit

### LINKED ISSUE #2079: [Feature Request] Add option to show stability in cals and as a percentage
I find my self frequently switching between stability calibers and the stability as a percentage.

Could there be an option in the units preferences to show cals and % at the same time in the rocket design tab?

image

SiboVG:
This PR fixes #2079 and adds the stability in percentage length in addition to the stability value in the currently selected stability unit. If the selected unit is already the percentage, or the stability value is N/A, the stability percentage is not included.

https://github.com/openrocket/openrocket/assets/11031519/0222cefd-01eb-49aa-ae48-0b6cb8399227


neilweinstock:
Two things:
1) I don't think there should be a space between the number and the percent sign.  You don't say 100 %, you say 100%.
2) It seems weird to me that if you select Calibers, it also shows percentage.  But if you select Percentage, it does not show caliber.  Why does it work this way?

SiboVG:
> Two things:
> 
> 1. I don't think there should be a space between the number and the percent sign.  You don't say 100 %, you say 100%.
> 2. It seems weird to me that if you select Calibers, it also shows percentage.  But if you select Percentage, it does not show caliber.  Why does it work this way?

1. Well, in theory, there should always be a space between the value and the unit. People are used to writing the percentage sign without a space, but according to scientific notation, the space is correct. Also, how it's currently displayed is just standard unit formatting of OR, it would require a bit more (ugly) coding to change it to exclude the space.
2. Idk, but this made me think that we should maybe add a "Stability secondary" unit in the preferences so users can choose which extra stability value to show, instead of having it be fixed to the percentage unit?

neilweinstock:
1. OK, but a space before the percent sign will not stop looking weird/wrong to me.  If it takes some new weird code to remove it, though, then forget it for now.
2. Simplicity is key.  One option would be to always show both, but let the user select which one comes first via the preferences.  E.g., "2.0 cal (15 %)" vs. "15 % (2.0 cal)", so they can choose which one is "primary", but we always show both.  Is there any significant downside to always showing both?  It certainly wouldn't bother me, and given how difficult it is to switch back and forth by going into preferences, leaving both visible all the time seems like a good idea to me, even with the extra space. :)

neilweinstock:
Another alternative: instead of the parentheses, display like this:   "2.0 cal / 15 %"  This way, neither is really primary or secondary.  You always show both, and remove the preference entirely.

neilweinstock:
For what it's worth: https://www.rocketryforum.com/threads/announcement-openrocket-version-22-02-final-is-now-available-for-download.177835/post-2437626

neilweinstock:
OK, I yield.  Let's have a primary and secondary preference, as you suggested.  There should be an option on secondary to disable it, if reasonably possible.

I am thinking that slash-separating the two values might be better than putting the second one in parens.  What do you think?

SiboVG:
> OK, I yield. Let's have a primary and secondary preference, as you suggested. There should be an option on secondary to disable it, if reasonably possible.

One possibility is to disable the secondary unit when it is the same as the primary unit, but that is probably not straight-forward to users. Probably better to add a checkbox under the secondary stability unit preference. I would actually implement both; hide the secondary unit when the checkbox is unchecked in the preferences **or** the primary and secondary unit are equal.

> I am thinking that slash-separating the two values might be better than putting the second one in parens. What do you think?

When using the secondary preference idea, we should indeed use a slash.

neilweinstock:
> One possibility is to disable the secondary unit when it is the same as the primary unit, but that is probably not straight-forward to users. Probably better to add a checkbox under the secondary stability unit preference. I would actually implement both; hide the secondary unit when the checkbox is unchecked in the preferences **or** the primary and secondary unit are equal.

That sounds perfect.

hcraigmiller:
Functions as described, with no anomalous behavior found.

OR Build: 1761

SiboVG:
Secondary stability unit can now be chosen in the preferences + can be disabled.

https://github.com/openrocket/openrocket/assets/11031519/d945fbe8-d3a0-441a-898a-73128853e59e

neilweinstock:
Looks great.

hcraigmiller:
Very nice.

OR Build: 1765

PR #2236, opened 2023-06-27
[#2225] Copy simulation data from sustainer to other stages for CSV exporting

### LINKED ISSUE #2225: [Enhancement] Include data before stage separation in sim export
Update: this is not a bug as described in this comment. See comments below for more info. What the enhancement should be is to include the sustainer data before separation (which is common to all stages) in the sim export.

### Describe the bug

Reported by user david314 on our Discord server:
I found a bug, when exporting a multi-stage rocket simulation, the lower stage has no data until separation.

### To Reproduce

1. Open 'Three-stage rocket' example design
2. Go to the simulations tab
3. Export one of the simulations
4. Select export variables, e.g. export time and altitude
5. Select one of the bottom stages (e.g. booster 2)
6. Export
7. You'll see that the exported CSV starts its time at the stage separation time of the exported stage

#### Screenshot(s):
Example export of the three-stage rocket example design.
Sustainer export:
image

Booster 1 export:
image

Booster 2 export:
image

### OpenRocket version

22.02, unstable

macOS

### Additional context

_No response_

SiboVG:
This PR fixes #2225. Common data from the sustainer stage is now copied to lower stages.

For instance, export of 'Three-stage rocket' booster1 before this PR:
image

And after:
image

---

**NOTE**: Whilst working on this PR, I noticed a code issue. When simulation results were loaded from the .ork file, the source object of each event type was not loaded. The event source was simply not stored in the .ork file. To fix this, I modified the .ork file format to include the ID of each component, and then used that same ID in a newly added 'source' attribute for the flight event section.

So, an example of saved rocket component content:
```XML
<stage>
  <name>Sustainer</name>
  <id>38559a9d-45ba-42c8-9291-68d45875ba4b</id>
```
and a corresponding event:
```XML
<event time="7.446" type="ejectioncharge" source="38559a9d-45ba-42c8-9291-68d45875ba4b"/>
```

The reason I'm telling this is that copying flight events from the sustainer data will not work for simulations saved before this PR. You'd have to **rerun the sim first**. If you then save the file and re-open it, everything should work fine, because then the changes from this PR are applied.


JoePfeiffer:
Would it make sense to accomplish this by modifying the FlightDataBranch constructor so it copies the entire parent data branch, instead of just the last datapoint?  That would also simplify SimulationPlot, since it wouldn't be necessary to copy data from the sustainer branch to the other branches when plotting.

SiboVG:
> Would it make sense to accomplish this by modifying the FlightDataBranch constructor so it copies the entire parent data branch, instead of just the last datapoint? That would also simplify SimulationPlot, since it wouldn't be necessary to copy data from the sustainer branch to the other branches when plotting.

Good point, I'll try to do that instead.

SiboVG:
As per Joe's suggestion, the flight data is now copied in the FlightDataBranch constructor, thus eliminating my patchwork code, and the one in the simulation plotting.

PR #2237, opened 2023-06-27
[#1629] Show calculated mass, CG & CD in override tab if not overridden

### LINKED ISSUE #1629: Show calculated Mass, CG, and CD values for non-overridden components
In the Override tab, if a component isn't overridden then an override value of 0 is shown for all three values unless changed by the user. Showing the calculated values would give the user a starting point for overrides.

SiboVG:
This PR fixes #1629. When the mass, CG or CD of a component is not overridden, the override values will be the calculated values.

Demo:

https://github.com/openrocket/openrocket/assets/11031519/4f1ead16-5b26-4c78-96e0-189f8b773177

Note: when toggling the override CD the simulation results can slightly change. This is because in the overridden state, the individual CD components (pressure, base & friction CD) are set to 0 and only the total CD is decisive, so I think that's the reason for the very small difference.

---

Additionally, I also improved the formatting of the component analysis stability, drag characteristics and roll dynamics tables. The number formatting is now more standardized (better decimal behavior and decimal separator of locale) + when selecting a row, the row now also gets highlighted. You always had the ability to select a row and then copy the values of that row with ctr/cmd + c, but this wasn't visually clear. Finally, in the roll dynamics table, the first row isn't in bold anymore (there was no reason for that - just a consequence of bad table formatting), and I changed the "Total" label to "Total (Rocket)" to make it clear that e.g. the total CD is the CD of the rocket.

### DIFF
--- a/core/resources/l10n/messages.properties
+++ b/core/resources/l10n/messages.properties
@@ -860,7 +860,7 @@ componentanalysisdlg.lbl.reflenght = Reference length:
 componentanalysisdlg.lbl.refarea = Reference area:
 !componentanalysisdlg.But.close =Close
 componentanalysisdlg.TabStability.Col.Component = Component
-componentanalysisdlg.TOTAL = Total
+componentanalysisdlg.TOTAL = Total (Rocket)
 componentanalysisdlg.noWarnings = <html><i><font color=\"gray\">No warnings.</font></i>
 
 ! Custom Material dialog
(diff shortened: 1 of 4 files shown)


neilweinstock:
This looks very nice.  The only minor downside (at least, based on the video grab above, I haven't had a chance to try it yet) is that you can't toggle the override on and off, since it appears to revert to the calculated value whenever you disable it.

I'm trying to think of whether this actually matters; I feel like I've used the ability to toggle override on and off in the past but for the life of me I can't think of what the scenario was.  Most likely it is unimportant and this implementation will be perfectly fine (or really, better than fine).

I'll keep thinking if there's some sort of use case that needs to be accommodated here, but if I can't come up with it then we can just stay with this.

PR #2265, opened 2023-07-21
[#2252] Support for canted fins in fin marking guide

### LINKED ISSUE #2252: [Bug] Fin marking guide doesn't account for canted fins
### Describe the bug

If the "Print design info" menu item is used to print a fin marking guide, the fins are shown as straight regardless of whether they should be canted.

### To Reproduce

1. create a new design
2. add a body tube
3. add canted fins
4. print a body tube marking guide

### Screenshots / .ork file

guide

### OpenRocket version

unstable

Linux

### Additional context

_No response_

SiboVG:
This PR fixes #2252 and supports canted fins in the fin marking guide. I also changed the front indication drawing (which is drawn for instance when you have a launch lug and fins in your design) from a simply vertical text to the right of the marking to a tab on the bottom of the marking guide. It now looks cleaner than the Front text overlapping with the marking arrows.

The finished result for a 15° canted 3-fin design looks like this:
image

An analysis:
255275575-06814a05-e92e-4729-adec-9ebea3710107

So, there are 4 elements:
1. Double-arrow indicates how the fin is rotated.
2. The left side of the arrow indicates the location of the fin aft end. Note that the right side of the arrow does not indicate the true location of the fin fore end. The true fin fore end is located somewhere along the double arrow.
3. A helper dashed line is drawn to mark the fin aft end. It has similar functionality as the double arrow for a non-canted fin, namely to help you with the radial positioning of the fins. But now instead of positioning your entire fin to the line you drew on the body tube using the marking guide, you position the fin aft end to that line (see element no. 2).
4. The center of the fin is marked by a light-grey cross (in case the user wants to know this).

Here is another example that hopefully clears things up some more:
image


SiboVG:
Of course I'd like some feedback from actual rocketeers on whether this workflow makes sense (@neilweinstock, @JoePfeiffer, @hcraigmiller).

JoePfeiffer:
If I were to build a rocket with canted fins (I never have), I'd want the fore end marking to also be at the true fore end of the fin.  The way I'd use it would be to wrap it around the body tube, and mark the fore and aft locations for each fin.  Then take the template off, and use those marks to position the fins.

SiboVG:
> If I were to build a rocket with canted fins (I never have), I'd want the fore end marking to also be at the true fore end of the fin. The way I'd use it would be to wrap it around the body tube, and mark the fore and aft locations for each fin. Then take the template off, and use those marks to position the fins.

I agree that would be a logical workflow, but the issue is that the width of the marking guide would then be `fin_length * cos(cantAngle)`. So that'd be quite small for smaller rockets and the text size would have to be adjusted etc.

JoePfeiffer:
> > If I were to build a rocket with canted fins (I never have), I'd want the fore end marking to also be at the true fore end of the fin. The way I'd use it would be to wrap it around the body tube, and mark the fore and aft locations for each fin. Then take the template off, and use those marks to position the fins.
> 
> I agree that would be a logical workflow, but the issue is that the width of the marking guide would then be `fin_length * cos(cantAngle)`. So that'd be quite small for smaller rockets and the text size would have to be adjusted etc.

Maybe put a line at the fin fore end, so users could cut there if they preferred?

SiboVG:
> Maybe put a line at the fin fore end, so users could cut there if they preferred?

Hm, maybe this clutters the guide too much?
image

I guess we need our UI expert @neilweinstock 😉.

JoePfeiffer:
Yeah, I see the problem with clutter.  It's just that marking the true fore and aft fin locations makes it so much more useful...

neilweinstock:
The canted marking guide is a nice touch. 

Personally, I wouldn’t feel the need for fore or aft edge marks, I’ll handle that myself. 

For the real cherry on top, though, apply the necessary (very slight) curve to the fin roots in the fin templates. :)

JoePfeiffer:
> The canted marking guide is a nice touch.
> 
> Personally, I wouldn’t feel the need for fore or aft edge marks, I’ll handle that myself.

How do you imagine the canted fin workflow?  My thoughts on how I'd handle it (which are very close to how I do non-canted fins) are above.

> For the real cherry on top, though, apply the necessary (very slight) curve to the fin roots in the fin templates. :)

Oh, that cherry is in @SiboVG 's PR #2243, which has already been merged!

neilweinstock:
LOL, should have guessed.  

As to how I would handle it: mark the canted fin lines. Draw a ring around the rocket to mark fore or aft edges of fins. 

Or, even simpler, just position the marking guide so that the back edge matches the back edge of fins

SiboVG:
So, in conclusion, does this PR need any adjustments?

neilweinstock:
I'm still not quite sure what the "helper lines" are accomplishing?

neilweinstock:
> It's for radial positioning. Let's say you put a launch lug at one end of a body tube. You draw the lug position and helper lines. Then, on the other end of the body tube you can use the helper lines to radially position the fin marking guide and correctly position the fin.

I'm not all that convinced they're really all that helpful; they wouldn't be for me.  But you might as well leave them in, they're not hurting anyone.

In which case, no further questions your honor.

JoePfeiffer:
I'd really still prefer the version with the extra line for the true fore end.

SiboVG:
> I'd really still prefer the version with the extra line for the true fore end.

Okay, I'll add it.

SiboVG:
Fin fore end now also marked:
image

neilweinstock:
LOL, I still totally don't get all the additional markings or what I would do with them.  What is the little gray "x"?  What is the vertical dotted line?  When the guide is fully wrapped around the tube neither of those have any exposed bits to enable using them for either marking the tube or aligning with existing marks.  I remain confused.

SiboVG:
> LOL, I still totally don't get all the additional markings or what I would do with them. What is the little gray "x"? What is the vertical dotted line? When the guide is fully wrapped around the tube neither of those have any exposed bits to enable using them for either marking the tube or aligning with existing marks. I remain confused.

The x is informative, in case someone needs it, you never know. I don't think it hurts too much if it's present. The vertical dotted line is not for marking, it's for cutting the marking guide if you only want the guide to span the fin width (a solution to Joe's discussion on having the fore end as well).

neilweinstock:
OK, so the vertical line is only present if the fins are smaller than the marking guide.  In any case, i don't think it's at all clear what it is.  That may not be a reason to leave it out, though.

What are the Xes?

wolsen:
One comment just passing by - but what if the arrows were directional rather than bi-directional. Just having the arrow point towards the fore-end for each of the fin slots would convey the direction as well as be minimally intrusive on the footprint front.

SiboVG:
> One comment just passing by - but what if the arrows were directional rather than bi-directional. Just having the arrow point towards the fore-end for each of the fin slots would convey the direction as well as be minimally intrusive on the footprint front.

Hm, that's a valid point.

PR #2292, opened 2023-08-15
New tube fin icons

neilweinstock:
PNG files for only now, source not yet converted to GIMP format, and there may be some transparency issues.  Just for preview and evaluation.


neilweinstock:
OK, should be ready to try out.  Only tube fin icons have been touched.

JoePfeiffer:
That looks really nice -- huge step up.

neilweinstock:
Do you think I need to have GIMP source files in place before we can merge this?

JoePfeiffer:
What did you use to create it? My first thought is the native format of that tool, whatever it is, should be good enough.

Is there something buried in our license terms about using GIMP?

neilweinstock:
No, I think it's just tradition.

I used Paint.net, which is free but not open source, and Windows-only.  The current source file is a mess, a quick and dirty hack to get where I wanted as quickly as possible.  I could make a clean version of it.

GIMP (despite how much I hate it) does have the benefit of being FOSS and cross-platform.

JoePfeiffer:
> No, I think it's just tradition.
> 
> I used Paint.net, which is free but not open source, and Windows-only. The current source file is a mess, a quick and dirty hack to get where I wanted as quickly as possible. I could make a clean version of it.

Of course I'm not familiar with Paint.net -- how complicated is the source image?  If it's really just the same data as the .png in another format (as opposed to, I don't know, several layers with a tube in each or some sort of vector graphics) I'd say just the .png ought to be fine.

SiboVG:
We'll merge this and add the source files later.

neilweinstock:
Thanks

PR #2294, opened 2023-08-16
Auto-detect UI theme based on OS preference

SiboVG:
This PR adds an Auto option to the UI themes, which auto-detects whether the operating system uses a dark theme or light theme, and uses the corresponding theme for OR.

image


JoePfeiffer:
Seems to work fine under Linux (xfce4 desktop).  Looking at what happened to nearly all my open windows when I changed to a dark theme, is it most common to autodetect by default?  If so, should we?  It'll be a bit jarring for people who've been using a dark theme to suddenly have the new version display dark, but I suspect they'll welcome the change.

SiboVG:
> Seems to work fine under Linux (xfce4 desktop). Looking at what happened to nearly all my open windows when I changed to a dark theme, is it most common to autodetect by default? If so, should we? It'll be a bit jarring for people who've been using a dark theme to suddenly have the new version display dark, but I suspect they'll welcome the change.

Auto dark mode detection is somewhat common in major programs, although they often display some sort of welcome slide show about setting everything up for a first-time user and whether you want to use auto, light or dark theme. But we shouldn't spend our time on something like that...

My main driver for this PR was so people don't overlook that the next release will have a dark mode (not that many people read our release notes extensively...).

PR #2299, opened 2023-08-19
[#2295] Use component color for sticker edge mode

### LINKED ISSUE #2295: [Enhancement] Appearance Sticker repeat mode should use component color for non-texture areas
When the repeat mode of a component appearance is set to "Sticker", I would expect the parts of the component that are not covered by the texture to use the component color. However, currently, this is always black.

image

SiboVG:
This PR fixes #2295.

image

Drawback is that the opacity of the color has no effect, but I already spent too much time trying to fix it, so I'll leave that as is for now.


neilweinstock:
Imperfect fix is still far better than no fix!

PR #2301, opened 2023-08-23
[#604] Implement 3D component exporting to Wavefront OBJ

### LINKED ISSUE #604: 3D Print Export Feature
It would be nice to export the design to a CAD / 3D modeller software so as to be able to 3D print (or eventually cast) some parts.

I have an [on-going code repository for IceSL export](https://github.com/loic-fejoz/openrocket/tree/icesl-export).

For instance, one could 3D print the ogive and a lower tube+fins:

NB: Eventually another export to STL file format could be nice too.

SiboVG:
This PR fixes #604 and introduces the ability for users to export rocket components as Wavefront OBJ 3D objects.

### Why OBJ Format?

- **Editability**: OBJ files are easily editable in 3D software (e.g. Blender), compared to e.g. an STL file. Granted, it's less easy to use in CAS software than STEP-files, but STEP-files are difficult to generate without mesh errors.
- **3D Printing Compatibility**: Most 3D printing slicers directly support OBJ importing. If not, there are a ton of free and easy-to-use converters available to export OBJ files to STL files (which is the standard file format for slicers).

### Key Features

I kinda went all out on the features (been grinding on this for a couple of weeks now...)

- **Rocket Component Export**: All components, including motors, can be exported.
- **Optimized Geometry**: Ensures manifold geometry, correct vertex winding, realistic normals, and simplified editing in 3D software.
    - Manifold geometry is necessary for 3D printing.
    - Correct vertex winding is necessary for correct definition of inside and outside faces.
    - Realistic normals are needed for smooth lighting of the model.
    - Issues like multiple vertices converging at a single point and duplicate vertices are avoided.
- **Geometry Options**: Users can triangulate the geometry or modify offsets relative to the rocket's origin.
- **Vertex Groups**: Each rocket component gets its unique vertex group, allowing easy separation in 3D software and ensuring each part remains distinct.
- **Export Flexibility**: Export components collectively or individually (separate files).
- **Scaling Adjustments**: Default scaling uses SI units (meters). To cater to 3D printing applications, users can shift to mm scaling (set scaling to 1000). The OBJ format lacks standardized units, hence the need for scaling.
- **Export Quality**: Customize your export's detail level: low, normal, or high quality.
- **Easy Accessibility**: Access the export feature via file or context menu. For instance, right-click on a nose cone to export it on its own.
- **Intelligent Exports**: Inactive components are left out.
- **Component Appearance Exports**: Perfect for users aiming for realistic visualizations in 3D software. Enables both color and decal exports. The appearance is stored in an MTL file. Decals from the design file are stored in a unique directory alongside the OBJ. The MTL file has limitations, so some post-processing might be necessary for optimal visuals. I may post a YouTube tutorial for this someday.
   - **Color Consistency**: There's an option to export in sRGB over the conventional linear color space for color consistency in platforms like Blender. Blender uses sRGB, so your colors will look off if you use normal coloring.

---

Some nice photos:

Export panel:
Screenshot 2023-08-23 at 17 39 00

Exported OBJ imported in Blender (after 2 minutes of post-processing the textures):

Screenshot 2023-08-23 at 17 38 31

---

You can choose to show/hide advanced export features:

https://github.com/openrocket/openrocket/assets/11031519/3a0999bd-585b-4491-be38-56a4c5726102

---

The export feature is accessible through the context menu:

Screenshot 2023-08-23 at 17 39 33

---

Even works with pods:

Screenshot 2023-08-23 at 19 32 41

---

Phew, this has been quite the PR, I think I can start my OR retirement now?


neilweinstock:
Wow.

Regrettably I am completely unable to test this in any meaningful way.  We'll need some beta testers.

How do you access this feature *other than* the context menu?   What did you do (if anything) with shoulder shaping, shock cord anchors, etc.?  Leave that to post-editing?

Should the export dialog show somewhere (titlebar or somewhere else near the top) the name and/or type of component you're exporting?

SiboVG:
> Wow.
> 
> Regrettably I am completely unable to test this in any meaningful way. We'll need some beta testers.
> 
> How do you access this feature _other than_ the context menu? What did you do (if anything) with shoulder shaping, shock cord anchors, etc.? Leave that to post-editing?
> 
> Should the export dialog show somewhere (titlebar or somewhere else near the top) the name and/or type of component you're exporting?

As mentioned, you can access it through the file menu (File > Export as > Wavefront OBJ).
image


Shoulders are exported. No idea what shock cord anchors are. Mass components are just exported like you see them in the 3D view.

Hm, idk about showing the components that are to be exported, seems redundant to me. You'll also end up with a long string  when exporting the entire rocket...

neilweinstock:
> Hm, idk about showing the components that are to be exported, seems redundant to me. You'll also end up with a long string when exporting the entire rocket...

If there is a single parent component (possibly with any number of descendents) then I would just name the parent component.

If it is a set of multi-selected components (is this allowed/possible?) then I would just say "multiple components".

neilweinstock:
Argh, stupid "close with comment" button. :(

SiboVG:
> If there is a single parent component (possibly with any number of descendents) then I would just name the parent component.
> 
> If it is a set of multi-selected components (is this allowed/possible?) then I would just say "multiple components".

Fair enough, I'll add it.

SiboVG:
@neilweinstock added the components label. It also dynamically updates when you check "Export children". The tooltip text gives you a list of all the components that will be exported.


https://github.com/openrocket/openrocket/assets/11031519/11112b39-fb7f-4e23-bd9f-d044603780ee


I also added some other minor improvements, such as disabling the "export children" checkbox when there were no children.

neilweinstock:
Excellent.  I'll give it a workout tonight, even if I can't really evaluate the quality of the output.

neilweinstock:
Ah, didn't realize that.  Thanks for the heads-up.

JoePfeiffer:
This is officially COOL. Exported  as .obj, sliced by prusa-slicer, printed in PETG on a slightly modified ANet A8 using octoprint. Wow.  Just wow.

One tiny anomaly: prusa-slicer complains about an empty layer at what appears to be the highest layer in the model. Doesn't affect printing (obviously).

[3da.zip](https://github.com/openrocket/openrocket/files/12425168/3da.zip)
3da

SiboVG:
> 3da

That's awesome! I like the idea of printing small, souvenir-sized rockets; easier to display on your desk than the full-scale model 😉.

> One tiny anomaly: prusa-slicer complains about an empty layer at what appears to be the highest layer in the model. Doesn't affect printing (obviously).

Hm, I'll take a look at it.

JoePfeiffer:
A little more serious anomaly:  the attached .obj file can't be loaded into prusa-slicer (annoyingly, prusa-slicer gives no information beyond "Loading of a model file failed"). I had a hunch the inner motor mount tube might be the problem, but deleting it didn't help.
[3d.zip](https://github.com/openrocket/openrocket/files/12431405/3d.zip)

SiboVG:
> A little more serious anomaly: the attached .obj file can't be loaded into prusa-slicer (annoyingly, prusa-slicer gives no information beyond "Loading of a model file failed"). I had a hunch the inner motor mount tube might be the problem, but deleting it didn't help. [3d.zip](https://github.com/openrocket/openrocket/files/12431405/3d.zip)

Huh, interestingly enough, Cura does not complain about it.

Anyway, you can fix the issue by triangulating the mesh.
image

That makes me think: should we maybe add a button "Optimize for 3D-printing" that sets the scaling to 1000 and enables triangulation?

neilweinstock:
> That makes me think: should we maybe add a button "Optimize for 3D-printing" that sets the scaling to 1000 and enables triangulation?

That seems like a good idea, since optimizing for 3D printing is what is probably the most likely use case here.  I think.

JoePfeiffer:
Ah, OK -- I haven't looked into .obj files a lot, so I didn't realize they weren't all triangulated.

Rather than an "optimize" button, I suspect our infinitely more common use case calls for triangulated meshes and scaling, so I'd suggest just adding a scale slider to the Advanced options and set the default values there to True and 1000.

SiboVG:
Added an optimization button for 3D printing and rendering. By default, the settings are optimized for 3D printing (if you tested this PR before, the settings are still set you your previous preferences). So you'll have to click the "3D printing" button.
image

neilweinstock:
I gave it a tough test... and it's impressively good but not quite there.

Here's the rocket as rendered in OR:
image

I exported the whole thing, optimized for rendering.  Here's what I got, as displayed by Mac OS Quickview:
image

It appears that any component with a decal is rendered as transparent, other than the decals which look pretty good.  I think they're actually not quite rotated correctly but hard to tell in this transparent render.

Next I tried optimize for printing, and got this:
image

That's very close, but the red-circled part is not right; it actually looks like this:
image

While I'm here, I notice that in the Export menu, the RASAero extension is shown in caps but the others are lower case:
image

neilweinstock:
Here's the file.

[nibiru-grey-orange.ork.zip](https://github.com/openrocket/openrocket/files/12449140/nibiru-grey-orange.ork.zip)

I'm actually seeing some inconsistency regarding my above comments.  I have a couple of different exported files (all of the same rocket) that look very different in Quickview.  I need to sort this out and will report back.

neilweinstock:
Wait, there's more weirdness.  I switched to "light" mode, restarted OR, and now the export dialog opens like this:
image

Apart for the window being too small (if I resize, then close it and re-open it it returns to the small size), the whole thing is flipped horizontally relative to dark mode.  Is this intentional?

image

neilweinstock:
More:
I repeated the experiment exporting the whole rocket and got the same results as before.

I note that, in light mode at least, clicking the "3D printing" or "Rendering" buttons leaves no visual indicator of which is selected.  For a moment, focus is put on the selected button (so it is outlined) but as soon as you click anywhere else it goes away.  Similarly, when first opening the export window, there is no indication of which option is selected.

After doing this test, I decided to just try exporting a nose cone.  So I selected the cone in that same rocket, and it gave an exception.  See attached report.
[exportobj-exception.txt](https://github.com/openrocket/openrocket/files/12449209/exportobj-exception.txt)

SiboVG:
> While I'm here, I notice that in the Export menu, the RASAero extension is shown in caps but the others are lower case:
> image

This is on purpose, this is the official writing for each file type.



> I exported the whole thing, optimized for rendering. Here's what I got, as displayed by Mac OS Quickview:
> image

That's a limitation of the MTL file type, nothing I can do about that.

> Next I tried optimize for printing, and got this:
> image

Well, that's not on purpose of course 🙃.

SiboVG:
> Apart for the window being too small (if I resize, then close it and re-open it it returns to the small size), the whole thing is flipped horizontally relative to dark mode. Is this intentional?

That's not something I choose, on dark mode, the accessory is just rendered on the right instead of the left.



> Wait, there's more weirdness. I switched to "light" mode, restarted OR, and now the export dialog opens like this:
> image

That's a bug on light mode only on macOS, that the file chooser is smaller than it ought to be. I can work around it with a hack...

SiboVG:
Thanks for testing @neilweinstock!

SiboVG:
Fixed the exception and added visualization for optimization mode selection. The button is disabled when the settings correspond to 3D printing/rendering optimization.

https://github.com/openrocket/openrocket/assets/11031519/127f047d-a38e-407f-b23d-c5853def7556

neilweinstock:
That looks better.  I'll try to break it again tonight. :)

SiboVG:
> Next I tried optimize for printing, and got this:
> image

Okay, checked it, and the issue is the triangulation is not very smart. It just tries to split faces with more than 3 vertices to 3-vertex faces, but because of the complex fin shape, it doesn't work well. The export is perfect when you turn off triangulation. So, really, not much I can do about it. It would be extremely complex to have triangulation work for very complex poly-vertex faces...

Also added a hack for the file chooser size on macOS light theme. Doesn't look all too pretty, but it's the best I can do, this is a Java bug, not an OR bug.

image

JoePfeiffer:
I've got no idea how to convey this to the user, but of course all of our mass and CG estimates are completely at the mercy of the slicer.

SiboVG:
> I've got no idea how to convey this to the user, but of course all of our mass and CG estimates are completely at the mercy of the slicer.

Good point, but I guess there's not much difference to a normal rocket design process. You manufacture/buy the part, weigh it and adjust your design based on your weight measurements.

neilweinstock:
That's the sort of thing you could be discussed in forum that allows for long-form writing, which could be embedded help, a wiki page, a blog, or whatever.  Right now, we have almost no significant help embedded in the program, and no clear way to direct people to it.  This is something we should work on.

In the meantime, though, I don't think there's anything special that needs to be done here.

Perhaps we should add some common 3D-printing materials (at minimum, PLA, PETG, and ABS) to our materials database to go along with this feature?

neilweinstock:
Published values vary a bit but these seem representative:
PLA: 1.25 g/cm3
ABS: 1.05 g/cm3
PETG: 1.25 g/cm3

neilweinstock:
I’d leave out the word “assuming”

JoePfeiffer:
Though I just did hit another anomaly:  I'm trying to export the attached rocket in two parts: the nose cone, and everything else. The nose cone exports fine, "everything else" appears to hang OR.  Trying to export optimized for printing.
[3d.zip](https://github.com/openrocket/openrocket/files/12476374/3d.zip)

JoePfeiffer:
That got it -- thanks!

neilweinstock:
I couldn't crash the latest build with my (admittedly brief) testing.  It's getting close.  Still a few comments:

1) I need to do more isolation of this, but it does seem like the decals are not being rotated properly... that is, they are not obeying the positioning details in the Appearance panel.  If the .obj format is not capable of doing this, then I question the value of exporting the decals at all (to go along with the transparency issue, though that can be worked around).

2) Which optimization is selected is indicated by greying out and disabling the button.  That is extremely non-standard UI.  Selection should be indicated by showing the button depressed and/or a different color, but not "disabled".
image

3) the density values are not correct; you seem to have swapped the values for ABS and PETG (or PLA)

SiboVG:
> 1. I need to do more isolation of this, but it does seem like the decals are not being rotated properly... that is, they are not obeying the positioning details in the Appearance panel.  If the .obj format is not capable of doing this, then I question the value of exporting the decals at all (to go along with the transparency issue, though that can be worked around).

Rotation is not possible in an MTL file. I think you're overreacting a bit. The clean-up of the OBJ takes only a very small amount of steps in Blender, but it does save you a ton of time of not having to first export the decals in OR, then add a texture to your material, then having to correct the scaling and positioning of the texture (which is not trivial, OR uses some unconventional rules for this).


> 3\. the density values are not correct; you seem to have swapped the values for ABS and PETG (or PLA)

My bad, fixed now.

SiboVG:
> 2\. Which optimization is selected is indicated by greying out and disabling the button.  That is extremely non-standard UI.  Selection should be indicated by showing the button depressed and/or a different color, but not "disabled".

A depressed button indicates that you can repress the button to get it back to a deselected state. This it **not** what we want here. After either buttons is pressed, it should not be possible to repress it. I also don't know what benefit changing the color adds.

neilweinstock:
> > 1. I need to do more isolation of this, but it does seem like the decals are not being rotated properly... that is, they are not obeying the positioning details in the Appearance panel.  If the .obj format is not capable of doing this, then I question the value of exporting the decals at all (to go along with the transparency issue, though that can be worked around).
> 
> Rotation is not possible in an MTL file. I think you're overreacting a bit. The clean-up of the OBJ takes only a very small amount of steps in Blender, but it does save you a ton of time of not having to first export the decals in OR, then add a texture to your material, then having to correct the scaling and positioning of the texture (which is not trivial, OR uses some unconventional rules for this).

You're right.  I was thinking of the OBJ as an end product, because I don't have any tools to import and then work further.  But as an intermediate product, what you have there right now makes perfect sense.  Objection withdrawn.

>A depressed button indicates that you can repress the button to get it back to a deselected state. This it not what we want here. After either buttons is pressed, it should not be possible to repress it. I also don't know what benefit changing the color adds.

I understand your thinking here but don't agree with it.  However, I will not object to keeping it the way it is for now.

SiboVG:
> Just to summarize my objection: a disabled button does not convey that that option is selected. That's all. But again, I can live with it for now.

Valid point, but it's not really a selection, it's more applying something. I'll think about it, maybe I'll apply some highlighting to the applied button or something, I'll experiment with it.

neilweinstock:
> > Just to summarize my objection: a disabled button does not convey that that option is selected. That's all. But again, I can live with it for now.
> 
> Valid point, but it's not really a selection, it's more applying something. I'll think about it, maybe I'll apply some highlighting to the applied button or something, I'll experiment with it.

It's applying something but it should also be an indicator.  It works for the former but not for the latter.

Feel free to fiddle, but I'd be fine deferring this issue until the next release.  Don't want to inject any more delays into this release than necessary at this point.

neilweinstock:
Perfect.

SiboVG:
I'll take the liberty of finally merging this sucker. 😮‍💨

neilweinstock:
Nice work on this one, that's a heck of a tentpole feature for the upcoming release.

PR #2317, opened 2023-09-09
[#2316] Add warning for zero-thickness component OBJ export  @SiboVG

### LINKED ISSUE #2316: [Bug] obj export should warn on 0-thickness walls
### Describe the bug

I just accidentally created a nose cone with a wall thickness of 0; obj export worked just fine but (of course) prusa-slicer complained about it.  The user ought to be warned about this, but allowed to proceed if that's what they really want. The obj file is attached.
[nose.obj.zip](https://github.com/openrocket/openrocket/files/12564377/nose.obj.zip)

### To Reproduce

1. 
2. 
3. 
...

#### Screenshot(s):

#### .ork file:

### OpenRocket version

unstable

Linux

### Additional context

_No response_

SiboVG:
This PR fixes #2316. Example of warnings generated after setting the shoulder thickness of a nose cone to 0, and the thickness of a body tube to 0:
image

I also fixed another issue where the inside mesh of a nose cone was still generated when the nose cone shoulder caused the mesh to be filled.


JoePfeiffer:
Nice -- but would it be possible to raise the warning as soon as the user clicks the "export as obj" button?

SiboVG:
> Nice -- but would it be possible to raise the warning as soon as the user clicks the "export as obj" button?

Well, everything's possible, but:
1. The current method is how all export warnings work in OR (including RockSim export and RASAero export)
2. The current method is the most efficient and clean one code-wise.

neilweinstock:
I think issuing the warning earlier would definitely be nice but not worth excessive effort, given that it's a relatively uncommon scenario.

JoePfeiffer:
Yeah, we've got bigger fish to fry.  I'd like this to be revisited some day, and try to have all the exports alert the user as early as possible.

SiboVG:
Uh, wait, why did you close this @JoePfeiffer?

JoePfeiffer:
Wrong button

Issue #2103, opened 2023-03-07
[Bug] Way too low atmospheric pressure for new simulations


SiboVG:
**Describe the bug**
When a new simulation is added with the atmospheric conditions set to ISA, the pressure is set to 2.88 mbar instead of 1013 mbar.... I've had a number of users on the r/rocketry Discord server complain about unrealistically low pressures in OR.

**To Reproduce**
1. Create a new design
2. Add a new simulation
3. You'll see that the pressure is set to 2.88 mbar

**Expected behavior**
Pressure should be 1013 mbar at sea level.

**Screenshots**
<img width="1555" alt="image" src="https://user-images.githubusercontent.com/11031519/223288129-29248dc8-a5cb-4bec-8309-55033b72e5b5.png">

**Platform**
 - OS: macOS
 - OpenRocket version: 22.02 & unstable

hcraigmiller:
Functions as expected, no anomalous behavior found.

OR Build: 1605
Windows 11 Pro; Version 22H2; Build 22621.1265; Windows Feature Experience Pack 1000.22638.1000.0
Java version "11.0.18" 2023-01-17 LTS; Java(TM) SE Runtime Environment 18.9 (build 11.0.18+9-LTS-195)

PR #2320, opened 2023-09-13
Release notes for 23.09 beta 1

neilweinstock:
For review.  I'm not sure if these are too long, or if they should be better categorized, or what.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -37,8 +37,8 @@ This is the first beta release of version 23.09.
 
 ### Other New Features
 * Bumped app to Java 17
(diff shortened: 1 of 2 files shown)


SiboVG:
Thanks Neil! Haven't checked them carefully yet, but I like what I see so far.

One thing though: I would add the fix for #2103 to the major bug fixes list. I saw **a lot** of users with this issue, and it had a huge impact on their simulations.

neilweinstock:
Updated, please check again.

SiboVG:
Made some final changes for consistency @neilweinstock. LGTM now.

PR #2332, opened 2023-09-16
Add David Cain's 3d printed model rocket example

JoePfeiffer:
David Cain's example rocket to show how to modify a rocket for 3D printing


SiboVG:
@JoePfeiffer I think you didn't include the actual design file + you should update your dbcook components, your PR is trying to revert the latest update.

SiboVG:
Not trying to be rude, but how is this different from the simple model rocket example? I also don't know why we would want a special example rocket for 3D printing, what does it showcase? In theory, every design is 3D printable.

JoePfeiffer:
Thanks on my errors (sigh)

David's modifications to the design for printing separate the fin can as a coherent assembly to be printed as a single part  without the main section of body tube, adds a coupler to be printed with the fin can that will mate with the body tube, and modifies the dimensions of the nose cone shoulder to match the body tube when printed.

SiboVG:
> David's modifications to the design for printing separate the fin can as a coherent assembly to be printed as a single part without the main section of body tube, adds a coupler to be printed with the fin can that will mate with the body tube, and modifies the dimensions of the nose cone shoulder to match the body tube when printed.

Ah, alright, thanks for the clarification.

PR #2347, opened 2023-09-22
[#2344] Optimize UI color fetching

### LINKED ISSUE #2344: 23.09b1: Freeform fin editor extremely slow
In 22.02 dragging around a point in a freeform fin is pretty fluid.  In 23.09b1 it is extremely slow (1-2fps).

Tested with the Simple Model Rocket example with the trapezoidal fins converted to freeform.  Saw similar behavior on Mac and Windows.  On Windows, multi-editing two freeform fins almost completely froze the program.

https://github.com/openrocket/openrocket/assets/243872/3c7af1be-5214-4aef-8cdd-59db565742fe

SiboVG:
This PR fixes #2344. The issue was that the UI colors were fetched dynamically, which apparently caused a lot of computation (I think because the get color methods returned a new Color instance, so new Color objects kept being generated). With this PR, colors are cached.

So, everything works smooth again:

https://github.com/openrocket/openrocket/assets/11031519/e89387d4-def1-48bb-a335-71f4b0fcc511


neilweinstock:
Bless you sir.

On Windows this is now usable again.  I'm not convinced it's *quite* as fast as it was in 22.02, but I don't have an easy way to measure.  I will try on Mac tonight.

PR #2352, opened 2023-09-27
Bug fixes in RK4SimulationStepper and RK4SimulationStatus

JoePfeiffer:
1. When calling calculateAverageThrust(), use the atmospheric conditions from the current timestep, not the previous one. This also let us get rid of previousAtmosphericConditions.  Note -- we don't actually use the atmospheric conditions in calculating thrust; that's presumably in case we ever want to compensate a motor for altitude.

2. Instead of reallocating store every step, let it persist. That let us use the acceleration from the previous time step for free.

3. Clean up flight events unit tests, and tie precision of tests that depend on stepper to stepper timestep. This lets us check time on all events.

4. Properly pass acceleration data to post listeners. Also correct bug where calling pre listeners kept post listeners from being called.

Marking this a draft as it shouldn't go in until after next release.


SiboVG:
Nice!

PR #2362, opened 2023-10-08
Change fin can body tube thickness from .33 mm to .8 mm

JoePfeiffer:
This makes the tube two "lines" wide for a 3D printer with a .4mm nozzle


SiboVG:
Same thing should happen for the paper body tube... It causes a bunch of issues when slicing that tube.

JoePfeiffer:
The body tube isn't intended to be printed -- that's just a standard paper BT-50 body tube.

SiboVG:
> The body tube isn't intended to be printed -- that's just a standard paper BT-50 body tube.

Okay, but here's my use case for the 3D printable rocket example (and I imagine for other users as well): "Oh cool, I can 3D print my rockets now. Let me try to print the 3D printable rocket design to get some practice and see how that goes. Oh jeez, I can't even slice this design properly, the OBJ export feature must be broken.". I just want to print the entire rocket, have it as a small souvenir or a calibration test for future prints or whatever. The example design really should be foolproof...

Once again, I really don't see any reason why the width was chosen that small to begin with. Sure you could argue "it's the nozzle width or width of a piece of paper", but even then, if you print it, you'll end up with something that will break in the blink of an eye when you even try to scrape it of your print bed.

SiboVG:
I apologize for being so adamant about this, but there is a high chance that this example design is the first contact for users with OR's 3D printability feature, and for that it needs to be absolutely foolproof.

JoePfeiffer:
Trying to print it in one piece also would also be a real mess of required supports...  maybe a second example that is intended to be all printed, and relabel this one as "printable nose cone and fin section"?

jimmiedave:
Here's my take:

I understand the desire for a "spit it out of OR and print the whole thing: rocket. It's sort of the promise of 3D printing we've all been sold. But unless a model has been designed to do just that (and designing it so may compromise it as a flying model), it's usually not that easy.

Generally speaking, printing a body tube is a bad idea, for many reasons.

- a paper tube is available instantly and replaceable/repairable.
- a paper tube is smoother on sliding and painted surfaces.
- a paper tube's chemistry accepts paint better.
- a paper tube may be heavier than printed material (depending...).
- printing the tube increases print time and unreliability without offering appreciable benefit.
- printing a rocket intact means XY force on the topmost layers will exert greater Z forces on the adhesion layer at the bottom (b/c the intact rocket is taller - leverage), increasing the likelihood of a detached and failed print.
- printing a rocket intact means the NC may be difficult to impossible to remove from the finished print. If it's easy to remove, it's then probably too loose for flight without taping.
- printing a rocket intact commits user to a completely good print (vs being able to reprint a smaller section if a problem).

I saw setting the model up as I did (Print NC and Fin Can, use paper tube) as giving users a start on the way they'd most likely use OR for 3D printing.

Certainly thickening up the fin can and NC wall to at least .8mm is the right thing to do. The current tube wall measurement is the measurement of a paper tube from OR (I think inherited from ASMR). I didn't size it to be too thin or something. It's a paper tube, and supposed to be that thick.

There is a fair bit of interesting rocket printing going on. In particular, Nerys project to print a 6-foot rocket in sections you could carry on in a suitcase is pretty cool. But a project like that needs lots of special intervention downstream (in the slicer) of an imagined OR export, if it's going to print right. So still no spit it out and print.

Overall, do what you like with the model. It's a collaborative project, after all. 

But for my part, after 3D printing since like 2015, initially on a printer that was mostly plywood, I've failed a bunch, and learned a lot, and can't see printing the tube as a likely path to first-try success.

neilweinstock:
I agree with everything Dave wrote above. 3D printed body tubes are possible, but not the most common approach to 3D printing rockets. 

If not already done, set the name of the body component as “paper body tube”, and the other parts as “printable nose cone” and “printable fin can”, or something like that. That leaves it pretty clear to the user.

PR #2365, opened 2023-10-11
Limit time step for AbstractEulerStepper to time until next event

JoePfeiffer:
A maxTimeStep is passed in to the simulation steppers; this is intended to make sure a simulation time step doesn't extend past the next event. RK4SimulationStepper.step() honors this; AbstractEulerStepper.step() does not (looking back at 15.03, BasicLandingStepper.step() and BasicTumbleStepper.step() did not then, either). The result is that the effect of something like a main parachute deployment can appear after the actual event; this can be seen in Simulation 1 from the dual deploy example, in which the main chute deploy occurs at time 30.444, but the next time step is at 30.943; this is plotted like this:
timesteptoobig
The orange-ish vertical line is the acceleration.

This could easily give a user the impression that we are modelling the time for the parachute to open or something, and of course we're not. Limiting the time step to the next event, both main deploy and the next time step occur at 30.452, better showing the user what's really happening.
step-size-ok

Since it's a one-liner it's really tempting to sneak it in for the release, but let's keep to the freeze.


neilweinstock:
I dunno, that is very tempting to get in this release. We should think about it.

JoePfeiffer:
The more I think about it, the more this idea can be refined. A new PR will be arriving shortly.

PR #2372, opened 2023-10-16
Clarify that only nose cone and fins of 3D printed example rocket are to be printed

JoePfeiffer:
While I agree with @neilweinstock and @jimmiedave that it makes no sense to print a body tube, @SiboVG has a point that a new user is likely to try anyway. This modification to the 3D printable example tries to clarify the intent.


SiboVG:
Thanks @JoePfeiffer!

PR #2389, opened 2023-10-28
[#2357] Remember column width, order, and visibility in preset table

### LINKED ISSUE #2357: Preset browser needs to remember column order in addition to width
Right now the preset browser allows you to adjust both the width and the order of the columns.  But only the widths are saved; the column order reverts to default.  Both should be saved together.

SiboVG:
This PR fixes #2357. The component preset table now remembers the following settings:
- column widths
- column order (when you click and drag columns around)
- column visibility (when you right-click on the table column header and hide a column)

Previously, only the column width was remembered. Additionally, this PR also cleans the column width remembering code.

https://github.com/openrocket/openrocket/assets/11031519/659678ef-c345-461e-997a-a27f9489488c


neilweinstock:
Thank you thank you thank you.

Will test tomorrow night.

neilweinstock:
Fantastic.  All worked correctly in my tests.  Makes the parts library *so* much more usable.  Thanks again!

PR #2396, opened 2023-11-14
Also check color when using default material

SiboVG:
This PR fixes a confusion from TRF [here](https://www.rocketryforum.com/threads/wanted-beta-testers-for-the-next-release-of-openrocket.181945/post-2496881) regarding the "Select all of same color" feature. Previously, when the currently selected component had its appearance set to default, using "Select all of same color" would only select other components **of the same type** that also had the default appearance checked. In the post on TRF, both the nose cone and transition used the default appearance and had the same white color, but the same-color selection did not work.

This PR now ensures that when the currently selected component has the default appearance, it will check for all other components that **also have the default appearance** and have the same default appearance, regardless of whether its the same component type as the selected component or not.

### DIFF
--- a/swing/src/net/sf/openrocket/gui/main/RocketActions.java
+++ b/swing/src/net/sf/openrocket/gui/main/RocketActions.java
@@ -19,6 +19,8 @@ import javax.swing.KeyStroke;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 
+import net.sf.openrocket.appearance.Appearance;
+import net.sf.openrocket.appearance.defaults.DefaultAppearance;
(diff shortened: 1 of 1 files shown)


SiboVG:
copilot:poem

JoePfeiffer:
Should we call this a bug fix to squeeze in before release, or an enhancement for right after release?

SiboVG:
> Should we call this a bug fix to squeeze in before release, or an enhancement for right after release?

Not sure; it would make sense to merge it before the release, since the same-color selection was introduced in this release. The code change itself isn't all too major either.

neilweinstock:
> Not sure; it would make sense to merge it before the release, since the same-color selection was introduced in this release.

This is exactly why I was going to recommend merging.  Good job Sibo.

PR #2402, opened 2023-11-17
[#2388] Sync wind average speed, standard div & turbulence widgets

### LINKED ISSUE #2388: [Feature Request] Simulation wind conditions editor needs improvement
Several things noticed while working on #2386 

1. When the user changes the average windspeed, neither the standard deviation nor the turbulence intensity changes. Since the turbulence intensity is changed. Since the turbulence intensity is defined as st dev / average, at least one of them simply has to. My suggestion would be to leave turbulence intensity constant as average windspeed changes, and adjust st dev. This would automatically take care of the case that the standard deviation of wind speed has to be 0 when the average is 0. I haven't checked whether it actually changes one or the other and just doesn't show it in the UI (see below).
2. When the user adjusts standard deviation, again neither average nor turbulence is adjusted in the UI. In this case, if average is clamped to be at least 0.1 m/s (though the user can manually set the average speed to 0). Again, don't know whether turbulence is adjusted internally (I think turbulence is really only for the user, and standard deviation is used internally. So changing turbulence internally wouldn't matter. Turbulence should be updated to match standard deviation. It should be impossible to change standard deviation when average is 0.
3. When the user adjusts turbulence, same story.

In some of these cases, when you adjust one of the parameters and then go to change the other, it updates to match the correct value from the other (example: set average to 2, st dev to 10. No change to turbulence. Adjust turbulence and it jumps to 5, then starts adjusting as one would expect).

SiboVG:
This PR fixes #2388. The wind speed, deviation, and turbulence widgets are now synced to each other. In addition, I did a refactor so the launch tab in the preferences dialog and the launch conditions in the simulation edit dialog now use the same code, instead of having literal duplicate code for both of them.

https://github.com/openrocket/openrocket/assets/11031519/3b4ae41e-bec6-4e02-ade7-92ee32c112ad


JoePfeiffer:
Nicely done. Too bad this isn't number 2042 instead of 2402 (shhh).

SiboVG:
> Nicely done. Too bad this isn't number 2042 instead of 2402 (shhh).

🤫

PR #2413, opened 2023-11-25
[#2338] Organize axis plot types selector in categories + add search function

### LINKED ISSUE #2338: [Enhancement] Organize sim plot y axis types selection box
The simulation plot y axis types selector is an utter mess. You need to scroll through a giant list of items and search carefully for the item you want.

We should organize everything in categories, such as "Angular motion", "Stability information"... I'm thinking something like a multi-level dropdown, such as this:
image

Additionally, it would also be very neat to have a search function when you click the combobox, such as this:
image

SiboVG:
This PR fixes #2338 by performing a **well-needed** UI enhancement of the x- and y-axis flight data type selectors.

The current UI is a way-too-long list of all the available items:
image

This PR categorizes all the data types into dedicated groups. Even more useful, there's also a search field where you can type the desired data type.

Demo:

https://github.com/openrocket/openrocket/assets/11031519/72741d83-8bc4-446f-8296-c8906695621f

How it works:
- When no text is entered in the search box, the items are displayed in their group.
- When the search field is used, the popup switches from the category display to a search display. The search display lists all the data types that match the search word. The matching text is underlined. If you click on one of the items in the list, it will be selected.
- When typing a search word, the first filtered item will be highlighted. If you press the "Enter" key, you will select the highlighted item.
- You can change the highlighted item by using the up/left and down/right arrow keys. Pressing the "Enter" key selects the highlighted item.

---

I wrote the custom combobox UI in a new, general widget, called `SearchableAndCategorizableComboBox` (long name, I know), so it can be used in other parts of the UI if we wanted to. I feel like there may be other areas of OpenRocket where this could be useful.


neilweinstock:
That is beautiful.  Big improvement.

PR #2426, opened 2023-12-24
Set Balsa as default material for fins

SiboVG:
It doesn't make much sense for Cardboard to be the default material for fins. This PR sets the default fin material to Balsa.
image


JoePfeiffer:
Thank you! It would also be a good idea for the default material for nose cones to be polystyrene.

SiboVG:
> Thank you! It would also be a good idea for the default material for nose cones to be polystyrene.

Done.

neilweinstock:
Thanks, I've been meaning to file an issue on this for.... well, years.

PR #2431, opened 2024-01-03
SIM_ABORT flight event

JoePfeiffer:
Instead of throwing an exception when an error in a .ork file makes it impossible to continue a simulation, define a SIM_ABORT flight data type which logs the problem and halts the simulation branch (but lets the rest of the simulation continue). Program errors such as NaN's will still throw an exception.

The abort is displayed in the upper right corner of the simulation plot window and added to the tooltip in the simulation panel.

The abort.ork file has two sims, each of which will trigger an abort. The first will abort at stage separation time since it's impossible to calculated the sustainer's center of pressure; the second will abort immediately on launch as no motors are defined in the sim. They both threw exceptions in 23.09 (there is a warning when you open abort.ork under 23.09, as 23.09 doesn't have the SIM_ABORT flight event).

Unit tests haven't been updated to look for aborts instead of exceptions yet.
[abort.zip](https://github.com/openrocket/openrocket/files/13820484/abort.zip)


SiboVG:
Great work @JoePfeiffer!

Just so other people can test this, [here's the JAR file from the latest build](https://drive.google.com/file/d/1MzpeYvdR213gwdSYqXQebGW_vYjX9w0c/view?usp=drive_link).

Here's the plot for sim 1:
image
image

sim 2:
image

I'm not entirely convinced about plotting the abort text directly on the plot graph though. You can see in the first screenshot that it overlaps with the plot of 'Stage'. We can certainly stick with this approach, but in that case the y axes should change their range so the data doesn't overlap with the abort text.

JoePfeiffer:
Differences in displays are always interesting! On mine, all the added text is far enough to the right that the overlap issue doesn't happen (though it does overlap the ground hit event if using the vertical line marker style).  So I thought I was putting it on unused space!  I wonder if it could be put in an opaque box with a boundary, so it's visually in front of the plot? Maybe in the lower left, also?

JoePfeiffer:
Latest commits put the abort message in a box in the lower left corner.  The only time this area of the graph is used is if there's a high-velocity deployment
nocp
jar file is [here](https://drive.google.com/file/d/17beSGw8CWDF04-_j4SD4fkKQ_lWifObD/view?usp=drive_link)
(apologies for not including the link last time; for some reason I was thinking the artifact still got created if the checks failed)

SiboVG:
Thanks for the adjustments @JoePfeiffer!

It's indeed better to have it in the bottom-left corner, but if I only plot the altitude, the box still overlaps with the data.
image

It may be wise to add a "Hide errors" checkbox next to the "Export image" button, which is only visible when there is a sim abort situation and is unchecked by default.

JoePfeiffer:
Maybe put the check box between "Show data points" and the branch display pulldown; for consistency (especially with "Show warnings" in the design window) call it "Show errors" and have it enabled by default?

SiboVG:
> Maybe put the check box between "Show data points" and the branch display pulldown; for consistency (especially with "Show warnings" in the design window) call it "Show errors" and have it enabled by default?

Ah, yes, good point.

SiboVG:
Thanks for the addition @JoePfeiffer! One final detail: could you only add the checkbox when there is an error in the simulation? E.g. for 'A simple model rocket', the checkbox does not need to be there and may confuse users.
image

JoePfeiffer:
OK, though in that case we should do the same with the "hide warnings" checkbox in the design window.

SiboVG:
> OK, though in that case we should do the same with the "hide warnings" checkbox in the design window.

Fair remark! I'll remember to implement that at some point.

JoePfeiffer:
[New .jar file](https://drive.google.com/file/d/1KYDWkyj1OpS0_ttmRU0zX5h6n-L6_xok/view?usp=sharing)

JoePfeiffer:
[Jar file](https://drive.google.com/file/d/1UZ8QiisfhlKZ6bPiC_xP9wkQhNHpGX6a/view?usp=sharing)

SiboVG:
> [Jar file](https://drive.google.com/file/d/1UZ8QiisfhlKZ6bPiC_xP9wkQhNHpGX6a/view?usp=sharing)

FYI: GitHub build works again, so no need to upload the JAR file anymore 🙂.

Will do some testing later.

JoePfeiffer:
Yeah, I thought about saying that, but figured there's been enough jars uploaded that one more wouldn't hurt.

SiboVG:
Looks very good @JoePfeiffer, thanks for your effort 💪. I've made some minor final adjustments: I've used a semicolon as the separator for the sim abort annotation text instead of a comma because the locale of some users is set to display numbers with a comma. So you could have "Time: 0,0, Cause: ..." which isn't pretty. I also added the seconds unit to the time value. Finally, I update the enable state of the "Show errors" checkbox based on the currently selected stage:

https://github.com/openrocket/openrocket/assets/11031519/2ed45389-5c5c-430d-b793-b2e06b1fc635

PR #2432, opened 2024-01-04
More accurate thrust

JoePfeiffer:
The existing code doesn't take the thrustcurve datapoint times into account when calculating RK4 time steps. Then, instead of using the actual thrust at the times of the RK4 computation, the average thrust over a timestep is computed and used for the whole timestep computation (given the first choice, this is arguably the best that could be done). This is slightly less accurate than using the thrust at the actual data points, but doesn't cause a real issue except in combination with #2431 when simulating a rocket with CHAD staging.

With #2431 merged, the booster will immediately tumble, and, because thrust is averaged over a time step, will falsely trigger an abort on tumble under thrust (until #2431 is ready for merging, the inability to calculate CP will result in an exception just as it does now).

This code inserts an event for every thrustcurve datapoint, and then adjusts simulation timesteps so they can't go past a scheduled event. It also uses the thrust at the time being evaluated for its calculations instead of an average.

This is enough of a separate issue that I'm making it a separate PR instead of folding it in with  #2431.


SiboVG:
Once again, great effort @JoePfeiffer!

PR #2449, opened 2024-01-28
Change recovery deployment under thrust from Warning to SimulationAbort

JoePfeiffer:
At present, if a recovery device deploys while the rocket is under thrust, a Warning is triggered (and the rest of the flight is treated as if the motor shut off when the device deployed). In reality, we aren't able to simulate what happens in this case, so we should abort the simulation.


SiboVG:
Thanks Joe!

Issue #242, opened 2015-09-26
Export design profiles to SVG or other vector formats


wcravens:
It would be handy to export the design profiles (centering rings, fins, etc) as an svg file, or other vector format, that could be used in a workflow towards a laser cutter, cnc, etc.

wcravens:
May want to contact author: http://www.rocketryforum.com/showthread.php?120485-Needed-Open-Rocket-Fin-Design-Tutorial&p=1387256#post1387256

dkingsley:
Free Form fins profiles can be exported to CSV files directly from the dialog.

snrkl:
+1 
Useful for fins, but REALLY useful for transitions and complex CRs (such as cluster CRs)

SiboVG:
Fixed in #2938.

PR #2452, opened 2024-01-31
Allow fins to be exported to an SVG file

SiboVG:
This PR adds the functionality to export fin profiles (any kind) to an SVG file. Since this functionality will likely primarily be used for laser cutting, I've kept the functionality simple, allowing only to configure the stroke width and color:

https://github.com/openrocket/openrocket/assets/11031519/4d5021ba-3bc3-47cd-820e-98aae48aaa05

The dimensions of the fins are respected in the SVG file.


SiboVG:
I guess this partially fixes #242, although this PR only exports fins, not other components - though this would not be terribly difficult; the SVG file format is very simple. But that requires some thought on how to implement it in the UI.

neilweinstock:
This looks terrific.

I think centering rings are the other component of interest; SVG generation should be very simple for those (UI should be identical... export button at bottom).

There's probably also a need for an "export all" function so you don't need to go through and pick out each one individually... but that can be a separate feature.

JoePfeiffer:
The hard things about centering rings is we'll really need to do something about centering rings for clusters, and people will want slots in the centering rings for fin tabs.

As always, this is really nice.

PR #2457, opened 2024-02-13
Allow sustainer tumble before apogee

JoePfeiffer:
When the tumble stepper was first added, a stage was only allowed to tumble after apogee was reached (I've got no idea why this limitation was imposed). Later, booster stages were allowed to tumble before apogee but the sustainer was not; if the conditions for tumbling were reached for the sustainer a warning was raised but the sustainer continued to use the RK4 stepper (see #181). At some point (I'm not sure when) the warning was removed.  Nowhere do I get a clear picture of why the sustainer is prohibited from tumbling before apogee.

This PR allows the sustainer to tumble under the same conditions as other stages. If the sustainer is still under thrust when it begins to tumble, the simulation stops with a SIM_ABORT event. There are two main reasons to do this:

1. We get not-infrequent questions asking why a simulation gives an unexpectedly low apogee from users who don't realize their design is unstable and don't understand the meaning of the high Angle of Attack warning. Hopefully an explicit "tumble under thrust" abort may be clearer.

2. When a rocket tumbles under thrust, we can get a large number of high angle of attack warnings, basically for every angle encountered. While there is probably a way to consolidate them, it seems cleaner to simply stop when we don't have good data to present any more.

Here's a .ork for the Simple Model Rocket example, with its fins removed so it will be unstable:
[simple-nofins.zip](https://github.com/openrocket/openrocket/files/14271032/simple-nofins.zip)

Simulating the A8-3 sim with current OR unstable and plotting the flight:
unstable-flight

Note that it doesn't report the rocket as tumbling until apogee, even though the fact that it's actually tumbling under thrust is causing the low apogee.

The warnings from that sim (they actually continued past the bottom of my screen):
warnings

Now same rocket, same sim, this PR:
sim-abort

### DIFF
--- a/core/src/net/sf/openrocket/simulation/BasicEventSimulationEngine.java
+++ b/core/src/net/sf/openrocket/simulation/BasicEventSimulationEngine.java
@@ -258,21 +258,16 @@ public class BasicEventSimulationEngine implements SimulationEngine {
 				
 				// Check for Tumbling
 				// Conditions for transition are:
-				//  apogee reached (if sustainer stage)
-				// and is not already tumbling
+				// is not already tumbling
 				// and not stable (cg > cp)
 				// and aoa > AOA_TUMBLE_CONDITION threshold
-				// and thrust < THRUST_TUMBLE_CONDITION threshold
 				
 				if (!currentStatus.isTumbling()) {
 					final double cp = currentStatus.getFlightData().getLast(FlightDataType.TYPE_CP_LOCATION);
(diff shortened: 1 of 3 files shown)


SiboVG:
This is a good change, thanks @JoePfeiffer! Something I noticed from your screenshot though is that we should really round the time in the sim abort plot overlay to something reasonable, like 1 ms.

JoePfeiffer:
I noticed that too -- I was thinking I'd make that change in my next PR (Warning events), as there will be some other changes in that text box to report the warning events as well.

PR #2459, opened 2024-02-14
[#2445] Add ellipsis to export as in file menu

### LINKED ISSUE #2445: [Bug] "Export as" "Export simulation table as CSV file" menu options incorrect: should be "Export as...", etc.
### Describe the bug

Menu language generally interprets an ellipsis ("...") after a menu entry to mean that there's a following dialog or sheet, rather than immediate action.

In OpenRocket, we can see the correctly presented "New" (which immediately creates a new doc), and "Open..." (which opens a dialog allowing user to select a doc to open).

"Export as" implies immediate action upon its selection, while "Export as..." implies there's a following dialog box, which, in the case of OpenRocket, there is.

In OR there is the choice of file format presented in the "Export as" menu, but this does not obviate the Export dialog appearance, and it'd be kind of weird to follow the file formats with ellipses.

### To Reproduce

1. Select "Export as" from the "File" menu.
2. Choose a format (.rkt or .CDX1, or Wavefront .OBJ, if a prior part selection has been made)
3. Note appearance of a dialog box.

"Export simulation table as CSV file" same issue, simpler repro: select "Export simulation table as CSV file" menu item, note dialog.
...

#### Screenshot(s):

Screenshot 2024-01-25 at 7 15 11 PM
Screenshot 2024-01-25 at 7 15 21 PM

#### .ork file:

### OpenRocket version

23.09

macOS

### Additional context

_No response_

SiboVG:
This PR fixes #2445.


JoePfeiffer:
Wow, actual ellipsis character not "..."!

PR #2502, opened 2024-07-03
Combine some code between RK4SimulationStepper, AbstractSimulationStepper, and Euler steppers

JoePfeiffer:
The original goal of this PR was just to make more data available at t=0; it got some feature bloat along the way.

RK4SimulationStepper had a private DataStore, which was used to store simulation variables. These simply appeared as variables in AbstractEulerStepper. The main thing this PR ends up doing is moving the DataStore to AbstractSimulationStepper, and having both RK4SimulationStepper and the Euler steppers use it from there.

It also adds a method for the DataStore to save its contents to the FlightDataBranch, rather than having the data saved by outside code in RK4SimulationStepper and AbstractEulerStepper. This is calculated at the start of each step, and is saved at that time.

There was also an RK4SimulationStatus that added a little bit of information to SimulationStatus; it became possible to eliminate that class.

Oh, and it also saves more data at t=0 :)


SiboVG:
This is awesome, Joe!

PR #2511, opened 2024-07-16
[#2485] Hide components

### LINKED ISSUE #2485: [Feature Request] Hide components
Sometimes, designs can get visually cluttered by components. I'd like a way to visually hide components. The hidden components would still account for simulations etc., but would just not show in the rocket design view.

An option in the Edit menu and context menu to Hide the selected component. Once a component is hidden, the Edit menu item would change from "Hide selected component" to "Unhide selected component". There should also be an option to Unhide all components.

Components that are hidden should have greyed out text in the component tree and a hidden icon:
image

AhanuDewhirst:
This PR fixes #2485 by allowing component visibility to be changed in the component tree via the Edit and Context menus.

Side View

3D Figure


neilweinstock:
Pretty slick.  Not sure I've ever wanted this capability, but I can imagine some will find use for it.

swing/src/main/java/info/openrocket/swing/gui/theme/UITheme.java
```diff
@@ -85,6 +85,7 @@ public interface Theme {
 
         Color getComponentTreeBackgroundColor();
         Color getComponentTreeForegroundColor();
+        Color getvisibilityHiddenForegroundColor();
```
SiboVG, line 88: The 'v' should be capitalized

SiboVG (changes requested):

SiboVG:
Just tested it, and it works great, great job @AhanuDewhirst! The code is also high-quality 👍

Maybe one suggestion: other programs often have a "Show all" button, which will show all hidden components (it should be disabled if no hidden components), even if they are not selected.

JoePfeiffer:
Very nice indeed! @SiboVG it actually has hide all and show all buttons, attached to the stage and rocket. I like doing it this way; it seems to work well with our general tree structure. I wonder if the button names should just be Hide and Show, like everything else, though.

One small issue is hiding a motor mount, stage or the entire rocket doesn't hide configured motor(s) in that component.

@neilweinstock my use case for this is for complicated designs -- I sometimes find seeing just which component is selected to be difficult (possibly an issue for another time).  Toggling visibility a couple of times helps me find it.

neilweinstock:
Fine with me!  Not every feature is for everyone.

SiboVG:
> @SiboVG it actually has hide all and show all buttons, attached to the stage and rocket. I like doing it this way; it seems to work well with our general tree structure. I wonder if the button names should just be Hide and Show, like everything else, though.

Sure, it has a "Hide all" option in the stage and rocket, but there is no way to simultaneously show all components in the design that are hidden. I'd like this feature for the scenario where a user has accidentally hidden one or more components and doesn't know how to revert this. The "Show all in design" button in the edit dialog could prevent this. If there's one thing I've learning while working on OR, it's no never overestimate your users.

SiboVG:
> > @SiboVG it actually has hide all and show all buttons, attached to the stage and rocket. I like doing it this way; it seems to work well with our general tree structure. I wonder if the button names should just be Hide and Show, like everything else, though.
> 
> Sure, it has a "Hide all" option in the stage and rocket, but there is no way to simultaneously show all components in the design that are hidden. I'd like this feature for the scenario where a user has accidentally hidden one or more components and doesn't know how to revert this. The "Show all in design" button in the edit dialog could prevent this. If there's one thing I've learning while working on OR, it's no never overestimate your users.

Maybe we should change the current "Show all" text to "Show selected", to reduce confusion with the "Show all in design" button.

JoePfeiffer:
> > @SiboVG it actually has hide all and show all buttons, attached to the stage and rocket. I like doing it this way; it seems to work well with our general tree structure. I wonder if the button names should just be Hide and Show, like everything else, though.
> 
> Sure, it has a "Hide all" option in the stage and rocket, but there is no way to simultaneously show all components in the design that are hidden. I'd like this feature for the scenario where a user has accidentally hidden one or more components and doesn't know how to revert this. 

Checking... if I hide a component, then hide the stage, then show the stage, the hidden component is now visible (along with everything else, of course). Might be easier for the user if the hide/show stage and rocket buttons were "show" if any components were hidden, only "hide" if all the subcomponents are visible?

>The "Show all in design" button in the edit dialog could prevent this. If there's one thing I've learning while working on OR, it's no never overestimate your users.

Good point!

swing/src/main/java/info/openrocket/swing/gui/theme/UITheme.java
AhanuDewhirst: Good catch! I'll have that fixed.

AhanuDewhirst:
Hi all. Here are my thoughts on the topic:

Firstly, thank you @SiboVG and @JoePfeiffer for your input and feedback. It's helped me reconsider the logic from an end-user perspective. I'm now taking a scenario-based approach to hopefully ensure the functionality is consistent and intuitive.

These are the scenarios I've considered, and the visual feedback I'd expect as a user:

| Scenario | Feedback |
|--|--|
| A single component is selected (excluding stage or rocket) | Show/Hide selected |
| Multiple components are selected (excluding stage or rocket) | Show/Hide selected |
| All components are selected | Show/Hide all |
| A stage is selected | Show/Hide selected |
| The rocket is selected | Show/Hide all |
| A component and a stage are selected (component is a child of stage) | Show/Hide selected |
| A component and a stage are selected (component is not a child of stage) | Show/Hide selected |
| One or more components and the rocket are selected | Show/Hide all |

---
From an implementation perspective, I've separated behaviour from feedback:

Behaviour
* If the rocket or all components are selected, show/hide all components.
* If a stage is selected, show/hide the stage's descendants.
* If a component is selected, show/hide the component.
* Hiding takes precedence over showing, which eliminates situations of contrarying visibility.

Feedback
* If the rocket or all components are selected, display "Show all" or "Hide all", otherwise display "Show selected" or "Hide selected"
* If any of the rocket's descendants are visible, the rocket is shown as visible. Otherwise, the rocket is hidden.
* If any of a stage's descendants are visible, the stage is shown as visible. Otherwise, the stage is hidden.

Also note, hiding a motor mount now hides configured motors, and "Show all" is enabled in the Edit menu when any component is hidden.

Let me know what you think!

SiboVG (approved):

JoePfeiffer:
Very nice work -- thank you for your  contribution to OpenRocket!

SiboVG:
Thanks a bunch @AhanuDewhirst, we really appreciate your contribution! 🙂 If you ever feel bored, be sure to check out some of the other open issues 😉

PR #2513, opened 2024-07-18
[#2093] Allow 2D views to be rotated by click-dragging

### LINKED ISSUE #2093: [Feature Request] In side or top view, rotate rocket by click-dragging up and down
It's old news that many users don't notice the rotation slider on the left side of the rocket display.  Why don't we support a mouse drag up/down to rotate the rocket, just like in 3D view?  Ideally the slider should stay, and slide up and down in concert with the rocket rotation.

I frequently drag the mouse expecting the rocket to rotate, because I'm so accustomed to doing that in the 3D view.

SiboVG:
This PR fixes #2093. You can now rotate the 2D views by clicking and dragging in the design view. The behavior is identical to using the rotation slider in the design view.

The sensitivity is set so that moving the mouse upwards the same distance as the height of the design view will result in a 180° rotation. E.g. if you start your drag in the middle of the design view and drag up to the top of the design view, you will rotate 90°.

https://github.com/user-attachments/assets/e857c21f-e0be-43de-98c5-ee5eefcef051


neilweinstock:
Awesome.

PR #2522, opened 2024-07-26
Replace UniqueID class with ModID and UUID

JoePfeiffer:
Currently, there is a static UniqueID class which is used as a token server for two completely unrelated types of tokens:

One of its methods provides the string representation of a random UUID. This is used to provide a unique identifier for flight configurations, rockets components, and the like.

The other provides an int, which is used to monitor whether the contents of an object have changed to determine whether calculation results have changed and need to be recalculated, or if cached values can be used.

This PR removes UniqueID, and uses a UUID for a unique identifier as needed, and creates a ModID class for tokens marking changes to objects.

The prior implementation was problematic for several reasons. First, of course, it was used for two completely unrelated types of tokens. Second, the tokens it provided were either String or int, and so provided no type protection.

In the case of UUIDs, there is no reason the internal representation should be a string. Java's UUID class is perfectly appropriate (and was used within UniqueID to obtain new UUIDs), and comparisons can be faster than with the string representation. This PR uses UUIDs internally, and converts to and from the string representation as needed for IO.

In the case of modification ID's, things were worse. The class was documented as providing a modification ID that was non-negative, unique, and monotonically increasing. Because of how the IDs were being used, none of these were actually true, and only uniqueness matters.

In use, sometimes modID values were provided by the class, sometimes they were created by simply incrementing them in code, and sometimes they were created by applying arithmetic to either other modID values or values within the class (so they were not monotonic, and I really wouldn't want to try to prove they were really unique).

A modID value of 0 was sometimes used as an initialization value, and sometimes so classes that didn't actually use them could satisfy some inheritance requirements of abstract classes.

A modID of -1 was sometimes used as an initialization value.

Monotonicity didn't really matter, as the only comparisons ever actually used were for equality.

The code now has a ModID class; new ModIDs are only created by calling the constructor.
Two special static ModIDs, ZERO and INVALID are used for the purposes of -1 and 0 before.

ModIDs contain an int "serial number" called id which is incremented as each ModID is created; also, ModIDs implement Comparable. This really isn't necessary, but I can see it might be useful to be able to retrieve the id for debugging.


SiboVG:
Thanks Joe! I'd like to review this before we merge though, just to make sure we don't break anything like undoing history etc.

JoePfeiffer:
Please do!

SiboVG (approved):

PR #2529, opened 2024-08-09
Add more stage separation options for deployable payloads and a deployable payload example

### LINKED ISSUE #852: Add more stage separation options
Some rockets come down in multiple segments, for example a nosecone that comes down on it's own parachute. This can be approximated by putting the nose and the associated parachute in a separate stage, but there isn't a good way to have them separate at apogee besides adjusting it to separate at a certain time in flight. Adding stage separation options for apogee and at a certain altitude during descent would make simulating complicated recovery systems with multiple sections much easier.

### LINKED ISSUE #2519: Add an example design that demonstrates payload separation
We've had a lot of inquiries about how to model payload separation.  Would be good to add an example design that shows it, rather than needing to re-explain it every time.

JoePfeiffer:
This adds three new stage separation options: separate at a given altitude (ascending), separate at apogee, and separate at a given altitude (descending). All three can have a delay following the trigger event.

Separate at apogee and separate at altitude (descending) are to support deployable payloads, which descend on their own recovery system separate from the rest of the rocket. Separate at altitude (ascending) is to support a conventional booster/sustainer; I noticed recently that at least one flight computer (Altus Metrum EasyMega/TeleMega) can be programmed to fire an event in this circumstance.

Created an example rocket consisting of the existing Simple Model Rocket example with an added deployable payload.

One last thing while I was in there I reordered the fields for recovery device deployment to make the decision flow more obvious; the delay after the triggering event is now the last thing in that dialog.

Fixes #852 and fixes #2519


SiboVG:
Very nice!

PR #2530, opened 2024-08-12
[fixes #2521]

### LINKED ISSUE #2521: Sim Plot tooltips do not show name of "sustainer" stage
I do not know if this was a conscious choice or if it's a bug.  Either way IMHO it should be changed.

When rolling over one of the data lines in a sim plot, a tooltip appears showing details.  For booster stages the name of the stage is shown at the top of the tooltip.  For the sustainer stage (including single-stage rockets) the name of the stage is *not* given.

I believe the stage name should be shown for *all* stages in multi-stage rockets.  It's debatable whether it should be shown for single-stage rockets, but I'd be inclined to include it.

NCoppola25:
This PR fixes #2521 in which sim plot tooltips did not include the stage name for the main rocket. This included the sustainer stage and any boosters before they separated. 

There were two main problems:
1) When setting the description for the sustainer series, its corresponding branch name was not used, so only the data type and units were passed to the tooltip.
2) The main rocket's branch was modeled as the sustainer stage, but for multi-stage rockets, one or more boosters will be attached to the rocket for a portion of the flight. Therefore, the different portions of the flight needed to be labeled with all stages still attached to the rocket. 

This was fixed by splitting the sustainer series into multiple series (for multi-stage rockets), and setting the description of each series to include all stages currently on the rocket at the specific time. The point(s) of separation were found for each stage, and this was used to split the series. 
Additionally, since there are now more series than branches (for multi-stage rockets), the coloring mechanism had to be adjusted. The colors of the lines on the plot are now set based on the number of series, which is still a function of the number of branches.

Here is a [jar file for testing.](https://drive.google.com/file/d/1KTp8wRJlA5C8D6FQSxMo5cQhrhwv9hJA/view?usp=drive_link)


SiboVG:
Thanks for the PR! It works well if the "All Stage" combobox item is selected, but selecting individual stages is now broken.


https://github.com/user-attachments/assets/2c59ee7f-f7b4-4fbb-9774-30a717e4ce29

SiboVG (changes requested):
Plotting individual stages on a multi-stage plot is now broken.

JoePfeiffer (commented):
I'm curious why it was necessary to break the plot for the sustainer into multiple series -- my first thought looking at it was that it would be most straightforward to just keep track of the stages active on each data point, and either construct a String of the form "sustainer + booster" to pass to formatSampleTooltip(), or pass an array of Strings with the names of all of the currently active stages as elements of the array. So there must be something I'm missing here.

NCoppola25:
Joe, that is a good point. I went with the approach of creating new series because the generateToolTip() method currently takes a single series as its input. I wasn't sure if/how the active stages for each data point could be maintained in order to access that information when generating the tooltip text. So, I opted to use a series for each combination of active stages.

JoePfeiffer:
Ah, OK, I'd forgotten that we can't pass an event list into generateToolTip() (it's an override, not something we have control over the parameters).

JoePfeiffer:
@SiboVG, looks like it's all working now. Thoughts?

SiboVG (approved):

SiboVG:
Yup, everything works great. Thanks a bunch @NCoppola25 for your excellent PR, much appreciated!

PR #2544, opened 2024-08-27
[#2525] Add option to plot and export component analysis parameter sweeps

### LINKED ISSUE #2525: [Feature Request] Ability to plot/export component analysis parameters in terms of Mach (+ roll rate, AOA, ...?)
User alper/alpersam on Discord made the following request:
"We want to extract the data for the Cn alpha from a openrocket simulation for post treatment. For now, I think the only option was to go to Tools > Component Analysis and then observe the change in Cn alpha by increasing the Mach from 0 up to 1.2 (for our case)
But it is very difficult and time consuming task, we basically need to manually extract each value at a given point one by one
I was wondering if there is an easier way of directly exporting these (Mach - Cn alpha)"

Add a button in the Component Analysis window to Plot/Export variables. This would open a new dialog where you can select the X axis parameter (Mach would be the most useful, but we could also add roll rate, and angle of attack), and the Y axis variables. If e.g. Mach is selected for the x axis, the roll rate and AOA value will take the value configured in the Component Analysis window.

In addition to selecting the x axis variable, the user must also input the value range for that axis (be sure to implement appropriate limiting and sanitization).

For the y axis, the user should also select for which component the variable should be plotted, e.g. 'Total (Rocket)', 'Nose cone', 'Body tube'... These are just the different rows in the component analysis table(s).

Similar to sim plots, the user could then plot the variables, or export them.

SiboVG:
## Goal of this PR

This PR fixes #2525 and introduces the ability to perform a parameter sweep over a given domain type (wind direction, angle of attack, mach number, or roll rate), and then plot and export different component analysis outputs (CP, CNa, CD...).

### The UI

The Component Analysis dialog is converted to a tabbed pan with 2 tabs: `General` and `Plot / Export`. The `General` tab contains the already existing CA content.
image

The `Plot / Export` tab has a section on top to configure the parameter sweep domain type (the X axis). You can choose the minimum and maximum value for the parameter sweep, as well as the step size. In the `Plot / Export` tab is another tabbed pane for plotting (`Plot`) and exporting (`Export`).
 
plot_export_tab

### Configuring an analysis sweep

The main configuration for a component analysis sweep is which domain type to choose and the sweep range and step size. As mentioned earlier, this can be set at the top of the `Plot / Export` tab.

The other parameters for the CA sweep are taken from the `General` tab. For instance, if I do a sweep over Mach number, then the analysis uses the values in the `General` tab for wind direction, angle of attack and roll rate).

### Plotting

Configuring a plot is similar to configuring a simulation plot, with two differences:
1. You must specify the parameter sweep boundary and step size
2. You must specify the rocket component to plot the parameter for

I have added one plot preset: `Total CD vs. Mach number`, which does a parameter sweep over the mach number, and plots the total CD of the rocket. For instance, for 'A simple model rocket':
image

Here is an example plot of a CD plot for all rocket components of 'A simple model rocket':

combined_plots

### Exporting

Configuring an export is very similar to a sim export. The only difference is that you must specify for which rocket components you wish to export a data type. If no components are selected, a warning dialog will appear and that data type will be ignored during the export.

Here's is an example export for a Three-staged rocket:

```CSV
# Parameters:;Wind direction:;54°;Angle of attack:;0°;Roll rate:;0 r/s;Active stages:;Sustainer;Flight configuration:;[C6-5]
# 121 data points written for 4 variables.
# Mach number;CP (A simple model rocket) (cm);CP (Body tube) (cm);CNα (Trapezoidal fin set) (​);CNα (A simple model rocket) (​)
0;31.933;12.5;9.876;11.876
0.025;31.933;12.5;9.876;11.876
0.05;31.934;12.5;9.878;11.878
0.075;31.936;12.5;9.881;11.881
0.1;31.938;12.5;9.886;11.886
0.125;31.941;12.5;9.892;11.892
0.15;31.944;12.5;9.899;11.899
0.175;31.948;12.5;9.907;11.907
0.2;31.952;12.5;9.917;11.917
...
```

Imported in Excel:
image

## Demo

https://github.com/user-attachments/assets/07157f7b-5d19-413e-af6d-12c07747251c

## Additional notes
For this PR to work, I refactored all the plot and export UI code from the simulation plot/export, so we wouldn't have to maintain duplicate code. I also had to re-implement #2530 for the refactored code.


SiboVG:
Note: I think this PR will become very useful in conjunction with the RASAero export feature, because RASAero has a feature called "Aero Plots" which basically does the same as this PR's Component Analysis Plot. So, we could compare the results and see where OR differs.

neilweinstock:
This is a major tentpole feature for the next release.  Very impressive.

I confess I've never really needed this for my own use, so I'm not entirely sure how to exercise it well.  I was curious about one thing, though: I have a sim from someone else that is very large and complex.  Opening up the plot tab yields this:
image

Two things here:
1) Choosing from the very large list of components is pretty unwieldy.  I'm guessing they're in component tree order, but the end result in this particular case is that it would be virtually impossible to pick out what you want.  It is probably true that the people who use this feature will do so with much simpler models, so this might not be an issue that really needs to be addressed, at least not yet.

2) I enlarged the dialog as much as I good to spread out the list.... the bottom section (below the scrolling window) expanded as I enlarged the window.  That section should be fixed size, and all the expansion should go to the scrolling window.  Just a layout tweak I think.

SiboVG:
> 1. Choosing from the very large list of components is pretty unwieldy.  I'm guessing they're in component tree order, but the end result in this particular case is that it would be virtually impossible to pick out what you want.  It is probably true that the people who use this feature will do so with much simpler models, so this might not be an issue that really needs to be addressed, at least not yet.

Wow, it never even occurred to me to have that many components. :) I guess in the future we could change it to a button "Select Components" and a label underneath showing the already selected components. The button would open a new dialog with the component tree, where you can select which components to add. I would grey out and disable components in the component tree that are not relevant for the data type (e.g. only fins have roll characteristics), and have a click behavior where clicking once will toggle the component selection, instead of having to hold Ctrl/Cmd to do multi-component selection.

We could also change the component combobox in the plot panel with a button "Select Component" which has similar behavior, but with only single-component selection. (Or we can change the behavior to have multi-component selection in the plot panel instead of having to add a new Y axis type for each component, and then have the component series be split up automatically; don't know).



> 2\. I enlarged the dialog as much as I good to spread out the list.... the bottom section (below the scrolling window) expanded as I enlarged the window.  That section should be fixed size, and all the expansion should go to the scrolling window.  Just a layout tweak I think.

I'll see what I can do.

SiboVG:
> 2\. I enlarged the dialog as much as I good to spread out the list.... the bottom section (below the scrolling window) expanded as I enlarged the window.  That section should be fixed size, and all the expansion should go to the scrolling window.  Just a layout tweak I think.

Fixed.

neilweinstock:
> > 1. Choosing from the very large list of components is pretty unwieldy.  I'm guessing they're in component tree order, but the end result in this particular case is that it would be virtually impossible to pick out what you want.  It is probably true that the people who use this feature will do so with much simpler models, so this might not be an issue that really needs to be addressed, at least not yet.
> 
> Wow, it never even occurred to me to have that many components. :) I guess in the future we could change it to a button "Select Components" and a label underneath showing the already selected components. The button would open a new dialog with the component tree, where you can select which components to add. I would grey out and disable components in the component tree that are not relevant for the data type (e.g. only fins have roll characteristics), and have a click behavior where clicking once will toggle the component selection, instead of having to hold Ctrl/Cmd to do multi-component selection.

Yeah that's not a normal or typical sim. :)

I think what you have right now is probably good for initial release, and we can enhance in the future as needed.

JoePfeiffer:
Oh, this is great. As always!

Issue #2224, opened 2023-05-29
In appearance panel, add option to choose from colors used in current model


neilweinstock:
This is the last of the features needed to make appearance management easy and convenient in OR.  I don't know if I'll get it but I wanted to put in the request for posterity.

In the appearance panel, probably next to the existing color button, it would be great to have a pulldown of some sort that lets you choose from all the colors currently in use in the model.

For each entry I would show a color swatch and the HTML hex color code.

So, if a rocket were red/white/black, the pulldown would show those three colors in the pulldown.

The current color of the component would also be in the list, and shown as "selected".

I am not sure if "shine" parameter should be lumped in here.  It is probably what you'd *want*, but it is probably not be worth the extra complexity... unless someone could figure out a UI for it that is clean and straightforward.  Otherwise forget it.

The ideal functional implementation would probably be a label "Existing design colors", next to a pulldown.  But that wouldn't fit next to the existing color button, and the idea is that you can set the color *either* by clicking the color button *or* using the pulldown, and in that case having them next to each other seemed nice.  I guess you could add the new thing as another line in the appearance panel, don't know.

So I haven't figured out exactly what this should look like yet, but the function would be very useful.

![image](https://github.com/openrocket/openrocket/assets/243872/46b30e42-e9ad-44e0-85f4-09eca34dc201)

PR #2545, opened 2024-08-28
[#2224] Add hex color input field in appearance panel

SiboVG:
This PR fixes part of #2224, namely adding a hex color selection input in the appearance panel next to the color button.

https://github.com/user-attachments/assets/d152a3e2-f8e6-4052-a396-c40c70b8bb67


neilweinstock:
This is delightful, thanks so much!  I haven't tested it exhaustively but it seems to work exactly as described.

SiboVG:
> This is delightful, thanks so much! I haven't tested it exhaustively but it seems to work exactly as described.

I'm glad you like it 🙂

PR #2554, opened 2024-09-05
[#1081] Add ability to choose user-defined component presets location

### LINKED ISSUE #1081: Need to be able to configure location of user-defined parts libraries
Our Preferences editor needs to be able to let users select where to put their custom component libraries (it is  currently, under Linux, in ~/.openrocket/Components).  Among other things, this would advertise to users that they actually can have their own custom component libraries in addition to the system libraries.

SiboVG:
This PR fixes #1081.

image


neilweinstock:
What exactly is a "User-defined component preset"?

SiboVG:
> What exactly is a "User-defined component preset"?

You can let OR import .orc files with component presets that will be included in the parts database. For example, unzip and upload the following file to the component preset directory and you will see it in the nose cone parts library (search for "test").
[test.orc.zip](https://github.com/user-attachments/files/16905001/test.orc.zip)

Issue #922, opened 2021-05-16
New feature request: Provide a detailed wind profile for flight simulations


raketenolli:
For high altitude flights, I would like to provide a detailed wind profile (wind speeds and directions as well as their gradients) varying with altitude. These winds would have to be taken into account when determining the flight path and performance in a simulation.

jonnypjohnston:
I do not know if this would be beneficial to the developers of OpenRocket, at PSAS we use the [pyHWM2014](https://pypi.org/project/pyhwm2014/) Horizontal Wind Model 2014. I know it is a Python package. Maybe this would be of assistance in developing a wind profile for OpenRocket.

JP

raketenolli:
Do I understand correctly that the Horizontal Wind Model provides a wind profile for a given time, latitude and longitude that's based on historic data and measurements? Of course this could be used as a source for wind profiles, but this feature would also include providing a wind profile oneself (e.g. from recent observations) and using it to calculate the 3D flight path.

roguextech:
Correct, that is what PSAS is using it for during simulations of our newest launch vehicle. There is quite a bit more info along with SolidWorks models of LV3.1 which is scheduled to launch on July 27, 2021.

The CubeSat that we hope to get sub-orbital within a couple of years is [OreSat](https://github.com/oresat).

Here is the link to the GitHub repo: [LV4-MDO](https://github.com/psas/lv4-mdo)

Charles-McG:
I'd like something as simple as the old multi-level wind addon to be functional again.

SiboVG:
Fixed by #2559.

PR #2559, opened 2024-09-13
 [#2060, #2558] Implement multi-level wind input 

### LINKED ISSUE #2060: [Bug] Multi-level-wind plugin no longer works on 22.02
**Describe the bug**
Reported by user u/maxjets on the r/rocketry Discord server: The multi-level-wind plugin no longer works on 22.02. Plugin link: [https://www.rocketryforum.com/threads/new-openrocket-plugin-to-allow-different-wind-speed-direction-at-different-altitudes.140619/](https://www.rocketryforum.com/threads/new-openrocket-plugin-to-allow-different-wind-speed-direction-at-different-altitudes.140619/).

It worked in OR 15.03.

### LINKED ISSUE #2558: [Bug] Wind direction is not saved in .ork file
### Describe the bug

I can't believe this has not been reported or noticed before, but the .ork file does not save the wind direction data. When you open a design, it always defaults to 90°.

### To Reproduce

1. Open 'A simple model rocket'
2. Go to the simulations tab, edit one of the simulations, and change the wind direction
3. Close the design, reopen it
4. The wind direction is reset to 90°

### Screenshots / .ork file

_No response_

### OpenRocket version

23.09, unstable

macOS

### Additional context

_No response_

SiboVG:
This PR fixes #2060 and fixes #2558, and I guess also #922 (except for the wind gradient request). Besides the already existing pink noise wind model, you can now also use a multi-level wind model for a simulation.

## UI
In the simulation config, there are 2 radio buttons to choose to either use the pink noise model, or the multi-level wind model.
select

Click `Add level` to create a new level, or `Remove level` to remove an existing level. If you want to show a visualization of the levels, you can click `Visualize levels`.
image

You can edit the level values in the table, and change the units.

## Simulation behavior
If the altitude is within the levels range, then the wind parameters are linearly interpolated from the closest two levels. If the altitude is out-of-bounds, the parameters are given the same parameters as the lower/upper level bound.

An example (parameters the same as the demo below):
image

## .ork file

The .ork now includes a `wind model` object for both the pink noise and multilevel model. Additionally, there is a `windmodeltype` to indicate which model the simulation uses. So, both wind model settings are stored in the .ork in case the user wants to switch between either one.

```XML
<windaverage>2.0</windaverage>
<windturbulence>0.1</windturbulence>
<winddirection>1.5707963267948966</winddirection>
<wind model="average">
  <speed>2.0</speed>
  <direction>1.5707963267948966</direction>
  <standarddeviation>0.1</standarddeviation>
</wind>
<wind model="multilevel">
  <windlevel altitude="0.0" speed="5.0" direction="1.5707963267948966" standarddeviation="0.1"/>
  <windlevel altitude="100.0" speed="5.0" direction="4.71238898038469" standarddeviation="0.1"/>
  <windlevel altitude="150.0" speed="2.0" direction="0.7853981633974483" standarddeviation="0.1"/>
</wind>
<windmodeltype>MultiLevel</windmodeltype>
```

Note that the old wind attributes are also still stored (the first entries of the XML). This is to retain backward compatibility with previous versions of OR.

## Demo

https://github.com/user-attachments/assets/372b1d88-cd30-4e5f-96bc-0f64ced6e65c


JoePfeiffer:
Another great piece of work!

It isn't clear to my why pink noise and multilevel should be separate options: my naive expectation (knowing very little about wind) would be for the wind at each level to be pink noise. So you wouldn't need the two options, you'd just decide whether to add new levels, with mean, direction, and turbulence at each level.

If I'm using Imperial units, switch the wind model to multi-level, and create a new level, the wind speed is 16.4 ft/s (which is 5 m/s).  Now when I click on the wind speed to change it, it jumps to 5065617 ft/s. I can now edit the speed as I wish. If I add a second new level and go to change the wind speed, I get no surprises.

SiboVG:
> It isn't clear to my why pink noise and multilevel should be separate options: my naive expectation (knowing very little about wind) would be for the wind at each level to be pink noise. So you wouldn't need the two options, you'd just decide whether to add new levels, with mean, direction, and turbulence at each level.

Good point! I would however still make a distinction between "Average wind" and "Multi-level wind" in the UI (and internally). I feel like it will be confusing for users if they only have the multi-level wind input. I can have a common section in the beginning to specify the standard deviation. I'm not sure about wind turbulence, since it's calculated using the average wind speed. Maybe I'll include that only in the average wind section.

neilweinstock:
This is a great thing that many have inquired about.

This is way outside my own domain, though, and I'm curious how it might actually be used. I'm guessing this would be used at the field, based on info gotten from some website (or app)? Or is there any reason to ever set this up in advance?  I'm not sure how one would know the conditions to that level of detail in advance.

Does there need to be a way to copy these settings between simulations?  Does there need to be a way to import that data from a website in order to be useful?

Or does this somewhat mimic the existing plug-ins that do this?  If people are happy with those, and this works similarly, then I would imagine this is good for a first release.  But I really have no idea.

JoePfeiffer:
> > It isn't clear to my why pink noise and multilevel should be separate options: my naive expectation (knowing very little about wind) would be for the wind at each level to be pink noise. So you wouldn't need the two options, you'd just decide whether to add new levels, with mean, direction, and turbulence at each level.
> 
> Good point! I would however still make a distinction between "Average wind" and "Multi-level wind" in the UI (and internally). I feel like it will be confusing for users if they only have the multi-level wind input. I can have a common section in the beginning to specify the standard deviation. I'm not sure about wind turbulence, since it's calculated using the average wind speed. Maybe I'll include that only in the average wind section.

It seems (to me, anyway; like I said I know very little about wind) that it would make the most sense to have the values at each level include the turbulence; just as currently pink noise comes up with a default wind speed and turbulence, the new model would come up with default values for level 0. A user could just adjust those values, or could add new levels.

I'm not quite sure why we have both turbulence and standard deviation, since they're calculated from each other.

JoePfeiffer:
> This is a great thing that many have inquired about.
> 
> This is way outside my own domain, though, and I'm curious how it might actually be used. I'm guessing this would be used at the field, based on info gotten from some website (or app)? Or is there any reason to ever set this up in advance? I'm not sure how one would know the conditions to that level of detail in advance.
> 
> Does there need to be a way to copy these settings between simulations? Does there need to be a way to import that data from a website in order to be useful?
> 
> Or does this somewhat mimic the existing plug-ins that do this? If people are happy with those, and this works similarly, then I would imagine this is good for a first release. But I really have no idea.

I frequently see people launching over around 10K feet checking the winds aloft before deciding whether they actually want to launch that day. The way I see this being used is somebody, morning of launch, before heading to the field, going to a site like windy.com to pick up winds at several levels to run a simulation.

What would be *really* cool would be going to a weather site with an API (I'm told there are many of them) and grabbing wind conditions at multiple altitudes from there. But no need to do that on this release.

neilweinstock:
> I frequently see people launching over around 10K feet checking the winds aloft before deciding whether they actually want to launch that day. The way I see this being used is somebody, morning of launch, before heading to the field, going to a site like windy.com to pick up winds at several levels to run a simulation.
> 
> What would be _really_ cool would be going to a weather site with an API (I'm told there are many of them) and grabbing wind conditions at multiple altitudes from there. But no need to do that on this release.

Fair enough.

SiboVG:
> It seems (to me, anyway; like I said I know very little about wind) that it would make the most sense to have the values at each level include the turbulence; just as currently pink noise comes up with a default wind speed and turbulence, the new model would come up with default values for level 0. A user could just adjust those values, or could add new levels.

Sure, I can make it configurable for each entry. I hope I can get it all to fit in the UI though :) 

> What would be _really_ cool would be going to a weather site with an API (I'm told there are many of them) and grabbing wind conditions at multiple altitudes from there. But no need to do that on this release.

Yup, this PR was sort of an entry point for that feature. Once we have this back-end, it's "just" a matter of fetching the right values and plugging them in. A lot of users have asked for that feature, would be nice to have it included.

> Or does this somewhat mimic the existing plug-ins that do this? If people are happy with those, and this works similarly, then I would imagine this is good for a first release. But I really have no idea.

My initial intent was to have native support for the existing MultiLevelWind plugin, since a lot of people seem to like it. And I'm tired of users stating that the plugin does not work with OR > 15.03, because the plugin was never officially updated.

SiboVG:
Updated behavior:

https://github.com/user-attachments/assets/7fe0e91e-fc47-4910-a5a3-0813219dc78b

JoePfeiffer:
Very nice. I still think the "average" model should be replaced with just a multi-level model with a single level, but that's not crucial.

PR #2561, opened 2024-09-18
Use selectable component tree for CA plot/export components selection

SiboVG:
This PR improves the method for selecting components in the component analysis plot/export panels. See PR #2544 for the previous implementation.

Now, there is a button to select components. Once clicked, it opens a new dialog with the rocket's component tree. Components that are not suitable for the current CA data type are disabled and rendered greyed out.

With this, you also don't have to add a new plot type entry for every component you want to plot. You can just add one entry and then multi-select components in the dialog.

https://github.com/user-attachments/assets/3d3556e6-df91-4e33-9327-4e8e024ec9c9

For this to properly work, I also had to replace the simulation export panel to use a custom panel layout instead of the existing table.

Before:
image

After:
Screenshot 2024-09-19 at 00 35 19


neilweinstock:
That looks very very nice.

Issue #2553, opened 2024-09-01
[Bug] Unnecessary "Forward end of airframe is open" warnings in Booster stage of "Three-stage rocket"


SiboVG:
### Describe the bug

For some reason, the body tubes in the booster stages in the "Three-stage rocket" example design show 2 warnings that the forward end of the airframe is open. Interestingly enough, those warnings disappear when you disable the sustainer.

### To Reproduce

1. Open the "Three-stage rocket" example design
2. Go to the simulations tab
3. Rerun all sims
4. The sim warnings appear

### OpenRocket version

unstable

### What platform are you running on?

macOS

JoePfeiffer:
Question here is under what conditions the warning should appear?

1. Booster has no recovery system (as in the 3 stage example). This will be either a black powder staged rocket with a tumble recovery booster, or a high altitude shot with a ballistic booster recovery. In either case the flyer doesn't care what happens to the booster and the warning shouldn't appear.
2. Booster has a recovery system, and uses motor eject to deploy. The user is likely to be using OR to try to set the motor delay, and should be warned the sim after separation will be inaccurate. But maybe the warning should be specific about calculated optimal delay is likely to be wrong?
3. Booster has a recovery system, and uses electronics to deploy. Results will be inaccurate, but how much is the user likely to care? I could go both ways on this one.

SiboVG:
> Question here is under what conditions the warning should appear?
> 
> 1. Booster has no recovery system (as in the 3 stage example). This will be either a black powder staged rocket with a tumble recovery booster, or a high altitude shot with a ballistic booster recovery. In either case the flyer doesn't care what happens to the booster and the warning shouldn't appear.
> 2. Booster has a recovery system, and uses motor eject to deploy. The user is likely to be using OR to try to set the motor delay, and should be warned the sim after separation will be inaccurate. But maybe the warning should be specific about calculated optimal delay is likely to be wrong?
> 3. Booster has a recovery system, and uses electronics to deploy. Results will be inaccurate, but how much is the user likely to care? I could go both ways on this one.

I'm sorry, am I missing something? What does this have to do with the forward airframe being open?

JoePfeiffer:
> I'm sorry, am I missing something? What does this have to do with the forward airframe being open?

The airframe is actually open in all these cases. The question is whether the user is likely to be expecting to simulate the booster flight post-separation, in which case they should be reminded that they aren't going to get an accurate result.

JoePfeiffer:
[...]

Issue #2475, opened 2024-04-09
[Bug] Simulation with warning loaded from file does not match user unit


SiboVG:
### Describe the bug

When saving simulation data to a .ork file, the simulation warnings are also stored. However, any numerical values stored in the warning are stored using the units of the person who saved the sim. If person that saved the sim uses imperial units and I have metric units, then when I open the file, the warning message will have values in imperial units.

The warning message should always adapt to use the user's units.

### To Reproduce

1. Set your unit preferences to metrics
2. Open 'A simple model rocket'
3. Open the warnings tab of Simulation 3
4. The deployment speed unit is in mph instead of m/s
5. Rerun Simulation 3, open the sim warning dialog again
6. Unit is now m/s

#### Screenshot(s):
<img width="1292" alt="image" src="https://github.com/openrocket/openrocket/assets/11031519/971f999d-93e0-4ca2-9300-b240a012bb5d">

### OpenRocket version

unstable

### What platform are you running on?

macOS

senft-research:
Well this seems like as good an issue as any to put this on. 

I am going to start work on a potential refactor of the warning system (I'll put a general "assume this effort has been abandoned" time period of around 2 months on it haha) to try and address this problem and the several other warning related issues we've had over the last year or 2. 

Wish me luck chaps

PR #2562, opened 2024-09-23
Flight warning event

JoePfeiffer:
This PR adds a flight event of type SIM_WARN, which is added to the simulation event icons when plotting a simulation. It also increases the information provided in event icon tooltips.

The attached zipped .ork file is an example of this PR:

1. It was saved under 23.09. There are two warnings in the saved file, but no SIM_WARN events. Opening the file, going to the Flight simulations tab, and plotting the simulation shows the existing flight events with the enhanced icons.
2. Running the sim and plotting the results now shows the two SIM_WARN events: one for the open booster body tube (Issue #2553 noted; this PR does not address that) and one for the sustainer high angle of attack as it tumbles.
3. Saving and reopening the .ork file now includes the SIM_WARN events.
4. The saved .ork file can be opened with 23.09; as one would expect there are a bunch of warnings from unknown elements.

sim_warn

Marking it as a draft for the moment so I can get comments. My planned next step is to add the information that event tooltips now have to regular sample tooltips.

Couple of specific questions:
1. Should the tooltip include the time, even when our domain is some other variable?
2. Is there any particular reason we're displaying the sample number?

[unstable-sustainer-23.09.zip](https://github.com/user-attachments/files/17100576/unstable-sustainer-23.09.zip)


SiboVG:
This looks very promising, thanks Joe! I haven't tested it yet, I'll try to do it in the coming days.

> Should the tooltip include the time, even when our domain is some other variable?

I would always include time, and if the domain is some other variable also include the domain value.

> Is there any particular reason we're displaying the sample number?

I don't know actually, I guess there's no harm in including it?

---

I also have a question: let's say there is a period of 10 seconds where a large angle of attack is encountered, does this PR add a warning event for each sample in those 10 seconds or only for the first sample?

JoePfeiffer:
> I would always include time, and if the domain is some other variable also include the domain value.
OK, just put in a commit with that. Also fixed a bug that turns out to have been around since 23.09 -- when an x axis type has NaN's it caused an exception.

The more I work with x axis  types other than time the less sure I am they're a good idea. The plots I get are nearly impossible to interpret.

> > Is there any particular reason we're displaying the sample number?
> 
> I don't know actually, I guess there's no harm in including it?

It makes the tooltip bigger, and I can't see any way to use it except when data has been exported as a .csv file, and in that case the .csv file has it.

> I also have a question: let's say there is a period of 10 seconds where a large angle of attack is encountered, does this PR add a warning event for each sample in those 10 seconds or only for the first sample?

Just the first sample.

neilweinstock:
This seems like a good bit of foundational work for the future flight event log.  Haven't gotten to try it yet, just reacting to what I see here.

My thoughts:
1) I don't think sample number is useful.  It's just clutter.
2) As for non-time X axes... I could imagine doing velocity vs. stability margin, or something like that.  Don't know how often people actually do that.  Could be good to poll the users if interested.

JoePfeiffer:
stability vs velocity plot; like I said it's pretty much incomprehensible to me but it does show what the tooltips look like.
stability-vs-velocity

JoePfeiffer:
Huh. Thought I saw some mis-positioned event icons, but can't duplicate (note that at present OR doesn't display event icons for data using the right side vertical axis at all).

SiboVG:
This looks really good overall @JoePfeiffer!

Two suggestions though:
1. Use units for the warning message values (e.g. degrees for the angle of attack instead of simply "17.8")
2. _(This probably involves a fair amount of work, so don't worry if you ignore this)_ If a warning spans multiple data points, we could display the time span in the tooltip, e.g. "Time: 1.37 – 2.03 s"

image

SiboVG:
> You're right, but that's a subject for another PR. This one just takes whatever the warning text is, and puts it in the tooltip. If the warning contains units, so does the tooltip. Here's the main chute deploy event for a dual deploy rocket with the drogue disabled:

That's what I figured. Something we should keep in mind. I guess somewhat related to #2475.

Issue #2543, opened 2024-08-23
[Bug] Launch lug pressure CD is extremely high for Mach values very close, but over 1


SiboVG:
### Describe the bug

When you enter an extremely small number, that is just over 1 (e.g. 1.0000000000000007), the pressure CD of launch lugs is extremely high.
<img width="891" alt="image" src="https://github.com/user-attachments/assets/46e00fea-a0e2-4833-8fa2-e7a40888ea43">

Note that this does not happen when you input '1', or a small number just below 1 (e.g. 0.9999999999999993).

### To Reproduce

1. Open 'A simple model rocket'
2. Open the Component Analysis Tool
3. Copy this value to your clipboard: 1.0000000000000007
4. Paste the value in the Mach Number spinner 
5. Extremely high pressure CD

### OpenRocket version

unstable

### What platform are you running on?

macOS

JoePfeiffer:
I'm seeing this too.
Cut-and-pasting 1.0000000000000007 into Mach gives really high pressure CD
Change Mach value to .1; pressure CD is reasonable
Typing or cut-and-pasting 1.0 still gives a reasonable CD

it's important to not just go straight from 1.0000000000000007 to 0.9999999999999993; the values are close enough it doesn't get recalculated (which caused me a few minutes of confusion).

SiboVG:
> it's important to not just go straight from 1.0000000000000007 to 0.9999999999999993; the values are close enough it doesn't get recalculated (which caused me a few minutes of confusion).

Me too! I didn't know that was the reason for it; I just ignored it :)

JoePfeiffer:
> > it's important to not just go straight from 1.0000000000000007 to 0.9999999999999993; the values are close enough it doesn't get recalculated (which caused me a few minutes of confusion).
> 
> Me too! I didn't know that was the reason for it; I just ignored it :)

I was 99% sure we had two bugs before I finally realized what was going on!

SiboVG:
The error is caused by the Prandtl-Glauert correction to the friction coefficient in `TubeCalc.java`:
```java
// If we're supersonic, apply a correction
if (conditions.getMach() > 1) {
    f = f / conditions.getBeta();
}
```

Beta is calculated as `sqrt(abs(1 - Mach^2))`, which becomes very small when Mach moves towards 1. As a consequence, the division of f by beta becomes very large.

[...]

PR #2570, opened 2024-10-08
Clamp beta factor (compressibility correction) in transonic region + apply it in supersonic region

SiboVG:
This PR is related to #2543. It clamps the Prandtl-Glauert compressibility correction factor (beta) to 0.6 in the transonic region, between Mach 0.8 and 1.1, as described in "Active Guidance and Dynamic Flight Mechanics for Model Rockets", David Ketchledge (1993). This avoids beta reaching infinity when it nears Mach 1. Additionally, I also applied the compressibility correction in the supersonic region for fins, as described in literature.

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/FlightConditions.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/FlightConditions.java
@@ -287,17 +287,16 @@ public class FlightConditions implements Cloneable, ChangeSource, Monitorable {
 
 	/**
 	 * Calculate the beta value (compressibility factor/Prandtl-Glauert correction factor) for the given Mach number.
-	 * Source: "Active Guidance and Dynamic Flight Mechanics for Model Rockets", David Ketchledge (1993)
 	 * @param mach the Mach number.
 	 * @return the beta value.
 	 */
 	private static double calculateBeta(double mach) {
-		if (mach < 0.8) {
+		if (mach < 0.99999) {
 			return MathUtil.safeSqrt(1 - mach * mach);
-		} else if (mach < 1.1)
-			// Because beta reaches infinity for M=1, we clamp the factor to 0.6 in the transonic region (Mach 0.8-1.1)
-			return 0.6;		// sqrt(1 - 0.8^2)
-		else {
+		} else if (mach < 1.00001) {
+			// Clamp to avoid singularity near Mach 1
+			return MathUtil.safeSqrt(1 - 0.99999 * 0.99999);
+		} else {
 			return MathUtil.safeSqrt(mach * mach - 1);
 		}
 	}
--- a/core/src/main/java/info/openrocket/core/aerodynamics/FlightConditions.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/FlightConditions.java
@@ -22,6 +22,7 @@ import info.openrocket.core.util.ModID;
  * @author Sampo Niskanen <sampo.niskanen@iki.fi>
  */
 public class FlightConditions implements Cloneable, ChangeSource, Monitorable {
+	private static final double MIN_BETA = 0.25;
 
 	private List<EventListener> listenerList = new ArrayList<>();
 	private EventObject event = new EventObject(this);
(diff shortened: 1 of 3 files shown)


SiboVG:
Note: as far as I can tell, this has little to no effect on the simulations, but it's the correct way to use compressibility correction.

JoePfeiffer:
Every source I'm finding (other than Ketchledge) says it's only usable at speeds up to about Mach 0.7; see, for instance, _Fundamentals of Aerodynamics_ by Anderson (6th ed, 2017), p. 774, "... it is purely a subsonic theory and begins to give inappropriate results at values of M = 0.7 and above".

There are a bunch of places describing the Prandtl-Glauert singularity as causing vapor cones at Mach 1, but the more careful sources (even wikipedia!) say there's a big pressure effect there (but not like Prandtl-Glauert), but wave their hands about just what it actually looks like.

So if we incorporate it at all we need to only use it at speeds below Mach 0.7 where its impact will be much less than the variation between motors, and then we still have the question of what the right thing to do is in the transonic and supersonic regions.  At that point it seems better to just not use it, particularly since it doesn't seem to make any real difference to our results.

SiboVG:
> Every source I'm finding (other than Ketchledge) says it's only usable at speeds up to about Mach 0.7; see, for instance, _Fundamentals of Aerodynamics_ by Anderson (6th ed, 2017), p. 774, "... it is purely a subsonic theory and begins to give inappropriate results at values of M = 0.7 and above".
> 
> There are a bunch of places describing the Prandtl-Glauert singularity as causing vapor cones at Mach 1, but the more careful sources (even wikipedia!) say there's a big pressure effect there (but not like Prandtl-Glauert), but wave their hands about just what it actually looks like.
> 
> So if we incorporate it at all we need to only use it at speeds below Mach 0.7 where its impact will be much less than the variation between motors, and then we still have the question of what the right thing to do is in the transonic and supersonic regions. At that point it seems better to just not use it, particularly since it doesn't seem to make any real difference to our results.

I could have sworn that I saw more references using the Prandtl correction at supersonic flow, but looking deeper into it, I think you're right.

I have removed the correction for supersonic flow, and have put a clamping when mach is between 0.99999 and 1.00001, to avoid the singularity.

I'm just gonna name-drop the Kármán-Tsien and Laitone's correction. They are similar to Prandtl-Glauert, but more accurate. I don't think it will make a big difference though, so it may not be worth the extra calculations.

JoePfeiffer:
> I could have sworn that I saw more references using the Prandtl correction at supersonic flow, but looking deeper into it, I think you're right.

Maybe we saw the same thing? If I didn't, I can't imagine why I would have put in the correction in the first place.

> I have removed the correction for supersonic flow, and have put a clamping when mach is between 0.99999 and 1.00001, to avoid the singularity.

It would be simpler to just decide the clamping value, and use MathUtil.max(CLAMP_VALUE, MathUtil.safeSqrt(1 - mach*mach))

> I'm just gonna name-drop the Kármán-Tsien and Laitone's correction. They are similar to Prandtl-Glauert, but more accurate. I don't think it will make a big difference though, so it may not be worth the extra calculations.

Since Prandtl-Glauert doesn't seem to appreciably improve our accuracy, I don't really see that we need it.  We certainly don't need either of the others! (they actually both appear in the text I mentioned above -- it's amusing that the author carefully derives Prandtl-Glauert, and then just drops the other two without any derivation).

SiboVG:
> It would be simpler to just decide the clamping value, and use MathUtil.max(CLAMP_VALUE, MathUtil.safeSqrt(1 - mach*mach))

Fair enough! I'll adjust it tonight.

> Since Prandtl-Glauert doesn't seem to appreciably improve our accuracy, I don't really see that we need it. We certainly don't need either of the others! (they actually both appear in the text I mentioned above -- it's amusing that the author carefully derives Prandtl-Glauert, and then just drops the other two without any derivation).

Yes, I also like this passage from Hoerner's "Fluid-Dynamic Drag":

> A number of attempts have been made (2,c) of improving the rule by introducing second-order terms. One effect thus taken into ac¬ count simply concerns the difference between (CpC01Yl /Cp;nc) and (AV/V)com/(AV/V);nc. It is be¬ lieved, however, that the second-order term in the pressure coefficient as a function of the thickness ratio (t/c) is usually cancelled by progressive boundary layer losses. Another correctly suggested modifica¬ tion, is to replace in equation 5 the ambient Mach number by the somewhat increased local Mach num¬ ber at the surface of the solid considered. All modifi¬ cations proposed tend to make the compressibility factor larger when approaching the critical Mach number, particularly in higher thickness ratios. Inas¬ much as the increment is but small (for typical air¬ foil sections) and since certain increases in pressure and lift coefficients can also be explained on different levels (4), the author proposes that it usually does not pay, in engineering applications, to complicate the original rule by introducing higher order terms.

SiboVG:
I clamped it to an arbitrary beta value of 0.25.

PR #2571, opened 2024-10-09
Angle constraints in fin editor

neilweinstock:
In the fin editor, while dragging a point, if you hold down "shift" it will constrain the location of the point relative to the previous point.  If you also hold down "Ctrl", it will constrain relative to the following point.

It will constrain either horizontally, vertically, or at 45 degree increments.

I'm not sure if the 45 degree constraint is useful, and have debated removing it.  I'd welcome input here.

Video shows operation.  45 degree constraint is a bit wonky when moving one of the root points.

https://github.com/user-attachments/assets/32c04815-c4c9-4a27-8aef-0889e32ed406


neilweinstock:
Addresses #427, mostly.

SiboVG:
Looks nice @neilweinstock, thanks! 

I've made minor changes: a small code cleanup, removed the "Double-click to edit" label and just allowed a single-click in the table to edit the entry, improved tab traversing in the table, and properly set the snap highlight color for the different UI themes.

I'm still figuring out how to fix the root point snapping.

neilweinstock:
Thanks for those changes.

Root point snapping is a bit of a conundrum.  However, the problem only occurs at the 45 degree snap, and since I'm not really convinced that's useful, one easy solution is just to get rid of 45 degree snapping altogether.  Or you could disable 45 degree snapping when dragging one of the root points.  Or leave it as is.  ¯\_(ツ)_/¯

SiboVG:
> Thanks for those changes.
> 
> Root point snapping is a bit of a conundrum. However, the problem only occurs at the 45 degree snap, and since I'm not really convinced that's useful, one easy solution is just to get rid of 45 degree snapping altogether. Or you could disable 45 degree snapping when dragging one of the root points. Or leave it as is. ¯_(ツ)_/¯

Well, I know what the issue is, and I have an idea of how to fix it, but I'll have to think about an implementation.

SiboVG:
Okay, so the issue is that FreeformFinSet has code to clamp the first and last point to the parent body. This clashes with the snapping. The solution IMO would be to do the first and last point clamping according to the 45° line intersection of the lock line and the parent body if you're snapping. If someone wants to implement that, that's fine, but for now, I just disabled root point snapping.

neilweinstock:
In my opinion, retaining vertical/horizontal snapping for root points (really, vertical is 99.99% of potential usage there) is more valuable than retaining 45 degree snapping everywhere else.  Ideally, I would just remove 45 degree snapping from root points and leave everything else, although the current code partitioning doesn't make that easy.

But I don't feel very strongly about it.

PR #2578, opened 2024-10-17
[#2576] Fix Sorting Problem With Sides and Line Count columns on Parachute Preset Library

### LINKED ISSUE #2576: [Bug] Both the "Sides" column and the "Line Count" column in the Parachute Component panel do not sort correctly
### Describe the bug

The "Sides" and the "Line Count" columns sort lexically, not numerically ie: if say sorted increasingly they will display 10, 12, 14,...,3, 4, 6, 8.  The sort is treating those two fields as text, not as numbers.  Ironically both the "Diameter" and the "Line Length" columns sort correctly (numerically) despite the data in those fields containing text ie: "1524 mm"

### To Reproduce

1. Create a design and get to a point where you can add a Parachute mass component or open a design that already contains one
2. In the parachute component preset window, click on either the Sides or the Line Count heading.  The first click should sort by increasing values.
3. Note that the first entry is "10" and not "3".  Scroll through the list and note as the data goes through the 10s and 20s values before getting to 3.

#### Screenshot(s):
image

#### .ork file:

### OpenRocket version

23.09

Windows

### Additional context

_No response_

stoopipoopy:
This PR fixes #2576 where the "Sides" and "Line Count" column in the Parachute Component panel sort lexically instead of numerically.

The problem was in ComponentPresetTable. During column initialization, the sorter did not have a special case for Integers as it did for Doubles and Booleans. This caused it to sort Integers like Strings instead of numbers because there was no base case other than the default.

The issue was fixed by adding a simple special case, just like the Booleans or Doubles, that handles Integers separately than all other Objects.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/dialogs/preset/ComponentPresetTable.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/dialogs/preset/ComponentPresetTable.java
@@ -6,11 +6,7 @@ import java.awt.event.ItemEvent;
 import java.awt.event.ItemListener;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import javax.swing.ButtonGroup;
 import javax.swing.JCheckBoxMenuItem;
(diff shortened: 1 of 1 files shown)


JoePfeiffer (changes requested):
Thank you! Looks good.

One small thing is we do prefer to import the classes individually, rather than using a wildcard.

stoopipoopy:
Fixed.

JoePfeiffer (approved):

SiboVG:
Thanks for the change @stoopipoopy!

PR #2579, opened 2024-10-18
Separate motor nominal vs actual diameter

### LINKED ISSUE #2569: Include Loki 76mm motors in 75mm motor filter
Loki motors are listed as 76mm but apparently fit in the same motor mounts as 75mm motors.  When the motor filter slider in the lower right is set to 75mm only (or, match mount size), the Loki motors should be included.

JoePfeiffer:
In the motor diameter filter, we now separate a motor's "nominal" and "actual" diameter. The nominal diameter is displayed to the user, while the actual diameter is used for filtering. The intent here is to allow Loki 76mm motors to pass the 75mm filter.

Also adds 32mm motors (Aerotech has at least one listed on their website).

In the following screenshot, note that (in English, anyway) the motor diameter is explicitly stated to be nominal.

motorfilter

Fixes #2569


neilweinstock:
This is exactly the sort of thing I was hoping for.  Thumbs up.

PR #2585, opened 2024-10-21
Aborted sim status

JoePfeiffer:
Add a status to mark simulations as having aborted, and display a "stop sign" icon for those simulations. The screen shot shows an aborted sim, and the tooltip that comes up when hovering over the icon (the tooltip is unchanged in this PR)
abort-status


SiboVG:
Very good, thanks @JoePfeiffer! In the screenshot, in the warnings column, there is only a warning. Should a sim abort also be added as a simulation error?

JoePfeiffer:
> > That was my first thought; then decided simply marking the sim as aborted in the left column got the idea across at least as well as adding it to the Warnings column would have. It could certainly also be added there if you like.
> 
> I believe it could be useful to also have it as an error, so if you open the simulation configuration dialog, and go to the Warnings tab, that the sim abort message is there as well. Some users may not notice the stop icon.

Ah, missed that. Yes, I'll add it.

JoePfeiffer:
Would it make sense to combine the first two columns (status and Warnings) into a single Status column showing both the simulation status and warnings? Then the simulation configuration Warnings tag would become a Status and Warnings tab.

SiboVG:
> Would it make sense to combine the first two columns (status and Warnings) into a single Status column showing both the simulation status and warnings? Then the simulation configuration Warnings tag would become a Status and Warnings tab.

I would keep them separated, because:
- IMO they are different concepts.
- A combined column can get crowded, because you have a status icon, warning icon + warning count, and error icon + error count.
- Currently, you can double-click the warnings cell to open the warnings tab of the sim config dialog, or double click the status cell when it is out-of-date to update the sim. Combining both columns means you lose that behavior.

SiboVG:
Hey @JoePfeiffer, is this PR ready for review?

JoePfeiffer:
If the simulation aborts, report it in the design window.
design-windo

Put the simulation status in the sim warnings window. If it's an abort, make it red.
warnings-window

SiboVG:
Neat @JoePfeiffer! From very brief testing:
- Don't use `Color.RED`, but use the `UITheme` colors, otherwise the colors look messed up in the dark theme UI.
- I wouldn't use red text for "Out of date" text in the sim status label, I would use orange text. Again, look at the UITheme colors, I think there's already a color for that (probably something like warning color)
- I'm not sure if we should add the "Click Run simulations to simulate" text to the status. I get that you may be trying to help novice users, which I encourage, but they may end up looking for the "Run simulations" button inside the sim config dialog, instead of above the sim table. Maybe we should add a "Run simulation" button next to the "Out of date" label when the sim is outdated?
- If you change something in the sim configuration, e.g. change the wind speed, the sim status is still set to "Up to date".

JoePfeiffer:
>     * Don't use `Color.RED`, but use the `UITheme` colors, otherwise the colors look messed up in the dark theme UI.

Nice catch, thank you.

>     * I wouldn't use red text for "Out of date" text in the sim status label, I would use orange text. Again, look at the UITheme colors, I think there's already a color for that (probably something like warning color)

Yes, red made sense on that before we were using it for the abort status. Switching it to orange is a good idea.

We have a basic problem with color and formatting of many of our strings -- the formatting information is hard-coded in the string itself in the l10n files. Fixing this, both because it'll clean things up and because it'll let us use the UI theme colors, is something we need to pursue. I'm going to say that's a problem for another day, and just go ahead and hard-code the orange for now.

>     * I'm not sure if we should add the "Click Run simulations to simulate" text to the status. I get that you may be trying to help novice users, which I encourage, but they may end up looking for the "Run simulations" button inside the sim config dialog, instead of above the sim table. Maybe we should add a "Run simulation" button next to the "Out of date" label when the sim is outdated?

On reflection, you're probably right. I actually took that text straight from the tooltip; the advice to the new user made more sense there than in the status. I've removed it from the status now.

>     * If you change something in the sim configuration, e.g. change the wind speed, the sim status is still set to "Up to date".

A problem, but not one for this PR. This PR only deals with displaying the sim status; making sure the sim status is actually correct is a different issue.

SiboVG:
> We have a basic problem with color and formatting of many of our strings -- the formatting information is hard-coded in the string itself in the l10n files. Fixing this, both because it'll clean things up and because it'll let us use the UI theme colors, is something we need to pursue. I'm going to say that's a problem for another day, and just go ahead and hard-code the orange for now.

Good point. Not critical indeed, something to put on the never-ending ToDo-list.

JoePfeiffer:
It turns out that whenever a simulation's options are updated the simulation is rerun immediately in the background, so it is indeed still up to date.

SiboVG:
For some reason, I'm still getting the red color instead of the UITheme error color.
image

JoePfeiffer:
> For some reason, I'm still getting the red color instead of the UITheme error color.

One of those was because I somehow failed to notice I'd used Color.red, and the others because of the hardcoded colors in the l10n strings. Upon reflection, I went ahead and changed those to pull the colors from the UI theme (in spite of my comment before); I only moved the color out of the strings, and not the italics tag. The "not simulated" strings have both italics and roman, which should be resolved before moving that out of the strings. I note that's the only simulation status that includes helpful advice to the user, so I'm not sure it should be doing so... but that's a discussion for another PR.

PR #2593, opened 2024-11-15
Filter out events from components that are no longer attached immediately after stage separation

JoePfeiffer:
It is possible for a FlightEvent, scheduled by a component of one stage of a rocket, to end up on the event queue for another stage and still be on the queue after the second stage lands. For instance, in FlightEventsTest.testMultiStage(), the center booster doesn't burn out until after the side boosters are on the ground. This results in a spurious EventAfterLanding warning.

In addition, there is a bug in the test causing events from the side booster branch to not be checked, so the bug had not been noticed until now.

This PR fixes the bug in the test, and also filters the events in the queue immediately after stage separation to eliminate the unattached events.

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
+++ b/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
@@ -342,16 +342,6 @@ public class BasicEventSimulationEngine implements SimulationEngine {
 				}
 			}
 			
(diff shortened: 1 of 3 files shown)


SiboVG:
Don't know how you discovered this, but thanks for the fix Joe!

JoePfeiffer:
> Don't know how you discovered this, but thanks for the fix Joe!

When I fixed the unit test bug there it was!

PR #2594, opened 2024-11-17
Don't set open airframe warning if rocket is unstable, and clarify wording

### LINKED ISSUE #2553: [Bug] Unnecessary "Forward end of airframe is open" warnings in Booster stage of "Three-stage rocket"

JoePfeiffer:
This PR identifies rockets (and rocket components) that are aerodynamically unstable, and does not set an open airframe warning in that case. The intent is to avoid spurious open airframe warnings on typical low power boosters after separation. If the booster is stable (as is the case for some high power boosters) the warning is set.

It also clarifies the wording of the warning itself; rather than simply stating that the airframe is open it states that this will case inaccurate altitude predictions.

Here's the warning as it appears currently, for the two stage rocket example:
old-warning

And here it is with this PR (note this example rocket is one for which the warning is relevant, so it still appears).
new-warning

Fixes #2553


SiboVG:
Thanks @JoePfeiffer! The warning from #2553 is now gone, but if I disable the sustainer, I would expect there to be open airframe warning. There is a sim abort, which is good - though it's not displayed in the rocket design view, I guess you need to merge your branch with unstable?
image

JoePfeiffer:
I went back and forth on whether the warning should appear in that case, and don't really have a strong opinion. Since @neilweinstock seems to be dealing with the most confused users about this, we should get his thoughts.

Do you mean the branch that is #2585? You raised some good points on that one that need to be addressed before it gets merged. This was kind a detour for a couple of days.

neilweinstock:
When you're only enabling the booster stages, it is probably most "correct" to show the warning, but since no one is ever actually going to fly the rocket like that, I'd be most happy not to display it.

Note that this change, while nice, won't address all the TARC folks who are splitting the rocket in the middle for deployment and most likely have stable bottom halves (so it should show the warning.)  That said, the TARC example rocket currently in the program does *not* generate this warning, and I don't know why (the bottom half is certainly stable).  I remain confused about this whole thing.

JoePfeiffer:
> Note that this change, while nice, won't address all the TARC folks who are splitting the rocket in the middle for deployment and most likely have stable bottom halves (so it should show the warning.) That said, the TARC example rocket currently in the program does _not_ generate this warning, and I don't know why (the bottom half is certainly stable). I remain confused about this whole thing.

There's a separate heuristic already in place -- once the rocket's speed is down to 20% of its max, it stops recording warnings completely (presumably so we don't get high angle of attack warnings at apogee and the like). Setting separation to occur at motor burnout will turn the warning up (and also the high speed recovery device, of course!).

Now I'm curious -- do we have a .ork from a team having the issue? I wonder how they've got their separation set?

neilweinstock:
I think this one qualifies: https://www.rocketryforum.com/threads/openrocket-issue-forward-end-of-airframe-is-open-radius-is-0-transition.189506/post-2663594

JoePfeiffer:
> I think this one qualifies: https://www.rocketryforum.com/threads/openrocket-issue-forward-end-of-airframe-is-open-radius-is-0-transition.189506/post-2663594
I'm not finding the .ork file?

neilweinstock:
> > I'm not finding the .ork file?

You're right.  I'll ask for it.

SiboVG:
Looks great on first sight, but I found an anomaly.

Looking at a Three-stage rocket example, if I rerun all simulations, there are no warnings, as expected.
image

If I again disable the sustainer, I correctly get the warning in the design window, but if I then re-run all simulations, I don't get a warning in sim 2 - even though that's the flight config with the disabled sustainer. Instead, I now get two warnings in the other sims.
image

JoePfeiffer:
Those anomalies are also present in current unstable -- see #2595. But you're right, I'd missed before that the impossible warnings aren't in the configuration with the disabled sustainer.

SiboVG:
> Those anomalies are also present in current unstable -- see #2595. But you're right, I'd missed before that the impossible warnings aren't in the configuration with the disabled sustainer.

Oh okay. In that case, I propose to add this new anomalous behavior to #2595 and to merge this PR.

Issue #2584, opened 2024-10-21
[Bug] Material search test sometimes fails in unit tests


JoePfeiffer:
### Describe the bug

Under conditions I can't reliably duplicate, the materials database search sometimes returns materials named e.g. <[material:Aluminum]> instead of &lt;Aluminum&gt; which causes unit tests to fail.  When unit tests are re-run, it succeeds.  It does seem like every time one of these tests fails they both do.  Cutting and pasting from the unit test results (note -- edited to replace "less than" and "greater than" signs with \&lt; and \&gt; because otherwise the display of the log interprets them as HTML tags):

testFindMaterialByTypeAndName()

org.opentest4j.AssertionFailedError: expected: &lt;Aluminum&gt; but was: <[material:Aluminum]>
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at app/org.junit.jupiter.api@5.10.2/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
	at app/info.openrocket.core@23.09.SNAPSHOT/info.openrocket.core.database.MaterialDatabaseTest.testFindMaterialByTypeAndName(MaterialDatabaseTest.java:61)
	at java.base@17.0.13-ea/java.lang.reflect.Method.invoke(Method.java:569)
	at java.base@17.0.13-ea/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base@17.0.13-ea/java.util.ArrayList.forEach(ArrayList.java:1511)

testFindMaterialByTypeNameAndDensity()

org.opentest4j.AssertionFailedError: expected: &lt;CustomMaterial&gt; but was: <[material:CustomMaterial]>
[...]

PR #2605, opened 2024-12-03
New installer config for 24.12

JoePfeiffer:
, including both amd64 and aarch64 architectures, and accounting for changes from gradle build environment.

Assumes jar file is in build/libs/OpenRocket.jar

Note -- in testing, the Windows amd64 installer works. The Linux amd64 installer installs OR, but attempting to run it gives the following stack trace (apologies for the screenshot; I don't seem to be able to copy the text from the window):
stacktrace


neilweinstock:
What is the purpose and/or benefit of an AMD64 build?

JoePfeiffer:
> What is the purpose and/or benefit of an AMD64 build?

Do you mean an Arm (Install4j wants to call it aarch64, and I have no fervent objections) build? AMD64 is the only architecture we've supported on Linux and Windows up until now, and is the architecture supported by macOS Intel. It's aarch64 that's new.

neilweinstock:
> > What is the purpose and/or benefit of an AMD64 build?
> 
> Do you mean an Arm (Install4j wants to call it aarch64, and I have no fervent objections) build? AMD64 is the only architecture we've supported on Linux and Windows up until now, and is the architecture supported by macOS Intel. It's aarch64 that's new.

Oh, I'd have thought we were calling it X86 or X86_64 or somesuch.  Never mind.

JoePfeiffer:
A bit more info -- the system installed jvm (from Debian repositories) runs the command line constructed in the launcher with no issues. The jvm bundled in the installer shows the behavior described above.

joseph@valiant$ java --version
openjdk 17.0.13 2024-10-15
OpenJDK Runtime Environment (build 17.0.13+11-Debian-2)
OpenJDK 64-Bit Server VM (build 17.0.13+11-Debian-2, mixed mode, sharing)

joseph@valiant$ jre/bin/java --version
openjdk 17.0.13 2024-10-15 LTS
OpenJDK Runtime Environment (build 17.0.13+12-LTS)
OpenJDK 64-Bit Server VM (build 17.0.13+12-LTS, mixed mode)

SiboVG:
I have updated the `DS_Store` to the new version (so the macOS installer displays correctly) + added a helper script to automatically generate the new DS_Store in the future. I've also updated the install4j config to install4j version 11. @JoePfeiffer you will need to upgrade your install4j version, I've sent you a mail with the upgraded license key.

Still haven't looked into the amd64 bug.

SiboVG:
Ha, interesting, I actually think the amd64 bug is related to #2584. It looks like the unit tests are failing because of an issue with the rocket substitutor.

SiboVG:
Hm, I don't have an AMD64 Linux build available, I can only run ARM64 on my MacBook, so I can't really debug this issue...

@JoePfeiffer does the exception also occur when running directly from the JAR or only from the packaged installer?

JoePfeiffer:
> Hm, I don't have an AMD64 Linux build available, I can only run ARM64 on my MacBook, so I can't really debug this issue...
> 
> @JoePfeiffer does the exception also occur when running directly from the JAR or only from the packaged installer?

That's what I'm finding craziest about it. It's the java in the packaged installer.  If I duplicate the command line at the bottom of the installed script but use my system java, it works. If I use the Java from the installer and just run the jar file, i get the guice error.

SiboVG:
> That's what I'm finding craziest about it. It's the java in the packaged installer. If I duplicate the command line at the bottom of the installed script but use my system java, it works. If I use the Java from the installer and just run the jar file, i get the guice error.

Did you test different JRE configurations in install4j?

JoePfeiffer:
> > That's what I'm finding craziest about it. It's the java in the packaged installer. If I duplicate the command line at the bottom of the installed script but use my system java, it works. If I use the Java from the installer and just run the jar file, i get the guice error.
> 
> Did you test different JRE configurations in install4j?
That's next.

PR #2606, opened 2024-12-06
Allow for configurable maximum simulation time

SiboVG:
This PR allows you to set a custom max simulation time in the simulation config dialog.

An example for a max 7 seconds time in A simple model rocket:
image

image

I do notice that this works as expected for the altitude and acceleration, but the velocity stops one sample early. @JoePfeiffer do you have a clue why this is?

The max time is stored for each sim with the `maxtime` attribute in the .ork file.

```XML
<conditions>
  <configid>da326836-0959-4c94-bcd5-49dee07235a4</configid>
  <launchrodlength>1.0</launchrodlength>
  <launchrodangle>0.0</launchrodangle>
  <launchroddirection>90.0</launchroddirection>
  <windaverage>2.0</windaverage>
  <windturbulence>0.1</windturbulence>
  <winddirection>1.5707963267948966</winddirection>
  <wind model="average">
    <speed>2.0</speed>
    <direction>1.5707963267948966</direction>
    <standarddeviation>0.2</standarddeviation>
  </wind>
  <windmodeltype>Average</windmodeltype>
  <launchaltitude>0.0</launchaltitude>
  <launchlatitude>45.0</launchlatitude>
  <launchlongitude>0.0</launchlongitude>
  <geodeticmethod>flat</geodeticmethod>
  <atmosphere model="isa"/>
  <timestep>0.05</timestep>
  <maxtime>7.0</maxtime>
</conditions>
```


neilweinstock:
Is this intended to go into beta 1?

SiboVG:
> Is this intended to go into beta 1?

Not necessarily, though it is a fairly simple PR.

JoePfeiffer:
> Okay, I'm actually seeing the bug (?) of the acceleration missing the last sample also on unstable, so it's not a result of this PR. Again, pinging @JoePfeiffer, is this normal? Does the acceleration require data from the next time sample and is that the reason why it's missing the last sample?

I haven't had time to look at it, but it's probably (unfortunately) normal. A bunch of the data for a time step is written before calculation, and other data is written after.  There isn't a single atomic write for each step.

SiboVG:
I'm confident enough that this is a small, non-breaking change (also all unit tests pass) to merge this PR for the beta. I'll adjust the release notes.

HappyCanuck:
Great news, I needed this. What will be the max. allowed value we can enter? I've often exceeded the current limit of 1200 secs.

SiboVG:
> Great news, I needed this. What will be the max. allowed value we can enter? I've often exceeded the current limit of 1200 secs.

There's currently no upper limit.

PR #2612, opened 2024-12-15
First round of tweaks to Simulation Warnings tab

neilweinstock:
I am making some cosmetic changes (already discussed on Discord) to the Simulation Warnings panel, and one small tweak to the results table.  I don't think this has to go in before the beta, but it could, depending on when the beta goes out.

First, I left-justified the warnings column in the results table.  Previously it was right-justified, and that caused it to smush against the simulation name.  I would prefer it (and most other columns) to be centered, but BorderLayout.CENTER is pushing things to the left, not centering, when I try it.

Old on the left, new on the right:
image

Next, in the Warnings tab, I did two things:
1) Relabeled the warning levels as "Critical(s)", "Warning(s)", and "Informational(s)".  These words are all being used as nouns.  I've bounced back and forth a few times on how I like this, but I ultimately prefer not having the word "Warning" repeated over and over.  I'm open to suggestions on better approaches here.

2) When a warning category has zero entries, it is not shown at all.  

Once again, old on the left and new on the right:
image

What is missing (other than code cleanup) is a message that says "No warnings" or somesuch when all categories are zero, so it doesn't just show a blank tab.


SiboVG:
Cleaned up the code a bit and removed empty warning sections.

SiboVG:
Added descriptions beneath the warning sections

Screenshot 2024-12-16 at 02 30 57

SiboVG:
Adjusted the warnings tab so the warning lists are in a JList instead of a JScrollPane, and instead put the entire warnings tab content in a scroll pane:
image


https://github.com/user-attachments/assets/16901d5f-7ae3-43a0-a858-22c4a3f81ff7

PR #2634, opened 2024-12-24
[Typo] Align spellings for contributing translations

jefferyjohn:
This PR is a patch for spellings in the contribution guide. 

* his/her are possessive, so 'they' should be possessive as well
* therefor generally means 'in return for' while therefore generally means 'as a result', with the latter seeming to fit better

From a copy editing perspective, it may be more inclusive and flow more smoothly to shorten 'his/her/their' to 'their', as this would retain the meaning, but this would be a more stylistic choice for the author. 

Hope this helps! Happy to make any edits and look forward to contributing more.


SiboVG:
Thank you for the change!

PR #2635, opened 2024-12-26
Add option to export all component instances or only singles

SiboVG:
This PR adds a new option in the OBJ export dialog to export all component instances, or just a single instance. This is useful when wanting to 3D print a fin set.

Screenshot 2024-12-26 at 19 03 48

If the checkbox is checked:
image

If it is unchecked:
image


SiboVG:
I know it's not smart to add features after the beta release, but this is a really simple change (just filtering the component instance list) and a fair number of users have requested this.

PR #2661, opened 2025-01-08
directly test for non-zero thrust when recovery device deploys, instead of inferring from motor state.

JoePfeiffer:
Inference was wrong anyway, as it didn't check for a motor that hadn't started burning yet.


JoePfeiffer:
Marking it as a draft until I can see if we have a unit test for it

neilweinstock:
That looks much better, much more direct and straightforward (and correct :)).

PR #2665, opened 2025-01-11
Import wind data from CSV

CSutter5:
This PR adds a button on the multilevel winds panel to import wind level data from a .CSV file.
Right now its assuming that the units are meter, degrees, and knots as the the units provided by the [NOAA SPC's Observed Soundings Archive](https://www.spc.noaa.gov/exper/soundings/)


neilweinstock:
Thanks, let us know when you think it's ready.  Screenshots and/or videos of your feature are helpful if you can provide them.

CSutter5:
I think its 99% done.
I'm just trying to think of what units should be used. Here are my ideas
1. Use the user selected units
2. Use the units that are [called out here](https://openrocket.readthedocs.io/en/latest/dev_guide/codebase_walkthrough.html#units-used-in-openrocket)
3. Keep it how I currently have (meters, knots, degrees)
Let me know what you think would be best

I've also included a video how how the feature is used *(ignore the kind of weird rocket)*

neilweinstock:
That looks excellent, thanks for you contribution!

I am honestly not sure about the units question... will wait to see what others think.

JoePfeiffer:
Definitely use SI units, as that's what OR consistently does internally. It would be nice if each value could have a specified unit (like we'd do if the input were XML), but trying to do that in a CSV file would be a mess.

CSutter5:
Everything I can think of is done, let me know if there is anything else you think should be added

SiboVG:
I'd like to test and code-review this PR, I'll try to make some time next week.

SiboVG:
> I'd like to test and code-review this PR, I'll try to make some time next week.

I had a busy week at work. Will try to look at the end of the coming week.

SiboVG:
A couple of notes:
1.  Because of the extra button, the sim config is unnecessarily wide. Maybe the "Import levels" button should be below the "Visualize levels" button
image

2. I'm not a fan of the "Import levels" icon, I would prefer something like the 'File > Import' icon (the blue arrow coming in)
3. It is not at all clear to the user what the CSV format should be. We should at the very minimum add a description text in the CSV file dialog.

@CSutter5 could you please post an example CSV file so I can more easily test the feature?

CSutter5:
Thanks for taking your time to look at it
Here is a test file [wind levels.csv](https://github.com/user-attachments/files/18546838/wind.levels.csv)
I will make those changes and push them

CSutter5:
I have pushed all of the changes, let me know how they look

SiboVG:
> I have pushed all of the changes, let me know how they look

Thanks! I'll try to make some time to review it.

SiboVG:
@CSutter5 I've made some improvements to the code, but I also added some TODO's in `importLevels`:
- The CSV separator should be customizable instead of simply being a comma. We do this for all our other CSV file handling. You can check the `CsvOptionPanel` code.
- I would use degrees for the input of the direction, and then convert it to radians when you parse the double. Users will probably be more keen on working in degrees than in radians.
- I don't understand what the `altitude != -9999 && speed != -5143.9296 && direction != -174.51` code does
- Maybe we should make the `stddev` optional and use default values for when it's not present in the CSV?

SiboVG:
@CSutter5 could you please update your unstable branch on GitHub, so we can merge the changes from unstable into this branch?

SiboVG:
I added a few more improvements:
- added a field separator configuration in the file chooser dialog
- removed the value sanity check in the SimulationConditionsPanel, since that's handled downstream by PinkNoiseWindModel
- added a warning stating that existing wind levels will be overwritten

SiboVG (approved):

SiboVG:
I think we're good to merge this PR. I know it's a new feature and we're in pre-release mode for OR 24.12 right now, but we're shipping a completely new feature (multi-level wind) anyway, so there's less harm in adding to that feature with this PR I guess. I also added extensive unit testing.

SiboVG:
Thanks a lot for your contribution and your swift response and cooperation @CSutter5, we appreciate it!

HappyCanuck:
Late to the party, I just came across this post... Regarding altitude units:

I always use ft for altitude, it's the standard unit in aviation and amateur launch sites have their waivers listed in ft as well.
The Upper wind forecast sites I use (FAA, NavCan) are all listed in ft. They always use the same default levels:

3000
6000
9000
12000
18000
24000
30000
34000
39000
45000
53000

I modified the original MultiLevel plugin a while ago to reflect this: https://github.com/HappyCanuck/MultiLevelWind

In meteorology meters are used, but their reporting is not using standard altitudes; they use standard pressure levels instead. So the altitude listed will always differ from one day to another, making it harder to establish a baseline when comparing two different launches.

I think it's great to have the possibility to import a CSV as it will save users a lot of time, but I'd recommend the default altitudes to be in ft, wind speed in kts and directions in degrees, from 0 to 359.

For more in-depth analysis (e.g. dispersion analysis), this is where I would import data from meteorology sites (openmeteo is an awesome source and they have a very good API to customize the data). So meters would likely work be better then.

At least offer the user an option to select default units when importing a file?

SiboVG:
> Late to the party, I just came across this post... Regarding altitude units:
> 
> I always use ft for altitude, it's the standard unit in aviation and amateur launch sites have their waivers listed in ft as well. The Upper wind forecast sites I use (FAA, NavCan) are all listed in ft. They always use the same default levels:
> 
> 3000 6000 9000 12000 18000 24000 30000 34000 39000 45000 53000
> 
> I modified the original MultiLevel plugin a while ago to reflect this: https://github.com/HappyCanuck/MultiLevelWind
> 
> In meteorology meters are used, but their reporting is not using standard altitudes; they use standard pressure levels instead. So the altitude listed will always differ from one day to another, making it harder to establish a baseline when comparing two different launches.
> 
> I think it's great to have the possibility to import a CSV as it will save users a lot of time, but I'd recommend the default altitudes to be in ft, wind speed in kts and directions in degrees, from 0 to 359.
> 
> For more in-depth analysis (e.g. dispersion analysis), this is where I would import data from meteorology sites (openmeteo is an awesome source and they have a very good API to customize the data). So meters would likely work be better then.
> 
> At least offer the user an option to select default units when importing a file?

Could you please create a new feature request for selecting the units in the import wind level from CSV file dialog?

PR #2666, opened 2025-01-12
Sim warnings updates

neilweinstock:
A couple of small tweaks:

1) In the simulation warnings panel, don't show the number of warnings on each header.  You can see the warnings right in front of you, don't need to see the number

2) In the sim results table warnings column, only show the number if its greater than 1.  I am not sure about this change.

I am confident about (1), less so about (2).

image

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
@@ -1263,9 +1263,7 @@ public class SimulationPanel extends JPanel {
 			int nrOfInfoWarnings = warnings.getNrOfInformationalWarnings();
 
 			if (nrOfCriticalWarnings > 0) {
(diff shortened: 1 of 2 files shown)


SiboVG:
I like change (1), but I would prefer to revert change (2). I find it more consistent and less confusing to always show the count, even if it means slightly more visual clutter.

neilweinstock:
> I like change (1), but I would prefer to revert change (2). I find it more consistent and less confusing to always show the count, even if it means slightly more visual clutter.

I'm OK with that.  It sounded good in my head but when I saw it in front of me I was unsure.  I've reverted it back.

neilweinstock:
Hah, and now I have no idea what it's failing.  Good times.

SiboVG:
> Hah, and now I have no idea what it's failing. Good times. Compiles and runs fine on my local machine...

It's not the fault of this PR. There's an integration test that needs to have relaxed constraints, due to simulation variability.

PR #2672, opened 2025-01-16
Make all (most?) warnings take a standard syntax

### LINKED ISSUE #2669: "Zero volume body" warning shows incorrectly
The "zero volume body" warning has some issues:
1) Priority is wrong, should be "low", shows as "normal"
2) Format of the warning is different from others
3) Once moved to low priority, should simply say "Zero volume bodies", since the "may simulate inaccurately" part is already stated in the header above the Information warning box.
image

This ORK file shows the problem.
[plasma_dart_ii.zip](https://github.com/user-attachments/files/18411750/plasma_dart_ii.zip)

JoePfeiffer:
This PR makes all warnings look like
Problem: "component", "component"
both for uniformity and for clarity. Also properly sets zero-volume body tube priority.

Fixes #2669

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanCalculator.java
@@ -360,7 +360,7 @@ public class BarrowmanCalculator extends AbstractAerodynamicCalculator {
 						// Check for phantom tube
 						if ((sym.getLength() < MathUtil.EPSILON) ||
 							(sym.getAftRadius() < MathUtil.EPSILON && sym.getForeRadius() < MathUtil.EPSILON)) {
-							warnings.add(Warning.ZERO_VOLUME_BODY, sym.getName());
+							warnings.add(Warning.ZERO_VOLUME_BODY, sym);
 						}
 
 						// check for gap or overlap in airframe. We'll use a textual comparison to see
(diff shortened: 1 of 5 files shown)


neilweinstock:
Does it "fix" the "special" warnings as well (e.g. Large AOA, high deployment speed)?

JoePfeiffer:
> Does it "fix" the "special" warnings as well (e.g. Large AOA, high deployment speed)?

No, I started down that rabbit hole but decided to just fix the big one. I actually think the large AOA warning is OK; it gives the warning and then the angle of attack in parentheses. There is no component displayed since it's the whole rocket.

Actually, since it doesn't throw that warning any more when the rocket is about to tumble I'm not sure it's even possible to trigger it! I was able to do it before by setting a really high wind speed, but the wind speed seems to be capped at 10 m/s now.

But hmmm, there may be something to do with the high speed deployment...

neilweinstock:
> Actually, since it doesn't throw that warning any more when the rocket is about to tumble I'm not sure it's even possible to trigger it! I was able to do it before by setting a really high wind speed, but the wind speed seems to be capped at 10 m/s now.

I was able to trigger it in the same way, with a high wind and low-powered motor.  It's not easy, and probably low priority.

> But hmmm, there may be something to do with the high speed deployment...

It would definitely be helpful to identify the component for that one.

neilweinstock:
Looks good to me.

PR #2686, opened 2025-01-24
[#2662] Add File->Properties Menu

### LINKED ISSUE #2662: Add File -> Properties menu item
File -> Properties should take you to the same config dialog as editing the "Rocket".  It'll make it a bit easier to find, since that's a pretty standard menu item in many apps.

MiguelECL:
Added the "Properties" menu item in the "File" parent menu. Clicking on properties allows editing on the Rocket properties such as design name, designer, etc. Fixes #2662.


SiboVG:
Thank you for the change @MiguelECL! Some small recommendations:
1. I would move the menu item to above the "Close design" item because the top menu item group already has a lot of items, and in other software the Properties option is often lower in the dropdown list (from my experience)

image

2. Change the shortcut to `Ctrl/Cmd + i`, I feel like this is easier than the `Ctrl/Cmd +  ^` shortcut. `Alt + Enter` is another often used shortcut for File Properties on Windows and Linux, but I prefer `Ctrl/Cmd + i`.

SiboVG:
Thanks for the changes @MiguelECL, appreciate it!

PR #2693, opened 2025-01-30
[#2664] Additional rocket properties

### LINKED ISSUE #2664: Add additional fields to rocket properties
Additional fields in the Rocket Properties dialog would facilitate better file repositories.  In particular, the rocketreviews.com ORK file repository has many files with no additional metadata at all, because the files were mechanically gathered from other places.

I am using that repository as a model.

First there's "type":
image

If the type is anything other than "Original Design" then you are also prompted to enter "Kit Name".  In the case of "Kitbash" you are prompted to enter multiple kit names.  Nowhere does it ask for Kit manufacturer, I don't know how that goes in there.

Then you can also specify "Optimization":
image

I might do this a bit differently, with separate checkboxes for "Flight", "Appearance", and "Construction".  If flight were unchecked we could consider disabling the motor config and simulation tabs (there are pluses and minuses to doing this).

MiguelECL:
Added additional properties such as "Design Type", as well as optimizations (Flight, Appearance, Construction). Functionality that saves these properties is also added. Fixes #2664

### DIFF
--- a/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
+++ b/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
@@ -67,7 +67,7 @@ public class Rocket extends ComponentAssembly {
 	
 	private String designer = "";
 	private String revision = "";
-	private String designType = "";
+	private DesignType designType = DesignType.ORIGINAL;
 	private String kitName = "";
 //	private boolean optimizationFlight = false;
 //	private boolean optimizationAppearance = false;
@@ -127,19 +127,20 @@ public class Rocket extends ComponentAssembly {
 	 * design...).
 	 * @return the design type
 	 */
-	public String getDesignType() {
+	public DesignType getDesignType() {
 		checkState();
 		return designType;
 	}
 
 	/**
 	 * Set the design type of the rocket.
-	 * @param s the design type
+	 * @param type the design type
 	 */
-	public void setDesignType(String s) {
-		if (s == null)
-			s = "";
-		designType = s;
+	public void setDesignType(DesignType type) {
+		if (type == null) {
+			type = DesignType.ORIGINAL;
+		}
+		designType = type;
 		fireComponentChangeEvent(ComponentChangeEvent.NONFUNCTIONAL_CHANGE);
 	}
 
--- a/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
+++ b/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
@@ -68,9 +68,10 @@ public class Rocket extends ComponentAssembly {
 	private String designer = "";
 	private String revision = "";
 	private String designType = "";
-	private boolean optimizationFlight = false;
-	private boolean optimizationAppearance = false;
-	private boolean optimizationConstruction = false;
+	private String kitName = "";
+//	private boolean optimizationFlight = false;
+//	private boolean optimizationAppearance = false;
+//	private boolean optimizationConstruction = false;
 	
 	
 	// Flight configuration list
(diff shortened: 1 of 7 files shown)


MiguelECL:
Have yet to add conditional text field(s) for adding Kit name(s) but this first draft has almost all functionality ready.

core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
```diff
@@ -67,6 +68,10 @@ public class Rocket extends ComponentAssembly {
 	
 	private String designer = "";
 	private String revision = "";
+	private String designType = "";
+	private boolean OptimizationFlight = false;
```
SiboVG, line 72: Please use camelCase for variables (so `optimizationFlight`)

swing/src/main/java/info/openrocket/swing/gui/configdialog/RocketConfig.java
```diff
@@ -11,11 +11,7 @@
 import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 
-import javax.swing.JDialog;
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JScrollPane;
-import javax.swing.JTextArea;
+import javax.swing.*;
```
SiboVG, line 14: Please don't use wildcard imports (if you use InteliJ, you can change a setting to disable this automatic behavior)

swing/src/main/java/info/openrocket/swing/gui/configdialog/RocketConfig.java
```diff
@@ -59,7 +55,37 @@ public RocketConfig(OpenRocketDocument d, RocketComponent c, JDialog parent) {
 		GUIUtil.setTabToFocusing(designerTextArea);
 		designerTextArea.addFocusListener(textFieldListener);
 		this.add(new JScrollPane(designerTextArea), "wmin 400lp, height 60lp:60lp:, grow 30, wrap para");
-		
+
+		//// Design Type
+		this.add(new JLabel(trans.get("RocketCfg.lbl.Designtype")), "top, pad 4lp, gapright 10lp");
+		String[] options = {"Original Design/Other", "Commercial Kit","Clone of a Commercial Kit","Upscale of a Commercial Kit",
```
SiboVG, line 61: Please add these String options as a translation key

swing/src/main/java/info/openrocket/swing/gui/configdialog/RocketConfig.java
```diff
@@ -59,7 +55,37 @@ public RocketConfig(OpenRocketDocument d, RocketComponent c, JDialog parent) {
...
+				"Downscale of a Commercial Kit", "Modification of a Commercial Kit", "Kit Bash of Commercial Kits"};
+		JComboBox<String> designTypeDropdown = new JComboBox<>(options);
+		designTypeDropdown.setSelectedItem(rocket.getDesignType());
+		designTypeDropdown.addActionListener(e -> rocket.setDesignType((String) designTypeDropdown.getSelectedItem()));
+		this.add(designTypeDropdown, "growx, wrap para");
+
+		//// Optimization
```
SiboVG, line 68: You should use radio buttons for this

SiboVG (changes requested):
Thanks for the changes @MiguelECL! I haven't tested this yet, but looking at the code, I added some remarks.

A general remark is that I'm not a fan of the Optimization settings. It makes it look like selecting either of these options makes a change to the program, which it clearly doesn't. I would not include this functionality yet until we actually do something useful with it, otherwise it just gets confusing. We could also add a warning message, but that's just extra clutter. We don't need the optimization property. Pinging @neilweinstock.

MiguelECL:
Should I remove the "Optimization" completely or leave it commented for further development?

neilweinstock:
I won't have a chance to test this code for a little while... screenshots would be helpful.

This is all in service to the need for sites such as RocketReviews to be able to slurp up OR files and extract useful metadata.  It is also "nice to have", even if most folks won't fill it in, and it doesn't *hurt* to have extra unused fields that are almost never visible to the user.

The "Optimization" parameter originally proposed is exactly as implemented on RocketReviews.  I am unsure of the best way to implement it within the program (as discussed in the original github issue), and would not object to simply deferring it until we can properly figure it out.  If this were earlier in the release cycle I would say "go for it" and we'll clean it up before release, but since we're in the 11th hour (and 58 minutes) it's probably best not to insert something half-baked.

I would be fine with leaving it commented for further development, even if it's possible that we will never follow up on it.

Oh, and thanks for the contribution!!!

SiboVG:
> Should I remove the "Optimization" properties completely or leave them commented for further development?

As Neil also states, I would comment it out.

MiguelECL:
Have to resize the dialog when the text field for the kit name(s) appears, as users have to do it manually right now.

SiboVG:
I have modified the code to make the saving of the design type independent of the translation key. This ensures that when files are distributed to users with different language settings, the design type loading still works.

The rest looks good, thanks a lot for the change @MiguelECL!

SiboVG (approved):

PR #2716, opened 2025-02-10
Fix 2675

JoePfeiffer:


JoePfeiffer:
Breaking this apart into two PRs

PR #2719, opened 2025-02-14
 [#2676] Migrate old net.sf.openrocket plugins at startup 

### LINKED ISSUE #2676: [Feature Request] Create interface code for plugins/scripts that use the old `net.sf.openrocket` package
Since OR 24.12, we've renamed the codebase package from `net.sf.openrocket` to `info.openrocket.core` and `info.openrocket.swing`. This means that we break compatibility with plugins/scripts that reference the old `net.sf.openrocket` packages.

When loading plugins/scripts, check if there are `net.sf.openrocket` package references. If so, first try to find the equivalent file in `info.openrocket.core`. If the file exists, replace the `net.sf.openrocket` reference with  `info.openrocket.core`. If it does not exist, try to search in `info.openrocket.swing`. If the file exists, replace the `net.sf.openrocket` reference with  `info.openrocket.swing`. If it does not exist, throw an error.

Screw the old plugins, but that's not a friendly solution :)

**Additional context**
Marking this as a release blocker.

SiboVG:
This PR fixes #2676. When OR scans the Plugins directory at startup, it checks if the JAR uses the old package structure (`net.sf.openrocket`). If it does, and it does not find a migrated version of that JAR, it starts the migration process. The migration process involves creating a copy of the JAR file with suffix `-migrated-new`. E.g. for the `CDOverride.jar` plugin, a new file is created `CDOverride-migrate-new.jar`. The ASM library is then used to modify the byte-code if the `XX-migrate-new.jar` to migrate references from `net.sf.openrocket` to `info.openrocket.core/swing`. Finally, when the UI loaded in `SwingStartup`, it checks if there are any `XX-migrated-new.jar` files present in the Plugins directory. If so, it notifies the user and renames the JAR file from `XX-migrated-new.jar` to `XX-migrated.jar`. I had to use this workaround because I can not yet use Swing components while loading the Plugins at startup because for some reason this messes with the Guice injections.

image

In the end, you are left with both the original JAR (so it works with older installed versions of OR) and the migrated JAR (compatible with OR 24.12).
image

I also added some backward compatibility code for the popular MultiLevelWind plugin (which is now built-in in OpenRocket, but users will still have the plugin loaded in their old designs).

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -50,6 +50,9 @@ dependencies {
 
     implementation 'de.javagl:obj:0.4.0'
 
+    implementation group: 'org.ow2.asm', name: 'asm', version: '9.7.1'
+    implementation group: 'org.ow2.asm', name: 'asm-commons', version: '9.7.1'
+
     implementation group: 'jakarta.activation', name: 'jakarta.activation-api', version: '2.1.2'
     implementation group: 'org.glassfish.jaxb', name: 'jaxb-runtime', version: '4.0.5'
     implementation group: 'org.glassfish', name: 'jakarta.json', version: '2.0.1'
--- a/core/src/main/java/info/openrocket/core/plugin/JarMigrationHelper.java
+++ b/core/src/main/java/info/openrocket/core/plugin/JarMigrationHelper.java
@@ -1,0 +1,228 @@
+package info.openrocket.core.plugin;
+
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.ClassWriter;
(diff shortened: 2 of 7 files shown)


JoePfeiffer:
Wow. Going in and editing the jar never would have occurred to me!

SiboVG:
> Wow. Going in and editing the jar never would have occurred to me!

I tried other ways (e.g. mirroring the code to a net.sf.openrocket package), but they all failed or were inelegant. This works well for the 2 plugins I tested.

Issue #2146, opened 2023-03-27
[Enhancement] Focus rocket config dialog when opening a file


SiboVG:
When opening a file where the rocket config dialog pops up, the focus is not set on the config dialog, but on the BasicFrame. Hence, I can not press the escape key to close the config dialog (a habit I developed). This appears to have happened recently, as this is the first time I've noticed it...
You can see here that the config dialog is not focused (the selected text is grey).
<img width="1555" alt="image" src="https://user-images.githubusercontent.com/11031519/228079994-3ecc74f8-532e-4cb9-96ac-d91c94af8fa7.png">

SiboVG:
New information: this issue only occurs when I close all design files, so OpenRocket is in its "always remain open" state, and then open e.g. "A simple model rocket". I have the correct behavior when I already had a design file open and then open "A simple model rocket".

JonathanDeLaCruzEncora:
Hi, I gave this issue a try, so if there are any changes, please let me know.

Additionally to the information on this issue, I got the same problem when going full screen mode in a window with an example file and then opening another example.

PR #2720, opened 2025-02-19
[#2146] Focus Rocket Config Dialog when opening a file.

JonathanDeLaCruzEncora:
This change fixes the issue #2146 where the Rocket Config Dialog that opens up when opening an example is not focused.

The same was happening when opening it in full screen mode, so now it is focused when it opens and it can be closed with the 'esc' key or modified.

The change is just requesting focus from the Design Name text field component.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/configdialog/RocketConfig.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/configdialog/RocketConfig.java
@@ -144,8 +144,6 @@ public class RocketConfig extends RocketComponentConfig {
 
 		addButtons();
 		addEasterEgg();
-
-		SwingUtilities.invokeLater(componentNameField::requestFocus);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


neilweinstock:
This is a nice change, thank you for the contribution!

SiboVG:
Thanks for the change. Unfortunately, I'm still experiencing issue #2146 on my machine...

JonathanDeLaCruzEncora:
Sadly I am not able to replicate the issue 🤔, if there is any additional information to provide I'll look into it

SiboVG:
> Sadly I am not able to replicate the issue 🤔, if there is any additional information to provide I'll look into it

Not really, it seems very specific to my machine/setup. I have 2 identical MacBook Pros, but only one of them suffers from this issue...

SiboVG:
@JonathanDeLaCruzEncora I finally got around to testing this on my machine. The issue was that the `ComponentConfigDialog.showDialog` code in `BasicFrame` should have been wrapped in a `SwingUtilities.invokeLater(()` to ensure proper UI execution on the Swing thread. I pushed the code directly to your branch. Thanks for helping us fix this rather annoying bug! 🙂

PR #2729, opened 2025-02-27
Cleanup of CSVSave listener example

### LINKED ISSUE #2696: [Bug] CSV export extension crashes when it can't write to the directory where it's trying to save the file
### Describe the bug

Report from the mailing list:
---------- Bug report ----------
Please include a description about what actions you were performing when the exception occurred:
(You can edit text directly in this window)

If possible, please send us the .ork file that caused the bug.
Include your email address (optional; it helps if we can contact you in case we need additional information):
(Do not modify anything below this line.)
---------- Exception stack trace ----------
net.sf.openrocket.gui.main.SwingExceptionHandler$InternalException: An exception occurred during the simulation at net.sf.openrocket.gui.main.SwingExceptionHandler.handleErrorCondition(SwingExceptionHandler.java:120) at net.sf.openrocket.gui.simulation.SimulationRunDialog$InteractiveSimulationWorker.simulationInterrupted(SimulationRunDialog.java:454) at net.sf.openrocket.gui.simulation.SimulationWorker.done(SimulationWorker.java:109) at java.desktop/javax.swing.SwingWorker$5.run(SwingWorker.java:750) at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.run(SwingWorker.java:848) at java.desktop/sun.swing.AccumulativeRunnable.run(AccumulativeRunnable.java:112) at java.desktop/javax.swing.SwingWorker$DoSubmitAccumulativeRunnable.actionPerformed(SwingWorker.java:858) at java.desktop/javax.swing.Timer.fire

### To Reproduce

1. Reduced Simulation Time step to 0.01s
2. Added simulation extension to add CSV report

### Screenshots / .ork file

[2425_USLI_Peregrine Explorer - FRR_20250129.ork.zip](https://github.com/user-attachments/files/18629714/2425_USLI_Peregrine.Explorer.-.FRR_20250129.ork.zip)

### OpenRocket version

23.09

Windows

### Additional context

_No response_

JoePfeiffer:
We've had a user who didn't notice the CSV export feature try to use the CSV export extension to do it. This will hopefully point people in the right place.

Fixes #2696

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/extension/example/CSVSave.java
+++ b/core/src/main/java/info/openrocket/core/simulation/extension/example/CSVSave.java
@@ -220,7 +220,7 @@ public class CSVSave extends AbstractSimulationExtension {
 
 	@Override
 	public String getDescription() {
-		return "dump a CSV file with a predetermined set of flight variables to a CSV file while running";
+		return "dump a CSV file with a predetermined set of flight variables to a CSV file while running.  Note: this is an example of reading simulation variables from an extension, not a practical way of exporting them. If your actual goal is to dump the simulation variables, use the Export Data tab in the Edit simulation dialog ";
 	}
 
 	@Override
--- a/core/src/main/java/info/openrocket/core/simulation/extension/example/CSVSave.java
+++ b/core/src/main/java/info/openrocket/core/simulation/extension/example/CSVSave.java
@@ -1,9 +1,14 @@
 package info.openrocket.core.simulation.extension.example;
 
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.PrintStream;
+import java.lang.Exception;
 import java.util.Iterator;
+
+import javax.swing.JOptionPane;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import info.openrocket.core.rocketcomponent.FinSet;
 import info.openrocket.core.rocketcomponent.RocketComponent;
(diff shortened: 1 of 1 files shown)


SiboVG:
_(Without looking at the code)_ Wouldn't a proper fix for #2696 be to show an info dialog or something when there is an issue with file permissions?

JoePfeiffer:
> _(Without looking at the code)_ Wouldn't a proper fix for #2696 be to show an info dialog or something when there is an issue with file permissions?

See my comment on #2696. In short "yes", but does that add more complexity to the example than it's really worth? On the other hand, maybe that would also make an example of how to put up the dialog?

SiboVG:
> > _(Without looking at the code)_ Wouldn't a proper fix for #2696 be to show an info dialog or something when there is an issue with file permissions?
> 
> See my comment on #2696. In short "yes", but does that add more complexity to the example than it's really worth? On the other hand, maybe that would also make an example of how to put up the dialog?

I feel like users are quick to overlook the description + it doesn't explain why they got the exception in the first place. Indeed, it may a useful example to using a dialog.

JoePfeiffer:
> > > _(Without looking at the code)_ Wouldn't a proper fix for #2696 be to show an info dialog or something when there is an issue with file permissions?
> > 
> > 
> > See my comment on #2696. In short "yes", but does that add more complexity to the example than it's really worth? On the other hand, maybe that would also make an example of how to put up the dialog?
> 
> I feel like users are quick to overlook the description + it doesn't explain why they got the exception in the first place. Indeed, it may a useful example to using a dialog.

Wound up doing quite a bit of cleanup:
Better follows structure of a simulation: instead of doing all the file create/open/close stuff based on flight events, create/open in startSimulation(); close in endSimulation()

Create output file atomically so multiple sims can run simultaneously with each one reliably getting a new file (previously there was a race condition that could end up with only a single file with everybody trying to write to it)

Add logging code instead of println()'s

Lots more comments telling the user what's going on

PR #2738, opened 2025-03-08
[#2656 & #2654] Complete rewrite of the multi-level wind UI

### LINKED ISSUE #2654: [Feature Request] Scale multilevel winds tab with size of window/allow resizing
The default window size for the multilevel winds pane makes it difficult to read and edit wind parameters, and it does not scale vertically when resizing the window.
image
image
When using more than a handful of wind levels, this makes scrolling and editing them inconvenient.

Either scale the multilevel wind pane to be proportional in height to the Edit Simulation window, or add a grabber to resize the table.

### LINKED ISSUE #2656: [Bug] Missing truncation/rounding of long decimals in Multi Level Winds table
### Describe the bug

While "Altitude" in multi-level winds table shows as integers, clicking an altitude value shows a 12-place decimal value. It appears to be lack of rounding or truncation of the value adjusted to the user's preferred units. Same problem in Wind Speed and Deviation. It's assumed it's possible for any number in the table to be displayed with an over-long decimal component depending upon context.

### To Reproduce

1. Open "A Simple Model Rocket"
2. Change or ensure that units for "Distance" and "Velocity" are set to some imperial unit.
3. Go to "Flight simulations" tab
4. Click "Edit simulation"
5. Choose "Multi-level" in "Wind model to use"
6. Click "Add level" at least 3 times
7. Click in the table on some non-zero Altitude, Speed or Deviation
8. Note long decimal value
...

#### Screenshot(s):
Screenshot 2025-01-04 at 10 59 31 PM
Screenshot 2025-01-04 at 10 59 39 PM

### OpenRocket version

24.12.01 beta 1

macOS

### Additional context

macOS Sonoma 14.7.1 MacBook Pro M3 Max processor

Verified on Windows 10 too

SiboVG:
This PR fixes #2656 and fixes #2654 by completely rewriting the UI of the multi-level wind input. Previously, I attempted to put everything in a `JTable`. However, this solution was not flexible and required some hacks, which caused issues in units and decimal rounding. The new solution is to place the input in a dedicated dialog, which includes the wind profile visualization by default.

The multi-level panel in the sim dialog contains an overview of the wind levels:
image

Demo:

https://github.com/user-attachments/assets/c2132110-3fcf-478f-987d-c01cfa74013d

Notable changes:
- Unit selectors are in the headers and affect all rows, instead having one selector for each row (doesn't make much sense to want to change individual cell units)
- Profile visualization is embedded in the wind edit dialog and dynamically responds to the wind table (highlight the selected row dot, update the units)
- Added tooltip information to the profile visualization dots
- Make it more clear that rows are resorted according to altitude. The moved row and its previous and next row are highlighted in a yellow color. Also when adding a new row and deleting a row, the highlighting is applied.
- Added a "Reset levels" button to reset the table back to its default state

And a bunch of other improvements.

---

This PR also adds a new "Wind direction" flight data type. It didn't make sense before multi-level wind to include this (it remained constant), but for multi-level winds it can change throughout the flight.


SiboVG:
We received feedback from TRF that it was unclear whether the altitude was referenced to ground or to mean sea level. Some users also wanted AGL, others wanted MSL. Therefore, I added a commit to allow you to select MSL or AGL referencing:

https://github.com/user-attachments/assets/69ccd920-5e96-4db3-8d83-1592fa82baec

PR #2744, opened 2025-03-09
[#2739] Move sim launch conditions and options to main tabs

### LINKED ISSUE #2739: [Enhancement] Move "Launch conditions" and "Simulation options" simulation dialog tabs to the main window tabs
I don't like how nested the simulation dialog is. For instance, if you want to change the multi-level wind settings, you have to go to `Settings --> Launch conditions --> Multi-level --> Edit levels`. We could remove one layer by moving the "Launch conditions" and "Simulation options" tabs to the main tabs, and to dissolve the "Settings" tab:

Image

SiboVG:
This PR fixes #2739 by moving the sim launch conditions and options tabs to the main tabs of the sim edit dialog, which results (IMO) in a less confusing UI.

image


neilweinstock:
That appears to show *all* sim options under the "Launch Conditions" tab.  Just a work-in-progress screenshot I presume?

SiboVG:
> That appears to show _all_ sim options under the "Launch Conditions" tab. Just a work-in-progress screenshot I presume?

No, it's final. The goal is to not have the "Launch Conditions" tab and "Simulation options" tab be grouped in a "Settings" tab (as it is now), causing unnecessary nesting. I didn't intend to separate settings from the "Launch Conditions" tab into subtabs.

neilweinstock:
Sorry, my bad, I forgot how that all was before _even though I looked before I replied_.

I'm not sure that Simulator Options are touched very often, or if they warrant equal billing with all the other important stuff up there in the tab bar, but I agree this is clearer.

SiboVG:
> Should the geodetic calculations be with launch conditions? It seems like launch conditions in general are about our model of the outside world, while sim options refer more specifically to how we run the sim. The geodetic calculations seem (to me) to be in the first category.

I guess you're technically correct, but I would keep it in the simulation options, since 99.99% of users will never touch that option, so it's better to "hide" it in the sim options tab.

neilweinstock:
I presume there will be duplicate "Reset to default" and "Save as default" buttons on the "Simulation Options" tab?

SiboVG:
> I presume there will be duplicate "Reset to default" and "Save as default" buttons on the "Simulation Options" tab?

There is a "Reset to default", but no "Save as default".

neilweinstock:
> > I presume there will be duplicate "Reset to default" and "Save as default" buttons on the "Simulation Options" tab?
> 
> There is a "Reset to default", but no "Save as default".

Wouldn't a "Save as default" be just as appropriate there as in Launch Conditions?

I never touch that stuff so just assuming.

neilweinstock:
*sigh*

I got fooled again, I thought the existing buttons applied to both launch conditions and sim options.  

However, I *do* think the set as default button is a good thing to have.

SiboVG:
> However, I _do_ think the set as default button is a good thing to have.

Yup, I've added it in the last commit.

PR #2747, opened 2025-03-11
 [#2725] Include assembly copies in motor count display 

### LINKED ISSUE #2725: In motor config, report pod or booster motors with number of instances
When you have a pod or booster set with motors, it only shows it as a single motor in the name of the motor config (and in the table).  It should, just like with clusters, show _N_ x the motor name, where _N_ is the number of instances in the pod or booster set.

If there is an _M_-way cluster in each pod, then it should show the product of N*M.  

Here's an example of how it is now.  I slapped a couple of powered pods onto the simple rocket example:
Image

In the "Pod Engine" column, it only shows a single engine, which is really incorrect, should show 2x for each one.

When the center and pod engines are the same, you get weirdness like "2xA8-3" for the config name, even though it would more accurately be "3xA8-3".

This works correctly for clusters, it's only pods and boosters that show the misleading info.

SiboVG:
This PR fixes #2725.

Parallel booster staging example design:
image

Changed booster inner tube to 3-motor cluster:
image

No cluster and only one instance:
image

---

Note that there I discovered a bug while working on this (which was already present in 23.09). When opening the "Pods--powered with recovery deployment" example design, for the last flight config, the"A10-3" motor displayed as "217A10-3". The booster stage in that config has 2 A10-3 motors, which should display "2× A10-3". However, the '×' symbol is coded using `Chars.TIMES` (`u00D7`). In the motor config display code, there was this: `count + Chars.TIMES + " " + motorData`. But, since `count` is an integer, `Chars.TIMES` was also interpreted as an int (=215), instead of the '×' char. So you just did an integer sum... The solution was to force String summation: `"" + count + Chars.TIMES + " " + motorData`.

### DIFF
--- a/core/src/main/java/info/openrocket/core/formatting/MotorConfigurationSubstitutor.java
+++ b/core/src/main/java/info/openrocket/core/formatting/MotorConfigurationSubstitutor.java
@@ -174,40 +174,44 @@ public class MotorConfigurationSubstitutor implements RocketSubstitutor {
 		return combinations;
 	}
 
-	private static String getFinalSubstitute(List<String> substitutes, StringBuilder sbStageSub) {
-		if (substitutes.size() == 1 || !sbStageSub.isEmpty()) {
+	/**
+	 * Get the final substitute for a stage. If there is only one substitute, return that. If there are multiple
+	 * identical substitutes, return a string with the number of occurrences of each configuration (e.g. "3xA3-4").
+	 * If there are multiple different substitutes, return a string with all the configurations separated by commas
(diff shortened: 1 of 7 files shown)


neilweinstock:
Thank you!  You're on quite a roll lately.

SiboVG:
> Thank you! You're on quite a roll lately.

I have a paradoxical way of life; I'm very busy at work, so I need more distraction. That's how I ended up joining OpenRocket in the first place, when I was doing my uni exams. Also, it took a while to finish #2738, so now that I'm "freed" from that, I can tackle smaller issues.

Issue #2291, opened 2023-08-14
[Bug] Pods with nose and tail cone, without fins, drives CP ridiculously forward


JoePfeiffer:
### Describe the bug

A user on our discord server posted a design featuring side pods having a nose cone, a body tube, and a tail cone that results in CP ridiculously far forward.

The .ork as posted just has the CP a little ways forward of what you might expect, but as you increase the radius of the pods the CP keeps moving forward to well ahead of the rocket.

If I create a "rocket" with just the parameters of the pods, I get a N/A for the CP.  Maybe there's an "almost" divide-by-zero causing issues?
[POD_ROCKET.zip](https://github.com/openrocket/openrocket/files/12330980/POD_ROCKET.zip)

### To Reproduce

Open the .ork and look at the CP

#### Screenshot(s):
*(drag-and-drop the screenshot(s) here)*

#### .ork file:
*(drag-and-drop the file here as a .zip file)*

### OpenRocket version

22.02

### What platform are you running on?

Linux

JoePfeiffer:
Modifying the code to use the best available wind tunnel-derived boattail formulas from Moore, F. G. and L. Y. Moore, "Improved Aerodynamics for Configurations With Boattails", Journal of Spacecraft and Rockets 45(2), March-April 2008
doesn't make a significant difference in CP location.

JoePfeiffer:
A bit more comment here: while the CP of the pod was ridiculously far forward, its CNa was very small and so not a real issue for the rocket as a whole. After some other code improvements, the result of a pod like this is that it has no CP at all (it's NaN), and its CNa is 0.

PR #2753, opened 2025-03-22
Store moment calculated from CP and CNa instead of CP itself

### LINKED ISSUE #2751: Suspicious CP behavior
Tested in 24.12b1.

The following screenshots differ only in terms of the aft diameter of the tail transition.  Look at the positions of CP.  Something is clearly amiss.

Image

Image

[tailcone-cp.zip](https://github.com/user-attachments/files/19324312/tailcone-cp.zip)

JoePfeiffer:
The discontinuity observed in #2751 when a component has a tailcone is caused by a divide-by-zero error when merging CPs for the components in the rocket.

At present, AerodynamicForces contains a member variable cp containing the center of pressure of the component or assembly, with CNa as its weight. It also redundantly stores CNa as a separate variable.

The discontinuity when there is a tailcone comes from the fact that, according to the Barrowman equations, a nose cone has a CNa of 2 and a tail cone has a CNa of -2. When taking the weighted average of the nose and tail cone the weight becomes 0; when dividing by the weight the CP is NaN.

This PR solves this by storing the moment defined by CP * CNa, with a weight of CNa. Now instead of combining CPs by performing a weighted average, we can just sum the moments. When getCP() is called, the moment is divided by the weight to return the CP. If the weight happens to be 0 at this point, a CP of (0, 0, 0) is returned (it doesn't return (NaN, NaN, NaN) because this could cause issues with things like really tiny fins).

It also gets rid of the separate storage of CNa.

Fixes #2751

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/AerodynamicForces.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/AerodynamicForces.java
@@ -17,7 +17,7 @@ public class AerodynamicForces implements Cloneable, Monitorable {
 	private RocketComponent component = null;
 
 	/** CP and CNa. */
-	private Coordinate cp = null;
+	private Coordinate cpCNa = Coordinate.ZERO;
 
 	/**
 	 * Normal force coefficient derivative. At values close to zero angle of attack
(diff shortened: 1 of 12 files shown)


neilweinstock:
Outstanding.

SiboVG:
@JoePfeiffer in the following image, I compare the component analysis of a simple model rocket of your PR (left) with the unstable branch (right), and the CP for the body tube and launch lug is 0 cm. I suppose this is not desired?
image

SiboVG:
Also: this PR makes slight modifications to Sampo's original thesis calculations, right? If so, we should eventually update the technical docs. I won't ask you to do update the LaTeX doc in this PR, but can you please add a txt document in `/doc` that mentions changes that have been made to the aerodynamic calculations, and a reference to the PR that implemented those changes? Eventually, we should do a thorough update of the technical docs with all the changes that have happened (pods, boosters, tube fins...).

JoePfeiffer:
The change in CP for the body tube doesn't matter, since its weight (CNa) is 0. The change for the launch lug... I need to see if the CNa should really be 0. If so it also doesn't matter, if not there's an issue elsewhere. I'll make this a draft until I can check that.

This should be mathematically equivalent to the old code, except for finessing the divide-by-zero. Looking at equation 3.29 in the thesis, this is actually a more direct implementation of the equation. The numerator is the sum of the moments, and the denominator is the sum of the weights. The old code implemented this with a bunch of weighted averages; the new code accumulates the numerator in the moment and the denominator in the weight, and then does the division when getCP() is called.

JoePfeiffer:
I don't know why the launch lug was ever reporting a non-zero CP; the calculateNonAxialForces() method in LaunchLugCalc is empty. The launch lug's CP is also being reported (before this PR) several centimeters forward of the actual launch lug.

SiboVG:
Does this PR also fix #2291?

JoePfeiffer:
> Does this PR also fix #2291?

No, I don't think that one got "lucky" on the order of summing so it didn't trigger the bug.

The issue in #2291 is with Barrowman's treatment of tapering transitions, including boat tails and tail cones.It's just wrong.

PR #2771, opened 2025-03-29
Get rid of 10 m/s wind speed and 11km wind level limits

### LINKED ISSUE #2767: Remove 10 m/s wind speed limit
Title says it all.  The limit serves no useful purpose that I can think of, and severely limits the usefulness of the multi-level wind capability.

JoePfeiffer:
The 10 m/s max wind speed probably came from the NAR, TRA, and UKRA safety code limits of 20 mph max ground wind; it doesn't apply at higher altitudes where the winds are pretty much always higher than that, and we also shouldn't be preventing a user from setting whatever they want.

Note that the slider for the average wind speed now works analogously to the sliders in some of the component configuration dialogs: you can only use the slider to set the wind speed up to 10 m/s, but you can enter any speed you want by typing it in.

There should be a warning that the wind speed exceeds safety codes; I'll make an issue for that so we don't forget it, but I'd rather wait to add it until the upcoming Warning rewrite.

The Extended ISA atmosphere model prohibits setting a launch site above 11km (the limit of the troposphere); the reason for that isn't clear to me, but anything that high isn't realistic anyway. Though the launch site height is limited, the model allows flights above that. This PR leaves the limit on max launch altitude, but allows arbitrary altitudes for the wind levels.

Fixes #2767


neilweinstock:
That sounds perfect for now.

Issue #2777, opened 2025-04-08
[Bug] .ork saved with incorrect part UUID in warning


JoePfeiffer:
### Describe the bug

The attached .ork file fails to load due to an incorrect source UUID in a warning.

The saved warning text is "Body Tube, Nose Cone: Discontinuity in rocket body diameter" with UUIDs
&lt;source&gt;90919039-bcf7-4843-addb-88b9469ebe98&lt;/source&gt;
&lt;source&gt;378ad71b-1b5b-4701-889b-b920e901ad8c&lt;/source&gt;

However, there is no component in the rocket with a UUID of 378ad71b-1b5b-4701-889b-b920e901ad8c. The nose cone's UUID is 31c41721-d0fd-4d87-8d7a-8017455f7d03

The .ork has some other oddnesses, but it isn't clear whether those are failures in OR or if those are part of the design. In particular, the nose cone's length is 0.254 cm, and the inner radius of the forward body tube is 0.

I'm not calling this a release blocker because only one user seems to have encountered it, and I haven't been able to duplicate it. I will be submitting a PR with a workaround so a .ork file like this can be loaded.

### To Reproduce

1. No idea
...

### Screenshots / .ork file

[ballsrocket.zip](https://github.com/user-attachments/files/19652148/ballsrocket.zip)

### OpenRocket version

24.12.beta.01, 24.12.RC.02, unstable. Don't know about earlier

### What platform are you running on?

Linux

JoePfeiffer:
There has been another report of this issue. Interestingly, the new report also shows the problem with a discontinuity between nose cone and body.


[...]

PR #2778, opened 2025-04-08
Work around invalid IDs in .ork file

JoePfeiffer:
Under circumstances I haven't been able to duplicate, a .ork file can have a bad UUID for a component in a warning. This works around the error by checking that components are non-null before extracting their name (on file read) or UUID (on file write).

Works around (doesn't fix) #2777

### DIFF
--- a/core/src/main/java/info/openrocket/core/file/openrocket/OpenRocketSaver.java
+++ b/core/src/main/java/info/openrocket/core/file/openrocket/OpenRocketSaver.java
@@ -443,7 +443,9 @@ public class OpenRocketSaver extends RocketSaver {
 
 				if (null != w.getSources()) {
 					for (RocketComponent c : w.getSources()) {
-						writeElement("source", c.getID());
+						if (c != null) {
+							writeElement("source", c.getID());
+						}
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


JoePfeiffer:
Let's hope we hear from another user seeing the problem; if that doesn't happen before we're reading for final release we can merge this in at that point (I really don't want to, since an effective workaround will mean the actual bug is unlikely to ever get fixed). Still hoping maybe we can get more insights into how it happens.

SiboVG:
@JoePfeiffer I think we just received a new bug report on the mailing list from user Dan Barnes with the same issue as #2777.

JoePfeiffer:
Looks like it, thanks. It'll be tomorrow before I can look carefully.

senft-research:
What is the current state of this work around? We wanting a more definitive fix, or is this going to be a stop-gap measure until we address the root causes we discussed in #2777?

JoePfeiffer:
> What is the current state of this work around? We wanting a more definitive fix, or is this going to be a stop-gap measure until we address the root causes we discussed in #2777?

I think I've got enough to fix the underlying problems. I want to leave this one as a draft for now.

PR #2788, opened 2025-04-16
Improve simulation edit dialog UI for low-resolution screens

SiboVG:
This PR is a first attempt at improving OpenRocket's UI for users that have a low resolution screen setting. We recently received feedback from a user who simply couldn't access some items in the dialog because they got out of screen.

This PR makes the following changes:
- Overall reduces the default size of the sim edit dialog
- Add a scrollpane so that when you resize the sim edit dialog to make it smaller, you can still scroll through the rest of the contents. Note that the window size is saved in the preferences, so the user would only have to do this once

image

https://github.com/user-attachments/assets/b859de1f-ae62-4439-87c3-5afff54aa26e

---

The dialog scrollpane behavior should eventually be implemented in every dialog, especially larger sized ones. I'll make an issue for it so we don't forget.

In the future, I also envisage to replace the standard JDialog with a custom Dialog class that detects if the window-size is larger than the screen resolution. If it is, it resizes the window automatically. But that should not be a priority right now.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/plot/PlotPanel.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/plot/PlotPanel.java
@@ -62,7 +62,7 @@ public class PlotPanel<T extends DataType & Groupable<G>,
 
 	public PlotPanel(T[] typesX, T[] typesY, C customConfiguration, C[] presets,
 					 C defaultConfiguration, Component[] extraWidgetsX, Component[] extraWidgetsY) {
-		super(new MigLayout("fill"));
+		super(new MigLayout("fill, ins n n 0 n"));
 
 		this.customConfiguration = customConfiguration;
 		this.presetArray = presets;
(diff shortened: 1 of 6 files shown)


neilweinstock:
That looks very good.  Although the user didn't mention it, I'd think many of the other dialogs would be in the same boat (motor selection, rocket optimization, component analysis).  They seem to be similar size to the sim edit (before your changes).

SiboVG:
> That looks very good. Although the user didn't mention it, I'd think many of the other dialogs would be in the same boat (motor selection, rocket optimization, component analysis). They seem to be similar size to the sim edit (before your changes).

Yes, that's why I created #2789; there are indeed many other windows that can be improved. This PR is more of a first step and example to achieving this. I won't complete any other dialogs here though. I'll get on that once I have the time and energy to do so.

neilweinstock:
Sounds good.

PR #2796, opened 2025-05-07
 [#2795] Properly format and escape warnings in the CSV export 

### LINKED ISSUE #2795: Clean up Warnings column of sim table CSV export
The CSV export of the Warnings column is a mess.  See attached.

There are three problems:
1) Internal Java info is being inserted, rather than simply the warnings text
2) Commas are not being escaped.
3) It's not clear the Warnings column is even useful in exported CSV.

In the short term, it probably makes more sense to fix it rather than omit it.

[simtest.csv](https://github.com/user-attachments/files/20038301/simtest.csv)

SiboVG:
This PR fixes #2795. Example output:

```
Warnings,Name,Configuration,Velocity off rod (m/s),Apogee (m),Velocity at deployment (m/s),Optimum delay (s),Max. velocity (m/s),Max. acceleration (m/s²),Time to apogee (s),Flight time (s),Ground hit velocity (m/s)
"Recovery device deployment at high speed (27.9 m/s):  ""Parachute, plastic, preassembled, 10 in., PN 2262"" | Recovery device deployment at high speed (27.9 m/s):  ""Parachute kit, plastic, 8 in., PN 2260""",Simulation 3 - too short delay,[None; C6-3],15.442,230.485,27.934,4.87,77.228,151.709,5.377,63.695,4.587
```

Note that the warnings are correctly escaped with double quotes. Tested an import in Excel and it works fine.

### DIFF
--- a/core/src/main/java/info/openrocket/core/util/StringUtils.java
+++ b/core/src/main/java/info/openrocket/core/util/StringUtils.java
@@ -73,35 +73,34 @@ public class StringUtils {
 	/**
 	 * Returns an escaped version of the String so that it can be safely used as a
 	 * value in a CSV file.
-	 * The goal is to surround the input string in double quotes if it contains any
-	 * double quotes, commas,
-	 * newlines, or carriage returns, and to escape any double quotes within the
-	 * string by doubling them up.
-	 * 
+	 * The method follows standard CSV escaping rules:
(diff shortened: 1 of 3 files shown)


neilweinstock:
Excellent.

PR #2806, opened 2025-05-16
Motor Manufacturer determination via ThrustCurveAPI

### LINKED ISSUE #2721: [Feature Request] Use thrustcurve.org API to determine motor manufacturers
When new motor manufacturers are added to thrustcurve.org, we don't see them unless we manually add them to the manufacturers list in SerializeThrustCurveMotors.java

thrustcurve's API is able to provide a list of manufacturers (along with impulse classes, diameters, etc). We should use that to populate our options here.

It should be noted that we have a Manufacturer class which includes a bunch of motor manufacturers, some of which don't appear in thrustcurve's database, and many of which have various apparently random abbreviations of motor names. I don't see any real reason for this.

senft-research:
Retrieves the Manufacturers directly from an GET call to the ThrustCurveAPI. 

Removes need to hardcode a manufacturers String array within the Serialization Code. 

_Dev Note: Not sure why my other commits were not verified via my GPG key, issue seems to be resolved._

Closes #2721

### DIFF
--- a/core/src/main/java/info/openrocket/core/thrustcurve/ThrustCurveAPI.java
+++ b/core/src/main/java/info/openrocket/core/thrustcurve/ThrustCurveAPI.java
@@ -37,7 +37,6 @@ public abstract class ThrustCurveAPI {
         return SearchResponseParser.parse(is);
 	}
 
-	//TODO might be ideal to have a fallback, incase there are issues with the ThrustCurveAPI metadata?
 	/**
 	 * Utilises the ThrustCurveAPI to get the Manufacturer abbreviations, for the purpose of being used to obtain the
 	 * rest of the Motor Data per manufacturer.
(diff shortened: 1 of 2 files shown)


core/src/main/java/info/openrocket/core/thrustcurve/ThrustCurveAPI.java
```diff
@@ -1,11 +1,10 @@
 package info.openrocket.core.thrustcurve;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.*;
```
SiboVG, line 3: Please refrain from using wildcard imports. Note that some IDEs, like IntelliJ, do this automatically, but you can change the application preferences to disable this behavior.

senft-research: Noted and corrected, not sure how I missed that to be honest.

PR #2809, opened 2025-05-21
Removed code apparently copied and pasted from RocketComponentConfig.…

### LINKED ISSUE #2680: [Bug] TITLE CHANGE: Cancelling "Save design info" dialog during "Save as..." REVERTS MOST RECENT CHANGE to component/simulation
### Describe the bug

**Ignore just about everything below in this comment. Scroll down to 15 May, 2025**

<strike>
Adding (by accident) and deleting a simulation without first creating a new config makes bad things happen upon save.

#### Expected behavior

OR works reliably when deleting an accidentally-created simulation.

### To Reproduce

1. With a fresh instance of OpenRocket, File>Open example>TARC payload rocket.
2. Cancel "Rocket configuration" dialog.
3. Click to Flight simulations tab.
4. Click "New simulation", OK its dialog.
5. Click "Delete simulations" (the just-added sim is highlighted already)
6. OK the "are you sure" dialog
7. Click "Motors & Configuration" tab. Note the existing single config for [None;F50-9] - this is as it should be.
8. Click "New Configuration".
9. Select an Aerotech E30-4 from the dialog (probably not important which motor) and OK the motor selection dialog.
10. Note presence of configs for [None;F50-9] and [None;E30-4] - this is as it should be.
11. Click the "Flight simulations" tab. 
12. Note two simulations, one for each of the two configs - this is as it should be.
13. Save the file
14. Cancel the "Save design info" dialog, and click Yes for the "are you sure" warning.
15. _**BEFORE**_ selecting a file name, note that the Flight simulations tab has now lost the motor for Simulation 2
16. Save the file someplace convenient.
17. Choose a filename and save
18. Save "Only summary data"
19. Quit OpenRocket.

Note that items 20 and forward refer to a table resizing issue on macOS that's probably related to the problems with the OR file. But the file has already been saved without a motor, and if it's corrupt, it already is by this point.

20. Re-open OpenRocket _**by double-clicking on the saved-file.ork icon**_ (File>Open will not repro the macOS table re-sizing behavior from opening the (corrupt?) file).
22. Click on "Flight simulations" tab.
23. Click on "Simulation 2" (or [SHIFT]+click to add "Simulation 2")
24. Click "Run simulations".
25. Click on Motors & Configuration tab.
26. ~Note that there are no configurations.~ In actuality, on macOS, the Motors and Configuration window has (as a result of these steps) been sized down so only the column titles show. The configurations are there (one with no motor), visible after making more space in the Motors & Config window. Discovered this after more testing. Not resizing in this way on Win ARM64 and Linux ARM64.
</strike>
#### Screenshot: ~no configs~ configs area has been resized to column titles only:

Image

#### Screenshot: Config 2 motor gone before file named or written to disk:

Image

#### .ork file:

[junk2.ork.zip](https://github.com/user-attachments/files/18484108/junk2.ork.zip)

[Junk3.ork.zip](https://github.com/user-attachments/files/18484110/Junk3.ork.zip)

### OpenRocket version

24.12 b1

macOS

### Additional context

macOS 14.7.1 Apple silicon M3 MAX, 36GB RAM

jimmiedave:
Removed code from `SaveDesignInfoPanel.java` which was apparently pasted from `RocketComponentConfig.java`. The removed code was causing misbehavior from an unrelated preference setting (former lines 61-73 of `SaveDesignInfoPanel`) and causing spurious `document.undo()` actions upon dialog cancel (former line 81): there should be nothing to undo re: the `SaveDesignInfoPanel`. This should resolve #2680

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
@@ -58,19 +58,6 @@ public class SaveDesignInfoPanel extends RocketConfig {
         cancelButton.addActionListener(new ActionListener() {
             @Override
             public void actionPerformed(ActionEvent arg0) {
-                // Don't do anything on cancel if you are editing an existing component, and it is not modified
-                if (!isNewComponent && parent != null && (parent instanceof ComponentConfigDialog && !((ComponentConfigDialog) parent).isModified())) {
-                    disposeDialog();
-                    return;
-                }
-                // Apply the cancel operation if set to auto discard in preferences
-                if (!preferences.isShowDiscardConfirmation()) {
-                    ComponentConfigDialog.clearConfigListeners = false;		// Undo action => config listeners of new component will be cleared
-                    disposeDialog();
-                    document.undo();
-                    return;
-                }
-
                 // Yes/No dialog: Are you sure you want to discard your changes?
                 JPanel msg = createCancelOperationContent();
                 int resultYesNo = JOptionPane.showConfirmDialog(SaveDesignInfoPanel.this, msg,
@@ -78,7 +65,6 @@ public class SaveDesignInfoPanel extends RocketConfig {
                 if (resultYesNo == JOptionPane.YES_OPTION) {
                     ComponentConfigDialog.clearConfigListeners = false;		// Undo action => config listeners of new component will be cleared
                     disposeDialog();
-                    document.undo();
                 }
             }
         });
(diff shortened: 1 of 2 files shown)


senft-research:
When you say "should", was it checked if the issue in #2680 was resolved by the removal of this?

jimmiedave:
Heh. Good point: I hate seeing "should" in requirements docs and the like.

1. Yes, it does fix the steps-to-repro demonstration of the problem: cancelling the SaveDesignInfoPanel no longer reverts the last prior change to config or sim. Because of the way the bug operated, I'd initially suspected a rogue "undo", and yep, found one, removed it.

2. "Should" because upon removing code that refers to prefs for the RocketComponentConfig dialog (that is pretty clearly a copy-paste that didn't get tidied up: you're starting a new dialog, copy/paste from a working one...), I'm just about certain that code has no place in SaveDesignInfoPanel, but am less certain that it can't ever be called from elsewhere (some languages have GOTO, but what language has COMEFROM?). I did use the IDE to look for any usages and didn't find anything problematic. On this piece, I'd place my certainty higher than 99%.

senft-research:
Aye sound, and cheers for going into detail and humouring my pedantic nature 😅

jimmiedave:
Gotta pull this back to fix.

PR #2821, opened 2025-06-22
Improve core startup for using OpenRocket in external Java application

SiboVG:
I was recently trying to use the OpenRocket JAR for another project, but it was difficult setting up a simple OpenRocket instance. This PR adds helper methods to easily start new core instance of OpenRocket, using `OpenRocketCore.initialize();`. ReadTheDocs documentation is also provided.


JoePfeiffer:
Will it be possible to use this with orhelper?

SiboVG:
> Will it be possible to use this with orhelper?

Maybe, I don't really know how orhelper currently initializes OR.

PR #2827, opened 2025-06-30
Make lateral angle and lateral orientation angle use North 0

### LINKED ISSUE #2811: [Bug] OR > [Plot / Export] > [Export Data] > Lateral Direction is Incorrect
### Describe the bug

TL;DR - the arguments to atan2( ) are inverted for Lateral Direction in the exported Sim Data.

I am working on a bash script to parse Exported Sim.csv data for GPS Driftcast.

I want several columns from the .csv file as events occur.

The Events are **Apogee**, **Main Deployment**, **Ground Hit**

The data I want is for GPS DC is:

**Event**, **Altitude**, **Lateral Distance**, **Lateral Direction**, **Vertical Velocity**, **Position East Of Launch,** **Position North of Launch**, **Latitude** and **Longitude**.

It appears that when **Lateral Direction** is calculated the ( x,y ) arguments for atan2( ) are swapped.

Easy to do ... been there, done that :)

I have not checked for other direction fields ...

Thanks !

-- kjh

### To Reproduce

1. Run Sim
2. Export Data
3. Check Lateral Direction using Position East of Launch and Position North of Launch.
4. for Apogee at about 13 sec, if x = East = -201.00 and y = North = -449.00 then atan2( y, x ) = -2.72069 radians = -155.88 deg = 204.12 normalized deg
5.  OR reports -114.12 deg = 245.88 normalized deg
...

#### Screenshot(s):

[tp-h128w-c50524-0400-or.csv](https://github.com/user-attachments/files/20391222/tp-h128w-c50524-0400-or.csv)

#### .ork file:
*(drag-and-

[vul-48-2412-RC01.zip](https://github.com/user-attachments/files/20391167/vul-48-2412-RC01.zip)

drop the file here as a .zip file)*

#### screenshot of my q&d bash script results:

Image

### OpenRocket version

24.12 RC-01

Linux

### Additional context

_No response_

JoePfeiffer:
In a couple of places, azimuth angles were being computed using atan2(y, x) instead of atan2(x, y), which sets angle 0 facing right instead of north. This PR corrects that.

Also, replaces instances of atan(a/b) with atan2(a, b) to avoid potentiona divide by 0 errors.

Fixes #2811

### DIFF
--- a/core/src/main/java/info/openrocket/core/file/wavefrontobj/export/components/MotorExporter.java
+++ b/core/src/main/java/info/openrocket/core/file/wavefrontobj/export/components/MotorExporter.java
@@ -98,7 +98,7 @@ public class MotorExporter {
             final float z = (float) (innerRadius * Math.sin(angle));
             obj.addVertex(transformer.convertLoc(x, y, z));
 
-            final double slopeAngle = Math.atan(coneLength / innerRadius);
+            final double slopeAngle = Math.atan2(coneLength, innerRadius);
             final float nx = (float) Math.cos(slopeAngle);
             final float ny = (float) -Math.cos(angle);
             final float nz = (float) -Math.sin(angle);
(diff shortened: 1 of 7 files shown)


kjhambrick:
Thank You Joe !

Your changes work !!

-- kjh

Issue #2713, opened 2025-02-08
[Feature Request] Incorporate relative humdity in calculations


SiboVG:
OR's atmospheric models (e.g. `AtmosphericConditions`) use the specific gas constant of dry air. It would be good to add an option to specify the relative humidity at the launch site, which would then be taken into account in e.g. the air density and pressure calculations. The difference may not be large, but any improvement to our accuracy is a bonus.

SiboVG:
Fixed by #2834.

PR #2834, opened 2025-07-06
Incorporates relative humidity to atmospheric conditions

gdilbaz:
This PR adresses [#2713](https://github.com/openrocket/openrocket/issues/2713) incorporating the relative humudity to gas constant calculations. Also adding option to specify it simulation options.

image

When modeling the behaviour of RH as the altitude increased i opted to model it linearly decreasing by 6% per kilometer. I did this due to the lack of literature tackling the problem. The literature I've found was inconsistent but mentioned rates between 4-8%. Because the program doesn't currently take humidity to consideration and the relatively small effect it'd have I think this model is fine four our purposes. Any help is really appreciated in this part.

Method:
 Clausius-Clapeyron equation was used to determine the saturation vapor pressure 
image

Then due definition :
Relative Humidity
image
Mixing ratio (Epsilon being the ratio of molar masses)
image
Specific Humidity
image

Subtituting e into the mixing ratio we get:
image

We also know 
image

Finally after some algebra:
image


core/src/main/java/info/openrocket/core/models/atmosphere/AtmosphericConditions.java
```diff
@@ -5,6 +5,8 @@
 import info.openrocket.core.util.Monitorable;
 import info.openrocket.core.util.ModID;
 
+import java.security.PublicKey;
```
SiboVG, line 8: Please remove this.

core/src/main/java/info/openrocket/core/models/atmosphere/AtmosphericConditions.java
```diff
@@ -75,18 +94,54 @@ public void setTemperature(double temperature) {
...
+		this.humidity = humidity;
+		this.modID = new ModID();
+	}
+
+	/**
+	 * Calculate the saturation water pressure using the Clausius-Clapeyron equation.
+	 * @return The saturation vapor pressure in Pa
+	 */
+	public double vaporPressureSaturation() {
+		return 611.3 * Math.exp(5423 * (1/273.15 - 1/getTemperature()));
```
SiboVG, line 114: You can precalculate the 1/273.15 division to improve efficiency. I would put the full equation (like you have now) as a comment before the return function and then use this equation:
```java
// 611.3 * Math.exp(5423 * (1/273.15 - 1/getTemperature()));
return 611.3 * Math.exp(19.854 - 5423/getTemperature());
```

core/src/main/java/info/openrocket/core/models/atmosphere/ExtendedISAModel.java
```diff
@@ -199,6 +228,19 @@ private double calculatePressure(double alt1, double temp1, double alt2, double
 		}
 	}
 
+	/**
+	 * Calculate the relative humidity at any altitude.
+	 * Assumes relative humidity drops at a rate of 0,06/km
+	 * Assumes there is no more humidity after 16,5 km
```
SiboVG, line 234: Where does this come from, can you please add a source in the docstring?

core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
```diff
@@ -21,10 +21,7 @@
 import info.openrocket.core.models.atmosphere.ExtendedISAModel;
 import info.openrocket.core.models.wind.PinkNoiseWindModel;
 import info.openrocket.core.preset.ComponentPreset;
-import info.openrocket.core.rocketcomponent.FlightConfiguration;
-import info.openrocket.core.rocketcomponent.MassObject;
-import info.openrocket.core.rocketcomponent.Rocket;
-import info.openrocket.core.rocketcomponent.RocketComponent;
+import info.openrocket.core.rocketcomponent.*;
```
SiboVG, line 24: Please don't use wildcard imports. IntelliJ does this automatically. You can disable this: https://www.baeldung.com/intellij-disable-wildcard-import.

core/src/main/resources/l10n/messages.properties
```diff
@@ -529,6 +529,8 @@ simedtdlg.lbl.Temperature = Temperature:
 simedtdlg.lbl.ttip.Temperature = The temperature at the launch site.
 simedtdlg.lbl.Pressure = Pressure:
 simedtdlg.lbl.ttip.Pressure = The atmospheric pressure at the launch site.
+simedtdlg.lbl.Humidity =Relative Humidity:
```
SiboVG, line 532: please add a space after the equals sign :)

core/src/test/java/info/openrocket/core/simulation/SimulationConditionsTest.java
```diff
@@ -61,6 +61,7 @@ public void testDefaultSimulationOptionFactory() {
         assertTrue(options.isISAAtmosphere());
         assertEquals(288.15, options.getLaunchTemperature(), EPSILON);
         assertEquals(101325, options.getLaunchPressure(), EPSILON);
+        //assertEquals(0, options.getLaunchHumidity(), EPSILON);
```
SiboVG, line 64: Does this not work? Why is it commented out?

swing/src/main/java/info/openrocket/swing/gui/simulation/SimulationConditionsPanel.java
```diff
@@ -166,11 +167,38 @@ public static void addSimulationConditionsPanel(JPanel parent, SimulationOptions
...
 
+		// Humidity:
+		label = new JLabel(trans.get("simedtdlg.lbl.Humidity"));
+		////The humidity at the launch site.
+		tip = trans.get("simedtdlg.lbl.ttip.Humidity");
+		label.setToolTipText(tip);
+		isa.addEnableComponent(label, false);
+		sub.add(label);
+
+		humidityModel = new DoubleModel(target, "LaunchHumidity", UnitGroup.UNITS_NONE, 0);
```
SiboVG, line 178: Please us `UnitGroup.UNITS_RELATIVE`, since relative humidity is expressed as a percentage.

core/src/test/java/info/openrocket/core/models/atmosphere/ExtendedISAModelTest.java
SiboVG, line 1: You only test cases with relative humidity equal to 0. Can you please also add test cases that test different humidity levels, and that test the different calculations, e.g. vapor pressure....

SiboVG (changes requested):

SiboVG:
Great work @gdilbaz, thanks! I've added some comments. I still need to test it thoroughly though. We are currently in the process of releasing the 24.12 release. Once that's done, we can look into merging your PR.

SiboVG:
@gdilbaz did you find time to take a look at my remarks?

gdilbaz:
> @gdilbaz did you find time to take a look at my remarks?
Unfortunately I couldn’t. I’ve been working in a different city for a while now. I am returning home later today will look into this ASAP.

core/src/main/java/info/openrocket/core/models/atmosphere/ExtendedISAModel.java
gdilbaz: I've added the papers I used but when i was searching for the modeling of RH I couldn't find it's modelling wrt altitude. I am hoping someone more knowledgeable about the subject can give a more concrete and correct model.

gdilbaz:
Hello,
Sorry for the, very, late reply. I've been quite busy and couldn't really managed to come back to this issue until recently. After some literature review, I've found a paper by David Romps. In this paper, he tackles on the RH distribution wrt altitude. But his work only cover's the tropics(20 degree S - 20 degree N). Outside the regions some of the assumptions starts breaking down. 

I'm willing to implement his paper since I couldn't find any other papers that are this useful. But it'll be locked into the region I've mentioned which may lead us to exclude simulating it outside the region concerning me with the usefulnes of the implementation. I'd appreciate any guidance and further literature to review. Thanks!

Here is the link to the paper: https://journals.ametsoc.org/view/journals/clim/27/19/jcli-d-14-00255.1.pdf

SiboVG:
> Hello, Sorry for the, very, late reply. I've been quite busy and couldn't really managed to come back to this issue until recently. After some literature review, I've found a paper by David Romps. In this paper, he tackles on the RH distribution wrt altitude. But his work only cover's the tropics(20 degree S - 20 degree N). Outside the regions some of the assumptions starts breaking down.
> 
> I'm willing to implement his paper since I couldn't find any other papers that are this useful. But it'll be locked into the region I've mentioned which may lead us to exclude simulating it outside the region concerning me with the usefulnes of the implementation. I'd appreciate any guidance and further literature to review. Thanks!
> 
> Here is the link to the paper: https://journals.ametsoc.org/view/journals/clim/27/19/jcli-d-14-00255.1.pdf

Thanks for the reply! The paper you linked could be a start, but it's not very generic. I would be inclined to just use a constant RH throughout the flight. So, you can replace the content of `calculateHumidity` with `return humidity1;` and a comment `TODO: implement changing humidity based on altitude`. We can fill this in later, or leave it as is. From the paper, in the first 10-15 km, the RH doesn't change much, so it also wouldn't change much during flight.

SiboVG:
Thanks a lot for your contribution @gdilbaz, we appreciate it!

PR #2838, opened 2025-07-12
Adding 'clone' helper methods to WorldCoordinate, Quaternion, SimulationStatus

MDNich:
Here I add `clone` helper methods to the `WorldCoordinate`, `Quaternion`, and `SimulationStatus` classes. These are quite useful for a Python binding I am using locally (and hope to prepare as an addition), and I believe they would be quite useful for others developing `SimulationExtensions` or `SimulationListeners` that seek to tweak the simulation algorithm.

Sincerely,
Marc D NICHITIU

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/SimulationStatus.java
+++ b/core/src/main/java/info/openrocket/core/simulation/SimulationStatus.java
@@ -555,6 +555,30 @@ public class SimulationStatus implements Cloneable, Monitorable {
 		}
 	}
 
+
+	/**
+	 * Copies simulation status data from the specified {@code SimulationStatus} object <code>orig</code>
+	 * to this instance, in particular the physical kinematics as well as boolean flags of <code>motorIgnited</code>, etc.
+	 *
+	 * @param orig the {@code SimulationStatus} object from which to copy data
+	 */
+	public void copyProperties(SimulationStatus orig) {
+		this.position = orig.position.clone();
+		this.worldPosition = orig.worldPosition;
+		this.velocity = orig.velocity.clone();
+		this.orientation = orig.orientation.clone();
+		this.rotationVelocity = orig.rotationVelocity.clone();
+		// these are booleans, so no cloning (primitives).
+		this.motorIgnited = orig.motorIgnited;
+		this.liftoff = orig.liftoff;
+		this.launchRodCleared = orig.launchRodCleared;
+		this.apogeeReached = orig.apogeeReached;
+	}
+
+
+
+
+
 	@Override
 	public ModID getModID() {
 		return modID;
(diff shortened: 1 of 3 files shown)


MDNich:
Sorry about the Revert part there - I didn't realize it would send all commits in. The only three changes are the clone in WorldCoordinate, clone in Quaternion, and copyProperties in SimulationStatus.

JoePfeiffer:
Should WorldCoordinates and Quarternion now say "implements Cloneable"?
I'm not clear on the purpose of SimulationStatus.copyProperties(), given that a clone() method exists. Would it be possible (and better) to modify that method?

MDNich:
Hi, 
Sure ! I had not thought about that implementation, we can definitely add that.

As for the clone method, this was useful if one would want to preserve a reference but copy all properties. For example, suppose I would write a listener that receives as an argument a `SimulationStatus`, but would need to have all the attributes of some (let's say `static`) other `SimulationStatus` in the class of the listener: instead of copying one by one the parameters, I could invoke this method, and effectively clone it, but keep the object reference, so that the simulation can move on with this injected state.

Hope that helps !

JoePfeiffer:
On another note, I'm really skeptical that going to an RK6 simulation will add enough more accuracy to warrant the longer run time. If you've tested the code (I note you reverted the commit, so I haven't looked at it so I don't know if it's done), did it make enough of a difference to make any unit tests fail? Our thrustcurves are piecewise linear (and the code breaks timesteps at the thrustcurve datapoints), and drag is only a function of V^2, I'd be a little surprised to see something higher order than cubic make a real difference.

But it will run slower. Not enough to matter to the vast majority of our users, but we do have some edge cases out there -- last month I used orhelper to drive OR to do some monte carlo simulations, and had to limit the number of runs for a given scenario to 25 because of how long it took on my workstation.

MDNich:
This is all true - for small time steps its not going to be much better, especially for simple cases; however, when using the PI controller extension (already exists), it would improve the accuracy by integrating more accurately - I upload here a couple comparisons for a custom PID controller of mine; for the same control parameters, aggressive as they may be, the noise in the simulation is greatly reduced by the RK4 -> RK6 switch.

MDNich:
[RK4_dt_0p001.pdf](https://github.com/user-attachments/files/21201653/RK4_dt_0p001.pdf)
[RK4_dt_0p0001.pdf](https://github.com/user-attachments/files/21201654/RK4_dt_0p0001.pdf)
[RK4_dt_0p00001.pdf](https://github.com/user-attachments/files/21201655/RK4_dt_0p00001.pdf)
[RK6_dt_0p001.pdf](https://github.com/user-attachments/files/21201656/RK6_dt_0p001.pdf)
[RK6_dt_0p0001.pdf](https://github.com/user-attachments/files/21201657/RK6_dt_0p0001.pdf)
[RK6_dt_0p00001.pdf](https://github.com/user-attachments/files/21201658/RK6_dt_0p00001.pdf)

MDNich:
Side by side comparisons:
1e-5
1e-4
1e-3

MDNich:
In any case I would make a separate PR for RK6 - this PR doesn't have anything to do with it : )

Issue #2855, opened 2025-07-28
TEKNOFEST Rocket Competition and AeroTech engine incompatibility


mehmeterdengecti:
Teknofest gives its own dimensions for the AeroTech rocket engines used in the TEKNOFEST rocket competition( the rocket engine dimensions given by Teknofest itself.)fest gives its own dimensions for the AeroTech rocket engines used in the TEKNOFEST rocket competition( the rocket engine dimensions given by Teknofest itself.) The dTeknofest gives its own dimensions for the AeroTech rocket engines used in the TEKNOFEST rocket competition( the rocket engine dimensions given by Teknofest itself.) The dimensions provided by Teknofest do not match the worry.
@tibbe @neilweinstock @soupwizard @itshanney @UncleRus  

<img width="822" height="333" alt="Image" src="https://github.com/user-attachments/assets/37eea4d3-63ba-4c5b-8c37-e141d906ab00" />

<img width="1106" height="866" alt="Image" src="https://github.com/user-attachments/assets/8ef5c4f5-5c7d-419a-9638-3f6cbfd374d4" />

<img width="896" height="937" alt="Image" src="https://github.com/user-attachments/assets/cd6f92cd-67e5-4d40-9067-614ce242611f" />

 There is a video of the error on the drive link : [https://drive.google.com/file/d/13mhumyKaDXdOJVwadV0nEN6qdDHpzUjU/view?usp=sharing](url)

ProtonNumber:
You shouldn't be putting the .eng file in "user defined component presets".

Click reset on that and try it again?

mehmeterdengecti:
> Raporunuz için teşekkürler. Lütfen teknofest_l1256.eng dosyasını buraya yükleyebilir misiniz?
Hey friend, Github does not allow me to upload files with .eng extension. I'm leaving the Drive link:  
https://drive.google.com/drive/folders/1YYre34S3No6xlKA23t_vPwQj4m85WziE?usp=sharing

mehmeterdengecti:
I accidentally closed the discussion, how do I reopen it?

mehmeterdengecti:
> You shouldn't be putting the .eng file in "user defined component presets".
> 
> Click reset on that and try it again?

Can you explain a little bit how I can do what you call a friend?

SiboVG:
@ProtonNumber is right. You have added your thrust curve (.eng file) in the component presets instead of in the thrust curve section:

[...]

PR #2856, opened 2025-07-28
[#2855] Make user-defined thrust curve and component preset loading more foolproof

SiboVG:
In light of #2855, this PR makes user-defined thrust curve and component preset loading more foolproof by:
- Disabling the "All files" option in the file chooser
- When loading thrust curves or component preset files, that the file extension is correct


JoePfeiffer:
It would also be a good idea to put up a dialog when a file can't be parsed, rather than just having an uncaught exception.

PR #2889, opened 2025-09-17
[#2524] Separate BarrowmanCalculator into DragCalculator and StabilityCalculator

### LINKED ISSUE #2524: [Feature Request] Separate drag and stability aerodynamic calculators
At present, OR uses extended Barrowman equations for both drag and stability calculations. The code is written so it would be possible to substitute another calculator (though to the best of my knowledge this has never been tested). The Barrowman equations for non-axial (ie stability) and axial (ie drag) force have very little to do with each other, and indeed the code for them is also pretty much separate. However, in terms of the program structure, there is a single aerodynamic calculator containing the code for both.

An idea that comes up periodically is to have a lookup table for drag coefficient vs. speed, so flight data can be used to give more accurate simulations. This would (probably) involve having the lookup table based drag replace our current drag calculations, while leaving the stability calculations alone. Likewise, the Air Force's DATCOM code could be used as an alternate drag calculator (this might run into ITAR restrictions, but a version of the code has been made available in the companion CD for Hammond's _Design methodologies for space transportation systems_, so one would expect that version to have been approved for distribution).

Doing this would be much more straightforward if the stability and drag calculations were split into separate calculators, so either could be replaced at will.

Note that it is possible to write a simulation extension to override the existing calculator; this is a pretty inelegant solution however.

SiboVG:
This PR fixes #2524. `BarrowmanCalculator` is now separated into separate stability and drag calculators. Just for fun, and for further future implementation, I added a `LookupTableDragCalculator` and `LookupTableStabilityCalculator` that lets you map drag and stability data to Mach and AoA values.

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanCalculator.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanCalculator.java
@@ -1,1087 +1,147 @@
 package info.openrocket.core.aerodynamics;
 
-import static info.openrocket.core.util.MathUtil.pow2;
-
-import info.openrocket.core.logging.Warning;
-import info.openrocket.core.logging.WarningSet;
-import info.openrocket.core.rocketcomponent.AxialStage;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import info.openrocket.core.aerodynamics.barrowman.FinSetCalc;
-import info.openrocket.core.aerodynamics.barrowman.RocketComponentCalc;
-import info.openrocket.core.rocketcomponent.position.AxialMethod;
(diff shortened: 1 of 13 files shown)


JoePfeiffer:
Wow, this looks great! One quick question -- how does a user enable the lookup? If I go to edit the simulation options I still only see the "Extended Barrowman" calculation method.

SiboVG:
> Wow, this looks great! One quick question -- how does a user enable the lookup? If I go to edit the simulation options I still only see the "Extended Barrowman" calculation method.

Yes, I did not implement a UI for it yet; this PR was meant more as a starting block for such a feature. But I can convert this to a draft and implement it later.

SiboVG:
You can now load data from a file in the UI:
image

PR #2891, opened 2025-09-18
[#2877] Open new design window in a different position from existing windows

### LINKED ISSUE #2877: Open new design windows in a different position from existing windows
Right now each new OR window opens up in exactly the same position, so multiple open windows will stack perfectly on top of each other until you move them.  

This is generally not what a user expects.  Multiple open windows would usually be cascaded slightly (new ones below and to the right of existing) so you can see them stacking up.

SiboVG:
This PR fixes #2877:

https://github.com/user-attachments/assets/94b4548c-17ba-4660-ab0c-83ff17abf855

New windows are opened a certain x and y offset from the previous window. Note that in the video, sometimes the window goes to a position that is not to the bottom-right of the previous one. That is because there is built-in logic to not go outside the window bounds.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/main/BasicFrame.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/main/BasicFrame.java
@@ -1,6 +1,11 @@
 package info.openrocket.swing.gui.main;
 
 import java.awt.Dimension;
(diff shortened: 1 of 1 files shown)


neilweinstock:
Looks perfect, thank you!

Issue #2765, opened 2025-03-25
Shock Chord Material Dropdown (24.12.RC.01)


RocketKidAnthony:
This is just an idea i think will streamline material dropdowns a little bit.
Under the shock chord menu, where you select material, it is a bit cluttered. I propose that the dropdown is differentiated by material instead of threads and lines.
     ex:
Kevlar thread 800 could be found under a Kevlar thread category.
Flat elastic chord could be found under an elastic chord category.
     I believe this would help the ui / ux look a little cleaner as the menu is a bit cluttered. Overall kind of like how it is in the nose cone or body tube material dropdown.

PR #2899, opened 2025-09-30
Add category feature to organize item

Kruzk02:
This PR assigns a item to a category based on first word of its name. This make grouping and filtering easier.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/widgets/GroupableAndSearchableComboBox.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/widgets/GroupableAndSearchableComboBox.java
@@ -5,10 +5,13 @@ import info.openrocket.core.util.Groupable;
 import info.openrocket.swing.gui.util.GUIUtil;
 import info.openrocket.swing.gui.theme.UITheme;
 
+import java.util.LinkedHashSet;
+import java.util.stream.Collectors;
 import javax.swing.AbstractListModel;
 import javax.swing.ComboBoxModel;
 import javax.swing.DefaultComboBoxModel;
 import javax.swing.DefaultListCellRenderer;
+import javax.swing.JCheckBox;
 import javax.swing.JCheckBoxMenuItem;
 import javax.swing.JComboBox;
 import javax.swing.JLabel;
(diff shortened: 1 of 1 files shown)


JoePfeiffer:
If I'm understanding this correctly, you're adding another level to the material selection menus. So, instead of selecting Plastics->Polystyrene (1.05 g/cm^3) a user would now select Plastics->Polystyrene->Polystyrene (1.05 g/cm^3). Is this correct?

Kruzk02:
Yes, instead of showing all the material in one place, I want to make it smaller to make it easier to navigate.

JoePfeiffer:
> Yes, instead of showing all the material in one place, I want to make it smaller to make it easier to navigate.

Where I'm confused is we already have one level of categories -- Metals, Woods, Plastics, etc. Looking through the materials available, there are only two existing materials that would be combined in to a single category  (two kinds of styrofoam). I'm having a hard time seeing a benefit to another level of menu, when there's only one item in the new menu in nearly all cases.

JoePfeiffer:
> Sorry, I got tunnel vision on just the parachute and shock cord because i saw a lot of same type.

Looking at the shock cord configuration editor, I finally understand the problem! But I have to think a better approach would be to replace the "Threads and Lines" category with categories for "Braided nylon", "Elastic braided cord", "Elastic cord", etc (though this is just off the top of my head -- should elastic braided cord and elastic cord be in the same category, for instance?).

Thoughts?

Kruzk02:
Best to keep related thing together, and i wouldn't change "Threads and Lines" since that would mean updating the itemGroupMap field from different class

Kruzk02:
> Thanks for attacking this one... the material selection for shock cords is definitely something that needed improvement.
> 
> When proposing a UI change like this, it would be super helpful to include screenshots and/or video screen captures to show what you've done.

Here's the screenshot of UI after my change.

Screenshot 2025-10-02 082510

SiboVG:
Thank you for the contribution @Kruzk02! However, I'm believe that the implementation needs some refinements. Would this method work for users running a translated version of OpenRocket, due to the hard-coded (English) search string? Also for languages that are written right to left (e.g. Arabic)? I can't look at the code for now, so I can't verify.

Regardless, I propose a more long-term and flexible solution: modify the itemGroupMap to support subgroups (where group could be a sorted list of groups; idx 0 being the top parent ("Threads and Lines" in this case), idx 1 could be a new group "Braided" etc.

JoePfeiffer:
> Regardless, I propose a more long-term and flexible solution: modify the itemGroupMap to support subgroups (where group could be a sorted list of groups; idx 0 being the top parent ("Threads and Lines" in this case), idx 1 could be a new group "Braided" etc.

Since there is only one entry in the top level menu, I don't see a need for it. It makes a lot more sense to me to get rid of the THREADS_LINES MaterialGroup, and replace it with groups like KEVLAR, NYLON, etc.

SiboVG:
Linking #2765.

Kruzk02:
Are this good now?
Screenshot 2025-10-03 094156

JoePfeiffer:
> I could live with this as shown above, big improvement to what's there now.
> 
> However, I would personally change the categories slightly. I don't think the first word is a good sort. Rather I'd sort by material like this:
> 
>     * Elastic
> 
>     * Kevlar
> 
>     * Nylon
> 
>     * Misc
> 
> 
> What do others think?

It isn't being sorted on the first word. The MaterialGroups' names are just about the same as the first word (because of the history of the material names), but is actually a little different. So all the materials whose names start with Kevlar are in "Kevlars".

JoePfeiffer:
This is looking really nice -- but one further request, please match the 4-character indentation used elsewhere.

neilweinstock:
> So, apart from those 3 material, put everything else in the misc group?

There are only a handful left over, right?  If so then yeah, that's what I would do.

Kruzk02:
Screenshot 2025-10-06 081827

neilweinstock:
I apologize for the apparent confusion here, but… braided nylon and tubular nylon should be under “Nylons”

Kruzk02:
I just didn't know where to put those material, so i place them in Other group for placeholder .

Screenshot 2025-10-06 092332

neilweinstock:
That looks good, thanks.

PR #2900, opened 2025-09-30
Make thrustcurve selection clearer for users

JoePfeiffer:
One of our not infrequent complaints from users is that a motor in the database has an inaccurate thrustcurve, or that its thrustcurve has changed for the worse. This is typically because our motor selection dialog fails to make clear to the user that a motor may have multiple thrustcurves to choose from. This PR attempts to address this.

At present, when a motor is selected in the dialog, the only indication that the user has a choice of thrustcurves is the presence of the "Select thrust curve" combobox in the upper left corner of the dialog.  Notably the "Filter Motors" pane remains selected, as shown in the following screenshot:

image

In order to see the motor details, the user must select the "Show Details" pane:

image

Even now, the thrustcurve selection combobox is literally as far away from the thrustcurve plot as it can possibly get and be in the same dialog. Also, the label in the combobox doesn't reflect the selected thrustcurve, as it's black (through version 23.09, the color matched the color of the curve; in 24.12 it does not. I'm actually not sure how that changed; I haven't found where in the code it was set to match in the old versions).

This PR attempts to improve this situation. First, when a motor is selected, the motor detail pane is selected automatically (the user can switch back to the filter pane either by selecting the pane, or by entering text in the search field). The thrust curve selection combobox, "hide similar motors" checkbox, and ejection charge delay are all moved to the detail pane. The thrust curve plot is moved to just under the comboboxes, and the color of the thrust curve selection combobox matches the color of the currently selected thrustcurve.

image


neilweinstock:
I had to stare at it a bit but I think I like it.

I'm wondering if there's a way to improve the delay selection, while we're at it.  I'm thinking a visible list of available delays with radio buttons, plus a blank custom delay field that is also selectable.  The current combobox fools a lot of people.

I'll try to throw together a mockup, or maybe there's another better way to do it.

jimmiedave:
I generally like it. I'd like to see a color chip, vs. color text in the curve selection droplist. More color onscreen to assist in discernment. The color text is subtle for me.

Screenshot 2025-09-30 at 8 39 09 PM

thzero:
I'm not a fan of moving the ejection charge to under the Show Details as it is now.

If however when I highlight a motor the Show Details is automatically switched, then cool.... that makes sense.  The ejection charge, the different thrust curves, etc. is all right there.

I like the chip color, but for people who are color blind, that makes it hard.  

Still prefer dropdown with custom text box as opposed to radio buttons.

JoePfeiffer:
This is all why I requested comments!

Yes, the color chip makes the difference clearer -- I want to do that. Using color at all causes issues for color blind people (my father was red-green color blind. I've got stories....), but I don't have a better idea here.

I really like the radio buttons for delay (sorry @thzero). It seems like a clearer way of letting the user know their options.

I put the "hide similar" at the top of that section for consistency with "hide OOP", Not important?

Yes, when you select a motor it automatically switches to the motor detail pane. While I like all the other changes (I'd better, since I did most of them), that automatic switch is the single most important one.

neilweinstock:
> I put the "hide similar" at the top of that section for consistency with "hide OOP", Not important?

No, not in my opinion.  Once consistency starts making things worse then it gets tossed out the window.  I would venture to guess that use of the "hide similar" checkbox is extraordinarily rare, it should be moved out of the way of what they're really interested in.  I would probably also prefer that the "hide out-of-production" box also go below the table, but I don't think it's causing much harm being up top.

> Yes, when you select a motor it automatically switches to the motor detail pane. While I like all the other changes (I'd better, since I did most of them), that automatic switch is the single most important one.

It is fundamental to this whole PR; without it you could not possibly hide the controls in question in the details tab.

PR #2907, opened 2025-10-15
Avoid setting null source in Messages

### LINKED ISSUE #2873: Error during save of project
I was going to save a project, when an error message appeared, and i don't know why this happens, This doesn't happens with a new project, it does if i'm going to save antoher project.

[

Image

](url)

JoePfeiffer:
Stop passing a null in as a Message source in the one place it was being done explicitly

Throw a bug exception if code ends up doing it elsewhere, so the crash moves to the actual bug location instead of having it happen when writing .ork file. Also throw a bug exception identifying the Warning type if a null source ends up being saved anyway, instead of just having a null pointer exception.

I suspect there is still a bug, but hopefully the new tests in the code will help track it down. I'll close #2873, since the symptom will be different.

Fixes #2873


SiboVG:
@JoePfeiffer could you please undo the update of the openrocket-database? Your PR would downgrade the latest version.

JoePfeiffer:
Ah, I've had a user give working instructions for triggering the bug. New PR will be incoming with the real fix.

PR #2912, opened 2025-10-21
Use mutable coordinates to improve memory usage

SiboVG:
This is sort of a continuation of #2909. OpenRocket currently uses an immutable coordinate class. This is safe, but it does means that if you needs to do transformations with it (`add`, `multiply`...) you create a new Coordinate object for each operation. You can see this clearly in the `step` method of the `RK4/RK6SimulationStepper`.

This PR introduces a new class, `MutableCoordinate`, and applies it to some hot paths in the simulation calculations to reduce the creation of new coordinate objects. I also added a `CoordinateIF` interface to make it easier to switch between a mutable (`MutableCoordinate`) and immutable (`Coordinate`) coordinate.

Finally, I added unit testing on the memory usage of running simulation so we can keep track when a new change causes a spike in memory allocation.

---

Here are some results of the memory improvements in JProfiler. I do notice that the memory results can vary quite a lot between runs (don't know if that's due to OpenRocket, Java, or something else), so take these results with a grain of salt.
Screenshot 2025-10-21 at 14 50 39
Screenshot 2025-10-21 at 14 52 51
Screenshot 2025-10-21 at 15 14 30

Also note that the "committed size" is roughly the same before and after this PR (~ 1.4 GB). This is (mostly) the heap size 
that Java tells the OS it may need. The actual size it uses (in blue) is much lower. So, if you would look at the memory usage in e.g. Activity Monitor on macOS, there may not be a significant difference. I think you would notice it more when you really stress test OR and run it for a prolonged period. That's not something I've done now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,22 +28,19 @@ jobs:
         run: ./gradlew clean check build dist
       # If wanting to serialize all the engines during build
       #        run: ./gradlew clean serializeEnginesDist build dist
-      - name: JaCoCo Coverage Report
-        uses: madrapps/jacoco-report@v1.6.1
-        with:
-          paths: |
-            build/reports/jacoco/jacocoRootReport/jacocoRootReport.xml
-          token: ${{ secrets.GITHUB_TOKEN }}
-          min-coverage-overall: 30
-          update-comment: true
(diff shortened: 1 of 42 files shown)


neilweinstock:
170 files changed, wow. 8-o

PR #2915, opened 2025-10-24
[#2902] Generate preview image when saving ORK file

### LINKED ISSUE #2902: Generate preview image when saving ORK file
When saving a file, OR should generate a preview image and save it into the archive as "preview.png", so the various file previewing features on different OSes can easily pull it out and display it when necessary.

The exact size and composition of the preview image is TBD.  I think it should be based on the existing rocket figure display... I would lean towards the 3D finished view because the appearance settings of the rocket are one of the things that you would want to preview.  Which view to use could be a preference setting, but that is not necessary for initial implementation.

Size should probably be something like 1000 pixels wide, 600-800 pixels tall.

Ideally, this means that the preview image will be rendered to an off-screen canvas before being saved.  It would be easier to simply take a snapshot of the rocket figure display at the moment the file is being saved, and use that.  That would be less desirable (IMHO) for a whole host of reasons, but it could be an easy first step on the way to more rigorous implementation.

SiboVG:
This PR fixes #2902. When saving a design, a file preview image of the 2D side view is generated and included in the .ork as `preview.png`. I attempted to do the same for the 3D view, but that's more tricky... It works if you are already in the 3D view, but to get it working all the time, you would have to do some magic with off-screen buffers and what not. That's for a future implementation... I did already put in the code to change the file preview view type (the code in the preferences panel is commented out).

Unzipped .ork of 'A simple model rocket':
Screenshot 2025-10-24 at 11 45 13

The image (1000x600 px):

preview

Note that I thought it would be helpful to show the flight info too.

### DIFF
--- a/core/src/main/java/info/openrocket/core/document/StorageOptions.java
+++ b/core/src/main/java/info/openrocket/core/document/StorageOptions.java
@@ -16,6 +16,7 @@ public class StorageOptions implements Cloneable {
 	private boolean saveSimulationData = false;
 
 	private boolean explicitlySet = false;
+	private byte[] previewImage;		// File preview image data
 
 	public FileType getFileType() {
 		return fileType;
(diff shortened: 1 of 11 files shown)


neilweinstock:
That's great!  Can be enhanced as needed in the future, but this gives what's needed to make this work.  The flight info is definitely useful.

I wonder if the CG/CP marks should be included?  Not all that important either way.

SiboVG:
> I wonder if the CG/CP marks should be included? Not all that important either way.

Yes. It's added now.

PR #2917, opened 2025-10-30
[#2736] Add Open-Meteo API integration for automatic elevation lookup

### LINKED ISSUE #2736: [Feature request] obtain launch site data from open-meteo.com
At present, users are able to enter latitude, longitude, and elevation data for simulations. It would be a lot more convenient if they could simply enter latitude and longitude, and look up elevation.

https://open-meteo.com offers a free service with a web API  for looking up elevation based on latitude and longitude. We could use this service to fill in the elevation for a location.

They ask that users of their service be non commercial, and limit themselves to fewer than 10,000 API calls per day. While I expect our users (total)  wouldn't come anywhere near that rate, we should probably implement a cache stored on disk with the user preferences, The first time a user looks up a particular location it goes to open-meteo.com, after that a lookup of the same location comes from the cache.

AhanuDewhirst:
This PR fixes #2736.

Adds support for automatically fetching elevation data from the Open-Meteo Elevation API based on latitude and longitude. When the user enters coordinates for a launch site, OpenRocket will try to retrieve the correct elevation automatically instead of requiring manual entry.

A new utility class OpenMeteoAPI was added to handle this. It uses Java’s HttpClient to make asynchronous requests, parses responses using jakarta.json, and caches results using java.util.prefs.Preferences to avoid repeated network calls for the same location. The API is called in the background using CompletableFuture so the UI remains responsive while data is being fetched.

The launch site dialog was updated so that the altitude spinner is disabled during the lookup, then re-enabled once the request completes. Errors, network issues, or invalid responses are logged and result in Double.NaN, leaving the field unchanged. Cached values are reused whenever possible.

image
image


swing/src/main/java/info/openrocket/swing/gui/simulation/SimulationConditionsPanel.java
```diff
@@ -193,10 +197,10 @@ public void stateChanged(EventObject e) {
 
 		m = new DoubleModel(target, "LaunchLatitude", UnitGroup.UNITS_LATITUDE, -90, 90);
 
-		spin = new JSpinner(m.getSpinnerModel());
-		spin.setEditor(new SpinnerEditor(spin));
-		spin.setToolTipText(tip);
-		sub.add(spin, "growx");
+		var latitudeSpinner = new JSpinner(m.getSpinnerModel());
```
SiboVG: Please don't use `var` types; it serves no purpose here and I generally don't like using it in Java.

SiboVG:
Thanks for the PR. However, clicking the meteo button, I got this exception:

```
java.lang.NoClassDefFoundError: java/net/http/HttpRequest
at info.openrocket.core.communication.OpenMeteoAPI.getElevation(OpenMeteoAPI.java:59)
at info.openrocket.swing.gui.simulation.SimulationConditionsPanel.lambda$addSimulationConditionsPanel$2(SimulationConditionsPanel.java:258)
at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:405)
at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:297)
```

Also, there is currently no API throttle. I don't want us to swamp the OpenMeteoAPI when multiple users fiddle with the latitude-longitude. This may not be critical for altitude lookup, but it will once we also use OpenMeteo for weather data.

swing/src/main/java/info/openrocket/swing/gui/simulation/SimulationConditionsPanel.java
AhanuDewhirst: Noted. I will update accordingly.

AhanuDewhirst:
Hi SiboVG,

Thanks for your feedback!

The NoClassDefFoundError is caused by the absence of the java.net.http module. Since I'm unfamiliar with how modules are included in the packaged JRE, and the general rules around JPMS in this project, I'll rewrite the class using HttpsURLConnection.

Regarding API throttling, do you have a rate limit or throttling time frame in mind?

SiboVG:
> The NoClassDefFoundError is caused by the absence of the java.net.http module. Since I'm unfamiliar with how modules are included in the packaged JRE, and the general rules around JPMS in this project, I'll rewrite the class using HttpsURLConnection.

Hm. I can run the program just fine from IntelliJ, but if I build my own JAR, I get the error. Maybe it's an issue in the gradle configuration. I'll investigate it.

> Regarding API throttling, do you have a rate limit or throttling time frame in mind?

OpenMeteo requests a limit of 10,000 API calls per day (per user).

SiboVG:
Ok, found the issue. The JarInJarStarter did not use the JDK classloader as a parent, so the http module (included in the Java 11+ JDK) was not being included.

AhanuDewhirst:
> Ok, found the issue. The JarInJarStarter did not use the JDK classloader as a parent, so the http module (included in the Java 11+ JDK) was not being included.

I'm glad you were able to figure that out for me, thank you!

> OpenMeteo requests a limit of 10,000 API calls per day (per user).

That should be straight forward enough to implement. However, if the user was to change the query parameters in quick succession (e.g. using the sliders), and send the request to open-meteo, should we also throttle the number of calls within a shorter timeframe? If so, what should that number and timeframe be?

JoePfeiffer:
I have to question whether we need rate-limiting. When we pull weather from open-meteo we'll want to have a cache with a one hour expiration; it's really hard for me to imagine we'll have a user who somehow manages to come up with 10,000 API calls a day.

AhanuDewhirst:
> I have to question whether we need rate-limiting. When we pull weather from open-meteo we'll want to have a cache with a one hour expiration; it's really hard for me to imagine we'll have a user who somehow manages to come up with 10,000 API calls a day.

I'm happy to implement it if need be. In the mean time, I will see if I can tackle something else while I wait for the go/no-go :)

thzero:
> I have to question whether we need rate-limiting. When we pull weather from open-meteo we'll want to have a cache with a one hour expiration; it's really hard for me to imagine we'll have a user who somehow manages to come up with 10,000 API calls a day.

Maybe this advanced feature is off by default, and only available if the user supplies their own Open-Meteo key?

JoePfeiffer:
> > I have to question whether we need rate-limiting. When we pull weather from open-meteo we'll want to have a cache with a one hour expiration; it's really hard for me to imagine we'll have a user who somehow manages to come up with 10,000 API calls a day.
> 
> Maybe this advanced feature is off by default, and only available if the user supplies their own Open-Meteo key?

open-meteo API doesn't require a key. When they say "open" they mean it!

thzero:
> > > I have to question whether we need rate-limiting. When we pull weather from open-meteo we'll want to have a cache with a one hour expiration; it's really hard for me to imagine we'll have a user who somehow manages to come up with 10,000 API calls a day.
> > 
> > 
> > Maybe this advanced feature is off by default, and only available if the user supplies their own Open-Meteo key?
> 
> open-meteo API doesn't require a key. When they say "open" they mean it!

Ah, well then no concerns.  Requests would be coming from all sorts of ips.... And yeah who is gonna use 10k in that timeframe?  If they do that only limits them.

Issue #2789, opened 2025-04-16
[Enhancement] Improve UI for low-resolution screens


SiboVG:
See #2788. We should:

- Try to optimize window sizes
- Make dialogs scrollable when resized

Kruzk02:
Hi can I work on this?

SiboVG:
> Hi can I work on this?

Hi, sure, thanks! I will, however, first push some near-finished code I made a while ago for minimizing the motor chooser dialog. I'll do that in the next hour. After that, you can work on the other dialogs.

SiboVG:
@Kruzk02 I pushed PR #2919. So you can look at that, #2788, and #2850 as reference.

Kruzk02:
Should I keep most simple dialogs in **dialogs** package as they are, or still add a scroll pane ? I'm currently working on the **configdialog** package.

SiboVG:
> Should I keep most simple dialogs in **dialogs** package as they are, or still add a scroll pane ? I'm currently working on the **configdialog** package.

You should remove unnecessary blank space in the dialogs, but don't overdo it, and then make sure that when the user resizes the dialog to smaller than the dialog content, that the content becomes scrollable.

Kruzk02:
@SiboVG I saw Simulation dialogs expand their content when stretched, but Component configs dialog stays fixed. Do I need stretched content panel or keep the current?

SiboVG:
> [@SiboVG](https://github.com/SiboVG) I saw Simulation dialogs expand their content when stretched, but Component configs dialog stays fixed. Do I need stretched content panel or keep the current?

It depends on the content. For most of the config dialog content, the widgets don't need to be scaled and can remain static. It's mostly relevant for things like tables, figures (e.g. fin point editor in the freeform fin set)...

thzero:
So what is our min. resolution we are trying to keep to, 1080x720?

https://www.browserstack.com/guide/common-screen-resolutions

SiboVG:
> So what is our min. resolution we are trying to keep to, 1080x720?
> 
> https://www.browserstack.com/guide/common-screen-resolutions

Resolution is not a complete metric. Some people use high-resolution, but a high UI scale. In general, just be sensible and make everything as small as possible AND ensure that you can resize a window without it cutting off widgets (e.g. by adding scroll panes).

PR #2919, opened 2025-11-01
 [#2789] Make motor chooser dialog UI more compact 

SiboVG:
This PR is a partial fix for #2789, namely for the motor chooser dialog. The dialog is more compact and you can resize the window.

Size comparison, with the old dialog in the back:
image

Resizable:

https://github.com/user-attachments/assets/717035e9-8830-4ee8-b8c2-a877b4fb6f28


neilweinstock:
Shouldn't the left panel layout be locked, and only the size of the scrollable motor list changes? A smaller view of the motor list would be more usable than adding a second-level scrollbar.

JoePfeiffer:
Ouch! I wish I'd known you were working on this when I started working on #2900! 

@neilweinstock I'm not quite sure I understand what you're saying. I think of the "scrollable motor list" as being the left side panel. When resizing, I definitely prefer keeping the width of the right side panel fixed while the left gets a scrollbar. I tend to pay little attention to the dimension columns in the motor list, since I've already filtered the motors to dimensions that will work for me.

neilweinstock:
> @neilweinstock I'm not quite sure I understand what you're saying. I think of the "scrollable motor list" as being the left side panel. When resizing, I definitely prefer keeping the width of the right side panel fixed while the left gets a scrollbar. I tend to pay little attention to the dimension columns in the motor list, since I've already filtered the motors to dimensions that will work for me.

@JoePfeiffer I'm talking about height, not width.  Watch the above video at 2 seconds in, the motor list stops shrinking vertically and instead a new scroll bar is added on the right.  My guess is that the minimum size of the scrolling list is set too large; it would still be useful and functional at much smaller size than shown in the video.  Probably, the various minimum sizes should be set so that the second scrollbar *never* appears on the left panel.

JoePfeiffer:
> @JoePfeiffer I'm talking about height, not width. Watch the above video at 2 seconds in, the motor list stops shrinking vertically and instead a new scroll bar is added on the right. My guess is that the minimum size of the scrolling list is set too large; it would still be useful and functional at much smaller size than shown in the video. Probably, the various minimum sizes should be set so that the second scrollbar _never_ appears on the left panel.

Ah, OK -- now I finally see what you're talking about.  Yes, I agree; the motor list itself should be allowed to shrink down until it only has a very minimal number of entries -- on the order of 3 or so -- before the scroll bar on the left pane appears.

SiboVG:
> Shouldn't the left panel layout be locked, and only the size of the scrollable motor list changes? A smaller view of the motor list would be more usable than adding a second-level scrollbar.

It does shrink, but the minimum table height is just set a bit high. I can change that though.

SiboVG:
> Ouch! I wish I'd known you were working on this when I started working on #2900!

My bad! But I don't think there's much overlap, so I don't expect (many) merge conflicts.

SiboVG:
New behavior:

https://github.com/user-attachments/assets/ce397f19-e4ac-4393-96ee-91a86896b749

neilweinstock:
Better… but i would still make the minimum table size even smaller. Maybe 4 lines.

SiboVG:
This is the best I'm willing to do.
image

neilweinstock:
That looks good to me. Sorry to be a pain.

SiboVG:
> That looks good to me. Sorry to be a pain.

No problem — that’s your role on the team: to be a pain for the good of the program. 😉

PR #2932, opened 2025-11-13
Fix small dialog width and center slider alignment

### LINKED ISSUE #2928: [Bug] Some config dialogs are too wide
### Describe the bug

Due to #2925, some dialogs default to a too wide size. For example, the rocket config dialog should not be this wide:

Image

### To Reproduce

.

#### Screenshot(s):

#### .ork file:

### OpenRocket version

unstable

macOS

### Additional context

_No response_

Kruzk02:
this pr fixes #2928.
The commentTextArea had its cols and rows values set too high, which caused the dialog to expand. Additionally, when cols and rows were not set, the dialog could expand infinitely and could not shrink back to its normal size.

### Before
Screenshot 2025-11-13 122021

### After
Screenshot 2025-11-13 122133


SiboVG:
Much better, thanks!

PR #2933, opened 2025-11-13
[#2100] Allow for decal deletion + add decal previews

### LINKED ISSUE #2100: Can't Delete Unused Appearance files...
Since I can't successfully edit an appearance file anymore, I have some of them that are 11 files deep now, and the accidental carryover from a copied part...  I'd love to be able to select those extra files... and delete them.

image

SiboVG:
This PR fixes #2100. You can now delete the selected decal of a component. A drawback is that you do first have to select the decal before you can delete it. Maybe in the future, we can add delete buttons in the combobox items, or even better, add a decal manager tool where you can also swap decals from a new file (useful if you use the same decal for multiple components).

Additionally, I moved the "Select file" option from the combobox to a dedicated button, which should be more clear.

This PR also adds decal previews. When you hover over the texture combobox,, you get a preview of the currently selected decal and which components use it. If you open the combobox, you get the same preview window when hovering over the different items:

https://github.com/user-attachments/assets/c169ea38-4356-4fca-86a1-f926edf3d1f3

### DIFF
--- a/core/src/main/java/info/openrocket/core/document/DecalRegistry.java
+++ b/core/src/main/java/info/openrocket/core/document/DecalRegistry.java
@@ -119,6 +119,13 @@ public class DecalRegistry {
 
 		return decals;
 	}
+	
+	boolean removeDecal(DecalImage decal) {
+		if (decal == null) {
+			return false;
+		}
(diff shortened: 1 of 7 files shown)


neilweinstock:
Very nice.

PR #2934, opened 2025-11-14
[#1994] Add decal texture creation service

### LINKED ISSUE #1994: Feature request: generate default texture image file
To create a texture (decal) file for OR, you typically must 
a) Get the measurements of the component
b) Calculate the appropriate dimensions of the image, depending on the resolution you want
c) Create an image with those dimensions
d) OPTIONAL: if doing fin decals, it is also quite helpful to create a layer with the outline of the fin, so you know where your design should be.  This step can be a fair bit of work.
e) Add that image to the component appearance.

It should be possible for OR to save most of this effort.  From the appearance tab, there could be a "create new texture" button, which then asks for a desired DPI and file name, and then generates a blank image file of the appropriate name and dimensions and opens it in an image editor.  All by itself that would be a very nice feature.

To make it even better, when creating the file for a fin set, add an appropriately-oriented fin outline to the image.

I haven't tried to design the exact UI for this yet.

SiboVG:
This PR fixes #1994, allowing you to generate new textures based on the rocket component dimensions. For fin sets, the fin outline can be drawn in the texture template, if the "Draw component outline" checkbox is enabled.

https://github.com/user-attachments/assets/54bc9f9e-e164-4d05-88bd-847e6a470292


neilweinstock:
You did an amazing job on this.  I only have three small comments:
1) When creating texture for a "right side" fin, the fin outline should be flipped horizontally (front of fin on the left instead of the right)
2) I think the fin outline should be finer.  I'm not sure it needs to be thicker than 1 or 2 pixels (at most).  Extra thickness could get in the way a bit when doing the decal design
3) This one is a little more nuanced: I think the fin outline should be drawn "outside" the actual fin, so the decal design can be done inside it.  In other words, if I have a rectangular fin, I shouldn't see any outline at all because the outline is just outside the viewable area.  In the case of a trapezoidal fin, I wouldn't see the root or the tip outline.  This is how I try to do it when I do this by hand.

SiboVG:
> * When creating texture for a "right side" fin, the fin outline should be flipped horizontally (front of fin on the left instead of the right)
> 
> * I think the fin outline should be finer.  I'm not sure it needs to be thicker than 1 or 2 pixels (at most).  Extra thickness could get in the way a bit when doing the decal design
> 
> * This one is a little more nuanced: I think the fin outline should be drawn "outside" the actual fin, so the decal design can be done inside it.  In other words, if I have a rectangular fin, I shouldn't see any outline at all because the outline is just outside the viewable area.  In the case of a trapezoidal fin, I wouldn't see the root or the tip outline.  This is how I try to do it when I do this by hand.

Fixed those + added settings for outline width and color.
image

neilweinstock:
Almost perfect, just one last item: when creating a right side fin decal, you have the outline rotated 180 degrees from where it needs to be.  Front root of the fin should be in the upper left corner of the image.

Another thing I noticed: If I change a decal, then "cancel" out of the fin component editor, the decal changes stick.  Should that be the case?  My gut is that, if possible, decal changes should cancel along with everything else.  I don't know how much mess it adds to do that... conceivably it could be a future addition, don't necessarily need to do it right now if it's a lot of work.  Feature as it stands is incredibly useful.

SiboVG:
> Almost perfect, just one last item: when creating a right side fin decal, you have the outline rotated 180 degrees from where it needs to be. Front root of the fin should be in the upper left corner of the image.

Right. Should be fixed now I think.



> Another thing I noticed: If I change a decal, then "cancel" out of the fin component editor, the decal changes stick. Should that be the case? My gut is that, if possible, decal changes should cancel along with everything else. I don't know how much mess it adds to do that... conceivably it could be a future addition, don't necessarily need to do it right now if it's a lot of work. Feature as it stands is incredibly useful.

I'm not seeing this. If I add a fin, click "Ok" in the config dialog, then edit the fin again, make changes to the decal, and then cancel, the decal changes have reverted correctly.

SiboVG:
@neilweinstock could you check the fin decals + clarify the issue you experienced the cancel action?

neilweinstock:
Will do

neilweinstock:
The fin decal creation looks good now.

I am still seeing the cancel problem.  If I edit a decal, save it, then hit "Cancel" to exit out of the fin config dialog, the decal remains changed.  This is on Windows in case it matters.

SiboVG:
> I am still seeing the cancel problem. If I edit a decal, save it, then hit "Cancel" to exit out of the fin config dialog, the decal remains changed. This is on Windows in case it matters.

I see now. It happens when you try to edit an existing decal; not undo creating a new one. To be fair, it would take a bit of effort to fix that + it's not a result of this PR. If you're ok with it, we can create a new issue for it and merge this PR.

PR #2936, opened 2025-11-17
Check for null component when saving

### LINKED ISSUE #2873: Error during save of project

JoePfeiffer:
fixes #2873


JoePfeiffer:
Naw, this isn't the way to do it.

PR #2938, opened 2025-11-18
[#2870] Add support for other component SVG exporting

### LINKED ISSUE #2870: SVG export for centering rings, bulkheads, etc.
Thank you so, so much for adding SVG export of fins to the latest release. Big time saver for those of us that use laser cutters for custom fins and have been tracing them from the PDF. 

Is there any way you could extend that functionality to the centering rings and bulkheads? It would be amazing if I could just export an SVG for the whole rocket that lets me cut fins, centering rings, bulkheads… I could even cut and emboss the paper fin guide. And I know people cut out profiles of the nose cone to use with a lathe. Really, everything that prints out on the PDF would be amazing if we could get it in SVG format. Folks with laser cutters have been exporting and carefully tracing, which is a huge pain.

I think it would be perfectly acceptable to add SVG export as an option in the main menu. I typically arrange and cut all my fins, centering rings, bulkhead, etc. out of one piece of 3 mm ply or balsa, depending on the size of the rocket.

Alternatively, it could be an SVG export button on the individual parts, as you’ve done for fins.

Or both! That might be even better.

SiboVG:
This PR fixes #2870 by adding SVG export support for all applicable rocket components (_unless I missed some_).

The export functionality is available both in the `File` menu and in the context menu:
image
image

Because you can access it in those 2 places, I have decided to remove the "Export to SVG" button from the fin component config dialog.

---

Clicking on the export to SVG button opens a config dialog:

image

You can choose various settings, such as stroke color and width, but also to draw the center crosshair for ring components, and to include the component names in the SVG file. Finally, you can also configure the part spacing if you have multiple parts in one SVG sheet. On the right, you can select which components you which to export. By default, it selects the component that you have selected in the rocket design view. I wrapped this panel in a tabbed panel, so that if we want to support extra SVG exports in the future, for instance a fin alignment guide, we can add new tabs for this.

---

Here is an example export of the 3-stage example rocket:

Sustainer-profile 00 12 34

Note that for tube and ring components, both the side view and the back view is exported.


neilweinstock:
Very nice.  I'm not sure if the side view is useful for the tube and ring components, but I guess it can't hurt to include them.

SiboVG:
> Very nice. I'm not sure if the side view is useful for the tube and ring components, but I guess it can't hurt to include them.

Indeed, I was hesitating to add them, but could be useful to some.

PR #2944, opened 2025-11-24
[#2816] Add simulation branch listeners

### LINKED ISSUE #2816: [Feature Request] Need simulation listeners for beginning and end of sim branch execution
At the moment their are simulation listeners defined for the beginning and end of the simulation, but not for each simulation branch. Because of the order in which simulation branches are executed, for a multistage rocket this results in the listener being called for the first stage booster, not for the sustainer.

For the stuff I'm working on right now, I'm trying to get it to be called when the sustainer finishes. Having a listener that's called at the end of each simulation branch would be a more general solution.

There should be a startSimulationBranch() listener called when each simulation branch begins executing, and an endSimulationBranch() listener called at the end of the branch

**Additional context**

SiboVG:
Fixes #2816 by adding `startSimulationBranch()` and `endSimulationBranch()` listener methods to `SimulationListener` interface to allow listeners to be notified when each simulation branch begins and ends. This allows listeners to detect when specific branches (e.g., sustainer) finish.


JoePfeiffer:
Thank you!

PR #2946, opened 2025-11-26
Allow finset config subclass to expand and shrink

Kruzk02:
This is similar to my previous pull request, with one minor issue: when the dialog is resized smaller to the point where the button panel layout can no longer shrink, it forces the scroll pane’s vertical bar to stop at the minimum width of the button panel. This causes part of the dialog content to be cut off a part of dialog.

https://github.com/user-attachments/assets/55d4f2bb-945e-48bf-90c4-3be9e80a05d0

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/configdialog/EllipticalFinSetConfig.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/configdialog/EllipticalFinSetConfig.java
@@ -35,11 +35,11 @@ public class EllipticalFinSetConfig extends FinSetConfig {
 		DoubleModel m;
 		JSpinner spin;
 		
-		JPanel mainPanel = new JPanel(new MigLayout());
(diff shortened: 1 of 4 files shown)


SiboVG:
Great, thanks!

Kruzk02:
Now that the rocket component dialog is complete, do any other dialogs need changes? Most of them seem fine or small.

SiboVG:
> Now that the rocket component dialog is complete, do any other dialogs need changes? Most of them seem fine or small.

For the component configs, this PR appears to be the final puzzle piece. However, I do notice that the component analysis window (Tools > Component Analysis) and Custom Expressions window (Tools > Custom Expressions) don't show scroll bars when you resize.

Eventually, we would also need to find a way to compact the UI elements themselves (e.g. remove the sliders and create a new "draggable spinner" widget with a small progress bar below it).

Thanks a lot for your work on this!

PR #2947, opened 2025-11-26
[Fixes #2592] Make plot legend sensitive and embolden plot lines

### LINKED ISSUE #2592: [Feature Request] make plot legend sensitive and embolden plot lines
When a user is examining simulation results, the various types of flight data being displayed are color coded with a legend at the bottom of the window identifying each color. It is sometimes (particularly when many types of data are being displayed) difficult to determine which plot line is actually being displayed in which color.

If the user hovers the mouse over one of the plot types in the legend, make that plot line bold so it stands out.

First, of course, one can already hover over a data point; the tooltip displays which data type is being viewed. This seems sort of backward, as I end up hovering over several lines trying to find the one I'm looking for.

Also, we could display the color swatch in the legend as a small block instead of a line; the greater area will make the color of the swatch clearer. This might be a good idea in addition to the enhancement being requested here.

OPaivaHeitor:
This PR fixes [#2592](https://github.com/openrocket/openrocket/issues/2592), in which the user requested for lines in the plot window to be highlighted upon being hovered on by the mouse, while remaining highlighted upon a mouse click.

The proposed solution: I decided to implement a set that would highlight plot lines by replacing the plotted line with another one with double de width. The original plot line is stored in the baseStrokes map along with a string to identify it so it can later be recovered to remove the highlighting.

Additionally, I implemented the mouseMoved and mouseClicked methods to retrieve entities in order to properly manipulate them.

Finally, the current implementation only removes highlighting from plot lines if the user clicks on the same line to unhighlight it, which allows the user to highlight all of the lines on the plot if they want.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/plot/SimulationChart.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/plot/SimulationChart.java
@@ -184,8 +184,9 @@ public class SimulationChart extends ChartPanel {
 	}
 
     /**
-     * <p>Called whenever the mouse moves.</p>
-     * <p>Gets an entity at the location of the mouse, if this entity is a LegendItemEntity, calls updateHighlightingSet.</p>
+     * Called whenever the mouse moves.
+     *
+     * Gets an entity at the location of the mouse, if this entity is a LegendItemEntity, calls updateHighlightingSet.
      */
     @Override
     public void mouseMoved(MouseEvent e) {
@@ -204,7 +205,7 @@ public class SimulationChart extends ChartPanel {
     }
 
     /**
-     * <p>Responsible for adding and removing labels from the selectedLegendLabels set before calling updateHighlightingSet.</p>
+     * Responsible for adding and removing labels from the selectedLegendLabels set before calling updateHighlightingSet.
      */
     @Override
     public void mouseClicked(MouseEvent e) {
--- a/swing/src/main/java/info/openrocket/swing/gui/plot/SimulationChart.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/plot/SimulationChart.java
@@ -1,8 +1,6 @@
 package info.openrocket.swing.gui.plot;
 
-import java.awt.Color;
-import java.awt.Cursor;
-import java.awt.Point;
+import java.awt.*;
 import java.awt.event.MouseEvent;
 import java.awt.event.MouseWheelEvent;
 import java.awt.event.MouseWheelListener;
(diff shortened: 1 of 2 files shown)


neilweinstock:
Thank you!  Can you provide a screenshot or quick video capture of your implementation?

OPaivaHeitor:
Of course! I'll send a video as soon as I have some spare time to record it.

JoePfeiffer:
Very nice! A few thoughts:
1. I wonder if it should be boldened even further -- maybe triple width instead of double?
2. Another nice touch would be to put a border around the plot label color swatch to further indicate which ones are being highlighted.
3. We typically don't include HTML tags in comments. Please remove them.

OPaivaHeitor:
> Thank you! Can you provide a screenshot or quick video capture of your implementation?

Hello, you can watch a quick video showcasing the implemented feature here: 


https://github.com/user-attachments/assets/18777eaa-44b8-4c7a-bdcf-2465b665e169

> Very nice! A few thoughts:
> 
> 1. I wonder if it should be boldened even further -- maybe triple width instead of double?

Thanks for the feedback!

Indeed, just doubling the width sometimes is subtle enough for the change to go unnoticed. On the other hand, if the lines are excessively boldened it starts hindering readability for data points. See example with quadruple with: 

image


> 3. We typically don't include HTML tags in comments. Please remove them.

Noted. I'll remove them along the next few commits.

> 2. Another nice touch would be to put a border around the plot label color swatch to further indicate which ones are being highlighted.

Could you clarify what you have in mind for the border around the plot label color swatch? Would it be a box surrounding the highlighted elements in the legend?

JoePfeiffer:
> > 1. I wonder if it should be boldened even further -- maybe triple width instead of double?
> 
> Thanks for the feedback!
> 
> Indeed, just doubling the width sometimes is subtle enough for the change to go unnoticed. On the other hand, if the lines are excessively boldened it starts hindering readability for data points. See example with quadruple with:

Yes, I'm just sort of guessing what would be a good width. After seeing quadruple width, triple seems like a good bet.

> > 2. Another nice touch would be to put a border around the plot label color swatch to further indicate which ones are being highlighted.
> 
> Could you clarify what you have in mind for the border around the plot label color swatch? Would it be a box surrounding the highlighted elements in the legend?

Yes, something like this:
selected-plot
(that's got a 4 pixel border; not sure what the right width would be).

OPaivaHeitor:
I added a 4 pixel width border and now the plot window should look something like this:

https://github.com/user-attachments/assets/3e0ab6f5-c3fd-4258-b12a-0058080e3e7e

swing/src/main/java/info/openrocket/swing/gui/plot/SimulationChart.java
```diff
@@ -1,15 +1,20 @@
 package info.openrocket.swing.gui.plot;
 
-import java.awt.Color;
-import java.awt.Cursor;
-import java.awt.Point;
+import java.awt.*;
```
JoePfeiffer: Please import classes individually

JoePfeiffer:
Just one more thing -- please import classes individually, rather than using a wildcard. I left a note in the "files changed"

OPaivaHeitor:
> Please import classes individually, rather than using a wildcard.

That's fixed, apologies for having overlooked that!

JoePfeiffer:
Thank you for your contribution to OpenRocket!

PR #2949, opened 2025-11-28
Replace spinner + slider with draggable spinner

SiboVG:
In light of #2789, this PR replaces the spinner (+ unit selector) + slider setup in the component config dialogs and the simulation config dialog with a new widget, `SpinnerWithSlider` that incorporates a progress bar beneath the spinner, and draggable up/down buttons of the spinner. For most dialogs, this reduces the window size slightly.

Behavior:
- You can still click on the up and down buttons to increment/decrement the spinner value
- Clicking and dragging will enable _slider_ mode. If you drag up and/or right, the spinner value increases. If you drag down and/or left, the spinner value decreases.
- If you drag while holding the SHIFT key, you can enter fine-control mode, where the step increment is 1/10th of the normal value
- Note that the click-dragging only works on the spinner buttons; not the spinner text itself. I thought this may be confusing for users who also use click-drag to select text. Well, you would have to make a choice - disable text selection, use vertical dragging for slide mode, horizontal dragging for text selection mode, or only use slide mode when the text box is not selected. But that complicates things IMO. I also don't think a lot of users use the sliders, to it's not too bad that the functionality is slightly hidden. I did keep the sliders in PhotoStudio, since there, you probably use sliders more than text input.

You can set the drag sensitivity in the UI preferences.

Demo:

https://github.com/user-attachments/assets/71ff5c08-855c-4557-bbb5-42bd173131ec


neilweinstock:
I have to say I have mixed feelings about this one.  It's certainly interesting, and the compactness is nice, but I think it's less user-friendly than the existing sliders.  Maybe it's just me but I do not really enjoy using these new controls.

I have three specific issues:
1) When hovering over the up/down arrows, the new cursor icon almost completely obscures the buttons underneath.  So if you want to click an up or down arrow, (a) you can't see what you're clicking, and (b) it doesn't feel natural (or comfortable) to click on a button with that new cursor.

2) The blue progress bar is not intuitive to me.  Here's what a fin editor looks like:
image.  To my eyes, the array of blue progress bars doesn't convery information to me without my having to really stop and think about what I'm seeing.  To be fair, the current sliders don't either, but I generally don't concentrate on them because they're out of my view when I'm focusing on the column of spinners.  The blue progress bars, on the other hand, demand my attention, but I can't easily tell what I'm looking at.  The difference in behaviors of positive-only fields (extending from left to right) vs. +/- fields (which extend from the middle towards either edge) contributes to the lack of clarity.

3) I've never been a fan of controls that you control by dragging up and down "in the air".  That is, not directly manipulating a control, but just abstractly dragging up and down.  A lot of audio software uses virtual knobs that work this way, never been a fan there either but at least in that case it's in service of using a familiar affordance (the physical "knob").  Here I don't feel like it's anything familiar.

As a side note, I'll mention that the Photo Studio settings are not using these new controls.  That panel just looks so much more inviting with those sliders.

Let's see what others think.

jimmiedave:
It's an interesting idea, and a real good try, but I have reservations.

Fooling around with these, my initial takes:

1. The movement is kinda stuttery, not smooth, here on macOS 14.7.4 Sonoma, M3 MAX CPU.
2. It's hard to make sense of the growing/shrinking bar when it's not clear what the universe of values for the widget could be. If it were 0-100, that would be pretty easy: no bar you're at 0, full bar at 100, proportionally in-between. But it's not clear when one or both endpoints are unbounded. Is there a maximum length I could make the body tube? If so, how can I know what it is from the widget? If there's not, how can the growing bar show proportionality? And if the widget clips the bar at the widget width, if I slide back down, is the new visually-clipped value the max now, and the proportionality changes?
3. It's not clear whether the increment of the values from sliding the cursor can be controlled. Is it always the least-significant digit shown when the scroll starts? Always .1 of the widget's order of magnitude? Does the increment change by widget (i.e. .01 for tube thickness, 1 for tube length)? Is it speed-sensitive? (if speed sensitive, the choppiness/stuttering really matters).
4. Unusual to control a horizontally-growing bar with vertical movement. These motions should be in some kind of sync - at least one of cardinal direction: E-W movement to affect an E-W growing/shrinking bar. And if we DO sync horizontal movement to horizontal bar, that's at odds with the up-down arrows beside the number in the widget.

Is it worth thinking about keeping vertical motion and incrementing the numbers (smoothly) but losing the bar where it's unbound?

JoePfeiffer:
I've never been fully convinced that the sliders on these controls are actually useful, and seeing this change reinforces that. I'd suggest:
1. Get rid of the slider.
2. Only make the drag behavior sensitive to vertical motion, not horizontal.

Regardless of what happens with those, under Linux when the pointer is over the spinner the cursor changes to an 'n-resize' cursor, like this:
n-resize
This doesn't convey, to me anyway, what it's going to do.

neilweinstock:
The sliders are useful sometimes, *especially* in Photo Studio, where they are an absolute lifesaver.  In the config dialogs... not that often but I have used them sometimes.  But this change feels to me like a step back in clarity and usability, unfortunately.

JoePfeiffer:
> The sliders are useful sometimes, _especially_ in Photo Studio, where they are an absolute lifesaver. In the config dialogs... not that often but I have used them sometimes. But this change feels to me like a step back in clarity and usability, unfortunately.

Good point on photo studio. It's says something about how rarely I use it that their use there didn't occur to me!

Issue #2454, opened 2024-02-02
[Bug] Some component default wall thicknesses are 0


JoePfeiffer:
### Describe the bug

Default wall thickness of coupler and engine block, and (if a shoulder is added to a nose cone) shoulder diameter and thickness for nose cone are all 0.

### To Reproduce

1. Create a nose cone
2. Extend the shoulder 1 cm
3. Diameter and thickness are 0. It would be better if they were (nose cone diameter) - 2*default body tube diameter
4. Add a body tube
5. Add a coupler or engine block to the body tube. Their wall thicknesses are 0. Note this is particularly misleading for the engine block, since the length is non-zero and in side view it looks like there's something there.
...

#### Screenshot(s):
*(drag-and-drop the screenshot(s) here)*

#### .ork file:
*(drag-and-drop the file here as a .zip file)*

### OpenRocket version

22.02, 23.09, unstable

### What platform are you running on?

Linux

### Additional context

This is on the border between a bug report and a feature request. Either way, should be fixed.

SiboVG:
> 3\. nose cone diameter) - 2*(default body tube diameter), and thickness was (default body tube diameter)

Shouldn't this be `shoulder diameter = (nose cone diameter) - (nose cone thickness)` and `thickness = (nose cone thickness) / 2`? That way, the shoulder starts halfway the cone wall and ends at the inner wall. I'm not sure if that's the most common configuration in real life nose cones.

JoePfeiffer:
Yikes, what was I thinking? What I wrote up there is obvious nonsense. 

Looking at a low power nose cone from my parts box, there isn't any obvious internal "step", the thickness of the shoulder is the thickness of the nose cone minus the thickness of the matching body tube. So a PNC-50KA nose cone has a base diameter of 2.48cm and a wall thickness of .157cm; it mates with a BT-50A body tube with a .033cm wall thickness so the shoulder thickness should be .157-.033=.124mm. Unfortunately the shoulder thickness OR calculates for it is 0.

hazard3045:
Hello, I am a french student and I would like to contribute to this issue.

Is it still open?

I would like to implement two cases:
* **sole nose cone:** implementing `shoulder diameter = (nose cone diameter) /2 ` and `shoulder thickness = (nose cone diameter)/4` , this way by default the shoulder is filled and ends halfway through the nose cone diameter.

[...]

PR #2952, opened 2025-12-02
[#2454] Fix default wall thickness of the shoulder of the nose cone to 0.

hazard3045:
The issue happened when a shoulder was created on a nose cone.

Now, when the lenght of the shoulder is set to a non-zero, the diameter and the radius are automatically set.
Likewise, when the lenght is set to 0, the others parameters too.
_End Capped_ parameter is now set to true by default.

This PR **does not** adress the issue with the default wall_thickness of the engines.

This issue was resolved in the _NoseCone_ class, but the bug is coming from the _Transition_ class (of which it inherits), if you want me to do so, you reject this PR and I can implement it in the mother class.

Capture d’écran 2025-12-02 091632


core/src/main/java/info/openrocket/core/rocketcomponent/NoseCone.java
```diff
@@ -111,12 +113,45 @@ public double getShoulderLength() {
...
 	 * it works for both normal and flipped nose cones.
+     * If a shoulder is created, automatically set the thickness of the shoulder
+     * to the thickness of the nose cone, and the diameter of the shoulder to the diameter
+     * of the nose cone minus 2 times the thickness of the nose cone, likewise, if
+     * the soulder is deleted (length to 0), automatically set radius and diameter to 0.
 	 */
 	public void setShoulderLength(double length) {
 		if (isFlipped) {
-			setForeShoulderLength(length);
+            if (getShoulderLength() == 0.){
```
SiboVG: You don't have to add a decimal after a `0`.

SiboVG (changes requested):

SiboVG:
Thanks for the change! I left a small comment in the code.

JoePfeiffer:
> This issue was resolved in the _NoseCone_ class, but the bug is coming from the _Transition_ class (of which it inherits), if you want me to do so, you reject this PR and I can implement it in the mother class.

Yes, let's fix it where the bug is actually located (in ~~SymmetricComponent~~ _Transition_) and let NoseCone inherit the fixed behavior.

Unfortunately I can't try it out at the moment since it's getting what looks like failures in the unit tests. Please update the unit tests as well.

core/src/main/java/info/openrocket/core/rocketcomponent/NoseCone.java
JoePfeiffer: Really ought to be using MathUtil.equals(double a, double b) rather than a crisp equality test (also other places where there are comparisons against 0, if any).

hazard3045:
The change has been brought from NoseCone to Transition (I did not actually understand why you talked about making changes in SymmetricComponent as BodyTube -which inherits from SymmetricComponent- does not seem to have shoulders).
The new PR includes 2 commits:

- The first one is mandatory for the PR and is actually the change expected
- The second one is just making sure that the dimensions of the new shoulder adapts to the linked component if any (a liberty I took, this commit is optional and can be ignored)

I also made sure to update the unit tests accordingly (which I forgot last time, sorry for that).

JoePfeiffer:
>  (I did not actually understand why you talked about making changes in SymmetricComponent as BodyTube -which inherits from SymmetricComponent- does not seem to have shoulders).

Oops. Fixed now.

neilweinstock:
@SiboVG are there still a requested change here?  Github is saying there is but it's not clear if that's really still true given the comment stream.

JoePfeiffer:
This is frustrating -- I want to rerun the checks since so much has happened since December, but I'm not finding the "re-run all checks" button

SiboVG:
> This is frustrating -- I want to rerun the checks since so much has happened since December, but I'm not finding the "re-run all checks" button

Doesn't appear to be possible anymore. I think this branch needs to be updated to unstable.

SiboVG (approved):

JoePfeiffer:
Thank you for your contribution! I'm sorry it sort of wound up on the side for a while...

PR #2950, opened 2025-11-30
[Feature Request] Show More Details in Parachute Parts Library #2487

luisroliveira:
This PR adds support for new parameters in the parachute library, allowing them to be displayed as sortable and filterable columns in the component selection interface. The following attributes have been added:
- Mass
- CDArea
- DragCoefficient
- SurfaceArea
- PackedLength
- PackedDiameter

Issue #2487, opened 2024-05-27
[Feature Request] <Show More Details in Parachute Parts Library>


Aeroengineer1:
It would be nice to have the option to show things like Packing Dimensions, Weight, CD, and Mass as sortable and filterable columns for the parachute parts library.  I did take a look at the library source code, and it seems that some of these things get computed when the user adds one of these components to their model and others are direct atributes in the library.

The reason that this would be useful is that one could use this to optimize their selection criteria on these engineering parameters.  For example finding the lightest solutions that provide the smallest packing while maintaining a constant CD*Area figure of merit.  

That actually brings up another thought.  That would be to allow custom expressions to create a figure or merit calculation in the selection tool, or at least include a figure of merit of CD*Area.  This figure of merit would be useful because the terminal velocity equation would then only have rocket mass as the remaining variable, whereas, CD*Area are a function of the parachute all other aspects of drag being equal for the rocket.   This would then allow the user to select their parachute on this figure of merit along with the other aspects such as packing volume, cost, etc.

Adam

PR #2953, opened 2025-12-02
 [Feature Request] Show More Details in Parachute Parts Library #2487 #2950 

luisroliveira:
This PR adds support for new parameters in the parachute library, allowing them to be displayed as sortable and filterable columns in the component selection interface. The following attributes have been added:

    Mass
    CDArea
    DragCoefficient
    SurfaceArea
    PackedLength
    PackedDiameter

Some columns remained empty because, for certain types of parachutes, not all information is available in the database.

https://github.com/user-attachments/assets/e1a89d21-62b3-47f5-9e89-84f07f69c519


JoePfeiffer:
I see that the new columns are sortable, but how are they filterable? ~~Also, for the sake of people on small screens, would it be reasonable to make it possible to hide table columns?~~

neilweinstock:
> I see that the new columns are sortable, but how are they filterable? Also, for the sake of people on small screens, would it be reasonable to make it possible to hide table columns?

My strategy is generally just to put the columns I don't care about all the way to the right.  Out of sight, out of mind.

However, I just went into the parts library and it looks like if you right click on a header you can show/hide all the available columns.  Who knew!

JoePfeiffer:
> Who knew!

Not me, obviously! I just edited my comment to strike eliminate that suggestion.

SiboVG:
I would hide some of these options by default. The dialog is _really_ packed now...
image

SiboVG:
> I would hide some of these options by default. The dialog is _really_ packed now... image

And/or change the table to have a horizontal scroll bar.

PR #2958, opened 2025-12-05
Allow UI Theme to be dynamically updated

SiboVG:
No more restart required to change the UI theme.

https://github.com/user-attachments/assets/dea13283-0071-436a-b35f-6b5bcedc3514


neilweinstock:
Awesome!

Issue #2864, opened 2025-08-02
[Feature Request] Re-label, Re-design, or Remove "Reset to Default" and "Save as default" buttons from "Edit simulation">"Launch conditions"


jimmiedave:
**Is your feature request related to a problem? Please describe.**
At this writing, OpenRocket's 24.12 Launch Conditions tab has "Reset to Default" and "Save as default" buttons the don't appear to do anything, or at least do nothing effective.

**Describe the solution you'd like**

- At a minimum, determine what the buttons are _supposed_ to be doing, make them do that and re-label them so what affordance is being offered is very clear.
- If the buttons can be made to do something more useful for the user with clarity, a redesign of this feature may be warranted.
- If it's not possible to determine the original intent of the buttons, nor describe a reasonable use case for them, potentially remove the buttons from the panel.

**Describe alternatives you've considered**
Just plain pull the buttons off and see if anyone cares.

**Additional context**
It seems likely that a user changing multiple existing simulations (vs. creating new ones) would not like to continuously type the same datum in the same field for each sim. Likewise, user probably doesn't want to have to bounce between Preferences and sim edits to get the job done. There's probably some clear use case that makes sense, but there's got to be either a better way to address it in UI, or to label the existing buttons (after making them actually _do_ something).

<img width="500" height="398" alt="Image" src="https://github.com/user-attachments/assets/279e9ebb-f217-4f0e-b90d-f82a144bf7d1" />

jimmiedave:
Here's an idea for redesign, based on @JoePfeiffer 's comment:

1. remove current "Save as Default", "Reset to Default" buttons -> or rename and adjust them (see later)
2. Add to each simulation line a trio of icons: "Edit", "Duplicate", "Delete". Each executes the named operation on the line item clicked (not for any other selected items). These icons do not change or interfere with the top-of-list buttons, at least not for this design option. The top-of-list buttons can still execute operations on multiple selected items.
3.  The "Edit" operation begins a standard Edit on the clicked item, same as top-of-screen button with a single item selected.
4. The "Duplicate" operation makes a duplicate of the item clicked, with the exception of the simulation name, which will be named the same as the item clicked, suffixed with " (copy)" so that it's clear which simulation data sourced the duplicate.
5. The "Delete" operation offers to delete the clicked simulation, with an "Are you sure" dialog.

[...]

PR #2960, opened 2025-12-05
Add per-row action buttons and improve simulation management UI

mira194:
### Summary
This PR implements per-row action buttons in the simulation table and streamlines the simulation management interface by removing redundant default preference buttons, based on @JoePfeiffer's design proposal.

#### 1. SimulationPanel - Per-row action buttons
- ✅ Added "Tools" column with Edit/Duplicate/Delete icons on each simulation row
- ✅ Edit: Opens simulation editor for the clicked row
- ✅ Duplicate: Creates a copy with " (copy)" suffix appended to name
- ✅ Delete: Shows confirmation dialog before deletion
- ✅ Per-row actions operate independently of table selection
- ✅ Top toolbar buttons continue to work on multiple selected rows
- ✅ Hand cursor feedback when hovering over action buttons

**Technical implementation:**
- Custom `ToolsColumnRenderer` for displaying action buttons
- Custom `ToolsColumnEditor` for handling button interactions
- `MouseListener` with position detection for reliable click handling
- New methods: `editSimulationAtRow()`, `duplicateSimulationAtRow()`, `deleteSimulationAtRow()`

#### 2. SimulationPanel - Duplicate button in toolbar
- ✅ Added Duplicate button to the main toolbar
- ✅ Works on multiple selected simulations
- ✅ Properly integrated with existing actions

#### 3. SimulationConditionsPanel - Removed default preference buttons
- ✅ Removed `addDefaultButtons()` method
- ✅ Removed "Save as Default" and "Reset to Default" buttons
- ✅ Cleaner, less cluttered interface

#### 4. Translations
- ✅ Added English translations in `messages.properties`:
  - `simpanel.but.duplicatesimulation = Duplicate`
  - `simpanel.but.ttip.duplicatesim = Create a copy of the selected simulation(s)`

#### 5. Tests
- ✅ Verified per-row actions work correctly
- ✅ Verified toolbar duplicate button works on single and multiple selections
- ✅ Verified removal of default buttons doesn't break functionality

### Screenshots
image
image
image


swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -301,7 +286,37 @@ public void mouseClicked(MouseEvent e) {
 				int selectedRow = simulationTable.getSelectedRow();
 				int row = simulationTable.rowAtPoint(e.getPoint());
 				int column = simulationTable.columnAtPoint(e.getPoint());
-
+                    // ========== NOUVEAU CODE - GÉRER LA COLONNE TOOLS ==========
```
SiboVG: Don't use French in your code. We only use English.

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -29,32 +24,15 @@
...
-import javax.swing.ListSelectionModel;
-import javax.swing.SwingUtilities;
+import javax.swing.*;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableColumn;
 import javax.swing.table.TableColumnModel;
-
+import java.awt.Insets;            // Pour Insets
```
SiboVG: Don't add comments for imports.

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -29,32 +24,15 @@
...
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JPopupMenu;
-import javax.swing.JScrollPane;
-import javax.swing.JTable;
-import javax.swing.KeyStroke;
-import javax.swing.ListSelectionModel;
-import javax.swing.SwingUtilities;
+import javax.swing.*;
```
SiboVG: Don't use wildcard imports

SiboVG (changes requested):

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -1401,12 +1416,179 @@ private void updateBoxColors(WarningsBox box, Color foreground) {
 			}
 		}
 	}
+    /**
```
SiboVG: You are writing a lot of unnecessary duplicate code. You should reuse the SimulationAction actions from `swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java`.

SiboVG (changes requested):

SiboVG:
Thank you for the changes. I added some comments to the code.

mira194:
Thank you for your feedback, I will fix that right away.

SiboVG:
Btw, please change the title of this PR, as it has nothing to do with issue #2864.

SiboVG:
We also recommend that you create a new git branch each time you work on a new issue or feature. You are currently working on your unstable branch.

SiboVG:
A few more comments:
- Please rename the 'Tools' column to 'Actions'
- Please add tooltip texts to the action buttons, like "Edit simulation", "Duplicate simulation", and "Delete simulation" – you can reuse existing translation keys for this
- Please use English git commits instead of French :)

SiboVG (approved):

core/src/main/resources/l10n/messages.properties
```diff
@@ -499,7 +499,14 @@ update.fetcher.badResponse = Bad response code from server: %d
 update.fetcher.badConnection = Could not connect to the GitHub server. Please check your internet connection.
 update.fetcher.malformedURL = Malformed URL: %s
 
+
 ! Simulation edit dialog
+# Simulation Panel - Tools column
```
SiboVG: You may remove this comment

core/src/main/resources/l10n/messages.properties
```diff
@@ -499,7 +499,14 @@ update.fetcher.badResponse = Bad response code from server: %d
...
+simpanel.col.Tools = Tools
```
SiboVG: Please rename this to "Actions" (both the key and the value)

core/src/main/resources/l10n/messages.properties
```diff
@@ -499,7 +499,14 @@ update.fetcher.badResponse = Bad response code from server: %d
...
+simpanel.dlg.deleteSingle.title = Delete Simulation
```
SiboVG: Please remove these translations; they are not used and not needed

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -1665,7 +1782,27 @@ public Double valueAt(int row) {
 
 							return data.getGroundHitVelocity();
 						}
-					}
+					},
+                    //// Tools column - per-row actions
+                    new Column("Actions") {
```
SiboVG: Please use the `simpanel.col.Actions` translation key here

SiboVG (changes requested):
Thank you for the quick changes! I left some additional comments. Once that's done, this PR should be good to go.

JoePfeiffer:
Nice! Should the "Actions" column's default location be between "Warnings" and "Name", rather than over on the right?

neilweinstock:
Good point.

I don't know *exactly* where it should be, but it has to be somewhere on the left because otherwise the buttons can get scrolled off the right edge when the window is small.  That is the disadvantage of this sort of approach, which can be a bit problematic given the way the sim results table works.

On the left, this column will add width to the table and push everything else further right, another disadvantage when the window is smaller.

Nonetheless I like this and think it's worth a try.

thzero:
> Good point.
> 
> I don't know _exactly_ where it should be, but it has to be somewhere on the left because otherwise the buttons can get scrolled off the right edge when the window is small. That is the disadvantage of this sort of approach, which can be a bit problematic given the way the sim results table works.
> 
> On the left, this column will add width to the table and push everything else further right, another disadvantage when the window is smaller.
> 
> Nonetheless I like this and think it's worth a try.

Yeah, that's the rub with table output.   Ive seen action buttons (on left to right UIs) on the left, and on the right over the decades.

But doesn't this duplicate the functionality of right-clicking on the row?   Is there an option to turn off the button?

neilweinstock:
> Yeah, that's the rub with table output. Ive seen action buttons (on left to right UIs) on the left, and on the right over the decades.

I've seen it work better on tables that are fixed width, so the buttons can't scroll off the right edge.

> But doesn't this duplicate the functionality of right-clicking on the row? Is there an option to turn off the button?

Yes, it duplicates the right-click menu *and* the buttons above the table.  The current build still has the buttons on the right, which in my mind is a non-starter, since they'll very frequently be scrolled off the right.

I'm not generally a fan of the existing buttons above the table, either in the sim table or the motor config table (and the two should stay consistent; if the in-row buttons are added here they should be added there as well).  I make mistakes with them all the time.  Whether this will be an improvement I'm not sure, and won't know until I get to see a build with the buttons on the left.  I'm going to need to go look at some other apps with similar types of tables to see what generally works well.

The right-click menu is good but it's hidden, so there need to be some sort of visible functions to manage the table entries.  Whether that's in-row buttons or the existing above-table buttons I'm not sure yet.

thzero:
> Yes, it duplicates the right-click menu _and_ the buttons above the table. The current build still has the buttons on the right, which in my mind is a non-starter, since they'll very frequently be scrolled off the right.
> 
> I'm not generally a fan of the existing buttons above the table, either in the sim table or the motor config table (and the two should stay consistent; if the in-row buttons are added here they should be added there as well). I make mistakes with them all the time. Whether this will be an improvement I'm not sure, and won't know until I get to see a build with the buttons on the left. I'm going to need to go look at some other apps with similar types of tables to see what generally works well.
> 
> The right-click menu is good but it's hidden, so there need to be some sort of visible functions to manage the table entries. Whether that's in-row buttons or the existing above-table buttons I'm not sure yet.

Guess I've never had an issue with the top-row buttons, nor the right-click which is a pretty well known pattern for decades.

Most of the 'action' column I've ever seen has typically been on websites that spit out data in tabular format of some type.

But I'd agree, it being on the right is not really a good idea at all.

PR #2964, opened 2025-12-07
[#2955] Tailcone scaling sometimes fails

### LINKED ISSUE #2955: Tailcone scaling sometimes fails
Sometimes when scaling the entire rocket the aft diameter of a tailcone is not scaled.

It looks like the trigger is when the fore diameter of the tailcone is set to "automatic".  If I set the fore diameter normally, it scales correctly.

Note that the aft diameter is *not* set to automatic, so it should scale properly in all circumstances.

[tailcone-scaling.zip](https://github.com/user-attachments/files/23962802/tailcone-scaling.zip)

BaknineNouhaila03:
This PR fixes #2955


SiboVG:
Thanks!

PR #2967, opened 2025-12-12
Fix finset snap drag highlight color + add point context menu

SiboVG:
This PR fixes an issue in the freeform finset point figure. The issue:
1. Go into snap-drag mode by holding the Shift key, or Shift-Ctrl key
2. Start dragging
3. Release the key modifier from step 1, while continuing to drag
4. The snap behavior stopped, but the line segment is still drawn in the snap highlight color

I also added the ability to right-click a point in the fin editor to show a context menu to delete the current point. I intuitively always right-click a point in the editor, only to be disappointed that nothing happens...

Demo showing old behavior first and then the new behavior:

https://github.com/user-attachments/assets/290d5498-f206-458f-bf8e-f6664ad7edfc


neilweinstock:
Very nice.

PR #2968, opened 2025-12-12
 [#2956] Snap view rotation when holding shift key 

### LINKED ISSUE #2956: Add snapping feature to rotation in side and back views
It would be nice if the rotation slider (or click-dragging the rocket itself) supported snapping.  I would enable the snapping while holding down the control key (similar to freeform fin editor).

Which angles should it snap to?  Well, there are smart and dumb ways to do it.  I think the dumb way would serve the vast majority of uses: snap to the nearest multiple of 30 or 45 degrees.

The smart way would be to look at the design and figure out what angles would be relevant.  That seems needlessly difficult to get started.  I would start with the dumb way and it can always be enhanced in the future if desired (it won't be).

I made a brief attempt to implement this myself and immediately got bogged down in the weirdness of the Jslider control, so I gave up.

SiboVG:
Fixes #2956. You can snap the view to multiples of 30° or 45° by holding down the `Shift` key. The original issue proposed using the `Ctrl` key, but `Shift` is more standard in other software. The finpoint editor also uses `Shift` (and `Shift+Ctrl`).

https://github.com/user-attachments/assets/3e1f0ba7-86cc-4f18-9422-a4fbe53f55a9


neilweinstock:
Nice, thank you!

PR #2970, opened 2025-12-13
 [#2560] Add FlatLafOutlines helper and improve user feedback for excessive inputs

### LINKED ISSUE #2560: [Feature Request] Use FlatLaf's JComponent.outline for error and warning feedback
Currently, when users enter dubious values in input fields, there's no immediate visual feedback to indicate which field is problematic. This can lead to confusion and delays in correcting input errors.

Implement FlatLaf's JComponent.outline property to render input widgets with an "error" or "warning" look. This can be achieved by calling myComponent.putClientProperty("JComponent.outline", "warning") on the component when a dubious value is entered. This will provide immediate visual feedback to users, highlighting which input field requires attention.

image

I haven't explored other alternatives as this solution is already included in FlatLaf by default and seems to be the most straightforward implementation.

SiboVG:
This PR fixes #2560 and adds a new class, `FlatLafOutlines`, to highlight a widget in a warning or error outline if a certain condition is met.

---

To get the ball rolling, the following validations are already implemented:
- Simulation config
      - Pressure: warning outline if launch pressure is < 0.75× or > 1.25× ISA pressure at the selected altitude (sea-level equivalent: < ~760 mbar or > ~1267 mbar)
      - Temperature (only when ISA is off): warning if ≤ -40°C or ≥ 50°C, error if ≤ -80°C or ≥ 80°C
      - Launch rod angle: warning if |angle| ≥ 45° from vertical, error if |angle| ≥ 60°
      - Average wind speed: warning if ≥ 10 m/s, error if ≥ 20 m/s
- Multilevel wind CSV import dialog
      - Column name mappings (when “has header” is enabled): error outline if required names (altitude/speed/direction) are empty, or if any non-empty names are duplicated (after trimming)
      - Column index mappings (when “has header” is disabled): error outline if two mappings use the same index (std deviation optional)

The threshold values are of course a bit subjective, so let me know if they should be changed.

---

When a validation threshold has been reached, a small tooltip popup will also be shown with a message, e.g. "Pressure is too low".

---

Demo:

https://github.com/user-attachments/assets/50e68395-33c8-4939-ba2c-91ca12cb7b2c

---

If you have other areas of the program where this validation could be applied to, please let me know.


JoePfeiffer:
Would this be a better way to do the outlines in #2592?

SiboVG:
> Would this be a better way to do the outlines in ~~#2592~~ #2947?

That's for plotting, right? Which means it's related to JFreeChart, not FlatLaf.

JoePfeiffer:
> > Would this be a better way to do the outlines in ~#2592~ #2947?
> 
> That's for plotting, right? Which means it's related to JFreeChart, not FlatLaf.

I don't *think* the legends are done by JFreeChart. Would need to check....

PR #2972, opened 2025-12-13
 [#1838] Embed thrust curve points in .ork 

### LINKED ISSUE #1838: Export motors with .ork file
When a .ork file is opened, OR attempts to find any motors used in configurations in the motor database.  If users are using custom thrusturves and sharing .ork files, the result is the motors just get lost when the file is opened.  As an example, see Issue #1837, in which the user uploaded a .ork file with custom motors and I had to request that he also upload the motors (in this case I think I was able to figure out what was going on without seeing his motors, but the point remains).

We should save the thrustcurves with the .ork and load them without going to the database, like we do for other components.

SiboVG:
This PR fixes #1838 by embedding thrust curve points and standard motor delays of a motor in the .ork file:

Example:
```xml
<motor configid="05896b5a-71a7-48ac-b7ae-eadeb267975b">
  <type>single</type>
  <manufacturer>Estes</manufacturer>
  <digest>c15b9b96bf06e0ab896394787da3c47e</digest>
  <standarddelays>2,4</standarddelays>
  <thrustcurvepoint>0,0,0.035,0.019</thrustcurvepoint>
  <thrustcurvepoint>0.02,0.418,0.035,0.019</thrustcurvepoint>
  <thrustcurvepoint>0.04,1.673,0.035,0.019</thrustcurvepoint>
  <thrustcurvepoint>0.065,4.076,0.035,0.019</thrustcurvepoint>
  <thrustcurvepoint>0.085,6.69,0.035,0.019</thrustcurvepoint>
  <thrustcurvepoint>0.105,9.304,0.035,0.018</thrustcurvepoint>
  <thrustcurvepoint>0.119,11.496,0.035,0.018</thrustcurvepoint>
  <thrustcurvepoint>0.136,12.75,0.035,0.018</thrustcurvepoint>
  <thrustcurvepoint>0.153,11.916,0.035,0.018</thrustcurvepoint>
  <thrustcurvepoint>0.173,10.666,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.187,9.304,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.198,7.214,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.207,5.645,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.226,4.809,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.258,4.182,0.035,0.017</thrustcurvepoint>
  <thrustcurvepoint>0.326,3.763,0.035,0.016</thrustcurvepoint>
  <thrustcurvepoint>0.422,3.554,0.035,0.016</thrustcurvepoint>
  <thrustcurvepoint>0.549,3.345,0.035,0.015</thrustcurvepoint>
  <thrustcurvepoint>0.665,3.345,0.035,0.015</thrustcurvepoint>
  <thrustcurvepoint>0.776,3.345,0.035,0.014</thrustcurvepoint>
  <thrustcurvepoint>0.863,3.345,0.035,0.014</thrustcurvepoint>
  <thrustcurvepoint>0.94,3.449,0.035,0.013</thrustcurvepoint>
  <thrustcurvepoint>0.991,3.449,0.035,0.013</thrustcurvepoint>
  <thrustcurvepoint>1.002,2.404,0.035,0.013</thrustcurvepoint>
  <thrustcurvepoint>1.01,1.254,0.035,0.013</thrustcurvepoint>
  <thrustcurvepoint>1.03,0,0.035,0.013</thrustcurvepoint>
  <designation>B4</designation>
  <diameter>0.018</diameter>
  <length>0.07</length>
  <delay>4.0</delay>
</motor>
```

When loading a .ork file, the following sequence is followed:
1. First, try to load the motor from the database
2. If the motor was not found in the database, load it from the embedded motor data

This ensures that when the motor database is updated, this is reflected when opening your file (instead of first trying to load the motor data from the .ork file).

### DIFF
--- a/core/src/main/java/info/openrocket/core/file/openrocket/importt/MotorHandler.java
+++ b/core/src/main/java/info/openrocket/core/file/openrocket/importt/MotorHandler.java
@@ -54,22 +54,35 @@ class MotorHandler extends AbstractElementHandler {
 	 * <p>
 	 * Preference order:
 	 * <ol>
-	 *   <li>Embedded thrust curve data (one or more {@code <thrustcurvepoint>} elements), if present.</li>
 	 *   <li>Lookup via {@link DocumentLoadingContext#getMotorFinder()} (typically the motor database).</li>
+	 *   <li>Embedded thrust curve data (one or more {@code <thrustcurvepoint>} elements), if present.</li>
 	 * </ol>
 	 *
 	 * @param warnings warnings sink
 	 * @return the resolved motor, or {@code null} if not found/invalid
 	 */
 	public Motor getMotor(WarningSet warnings) {
+		// First try to locate an equivalent motor in the motor database.
+		// If that fails, fall back to the embedded thrust curve data (if present).
+		WarningSet databaseWarnings = new WarningSet();
+		Motor databaseMotor = context.getMotorFinder().findMotor(type, manufacturer, designation, Double.NaN, Double.NaN, digest,
+				databaseWarnings);
+		if (databaseMotor != null) {
+			warnings.addAll(databaseWarnings);
+			return databaseMotor;
+		}
+
 		if (!thrustCurveTime.isEmpty()) {
 			Motor embedded = buildEmbeddedThrustCurveMotor(warnings);
 			if (embedded != null) {
+				// Deliberately discard databaseWarnings here: the motor isn't missing since we loaded it from the file.
 				return embedded;
 			}
 		}
-		return context.getMotorFinder().findMotor(type, manufacturer, designation, Double.NaN, Double.NaN, digest,
-				warnings);
+
+		// Nothing worked: surface any database lookup warnings (e.g. missing motor).
+		warnings.addAll(databaseWarnings);
+		return null;
 	}
 
 	/**
--- a/core/src/main/java/info/openrocket/core/file/openrocket/importt/MotorHandler.java
+++ b/core/src/main/java/info/openrocket/core/file/openrocket/importt/MotorHandler.java
@@ -1,7 +1,9 @@
 package info.openrocket.core.file.openrocket.importt;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Locale;
+import java.util.List;
 
 import info.openrocket.core.logging.Warning;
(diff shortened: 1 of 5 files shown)


JoePfeiffer:
This is awesome! It'll be a great help to people using motors not yet in the database, and people flying research motors.

JoePfeiffer:
I'm late to the party here, but.... is it better to embed the points in the rocket.ork part of the file, or should there be a /thrustcurves directory added to the zip file, with the file added in its native format and read at startup? The advantage to the latter would be adding new thrustcurve formats (if that ever happens) would automatically add whatever new data is in the new format.

SiboVG:
> I'm late to the party here, but.... is it better to embed the points in the rocket.ork part of the file, or should there be a /thrustcurves directory added to the zip file, with the file added in its native format and read at startup? The advantage to the latter would be adding new thrustcurve formats (if that ever happens) would automatically add whatever new data is in the new format.

Fair point. However, with PR #2991 (and also the previously used `thrustcurves.ser`), the motor data is not contained in a single .rasp or .eng file, so you would have to reconstruct a new engine file. Not something we cannot overcome though.

I'm not sure what the best strategy is. We would need to come to a consensus before the next release though, so we don't need to deal with backward compatibility in next releases...

neilweinstock:
My gut reaction was to agree with Joe… throwing motor files into the ORK archive just seems cleaner. But if the files need to be generated… I dunno. 🤷‍♂️ 

in general I’m a little confused by how this one will behave… I haven’t had a chance to try it out and understand how different corner cases are handled. 

One factor in deciding this particular question might be how older versions of OR will react to these files. My guess is that old versions will ignore unrecognized files in the archive, but will complain about extra stuff in the XML. Which is preferable?

JoePfeiffer:
I'd expect generating a .rse file to be similar in effort to putting the motor data in the rocket.ork as is currently happening (except, of course that the current code already exists).

Something I'd like to see down the line is the ability to export a .rse file, so people working with experimental motors could more easily share them. Putting them in the zip directory would be most of the work for that task, all by itself.

PR #2975, opened 2025-12-15
Increase fin set shape editor handle sensitivity, clear selection when clicking empty space, and improve performance 

SiboVG:
This PR fixes a few pet peeves I have with the fin set shape editor.

Firstly, when I want to drag a point, I often find myself clicking just outside the point handle box, thus unintentionally adding a new point. This PR increases the sensitivity of the handle by 25%, which is not too large so you can still add a new point close to the point, but noticeable enough to fix my miss-clicks issue.

Secondly, when I click empty space in the shape editor, I expect the currently selected point to become deselected - as is the case in the rocket view. This PR fixes that.

Lastly, I feel like the performance of the shape editor is sluggish... This PR fixes that by freezing the rocket while you are dragging a point in the shape editor, and to thaw the rocket after you finished dragging the point. In other words: the rocket design view will not be update in real-time _while_ dragging a point, but only _after_ you finished dragging.

For users who still need the real-time rocket view updates (maybe this is used for advanced rocket designs, idk), there is a checkbox in the Design preferences to re-enabled the old behavior. By default, the checkbox is unchecked.

image

---

Demo of the old behavior first and then the new one:

https://github.com/user-attachments/assets/985e0e8f-6242-4ff7-8e90-c5dc88786587


neilweinstock:
I have definitely used live-updating of the rocket view while adjusting fin shapes, but I don't need it all the time.  Might there be a way to use a modifier key to enable live update?

In that case I'd get rid of the preference and just have live update off except when modifying.  

Or else, maybe put the checkbox in the editor itself.  I'd prefer not to have to go back to preferences to toggle it on and off.

But I am definitely looking forward to the improved performance; looks great in the screenshots above.

SiboVG:
> Might there be a way to use a modifier key to enable live update?

Perhaps. But there are already quite a few key modifiers in the shape editor.

> Or else, maybe put the checkbox in the editor itself. I'd prefer not to have to go back to preferences to toggle it on and off.

That's what I did first, actually :). I could add it back in, but do you have a suggestion on the checkbox text and/or where to place it? I first used "Update rocket while dragging". In general, I'm not a huge fan of the button mess on the bottom section of the dialog.

neilweinstock:
Hmm, I'll need to think about it.  I agree that the modifiers are heavily loaded, and the buttons are a mess down there.  Functionally I like a keyboard modifier for this sort of thing but it is probably just too much (especially since you might want to do it while doing the lock angle, in which case it becomes crazy).

Maybe Scale Fin, Import from image, Export CSV, and Split fins could be condensed into a pulldown of some sort?  That's not really a standard way to do an *action* but I'm trying to think of a more compact way to present those functions.  Each of those tends to be an operation that you'll only do once, if at all so it seems OK to throw them into some type of junk drawer.  Github kind of does that with their buttons, e.g.:
image
Pulldown to select a function, then click the button to execute.

I suppose "Split Fins" should stay where it is for consistency with Pods, so it's really the other three that could condense, and then there'd be room for a "live update" checkbox.

I'll think about it.

SiboVG:
New UI:

https://github.com/user-attachments/assets/6dabd7c5-592d-43f5-bbe7-090f7adf00cb

neilweinstock:
Nice, thank you!

PR #3002, opened 2026-01-10
Add damping moment coefficient, corrective moment coefficient, damping ratio, and natural frequency to simulation data

SiboVG:
This PR adds computation of the damping moment coefficient (there was already a simulation extension for this, but now it's directly supported in the simulation), corrective moment coefficient, and damping ratio to the simulation. The latter is useful to investigate how the rocket oscillates - see peak of flight issue 197.

The new coefficients:
image
image

Comparison of A simple model rocket in OR and RockSim:
image
image


SiboVG:
I also added the natural frequency computation (oscillation frequency)

PR #3013, opened 2026-01-18
Fix RK6 stepper k-value accumulation bug

abwdn:
The RK6 algorithm requires accumulating multiple weighted k-values when computing intermediate positions. The previous code called setRocketPosition multiple times for k4, k5, k6, and k7 calculations, where each call overwrote the previous result instead of accumulating.

For example, computing k4 position requires:
  y + (1/12)*k1 + (1/3)*k2 + (-1/12)*k3

The fix chains the addScaled() calls together so all terms are accumulated in a single coordinate, matching the correct RK6 formulas.

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/RK6SimulationStepper.java
+++ b/core/src/main/java/info/openrocket/core/simulation/RK6SimulationStepper.java
@@ -269,23 +269,16 @@ public class RK6SimulationStepper extends AbstractSimulationStepper {
         double weightk3 = -1.0/12;
         status2 = status.clone();
         status2.setSimulationTime(status.getSimulationTime() + store.timeStep*1/3);
+        // Chain addScaled calls so all k-values are accumulated (not overwritten)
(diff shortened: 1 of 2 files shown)


SiboVG:
Hm, do you think the RK4 stepper suffers from the same issue @abwdn ? We did a refactor to improve the efficiency, hence the toImmutable calls. Can the same calls be reintroduced here, but using the correct equation?

abwdn:
Checked RK4 - it's not affected. The RK4 equations only use one k-value per intermediate step (k2 uses k1, k3 uses k2, k4 uses k3), so there's nothing to accumulate. RK6 is different because its intermediate steps combine multiple k-values together.

As for reintroducing the toImmutable calls - I don't think that's possible without breaking the accumulation. Each call resets from the base position, so the previous terms get lost. The chained approach also ends up being more efficient since we're only creating one immutable object instead of three.

SiboVG:
Thanks for the change @abwdn!

PR #3012, opened 2026-01-17
 [#2734] Add design view capturing to file and clipboard 

SiboVG:
Fixes #2734.

https://github.com/user-attachments/assets/c2d18ee4-5ec0-4edd-ae69-fa011fdc001e

PR #3014, opened 2026-01-18
[#2106] Allow user to configure the background and text color of the design view

### LINKED ISSUE #2106: We really need to be able to choose a background color in the build window...
It's hard to see white on white with the shine at 100%... (How many white fins do I have there?  Are any missing?)

image

And the photo studio tool, which is nice, doesn't help when I'm trying to show off my work the way I want it to be seen...

image

I'd really love to see a background colour change option.

Thanks!

SiboVG:
This PR fixes #2106. You can now specify a custom color for the background and text in the 2D and 3D view. This setting is saved for each document separately (in document preferences), or can be configured as an application default.

Note that this PR is branched from PR #3012, so that would need to be merged first.

https://github.com/user-attachments/assets/174e4c06-cd78-4ec7-ba18-73d898fd49e2


neilweinstock:
This looks good.  I don't foresee using it much but it will make some people happy.

PR #3021, opened 2026-01-22
Set AoA to 0 until the launch rod is cleared

SiboVG:
When the rocket has not yet cleared the launch rod, the angle of attack is 90° (because `airspeed.getZ() = 0`, causing `Math.acos(airspeed.getZ() / velocity)` to be 90°). This doesn't make sense because while the rocket is on the launch rod, the rocket centerline and velocity vector are parallel.

This PR sets the AoA to 0 if the launch rod has not yet been cleared.

Before:
image

After:
image

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/AbstractSimulationStepper.java
+++ b/core/src/main/java/info/openrocket/core/simulation/AbstractSimulationStepper.java
@@ -84,7 +84,7 @@ public abstract class AbstractSimulationStepper implements SimulationStepper {
 
 		double velocity = airSpeed.length();
 		store.flightConditions.setVelocity(velocity);
-		if (velocity > 0.01) {
+		if (status.isLaunchRodCleared() && velocity > 0.01) {
 			// aoa must be calculated from the monotonous cosine
 			// sine can be calculated by a simple division
 			store.flightConditions.setAOA(Math.acos(airSpeed.getZ() / velocity), len / velocity);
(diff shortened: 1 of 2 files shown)


JoePfeiffer:
But the angle of attack isn't the angle between the rocket's orientation and its velocity vector, it's the angle between its orientation and the relative airflow. It can easily be non-zero while it's on the rod.

It will be 90 before launch, and then decreases as the rocket accelerates.

SiboVG:
> But the angle of attack isn't the angle between the rocket's orientation and its velocity vector, it's the angle between its orientation and the relative airflow. It can easily be non-zero while it's on the rod.
> 
> It will be 90 before launch, and then decreases as the rocket accelerates.

That's what I first thought, but in Sampo's thesis states it is:
image

JoePfeiffer:
That's an error on his part (which I hadn't noticed before!). Note the next paragraph (correctly) says the fins provide a correcting moment when the rocket flies at an angle of attack; that can only be true if the AoA is relative to the airflow.

SiboVG:
> That's an error on his part (which I hadn't noticed before!). Note the next paragraph (correctly) says the fins provide a correcting moment when the rocket flies at an angle of attack; that can only be true if the AoA is relative to the airflow.

Ok. Good to know my intuition was right. Bad to know I didn't follow it.

SiboVG:
Actually, this brings up another issue: the AoA when on the launch rod should only be 90° if the launch rod is vertical. If the launch rod is angled, the rocket is not a perfect 90° wrt the wind. So, the current calculation still gives wrong results.

JoePfeiffer:
> Actually, this brings up another issue: the AoA when on the launch rod should only be 90° if the launch rod is vertical. If the launch rod is angled, the rocket is not a perfect 90° wrt the wind. So, the current calculation still gives wrong results.

Checking.... while my statement was incomplete as you point out, I just tried pointing the simple model rocket example 6 degrees into the wind, and the angle of attach before launch was 84.

Though this does bring up a small annoyance I've had: the AoA is always reported as being positive making it look like it's "bouncing" off 0. Reporting both positive and negative would show it oscillating around 0.

PR #3031, opened 2026-02-01
[#3027] Use AppleScript clipboard image copying on macOS

### LINKED ISSUE #3027: Regarding PR #3012: Cannot paste copied image to TRF on Mac
Copying the design area to clipboard seems to use a format on the Mac that the TRF software can't handle.  When I try to paste into a post (MacOS 26.2, MS Edge) it just spins forever.

I'm guessing this is not a problem on Windows but I haven't been able to test it yet.

SiboVG:
Fixes #3027. For macOS, the design view image clipboarding is now handled by an AppleScript subprocess. This appeared to be the only way to get it to work. I tested different Java native implementations with different clipboard image data flavors, but none of that worked.


thzero:
Seems less than ideal to put in more OS specific work arounds.

But if doing this, shouldn't this RocketPanel.java be updated too?

Issue #2757, opened 2025-03-22
[Enhancement] Don't allow a sim to be created with no motors


JoePfeiffer:
This is just barely different enough from #2755 that i'll make it a new feature request instead of reopening that one: it shouldn't be possible to create a simulation with no motors in the first place.

Similar to the "Motors & Configuration" tab not permitting a new configuration to be created if there are no motor mounts defined, the "Flight simulations" tab should display a similar message when there are no configurations defined to guide the user to create at least one configuration (and the "New simulation" should be disabled in this case).

When a new simulation is being created, the "[No motors]" configuration should not be displayed as an option.

PR #3039, opened 2026-02-11
Prevent creating simulation without motor and configuration

Kruzk02:
## Description
Fixes part of #2757: Prevent users from creating a simulation without selecting a configuration, avoiding invalid simulations.

### Simple tube

https://github.com/user-attachments/assets/3985872b-6ddf-4956-809d-fbc0f8d45b8e

### with example

https://github.com/user-attachments/assets/4bdcb829-0213-40f1-85f5-5e45424ab320

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
@@ -54,8 +54,6 @@ import javax.swing.SwingConstants;
 import javax.swing.SwingUtilities;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
-import javax.swing.event.TableModelEvent;
-import javax.swing.event.TableModelListener;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.TableColumn;
 import javax.swing.table.TableColumnModel;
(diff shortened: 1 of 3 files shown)


neilweinstock:
Seems reasonable.

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -1,13 +1,7 @@
...
 
 
-import java.awt.BorderLayout;
-import java.awt.Color;
-import java.awt.Component;
-import java.awt.Container;
-import java.awt.Dimension;
-import java.awt.Toolkit;
-import java.awt.Window;
+import java.awt.*;
```
SiboVG: Please do not use wildcard imports (it's less efficient than explicit imports). I know IntelliJ tends to to this automatically. You can disable this in the settings though.

swing/src/main/java/info/openrocket/swing/gui/main/SimulationPanel.java
```diff
@@ -30,28 +24,11 @@
...
-import javax.swing.JLabel;
-import javax.swing.JOptionPane;
-import javax.swing.JPanel;
-import javax.swing.JPopupMenu;
-import javax.swing.JScrollPane;
-import javax.swing.JTable;
-import javax.swing.KeyStroke;
-import javax.swing.ListSelectionModel;
-import javax.swing.SwingUtilities;
+import javax.swing.*;
```
SiboVG: See previous comment.

SiboVG (changes requested):
Thanks for the changes! I've left a few small comments.

SiboVG:
I guess it would make sense to disable all buttons in the sim panel, not just the "New simulation" button.

Kruzk02:
Sorry, i have to revert and force push the branch after accidentally commiting .idea/ files.

SiboVG:
Thanks for the changes @Kruzk02! One small issue we still have is that in a sim config, we can still select a "No motors" config, but that's not something I want to fix in this PR, as it requires a bit more thought to also keep backward compatibility with design files that have such a config selected.

EDIT: I've split up #2757 into two tasks to address my concern.

PR #3045, opened 2026-03-05
Tests/add disable stage tests

### LINKED ISSUE #1693: Write unit tests for disable stage in BarrowmanCalculator and MassCalculatorTest
There are now unit tests for verifying correct simulation results when disabling a stage, but it would be better to also add some unit tests for it in `BarrowmanCalculator.java` and `MassCalculatorTest.java`. Maybe it would even be reasonable to replace the current `DisableStageTest.java` with the previously mentioned tests, since `DisableStageTest.java` can still sometimes fail due to simulation inaccuracies.

shawnmcmanus:
Closes #1693

### Description

This is my first open source contribution, and I chose this issue because I have a personal interest in rocketry and wanted to contribute to a project I find genuinely exciting.

This PR adds deterministic unit tests for the disable-stage functionality in `BarrowmanCalculatorTest.java` and `MassCalculatorTest.java`, as suggested in issue #1693.

The existing `DisableStageTest.java` verifies disable-stage behavior through full simulations, which can occasionally produce flaky failures due to simulation inaccuracies. The new tests avoid this by asserting on analytical outputs instead:

- **`BarrowmanCalculatorTest.testDisabledStageAerodynamics()`** — verifies that disabling the core and booster stages correctly reduces CNa, shifts the center of pressure forward, and that re-enabling all stages fully restores the original values
- **`MassCalculatorTest.testDisabledStageMassAndCG()`** — verifies that disabling the core and booster stages correctly reduces structural mass, produces the expected center of gravity, and that re-enabling all stages fully restores the original values

Both tests use the existing Falcon 9 Heavy test rocket and its established constants, so the expected values are cross-validated against the existing test suite.

### Testing

All existing tests in both `BarrowmanCalculatorTest` (20 tests) and `MassCalculatorTest` (33 tests) pass with the new additions included.

### DIFF
--- a/core/src/test/java/info/openrocket/core/aerodynamics/BarrowmanCalculatorTest.java
+++ b/core/src/test/java/info/openrocket/core/aerodynamics/BarrowmanCalculatorTest.java
@@ -766,4 +766,46 @@ public class BarrowmanCalculatorTest {
 		}
 		assertEquals(11, boosterOnlyForceMap.size(), "Force map should contain 10 components");
 	}
+
+	@Test
+	public void testDisabledStageAerodynamics() {
+		Rocket rocket = TestRockets.makeFalcon9Heavy();
+		FlightConfigurationId fcid = new FlightConfigurationId(TestRockets.FALCON_9H_FCID_1);
+
+		Simulation simulation = new Simulation(rocket);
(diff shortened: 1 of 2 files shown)


SiboVG:
Thanks a lot for your contribution!

PR #3052, opened 2026-03-14
[#1029] Allow for custom plot line color and style 

### LINKED ISSUE #1029: New feature request: Ability to edit plot appearance
It could be a nice feature if you were able to edit the plot appearance for a certain simulation. So you would be able to change the line color of e.g. the altitude data set or the line style (solid, dotted, dashed...), or even the background color and grid lines color. In this way, users have more freedom over how their plots look, also allowing to edit the plot for better contrast between data sets.

SiboVG:
Fixes #1029. You can now configure the plot line color and style for each data type. These settings are stored in the .ork file.

image

image


neilweinstock:
When I just tried this out, the plot dialog opened too small to fit all the new UI:
image

The plot legend gets a bit wonky with different line styles but I think it's probably OK:
image

I tend to think that this would make more sense as a global app preference rather than a per-rocket setting.  Would be interested in hearing others' opinions.

JoePfeiffer:
Which plots I want to see depend on what I'm trying to figure out for a particular rocket. My guess is it's more useful as it is here.

SiboVG:
> When I just tried this out, the plot dialog opened too small to fit all the new UI:
> image

I don't think that's a result of this PR. I've noticed it already several weeks before this PR. I haven't looked into fixing it yet.

> I tend to think that this would make more sense as a global app preference rather than a per-rocket setting. Would be interested in hearing others' opinions.

I believe there are cases where you do want per-sim specific appearances, for instance if you want Plot A to use a solid line for altitude, but in Plot B a dashed line. But I also like the idea of a global app preference. Color-blind users could then customize the colors to their needs.

Would a "Set as default" button in the plot appearance row achieve the best of both worlds?

neilweinstock:
> Would a "Set as default" button in the plot appearance row achieve the best of both worlds?

Probably, yeah.  So in that case, when would you save the settings with the ORK file?  Only when they're different from default?

JoePfeiffer:
@SiboVG do you want to add the "set as default" or should we merge it now?

SiboVG:
> > Would a "Set as default" button in the plot appearance row achieve the best of both worlds?
> 
> Probably, yeah. So in that case, when would you save the settings with the ORK file? Only when they're different from default?

Yes, something like that.

SiboVG:
> @SiboVG do you want to add the "set as default" or should we merge it now?

I'll take a look at it this week.

SiboVG:
Done. I added a "Set as default" and "Reset to factory default" button. To not clutter the UI, I used icon buttons (this SVG icon replacement is really paying off).

https://github.com/user-attachments/assets/5fdce654-8b25-49eb-ac36-9f92dd89aefb

I do still want to change SwingPreferences to not use `type.getName()`, but `type.getSaveKey()`. But for that , #3109 needs to get merged first.

neilweinstock:
What are the tooltips on those two new buttons?

SiboVG:
> What are the tooltips on those two new buttons?

For set as default:

```
Set as default
Save the current color and line style as the application default for this data type
```

For reset to factory default:

```
Reset to factory default
Revert the color and line style to their built-in defaults for this plot
```

neilweinstock:
That sounds good.  I'm not sure I'm totally on board with the choics of icons for those buttons, but that's a good enough starting point (the "save as default" button looks like a bookmark icon, which doesn't map to the function in my mind... is that a common usage?)

SiboVG:
> That sounds good. I'm not sure I'm totally on board with the choics of icons for those buttons, but that's a good enough starting point (the "save as default" button looks like a bookmark icon, which doesn't map to the function in my mind... is that a common usage?)

There really is no standard icon for this functionality. This is the closest I could find.

PR #3061, opened 2026-03-20
fixed bug

AdithiMK:


JoePfeiffer:
Whatever changes you tried to make don't appear to have been committed before you opened your PR. All I see here is a single new, empty HTML file. Also, you'll need to be more specific about what bug you are fixing and how your PR fixes it.

SiboVG:
Maybe an AI leftover? 🙂 And is the issue only in the rendering or also in the calculations?

neilweinstock:
I'm inclined to close this one and let the submitter start again, if in fact there is anything to submit.

JoePfeiffer:
My guess is he came across a tutorial somewhere on how to submit a PR on github and used us to try it out.

PR #3066, opened 2026-03-23
Store embedded thrust curves as .rse files in .ork zip archive

SiboVG:
See the discussion in #2972. Instead of embedding motor thrust curve data as `<thrustcurvepoint>` XML elements, motors are now stored as .rse files in a `thrustcurves/` directory within the .ork zip archive, keyed by motor digest (e.g. thrustcurves/<digest>.rse). This is a more flexible way of storing thrust curves for future motor export features.

image

Motor loading precedence:

  1. Motor database lookup (by manufacturer, designation, digest)
  2. Embedded .rse file from thrustcurves/ zip directory


JoePfeiffer:
Very nice -- this is going to help people share designs with experimental motors a lot.

Now we just need to create a motor export function!

Issue #3065, opened 2026-03-23
Open airframe warning in ARC payload example rocket


SiboVG:
PR #3024 refactored the open airframe warning. However, I am confused about the current implementation, because now the ARC payload example rocket gets the warning, even though the airframe looks pretty manifold to me. Tagging @JoePfeiffer.

<img width="1474" height="997" alt="Image" src="https://github.com/user-attachments/assets/d7e34130-9d4b-4646-b18f-24c9281fc998" />

JoePfeiffer:
That shouldn't be happening....

SiboVG:
Closed by #3081

PR #3067, opened 2026-03-24
[#3062] Fix decal edit changes affecting both fin surfaces

### LINKED ISSUE #3062: Decal file change updates both left and right side of fin even if using separate appearance for each
Scenario:

1. You have a fin set with separate appearance for left and right.
2. Different decals are assigned to the left and right appearance
3. Edit one of the decals in an image editor (using the "Edit" button), and save it.
4. In the 3D view, _both sides of the fin get the updated image_, even though that image only should apply to one of the sides
5. This seems to be only a temporary display glitch; the other side of the fin still has the correct decal assignment according to the appearance panel.

SiboVG:
Fixes #3062 (I think). @neilweinstock can you verify? I wasn't able to reproduce the original error.


SiboVG:
FYI: the build is failing due to #3065. You may need to download my branch and build the JAR from there...

neilweinstock:
This fixes the problem on both Windows and Mac.

PR #3068, opened 2026-03-24
[#3063] Disable texture creation button for unsupported components & Support inner tube/tube coupler texture creation

### LINKED ISSUE #3063: Cannot create decal file for inner tube (and UI controls are a bit confused)
I wanted to create a texture file  to use the "Create Texture" button to create a decal file for an inner tube:

1. The button was enabled, so I clicked it.
2. I went through the dialog setting parameters
3. *Then* it said it couldn't create an image file for an inner tube.

Two problems here:

1. For unsupported component types, the "Create Texture" button should be disabled
2. Inner tubes should be supported.  Visually, they behave exactly like body tubes, and are often used for ring fins.

SiboVG:
Fixes #3063. Unsupported texture creation components have their "Create texture" button disabled. Inner tubes and tube couplers now support texture creation. Other ring components, like bulkheads, don't, since they are thin disks and require the top and bottom faces to be texture created instead of the side faces.

### DIFF
--- a/core/src/main/resources/l10n/messages.properties
+++ b/core/src/main/resources/l10n/messages.properties
@@ -1369,6 +1369,7 @@ AppearanceCfg.lbl.Appearance = Appearance
 AppearanceCfg.lbl.Usedefault = Use default
 AppearanceCfg.but.edit = Edit
 AppearanceCfg.but.createTexture = Create texture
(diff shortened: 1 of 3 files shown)


neilweinstock:
This all sounds good.  I had the same thoughts about centering rings and bulkheads.

PR #3092, opened 2026-03-29
Close BufferedWriter in marshalToRASAero

tejasae-afk:
I think marshalToRASAero might be doing a bit more work than it needs to here, but I could be wrong. The resource opened there can leak if marshalToRASAero exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.

### DIFF
--- a/core/src/main/java/info/openrocket/core/file/rasaero/export/RASAeroSaver.java
+++ b/core/src/main/java/info/openrocket/core/file/rasaero/export/RASAeroSaver.java
@@ -1,6 +1,7 @@
 package info.openrocket.core.file.rasaero.export;
 
 import info.openrocket.core.document.OpenRocketDocument;
+import info.openrocket.core.util.BugException;
 import info.openrocket.core.document.StorageOptions;
 import info.openrocket.core.file.RocketSaver;
 import info.openrocket.core.logging.ErrorSet;
@@ -69,9 +70,13 @@ public class RASAeroSaver extends RocketSaver {
             ErrorSet errors) throws IOException {
         log.info("Saving .CDX1 file");
 
-        BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(dest, StandardCharsets.UTF_8));
-        writer.write(marshalToRASAero(doc, warnings, errors));
-        writer.flush();
+        try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(dest, StandardCharsets.UTF_8))) {
+            writer.write(marshalToRASAero(doc, warnings, errors));
+            writer.flush();
+        } catch (IOException e) {
+            log.warn("Failed to write RASAero export: " + e.getMessage());
+            throw new BugException(e);
+        }
     }
 
     @Override


core/src/main/java/info/openrocket/core/file/rasaero/export/RASAeroSaver.java
```diff
@@ -69,9 +69,10 @@ public void save(OutputStream dest, OpenRocketDocument doc, StorageOptions optio
             ErrorSet errors) throws IOException {
         log.info("Saving .CDX1 file");
 
-        BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(dest, StandardCharsets.UTF_8));
-        writer.write(marshalToRASAero(doc, warnings, errors));
-        writer.flush();
+        try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(dest, StandardCharsets.UTF_8))) {
+            writer.write(marshalToRASAero(doc, warnings, errors));
+            writer.flush();
+        }
```
SiboVG, line 79: We should always include a `catch` statement, even if it's simply an ignore catch.

SiboVG (approved):

SiboVG:
Thank you for the contribution, but I don't fully understand why this change is needed. You say it eliminates a memory leak if the RASAero export failed due to an exception?

tejasae-afk:
Good question — here's the concrete scenario:

`marshalToRASAero()` can throw (e.g. a `JAXBException` wrapped in a `RuntimeException`). In the original code, if that happens, `writer.close()` is never called, leaving the `BufferedWriter` → `OutputStreamWriter` → `dest` `OutputStream` open. Whether that causes a real problem depends on what the caller passes as `dest`, but it means any buffered data won't be flushed and the caller can't detect a partial write cleanly.

Totally fair point on the `catch` block — happy to add one (even a simple rethrow or ignore catch to keep the pattern consistent with the rest of the codebase). Let me know if that makes sense and I'll update the PR, otherwise happy to close this here.

SiboVG:
> Good question — here's the concrete scenario:
> 
> `marshalToRASAero()` can throw (e.g. a `JAXBException` wrapped in a `RuntimeException`). In the original code, if that happens, `writer.close()` is never called, leaving the `BufferedWriter` → `OutputStreamWriter` → `dest` `OutputStream` open. Whether that causes a real problem depends on what the caller passes as `dest`, but it means any buffered data won't be flushed and the caller can't detect a partial write cleanly.


Right. You changed it to a try with resource.

> Totally fair point on the `catch` block — happy to add one (even a simple rethrow or ignore catch to keep the pattern consistent with the rest of the codebase). Let me know if that makes sense and I'll update the PR, otherwise happy to close this here.

Let's add a log warning and a BugException rethrow in the catch statement.

tejasae-afk:
Updated — added a `catch (IOException e)` block with a log warning and a `BugException` rethrow. Thanks!

SiboVG:
Thanks!

Issue #3089, opened 2026-03-29
Unit tests sometimes fail due to large AoA warning in example design


SiboVG:
Our unit tests sometimes fail due to a large AoA warning in the "Two stage high power rocket.ork" example design file:

```
> Task :core:test
    Test /home/runner/work/openrocket/openrocket/core/src/main/resources/datafiles/examples/Simulation scripting.ork PASSED (6.1s)
    Test /home/runner/work/openrocket/openrocket/core/src/main/resources/datafiles/examples/Three stage low power rocket.ork PASSED (7.9s)
    Test /home/runner/work/openrocket/openrocket/core/src/main/resources/datafiles/examples/Tube fin rocket.ork PASSED (1.3s)
    Test /home/runner/work/openrocket/openrocket/core/src/main/resources/datafiles/examples/Two stage high power rocket.ork FAILED

    org.opentest4j.AssertionFailedError: Warnings when simulating /home/runner/work/openrocket/openrocket/core/src/main/resources/datafiles/examples/Two stage high power rocket.ork (Simulation 1) (expected={informative=1, normal=0, critical=0}, actual={informative=1, normal=1, critical=0}):
    - NORMAL: Large angle of attack encountered (0°)
    - LOW: Open forward airframe (diameter > 0):  "Interstage"
     ==> expected: <{informative=1, normal=0, critical=0}> but was: <{informative=1, normal=1, critical=0}>
        at app/org.junit.jupiter.api@6.0.1/org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:158)
        at app/org.junit.jupiter.api@6.0.1/org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:139)
        at app/org.junit.jupiter.api@6.0.1/org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:201)
        at app/org.junit.jupiter.api@6.0.1/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:191)
        at app/org.junit.jupiter.api@6.0.1/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1211)
        at app/info.openrocket.core@26.xx.SNAPSHOT/info.openrocket.core.file.openrocket.ExampleFilesTest.openAndSimulateExampleFileHasNoWarnings(ExampleFilesTest.java:281)
```

JoePfeiffer:
[...]

PR #3093, opened 2026-03-29
Instead of computing stall margin in stability calculators, just make stall angle available to BasicEventSimulationEngine

JoePfeiffer:
Determining the margin in the stability calculators ends up calculating it on every update of the RK stepper parameters, so the large angle of attack test is made on the basis of an AOA that may not have actually occurred. Worse, the AOA reported in the warning is based on the last actual AOA recorded, so the reported AOA will typically not quite match any AOA in the data, and in some cases will be radically different (the cases that led to finding this were reporting 0 as a "large AOA"!).
    
By just making the stall angle available, we can test for large AOA using the angle that is actually recorded.

(note that the bug this addresses was exposed by #3089, but is not the actual bug from that report. PR for that one is incoming)


SiboVG:
Should we expose a stall margin calculation somewhere else, after the RK4 step? It could even be useful to expose this as a flight data type, no?

JoePfeiffer:
I'm not sure how useful it would be -- we could certainly do it. The place would be in BasicEventSimulationEngine when it's calculated.

PR #3096, opened 2026-03-29
Update icons with high-res version

### LINKED ISSUE #3022: Update app icon for MacOS Tahoe
Placeholder so we don't forget to do this before release.

When the app is not running, the current icon is placed in "squircle jail" on Tahoe, on an ugly gray background.  Once the app is launched the icon is replaced with the existing icon in its correct form.  We should do something to clean this up.

SiboVG:
Updates the app icons with the new improve high-res version, and adds a new squircle icon for macOS, to match macOS Big Sur and up icon styles. Fixes #3022.

This example does not show the macOS icon, but it shows the difference between the old logo, and the new crisp logo.
Screenshot 2026-03-30 at 01 50 47


neilweinstock:
When the app on Mac is running, it still shows the non-squircle icon (as in your screenshot above).  Is this intentional?  I'm not sure if running it from the app package will behave different from running it from the command line as I am doing now, but I suspect it will be the same based on past experience.

SiboVG:
> When the app on Mac is running, it still shows the non-squircle icon (as in your screenshot above). Is this intentional? I'm not sure if running it from the app package will behave different from running it from the command line as I am doing now, but I suspect it will be the same based on past experience.

It should be a squircle, but I haven't tested this yet. In the screnshot, it's not because I'm running the app straight from the source code.

PR #3119, opened 2026-04-13
Fix pitch/yaw rate discontinuity in local orientation rates

### LINKED ISSUE #2987: [Bug] Local Orientation Rates Show Mirroring Discontinuity
### Describe the bug

The pitch and yaw rates display an odd discontinuity, where it appears every quarter-cycle or so the pitch rate is exactly inverted, and the yaw rate reaches a vertical asymptote (Fig 1). 

Compiling this data into the online graphing calculator Desmos, I was able to find that taking the derivative of the Vertical Orientation parameter from OpenRocket yields a continuous function which nearly exactly matches what the pitch rate would be if it were not partially inverted. You can view a screenshot of the graph at (Fig 2), or go to this Desmos share link for more detail: https://www.desmos.com/calculator/rd9kj0lgli

This issue is persistent through a wide variety of rocket designs and simulation parameters, and is even evident in the example files provided with OpenRocket. 

I have found what I believe to be a solution to the issue. In the AbstractSimulationStepper file, line 92 is `store.thetaRotation.invRotateZInPlace(rot);`. However on line 91, `rot` was already acted upon by `status.getRocketOrientationQuaternion().invRotateInPlace(rot);`. By removing line 92, the issue goes away entirely (Fig 3). 
I do not understand for what mathematical reason Line 92 is present, given that the rotational velocity is already rotated into the local frame by the quaternion on Line 91. Removing it appears to fix the issue without introducing any others.

### To Reproduce

1. Open the "A simple model rocket" example file
2. Run Simulation 1
3. Plot the pitch rate of Simulation 1 and observe the discontinuity

#### Screenshot(s):

Fig 1 (the issue):
Image

Fig 2  _(pitch rate output is green, the derivative of vertical orientation is orange)_ :
Image

Fig 3 (the fixed pitch rate):
Image

#### .ork file:
Just use the "A simple model rocket" example file!

### OpenRocket version

24.12 and unstable

Windows

### Additional context

_No response_

tejasae-afk:
Fixes #2987.

The rotation velocity vector is rotated from inertial frame to body frame on line 98 via the rocket orientation quaternion. Line 99 then applies an additional `thetaRotation.invRotateZInPlace` on top of that, rotating the already-local vector into a theta-aligned frame. This extra step is what produces the mirroring discontinuity in the pitch and yaw rate plots.

Removing line 99 leaves the vector in the body frame, which is where pitch and yaw rates should be expressed. Comparing the output against the derivative of the vertical orientation parameter confirms this matches the expected continuous function.

The reporter tested this fix against a wide range of designs and found no regressions. I verified the same against the "A simple model rocket" example.

You know the codebase far better than I do, so happy to revisit if there's a reason thetaRotation was applied here.


senft-research:
Cant help but notice the failed tests in this PR. 

Have you investigated why those tests failed at all?

I know you noticed removing that line fixed the problem you identified, but I think checking the potential side effects of its removal would be wise, especially in the face of tests failing due to its removal.

Typically, if two unit tests end up failing due to the removal of a single line of code, that tends to be a red flag worth investigating.

tejasae-afk:
Makes sense! reviewing it again.

JoePfeiffer:
It's actually only one check failing, simulating an example rocket that has no business ever working (it's the "pods with recovery" rocket; the sustainer is unstable -- working on a replacement for it).

My question goes off in a different direction: should the rotation be in both the status and the datastore? Or would removing from one place or the other, and then pulling it from the one that still has it, be a more fundamental fix?

senft-research:
Oh? I thought I saw on the actions 2 tests? My bad if I misread. 

And I agree that is ultimately the fundemental question.

I think it goes deeper than that and more asking the elephant in the room question of "is the sim logic as a whole playing a bit too loose with the ability to alter key values in general".

I.e. is the theta issue here a symptom of how brittle the current way we toss the data around is done haha

JoePfeiffer:
> Oh? I thought I saw on the actions 2 tests? My bad if I misread.

It says two, but I believe that's a misreporting. The "pods -- powered with recovery deployment" example is aborting because it's tumbling under thrust; that's being counted both as a failure and also as a failure of "testSimulateAllExampleFiles()"

> And I agree that is ultimately the fundemental question.
> 
> I think it goes deeper than that and more asking the elephant in the room question of "is the sim logic as a whole playing a bit too loose with the ability to alter key values in general".
> 
> I.e. is the theta issue here a symptom of how brittle the current way we toss the data around is done haha

For the moment let's see where this leads us.

tejasae-afk:
The two uses of thetaRotation are operating on different quantities. The aerodynamic forces (CN, Cside) are computed in the theta-aligned frame and need rotateZInPlace to get back to body frame before the quaternion step — that's correct. But getRocketRotationVelocity() is a world-frame vector that's already in body frame after the quaternion invRotate, so the extra invRotateZInPlace was rotating something that was never in theta-space to begin with, which is where the discontinuity came from.

I don't think it's a status-vs-datastore issue — the asymmetry is correct because the quantities have different origins. But happy to be wrong if there's history here I'm missing.

JoePfeiffer:
I've replaced the failing example rocket. Please merge current unstable into your branch; you should stop having the failure.

tejasae-afk:
Thanks @JoePfeiffer — merged unstable in just now.

JoePfeiffer:
Thank you!

Issue #3120, opened 2026-04-15
[Bug] Motor database issues


JoePfeiffer:
### Describe the bug

Our new handling of the motor database is exposing some issues in the thrustcurve database; while the ones I've found may not technically be our bugs we'll be getting bug reports about them; they also make configuring rockets with these motors very difficult. I don't know whether prior devs deliberately inserted heuristics to work around the issues or if they got lucky, but here we are...

### To Reproduce

1. Open the Simple Model Rocket example
2. Go to Motors & Configuration tab
3. Select any configuration and click "select motor"
4. Select "Use common name"
5. Search for Estes B6 motor.
6. Observe there are two motors in the list named B6-0. The name should not include the delay; also, it should only appear once in the list.
7. Select "Use manufacturer's designation"
8. Observe there is now a B6 and a B6-0.
9. Selecting B6 gives an ejection charge delay of 0, 2, 4, 6, or plugged (this is correct)
10. Selecting B6-0 only gives an ejection charge delay of plugged (0 is not an option)
11. Search for C6
12. Select Estes C6
13. Observe the only ejection charge delay is plugged...

#### Screenshot(s):
*(drag-and-drop the screenshot(s) here)*

#### .ork file:
*(drag-and-drop the file here as a .zip file)*

### OpenRocket version

current unstable

### What platform are you running on?

Linux

### Additional context

The two motors I've found issues with have issues in the database. In the case of the B6, the motor name in the RASP file for the motor gives the motor name as B6-0, which is incorrect. In the case of the C6, the motor delay in one of the RASP files is P (plugged), which is also incorrect. I have sent John an email letting him know about these.

PR #3122, opened 2026-04-16
New "powered pods with recovery deployment" example

JoePfeiffer:
The existing "powered pods" example is problematic for several reasons:

1. It's overly complex. In addition to the powered pods it is intended to exemplify, it is also a two-stage rocket.
2. The booster has off-center throust, which OR does not model.
3. (most importantly) the sustainer is unstable. I'm not quite sure how it ever succeeds in not getting an abort because of tumbling under thrust.

This replacement example mitigates these concerns.


neilweinstock:
I'm showing this when I load it up, which I'm going to speculate is not what you intended although it is kind of cool:
image

Also, I get warnings on load-up:
image

I don't understand the thrust curve warnings.  They are a pitfall of the practice of saving the thrustcurves with the rocket.  But I don't understand why they don't match.  In either case, I would suggest saving the example out of 24.12, so the thrustcurves are not included.  I don't know if there's an option to *not* save them with the file (I didn't see anything obvious in preferences).

ADDENDUM: I do not get the thrust curve warnings when I open with the latest snapshot, so the second issue above is actually a non-issue.

JoePfeiffer:
That actually was the intent -- I supposed that means I should add some markings?

On the thrustcurve warnings, there are some issues there (see #3120 ). Saving with 24.12 a good idea for now.

neilweinstock:
You meant the tiny center tube?  Now that I look more closely it makes sense... seems a bit off-kilter for an example rocket though.

JoePfeiffer:
> You meant the tiny center tube? Now that I look more closely it makes sense... seems a bit off-kilter for an example rocket though.

I thought you meant the all-black! Yes, the short little center tube and long outer tubes are definitely intended. An earlier version had a tailcone  instead of a motor there, but that was just *too* weird.

Any time you're putting the streamers in the side pods you're already pretty far off-kilter...

PR #3124, opened 2026-04-18
Reduce profiler hotspots in simulation status, component transforms, and structure mass calculation

SiboVG:
This PR is an attempt to improve performance, based on a JProfiler scan of a minute of usage in OR (opening a design, running a few sims, plotting...).

The scan showed three main CPU hotspots in the current simulation flow:

1. `SwingPreferences.keyExists` in the simulation table render path (~3.0s CPU)
`Simulation.getStatus()` ends up cloning `FlightConfiguration`, which reads preferences and repeatedly hits `AbstractPreferences.keys()`.
2. `Transformation.getRotationTransform` / `RocketComponent.getComponentLocations` (~2.8s CPU)
Component locations were recomputed recursively for repeated mass and aerodynamic queries, especially for ring components.
3. `MassCalculation.calculateStructure` on every RK4 step (~166ms CPU)
     Structural mass was recalculated even when the flight configuration had not changed.

This PR addresses those hotspots with three targeted caches:

- cache preference node/key lookups so the simulation status path no longer repeatedly reads preference data from the backing store
- cache component locations and angles in RocketComponent, with invalidation on relevant component changes, so repeated transform-chain traversal is avoided
- cache structural mass between simulation steps and only recompute when the flight configuration mod ID changes

### DIFF
--- a/core/src/main/java/info/openrocket/core/rocketcomponent/LaunchLug.java
+++ b/core/src/main/java/info/openrocket/core/rocketcomponent/LaunchLug.java
@@ -213,6 +213,7 @@ public class LaunchLug extends Tube implements AnglePositionable, BoxBounded, Li
 		}
 		
 		this.radialOffset = parentRadius + radius;
+		clearCoordinateCaches();
 	}
 	
 	@Override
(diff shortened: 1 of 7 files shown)


neilweinstock:
These sound like good changes... have you observed a performance improvement?

SiboVG:
> These sound like good changes... have you observed a performance improvement?

Not really, but my computer is pretty performant, so I generally don't notice delays. A JProfiler rescan did show that the hotspots are gone.

Issue #2670, opened 2025-01-14
Eliminate "high speed deployment" warning in dual-deploy scenario


neilweinstock:
In a dual-deploy configuration (or equivalent like JLCR), a "high speed deployment" warning is inevitable when the main is deployed.  Somehow or other we need to get rid of this, since everything is operating normally.

I do not know the best solution for this, but it would be good to address it somehow.

JoePfeiffer:
It's not inevitable, and a cardboard rocket that sets it off is in some danger of a zipper -- 20 m/s is pretty quick. That said, it should be configurable.

neilweinstock:
Hmm... well, even the dual-deploy example built into OR shows the warning.  

The problem is that a configurable threshold is not really a great solution on its own, because I don't know what value to choose, and I don't want to have to change it for different rockets (again, wouldn't know what value to use).  I think there needs to be some sort of parameter attached to a recovery device and/or some heuristics that together help the app figure out how to treat it.  For instance, a plastic parachute in a BT20 rocket with apogee deploy vs. a 60 main deployed out of a fiberglass rocket while descending on a drogue.  Etc.  I don't know the answer here.

HappyCanuck:
> It's not inevitable, and a cardboard rocket that sets it off is in some danger of a zipper -- 20 m/s is pretty quick. That said, it should be configurable.

Agree with the configurable approach to avoid nuisance warnings. New users could use a default setting, but experienced users should be able to change it. Velocities between 23-30 m/s and even up to 45m/s (~150ft/s) can be acceptable for larger deployment systems.

thzero:
Gonna submit a PR for this because it was bugging me while at LDRS.

neilweinstock:
> Gonna submit a PR for this because it was bugging me while at LDRS.

What is your proposed solution?

thzero:
Configurable min/max values with ability to designate a recovery device as a drogue in the item dialog.

thzero:
https://github.com/openrocket/openrocket/pull/3129

PR #3129, opened 2026-04-20
Eliminate "high speed deployment" warning in dual-deploy scenario #2670

thzero:
# Recovery Device Speed Warnings

Adds support for drogue-aware recovery speed warnings, giving dual-deploy fliers more relevant simulation feedback.  This issue https://github.com/openrocket/openrocket/issues/2670.

## Motivation

The existing high-speed deployment warning fires on every recovery device regardless of role. On a dual-deploy rocket, this means the drogue — which is *intended* to deploy at high speed — constantly triggers a false warning. This change allows a recovery device to be marked as a drogue, enabling warnings that are actually meaningful for dual-deploy configurations.

## Changes

- **Drogue flag** — Parachutes and streamers can now be marked as a drogue chute via a new checkbox in the component config dialog.
- **Dual-deploy warnings** — When a drogue is present, the simulation engine evaluates four targeted warnings:
  - Drogue deploys below minimum speed (e.g. too slow to be effective)
  - Main deploys above maximum speed (high-speed main deployment)
  - Main deploys below minimum speed (main may not fully inflate)
  - A recovery device is marked as drogue but no main parachute exists in the configuration
- **Configurable thresholds** — All four speed thresholds are configurable per-simulation in the simulation options panel, and default values can be set application-wide in the preferences dialog. Previously the single threshold was hardcoded.
- **Backward compatible** — Rockets without a drogue flagged behave exactly as before, except the high-speed deployment threshold is now user-adjustable rather than hardcoded.
- **File persistence** — The drogue flag and threshold values are saved/loaded in the `.ork` file format.


neilweinstock:
I'll be testing this out tonight.  Screenshots showing the affected UI would be helpful.

thzero:
Recovery Item Dialog to mark a recovery item as a drogue or not.
droguesetting

OR Preferences for default speed limits
recoveryspeedwarningdefaults

Simulation Options (think I'm going to change this as its just too much unnecessary white space with this approach)
simulatoroptions

thzero:
Here is revised simulator options.  

image

neilweinstock:
I haven't had a chance to try this out yet, but in the meantime one quick UI comment: please try to get the spacing within your "Recovery Speed Warnings" box the same as the others in the program.  Your padding around the inside of the box (particularly, top and left) is too small.

I have a suspicion that eventually more configurable warnings thresholds are going to be added and they will have to get bumped to a separate tab somehow, but I guess for now it makes sense for these to live in the Simulation Options tab the way you have them.

thzero:
> I haven't had a chance to try this out yet, but in the meantime one quick UI comment: please try to get the spacing within your "Recovery Speed Warnings" box the same as the others in the program. Your padding around the inside of the box (particularly, top and left) is too small.
> 

Will look at that!

thzero:
> I haven't had a chance to try this out yet, but in the meantime one quick UI comment: please try to get the spacing within your "Recovery Speed Warnings" box the same as the others in the program. Your padding around the inside of the box (particularly, top and left) is too small.
> 
> I have a suspicion that eventually more configurable warnings thresholds are going to be added and they will have to get bumped to a separate tab somehow, but I guess for now it makes sense for these to live in the Simulation Options tab the way you have them.

Tweaked the UI to be consistent with padding and also where the editable fields are located.

JoePfeiffer:
This looks like a great improvement to the deployment warnings. One question, is "Recovery speed warning" a drogue high speed warning? It should probably be labelled that more clearly.

neilweinstock:
I have played with it a bit.  A few comments:
image
1. The box should be named "Recovery Warning Thresholds" and then remove the word "warning" from each of the line items.
2. I would name the items as follows, although I'm not necessarily completely happy with this:
a. High Speed Deployment
b. Low Speed Drogue Deployment
c. High Speed Drogue Deployment
d. Low Speed Main Deployment
e. High Speed Main Deployment
3. I'd probably like to see a bit more explanation directly included in the UI, and not just via tooltips.  I'm vaguely uncomfortable that the separation between dual-deploy settings and single-deploy settings is not entirely obvious.

Moving on to the preferences:
image

4. The layout of the thresholds box is still messed up.
5. "Reset to Defaults" should be available for the thresholds as well as the stuff on the left.  I don't know if it needs its own separate button for that or if it should be lumped in with the other stuff.  It does appear as if it's lumped in on the Simulation Settings page above, although the Reset button does not actually seem to do anything there right now.

And finally:
6. What is the intent when someone is not using a traditional drogue, either JLCR, using a streamer as drogue, or going purely drogueless? Apart from whatever is needed to tell the program that such a configuration is in use, they are likely to have different threshold requirements (.e.g, even if I configure a zero-size drogue when simulating a JLCR, there's going to be no such thing as a low-speed drogue warning... which the requires the user to manually tweak the sim settings for that rocket.)  Ultimately it seems like it's going to require a lot of manual fiddling by the user to handle typical situations.  There may be no good way around this.  I do think, though, that JLCR operation should be considered here.

thzero:
Thanks for the comments back.. let me work on some of those items.

> And finally: 6. What is the intent when someone is not using a traditional drogue, either JLCR, using a streamer as drogue, or going purely drogueless? Apart from whatever is needed to tell the program that such a configuration is in use, they are likely to have different threshold requirements (.e.g, even if I configure a zero-size drogue when simulating a JLCR, there's going to be no such thing as a low-speed drogue warning... which the requires the user to manually tweak the sim settings for that rocket.) Ultimately it seems like it's going to require a lot of manual fiddling by the user to handle typical situations. There may be no good way around this. I do think, though, that JLCR operation should be considered here.

I had pondered that too.   I thought of some ways, but I'm thinking about saying on the recovery device a 'drogueless' option and/or on the simulator options too.

I'm not quite sure what you mean by JLCR.  I use both JLCRs as well as traditional altimeters for DD.  You simply set the drogue for apogee, and you set the main parachute(s) for the altitude you expect them to deploy at.   With JLCR its basically same but drogue comes out based on the ejection charge, main is set for your JLCR altitude.

JoePfeiffer:
I still don't understand: what is "High speed deployment" if it isn't one of the other four?

neilweinstock:
> I'm not quite sure what you mean by JLCR. I use both JLCRs as well as traditional altimeters for DD. You simply set the drogue for apogee, and you set the main parachute(s) for the altitude you expect them to deploy at. With JLCR its basically same but drogue comes out based on the ejection charge, main is set for your JLCR altitude.

My apologies, you're correct, although my impression is that JLCRs are commonly used without drogues, so in those instances they would fit into the "drogueless" scenario.

>I still don't understand: what is "High speed deployment" if it isn't one of the other four?

Here's where I get a bit fuzzy on this. The "High Speed Deployment" warning is for single deployment scenarios.  My impression (as an LPR flyer) has always been that the max deployment speed for mains on HPR are somewhat higher than for the typical apogee-deployed LPR parachutes, and that is what calls for a different threshold.  Likewise, there is no "low speed deployment" warning in that situation.

At present there is insufficient explanation in the dialog to make the user understand this.  Even if there were, it's still messy and confusing... but I'm not sure how to fix it.

@thzero, where did you get the threshold values that you put in there as defaults?

Oh, one more issue: I was wondering if the "Drogue" checkbox should be moved higher up (like at the top of the deployment box).  It'd be easy to not notice it where it is now.

thzero:
> My apologies, you're correct, although my impression is that JLCRs are commonly used without drogues, so in those instances they would fit into the "drogueless" scenario.

Well true, I guess I've seen both drogue and drogueless if I stop and think about it

> > I still don't understand: what is "High speed deployment" if it isn't one of the other four?
> 
> Here's where I get a bit fuzzy on this. The "High Speed Deployment" warning is for single deployment scenarios. My impression (as an LPR flyer) has always been that the max deployment speed for mains on HPR are somewhat higher than for the typical apogee-deployed LPR parachutes, and that is what calls for a different threshold. Likewise, there is no "low speed deployment" warning in that situation.

Yup.  I just got annoyed because I was having to explain to someone about it.

> At present there is insufficient explanation in the dialog to make the user understand this. Even if there were, it's still messy and confusing... but I'm not sure how to fix it.

> @thzero, where did you get the threshold values that you put in there as defaults?

High end?  Frankly from what I use - from mentors, HPR rocketry, etc.

Current warning is 20ms, hardcoded.

I normally use 100fps max for drogues.  And 20fps for landing.

neilweinstock:
What about the "low speed" thresholds?  I was not aware that was even a thing.  I thought first deploy was usually set at apogee, so deployment velocity of the drogue could be very low.

thzero:
> What about the "low speed" thresholds? I was not aware that was even a thing. I thought first deploy was usually set at apogee, so deployment velocity of the drogue could be very low.

Yes, unless you screw up and say delay for 5 seconds or something then it could be much higher.  I.e. I often look at it, and say ok my backup charge is going to go off at 2 seconds after apogee, what speed is it at...

thzero:
ok, tried to break things out a bit more into single deploy (i.e. current) state and dual deployment (drogue and drogueless)....

single
dual

JoePfeiffer:
It's true that main deployment speed for high power is nearly always higher than deployment speed for low power. All the same, I think it's less confusing to treat the low power single deployment as the main deployment for consistency.

I've also never really heard of a deployment speed being too low, but at least for IREC we have a minimum speed at main deployment as we'll have occasional teams come through who are afraid enough of their recovery system that they want to bring the drogue down slower than the competition wants for the main. So I like the minimums for the selfish reason that it'll make reviewing them easier for me!

The idea of having a high speed landing warning has come up periodically -- would this UI be a good place to configure it?

thzero:
Well I removed that anyway, it was too hard to explain and even I don't know that I really was using it.

For normal usage, it works as it does today - bonus is that user can adjust instead of having a 20m/s.
Dual deployment - if you have a drogue recovery device then there is the high/low main warnings.   High is obviously your 'hey this is far too fast than I'm expecting', and low is really for playing with drogue chutes and setting a lower threshold that you don't want to go below... far easier than looking at graphs.

thzero:
> The idea of having a high speed landing warning has come up periodically -- would this UI be a good place to configure it?

Maybe?  You mean that if say your rocket lands at say 30fps, its probably needs a warning to be flagged?

neilweinstock:
I'm still not clear about the low speed drogue warning, but if it's to be included then the default threshold should be zero, so those deploying at apogee won't have the warning all the time unless they go in and explicitly set it.

image
Second one should be "Drogue High Speed warning".  Actually, looking at your new setup, I would word them as "Drogue low speed deployment warning" and "Drogue high speed deployment warning", and make the analogous changes for the other settings.  Wording should be consistent.

I also feel like the word "thresholds" should be in there somewhere, because that's what those numbers are.  You could add that word to the end of each setting, which is redundant but maybe OK (e.g.: Drogue High Speed Deployment Warning Threshold"). 

In the future there are likely to be more of these thresholds; in that case this all will likely need to be rethought and redone.  For example, these fall under the general category of "Warning thresholds", but I don't think it's worth putting another box wrapper around it all as long as all we're doing are the recovery thresholds.  Once we add anything else, though (including "high speed landing") then this probably needs to change.  It might even be best put into its own subtab under Simulation options, then we can load it up and organize it as needed.

thzero:
Ok, latest commit.

- One of things I noticed was that the warnings weren't stage specific
  - While I'm not much of a dual stage guy (yet) I wanted to cover this ground also.

After a bunch of configuration options, and not trying to add/modify too much, I've ended up with the following Recovery Warning System.

- Note I explicitly did not want to add any fake recovery devices to act as a 'droguless' configuration; it just felt counter intuitive and just wrong.

-----

# Recovery Warning System — Per-Stage Dual/Single/Drogueless Detection

## Overview

Recovery deployment speed warnings are evaluated **per stage**, independently, using the rocket's component configuration.

---

## Warning Paths

Each stage gets one of three warning paths when a recovery device deploys:

| Stage condition | Warning(s) fired |
|---|---|
| Has a drogue device (auto-detected) | `HighSpeedMainDeployment` + `LowSpeedMainDeployment` for the main chute |
| Marked as drogueless | `HighSpeedMainDeployment` only |
| Neither (no drogue, not drogueless) | `RecoveryHighSpeedDeployment` (current single deploy) |

**Drogueless** covers the case where a rocket freefalls or tumbles from apogee down to a lower deployment altitude, then deploys the main directly — no drogue chute is used to slow the descent. Because the rocket is in uncontrolled freefall until main deployment, a low-speed main warning doesn't apply (the main deploying too slowly is not a concern in this mode).

---

## How Detection Works

At every recovery device deployment event, the engine:

1. Checks if the **deploying stage** is flagged `isDrogueless()` → drogueless path
2. Otherwise, scans **active components in that stage only** for a `RecoveryDevice` with `isDrogue() == true` and a non-`NEVER` deploy event → dual-deployment path
3. If neither → single-deployment path

The scan is scoped to the deploying stage so a drogue in stage 1 of a multi-stage rocket does not affect stage 2's warning thresholds.

---

## Configuration

### Per-device: "Is Drogue" checkbox (`ParachuteConfig`, `StreamerConfig`)

Marks a recovery device as the drogue for its stage. Blocked if:
- The stage is already marked drogueless
- Another device in the same stage is already marked as a drogue

### Per-stage: "Drogueless main deployment" checkbox (`AxialStageConfig` → Recovery tab)

Marks a stage as using main-only deployment (no drogue). Blocked if:
- Any recovery device in the stage is marked as a drogue

The Recovery tab also shows a status line — either the name of the detected drogue device or "No drogue device found in this stage."

---

## Simulation Options Panel

The dual deployment tab shows two always-visible threshold fields (no radio buttons or conditional sections):

- **Main low-speed warning** — fires `LowSpeedMainDeployment` if main deploys below this speed in a dual-deployment stage
- **Main high-speed warning** — fires `HighSpeedMainDeployment` if main deploys above this speed in a dual-deployment or drogueless stage

---

## Warning Thresholds

Configured in simulation options (defaults):

| Threshold | Default | Used for |
|---|---|---|
| `RecoverySpeedWarning` | 20 m/s | Single-deployment high-speed deploy |
| `RecoveryDrogueMainHighSpeedWarning` | 30.48 m/s (100 fps) | Main deploy too fast (dual or drogueless) |
| `RecoveryDrogueMainLowSpeedWarning` | 15.24 m/s (50 fps) | Main deploy too slow (dual only) |

neilweinstock:
Your latest description sounds better, sorry I'm a bit slow to get around to testing this.  Will try to do so this weekend.

I'm a bit concerned about the potentially conflicting "drogueless" and "this is a drogue" settings.  I can't think of any other instance where component settings have been dependent on each other in this way (although I suppose the various "automatic" checkboxes are *kind of* like that).  But I'll see how it looks when I try it out.

Thanks for sticking with this one, it's tricky but should be valuable when finished.

thzero:
No problem.   :) The other thing  I was toying with is moving the 'is drogue' to Stage just like drogueless, but give them the option of selection a recovery device in a dropdown.   Then that can just show in the recovery device, but that where there is one place to go to manage it per stage.

JoePfeiffer:
Finally done with stuff from IREC so I can go back to being productive around here....

This looks really nice! But I do have some comments
1. The "Single deployment" "Dual deployment" tabs in the simulation options need some sort of title to identify them as deployment speed warnings. Maybe put them in a frame with a title?
2. What exactly does drogueless main deployment do? Does it tumble when the first deployment event occurs? Also, I'm not seeing where in the Recovery tab there is an option for "Drogueless main deployment"
3. For dual deployment, I'm only seeing a recovery minimum speed warning for the drogue. It seems like there should also be a maximum, and also that the default minimum should be 0.
4. I'd like to see a main recovery minimum speed warning as well.

thzero:
> Finally done with stuff from IREC so I can go back to being productive around here....
> 
> This looks really nice! But I do have some comments
> 
> 1. The "Single deployment" "Dual deployment" tabs in the simulation options need some sort of title to identify them as deployment speed warnings. Maybe put them in a frame with a title?

Good point.

> 2. What exactly does drogueless main deployment do? Does it tumble when the first deployment event occurs? Also, I'm not seeing where in the Recovery tab there is an option for "Drogueless main deployment"

Drogueless doesn't do anything different in the sim than it does today, this is just about the warnings.  It just doesn't expect there to be a drogue parachute.

What would be "Drogueless main deployment"?

> 3. For dual deployment, I'm only seeing a recovery minimum speed warning for the drogue. It seems like there should also be a maximum, and also that the default minimum should be 0.

I think there was some discussion on this above, but can do this too and set it to zero.

> 4. I'd like to see a main recovery minimum speed warning as well.

Sounds good. :)

JoePfeiffer:
> Drogueless doesn't do anything different in the sim than it does today, this is just about the warnings. It just doesn't expect there to be a drogue parachute.
> 
> What would be "Drogueless main deployment"?

I guess that's my actual question....in the second bullet point in the Dual Deployment pane, it says "To use drogueless main deployment: open the stage config (Recovery tab) and check Drogueless main deployment." So I'm not having any luck finding that option and I don't know what it would do if I did find it.

thzero:
> > Drogueless doesn't do anything different in the sim than it does today, this is just about the warnings. It just doesn't expect there to be a drogue parachute.
> > What would be "Drogueless main deployment"?
> 
> I guess that's my actual question....in the second bullet point in the Dual Deployment pane, it says "To use drogueless main deployment: open the stage config (Recovery tab) and check Drogueless main deployment." So I'm not having any luck finding that option and I don't know what it would do if I did find it.

Interesting.
My approach was to take the take of a) keeping each stage separate for multi-stage warnings, and b) decorating recovery devices as a drogue or not (within the stage).   The issue with drogueless is without a recovery device there is no where to decorate it.  Thus I eventually was like, can I just add that to the stage edit dialog and be able to set a stage as drogueless instead.

image


I've been debating with myself on this awhile and even going to instead of decorating the drogue on the recovery device, just put this in the stage dialog with a dropdown of the recovery devices to select from -> this would be basically mark the device as a drogue automatically.   I'm almost leaning towards this because it'd mean a) one place to go to set whether your stage is drogueless or has a drogue, and b) a place in the simulator to set the warning thresholds.   I think I'm going to try this out.

thzero:
Played with it a bit on the tab titles and ended up with this...

image
image

And the stage configuration dialog
image

I should say I think this approach feels much better, because its one place to go for each stage and no going between this and that control to edit things.

neilweinstock:
I tend to think this looks better, but what is the pull-down to the right of "Drogue and main deployment:"?  It's not at all clear to me what you're selecting there, or why.

thzero:
> I tend to think this looks better, but what is the pull-down to the right of "Drogue and main deployment:"? It's not at all clear to me what you're selecting there, or why.

Set which one is the drogue chute; need to know which one to queue the warnings off of.

thzero:
This makes it a bit more clear.

image

neilweinstock:
Better.  Let me suggest a simpler alternative, you can judge if this is better:
image

Then have the top option in the pulldown be "[none]" or "[drogueless]". (should use <<>> there instead of square brackets).

"Drogue device" maybe should be indented a bit more.

JoePfeiffer:
I am still unclear on what you mean by "drogueless". To me, it means there is a deployment charge at apogee, which destabilizes the rocket (typically by separating like you would for a drogue, but with no drogue parachute). At present we don't simulate this at all; our advice to users who want to do a drogueless deploy is to make a guess at the drag of the separated components and use a fake drogue chute to emulate it.

So when I see you've got an option for "drogueless" in dual deploy, what does that mean?

thzero:
image

thzero:
> I am still unclear on what you mean by "drogueless". To me, it means there is a deployment charge at apogee, which destabilizes the rocket (typically by separating like you would for a drogue, but with no drogue parachute). At present we don't simulate this at all; our advice to users who want to do a drogueless deploy is to make a guess at the drag of the separated components and use a fake drogue chute to emulate it.
> 
> So when I see you've got an option for "drogueless" in dual deploy, what does that mean?

It means you aren't using a drogue chute, and frankly it defaults to whatever the sim does today.

> our advice to users who want to do a drogueless deploy is to make a guess at the drag of the separated components and use a fake drogue chute to emulate it.

Dunno, the whole point for me at least was to have print warning out when my main parachute, in drogue/main recovery, opened at too high of speed.  And that it was different than the touchdown speed.   Keeps from having to go look at plots.

Then I think the idea of 'drogueless' came up because there was a guess people might complain at lack of this setting, even though the point OR doesnt really simulate this at all.

So maybe it should just say that...  if you want drogueless, you need to create  fake drogue recovery device because OR does not currently support the use case.

JoePfeiffer:
> > I am still unclear on what you mean by "drogueless". To me, it means there is a deployment charge at apogee, which destabilizes the rocket (typically by separating like you would for a drogue, but with no drogue parachute). At present we don't simulate this at all; our advice to users who want to do a drogueless deploy is to make a guess at the drag of the separated components and use a fake drogue chute to emulate it.
> > So when I see you've got an option for "drogueless" in dual deploy, what does that mean?
> 
> It means you aren't using a drogue chute, and frankly it defaults to whatever the sim does today.

OK, thanks -- that's not what I'd call drogueless, that's just a single deploy with a really late ejection charge (what OR does with it would depend on just what the rocket does at apogee -- rockets that have a nice vertical ascent tend to tumble so it'll tumble down; rockets with more of an arc will come in ballistic).
> 
> Dunno, the whole point for me at least was to have print warning out when my main parachute, in drogue/main recovery, opened at too high of speed. And that it was different than the touchdown speed. Keeps from having to go look at plots.

And don't get me wrong, I definitely want this enhancement!

> Then I think the idea of 'drogueless' came up because there was a guess people might complain at lack of this setting, even though the point OR doesnt really simulate this at all.
> 
> So maybe it should just say that... if you want drogueless, you need to create fake drogue recovery device because OR does not currently support the use case.

We shouldn't be giving options that we don't really simulate. I'd prefer to drop the "drogueless" option.

thzero:
> We shouldn't be giving options that we don't really simulate. I'd prefer to drop the "drogueless" option.

Works for me.. :)  So just Single Deploy (i.e. what it does already), and Dual Deploy (select your drogue) and get the updated warnings.

That work?

JoePfeiffer:
> > We shouldn't be giving options that we don't really simulate. I'd prefer to drop the "drogueless" option.
> 
> Works for me.. :) So just Single Deploy (i.e. what it does already), and Dual Deploy (select your drogue) and get the updated warnings.
> 
> That work?

Sounds good!

neilweinstock:
If we were going to include a drogueless option then I would definitely prefer just putting "< none >" in the first item, to be perfectly analogous to the way the texture selector works in the appearance panel.

The question is: do we really have a way to simulate drogueless?  Well, that would imply a separation at apogee and then some sort of simulation of the descent of the separated rocket without drogue.  We don't do that, and so therefore drogueless is really meaningless for us.  Also it would be necessary to specify separation without having a recovery device associated with it, which we do not have and I can no longer remember if this PR adds such an option.

Which leaves drogueless (including JLCR) users needing to include a "fake" drogue to simulate their flight, which kind of sucks because how are they to know how to size the fake drogue, and it looks hack-ish.

My guess is that the terminal velocity of a separated rocket falls within a modest range for pretty much all rockets... I could imagine us using a fairly simple formula to calculate it based on the rocket's design... if we had some good data to work from.  Which, for the moment, we don't.

So I don't really see any great solution for us right now, and removing the drogueless option is probably the right way to go.  But I think we should work towards having something that looks less hack-ish for drogueless users.

thzero:
Yeah, I agree, I think that this isn't the PR to deal with the drogueless option.. it needs a different PR focusing on it.

thzero:
Here is where we are now.

The general Edit->Preferences

image
image

The Stage configuration dialog
image

The simulation options overrides
image

image

JoePfeiffer:
Looks good to me.

thzero:
Ok, made these changes, few cleanups.  Going to test more after I get back from a long weekend.

JoePfeiffer:
I've held off merging this since you said you were going to do some more tests; is it ready to resolve conflicts and merge?

thzero:
> I've held off merging this since you said you were going to do some more tests; is it ready to resolve conflicts and merge?

I haven't had chance to do some more.  I've been out of town.

neilweinstock:
> > I've held off merging this since you said you were going to do some more tests; is it ready to resolve conflicts and merge?
> 
> I haven't had chance to do some more. I've been out of town.

Do you plan to do more on this?  If so, now's the time if you want it to be included in the upcoming beta.

thzero:
I haven't seen any more issues.  But I've also not gone and do a ton of extra testing.

JoePfeiffer:
I'd actually like to see this one in the upcoming beta; I think it's ready to merge, though some conflicts have arisen since @thzero last updated.

neilweinstock:
I would also like to see it.  If we're satisfied with the current state of functionality, then we can merge as soon as the conflicts are resolved.

thzero:
whew, that took some effort.

neilweinstock:
Nice work @thzero, way to stick with it. :)

SiboVG:
Why did we not move the UI for this to the `Warnings` tab?... There's plenty of room there. The Simulation options tab is really starting to take up a lot of space now, especially because there is all this dead padding.
image

thzero:
> Why did we not move the UI for this to the `Warnings` tab?... There's plenty of room there. The Simulation options tab is really starting to take up a lot of space now, especially because there is all this dead padding.

When I looked, the Warnings tab was just a list of warnings, there is other adjustments, etc.  So at least I didn't think to put it there for that reason.  Thought it was more of simulation options than anything.

Shouldn't be too bad to change, if we find in the beta that we really need or want it there instead.

neilweinstock:
Yeah, I had the same thought, but also Sibo is right that it's gotten a bit stacked in there.  As you say we can fiddle and adjust during the beta as needed.

JoePfeiffer:
Shouldn't we discussing this at #3239? My response will be over there.

PR #3137, opened 2026-04-25
support 3-character color entry in appearance panel

neilweinstock:
This small PR lets you enter colors on the appearance panel by using 3-character hex codes in addition to 6-character codes which were already supported.

3 character codes are automatically converted back to 6-character codes for display purposes.

The question is: should colors that *can* be coded as 3-character be displayed as such?  In other words, should white display as #FFF rather than #FFFFFF?  I'm not sure if this is good or bad.  I think I'd prefer it.  Marking this PR as draft until this is settled.  It'd be easy enough to add the 3 character display if desired.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/util/ColorConversion.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/util/ColorConversion.java
@@ -32,7 +32,17 @@ public class ColorConversion {
 		if (c == null) {
 			return null;
 		}
-		return String.format("#%02x%02x%02x", c.getRed(), c.getGreen(), c.getBlue()).toUpperCase();
+		int r = c.getRed();
+		int g = c.getGreen();
+		int b = c.getBlue();
+		if (((r>>4) == (r&15)) &&
+			((g>>4) == (g&15)) &&
+			((b>>4) == (b&15))) {
+			return String.format("#%01X%01X%01X", r&15, g&15, b&15);
+		}
+		else {
+			return String.format("#%02X%02X%02X", r, g, b);
+		}
 	}
 
 	public static Color brightenColor(Color color, int amount) {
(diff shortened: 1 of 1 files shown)


neilweinstock:
Current implementation:
https://github.com/user-attachments/assets/0bfa4384-94b2-4c69-9ac3-afdd3a163146

Proposed enhancement (maybe):
https://github.com/user-attachments/assets/93b41f9b-1ad4-44bc-8755-44d2dac71607

I cannot get these videos to reliably embed.  So frustrating.

neilweinstock:
I decided that displaying 3-character codes when possible is a good thing and have added that into this PR.

My only remaining question: are there unit tests that need to be updated for this?

JoePfeiffer:
> The question is: should colors that _can_ be coded as 3-character be displayed as such? In other words, should white display as #FFF rather than #FFFFFF? I'm not sure if this is good or bad. I think I'd prefer it. Marking this PR as draft until this is settled. It'd be easy enough to add the 3 character display if desired.

I wonder if a user would be surprised to see something they entered with six digits get displayed with 3.

neilweinstock:
> > The question is: should colors that _can_ be coded as 3-character be displayed as such? In other words, should white display as #FFF rather than #FFFFFF? I'm not sure if this is good or bad. I think I'd prefer it. Marking this PR as draft until this is settled. It'd be easy enough to add the 3 character display if desired.
> 
> I wonder if a user would be surprised to see something they entered with six digits get displayed with 3.

Maybe brief surprise but no alarm I don't think.  No more so than they would get when they enter 3 characters and it displays with 6.  Could go either way, but I personally prefer 3 character display when possible.  I am open to changing it if that's what everyone wants.  I think waiting for beta feedback makes the most sense for such a minor issue.

PR #3143, opened 2026-04-29
[#3142] Increase storage precision of simulation data points

### LINKED ISSUE #3142: [Feature Request] Insufficient Precision of Several Data Elements in the <datapoint> structure in Rocket.ork
Yes.  I've written a few scripts to unzip and then parse a raw Rocket.ork file.  However, it appears that mostl of the data elements in the <datapoint> tuples are stored as "%.3f" ( precision 2 format which is much too low for many attributes.

This precision is much too low for example when trying to back-calculate Rocket Diameter from Reference Area or even for Reference Lengths in my small rockets.

Increase the precision of the <datapoint> elements in Rocket.ork

Interestingly, when I export a specific simulation via [Edit Simulation]->[Export Data], most of the exported values are "%.6f" so the precision is simply not being written to the Rocket.ork file.

**Additional context**
Thank you !

SiboVG:
Fixes #3142 by increasing the decimal precision of simulation <flightdata>, <event>, and <datapoint> values to 6 digits after the decimal point. Do note that this increase the .ork file size. For example, 'A simple model rocket' with 15 simulations (duplicated the existing sims twice) went from 777 kB to 1.4 MB. Still very small compared to today's standards, but just something to note.

### DIFF
--- a/core/src/main/java/info/openrocket/core/file/openrocket/OpenRocketSaver.java
+++ b/core/src/main/java/info/openrocket/core/file/openrocket/OpenRocketSaver.java
@@ -433,25 +433,25 @@ public class OpenRocketSaver extends RocketSaver {
 		if (data != null) {
 			String str = "<flightdata";
 			if (!Double.isNaN(data.getMaxAltitude()))
-				str += " maxaltitude=\"" + TextUtil.doubleToString(data.getMaxAltitude()) + "\"";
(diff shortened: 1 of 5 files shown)


neilweinstock:
I don't know if the request was to increase the precision of *all* data points.  That might be OK but I don't think it's what was being asked for.

[edit: never mind]

PR #3173, opened 2026-06-23
Add triangular leading-edge fin cross sections

MDNich:
## Abstract

This PR adds a triangular leading-edge option for fin cross sections. The new section is based on the existing square fin cross section, with only the leading edge replaced by an isosceles triangular bevel.

## Problem

OpenRocket currently supports square, rounded, and airfoil fin cross sections. There is no way to model a fin that keeps a square/blunt trailing edge while using a sharp triangular leading edge, which is a common physical fin treatment for reducing leading-edge pressure drag while preserving a simple plate-style fin.

This also affects roll calculations: the projected leading-edge wedge area should be included in the fin area seen by roll flow, while the normal-force planform remains the existing fin planform.

## Implementation

Added a new `FinSet.CrossSection.TRIANGULAR` option with an included leading-edge angle parameter. The stored parameter is the full angle between the upper and lower leading-edge surfaces. The UI also lets the user edit the equivalent vertex distance from the original square leading-edge face; angle and distance are two views of the same stored value.

Core behavior:

- Adds triangular leading-edge angle/distance conversion helpers to `FinSet`.
- Treats the triangular section as square-plate based, with leading-edge wedge material removed for mass/volume approximation.
- Includes the triangular leading-edge projection in fin roll forcing and damping calculations.
- Keeps pitch/yaw normal-force planform behavior tied to the original fin planform.
- Keeps triangular trailing-edge/base drag behavior aligned with square fins because only the leading edge changes.

Aerodynamics:

- Subsonic triangular leading-edge pressure drag uses the existing rounded-leading-edge pressure coefficient correlation applied to the leading-edge-normal Mach number.
- Supersonic triangular leading-edge pressure drag uses the attached wedge shock relation via the theta-beta-M equation, falling back to the square-edge stagnation limit if an attached weak-shock solution is not available.
- Transonic behavior is smoothly blended between the subsonic and supersonic treatments.

Swing UI:

- Adds triangular leading-edge controls to trapezoid, elliptical, and freeform fin configuration dialogs.
- Shows the controls only when the triangular cross section is selected.
- Lets the user edit either included angle or vertex distance, one mode at a time.
- Adds a dynamic cross-section preview that preserves the selected included angle.

File format and interoperability:

- Writes `<crosssection>triangular</crosssection>` to `.ork` files.
- Writes optional `<leadingedgeangle>` in degrees for triangular fins.
- Loads `<leadingedgeangle>` back into `FinSet`.
- Documents the new `.ork` field in the developer file specification.
- RockSim export degrades triangular leading edges to square, since RockSim tip shape codes do not include a triangular-leading-edge-only option.
- RASAero export degrades triangular leading edges to square with a warning, since RASAero does not support this cross-section type.

## Testing

Added unit coverage for:

- triangular leading-edge pressure drag in subsonic and supersonic regimes;
- square-like base drag behavior for triangular fins;
- angle/distance conversion;
- increased projected roll damping area from the triangular leading edge;
- `.ork` save/load round trip for triangular cross section and leading-edge angle.

Validated locally with:

```bash
./gradlew core:compileJava swing:compileJava
./gradlew core:test --tests "info.openrocket.core.aerodynamics.FinSetCalcTest"
./gradlew core:test --tests "info.openrocket.core.file.openrocket.OpenRocketSaverTest.testTriangularFinLeadingEdgeAngleRoundTrip" swing:test --tests "*FinSetConfig*"
git diff --check
```

## Reviewer Notes

The triangular leading edge is intentionally modeled as a modification of the square cross section, not as a full symmetric airfoil. This preserves square trailing-edge/base-drag behavior while allowing the leading-edge wedge to affect pressure drag, mass/volume, and roll projected area.

## Contact

Please contact at `nichitiu@mit.edu` for more information.


neilweinstock:
Thank you for submitting this.  We're a bit backed up right now but will try to get to reviewing this as soon as we can.

MDNich:
I had found a small inconsistency - its patched in the second commit.

neilweinstock:
@MDNich do you plan to fix the conflicts on this one?

MDNich:
To be entirely fair I'm not sure how to - it says "too complex to resolve in the web editor" and its the effect of people contributing onto the same core files after I created the pull req - would appreciate some help with the resolution of those conflicts. Thanks

JoePfeiffer:
Hmm, I've got them resolved locally, but I'm getting an error trying to push. Going to try something that ought to override (explicitly pushing to https://github.com/MDNich/openrocket.git HEAD:unstable). I've got a hunch the underlying cause is you built the PR working from your unstable branch instead of creating a new branch for it, though I'm not sure about that.

PR #3178, opened 2026-07-05
Allow mass components to contain internal components

midasavocado:
This is a small change but it allows for mass components to have their own subcomponents. Mostly just for organization of rocket assemblies


neilweinstock:
Screenshots and/or recordings are very helpful when posting PRs such as this.

midasavocado:
> Screenshots and/or recordings are very helpful when posting PRs such as this.

I should have added a screenshot in the PR, sorry. Here is what it looks like:

image

This change was mostly made when I was working on a rocket that carried eggs, and I just thought it would be cleaner if I could put the "Egg" mass component into the "EggCasing" mass component...

neilweinstock:
Thanks.  Not something I've ever wanted but it's reasonable.  Will check it out.

PR #3186, opened 2026-07-22
[#3183] Gate tumble detection on recordWarnings() to skip rod-exit tr…

### LINKED ISSUE #3183: [Bug] Stable rocket randomly aborts (TUMBLE_UNDER_THRUST) at launch rod clearance in wind — silent ~1 m apogee, non-reproducible results
## Summary

A stable rocket (1.9 cal static margin) simulated in windy conditions with the
default 10% turbulence intensity **randomly** aborts with
`SimulationAbort.Cause.TUMBLE_UNDER_THRUST` on the first time step after
launch rod clearance. Identical settings produce a full ~590 m flight on one
run and a ~1 m apogee on the next, depending only on the turbulence seed. The
abort is not surfaced through the simulation warnings, and the tumble decision
is made in an angle-of-attack regime where OpenRocket's own technical
documentation states the aerodynamic model is no longer valid.

This appears to affect real users: see the Rocketry Forum thread
"Open Rocket warning - Stage began to tumble under thrust"
(rocketryforum.com/threads/139880), where users describe the same message
appearing and disappearing between re-runs of an unchanged simulation, with
"turn off wind" circulating as the workaround.

## Expected behavior

- A stable rocket launched in wind passes through a brief high-AOA transient
  at rod exit (AOA is geometrically `atan(wind / rail_exit_velocity)` at that
  instant) and weathercocks — the response documented in hobby literature and
  implied by NAR's guidance that rockets be guided until reaching **4x wind
  velocity**, i.e. a worst-case rod-exit AOA of `atan(1/4) ≈ 14°`.
- If conditions genuinely exceed what the solver can model, the result should
  say so explicitly — not report a ~1 m apogee with an empty warning set.
- Identical settings should not nondeterministically produce 590 m or 1 m.

## Evidence

All runs: attached `tumble_abort_repro.ork` — classic 3FNC cardboard/balsa
rocket, 79 g, 1.9 cal margin, AeroTech E15 (bundled RASAero motor file),
10% turbulence (GUI default), N = 24 random seeds per data point.

**Experiment 1 — abort probability vs. average wind (1.0 m rod):**

| wind (m/s) | deterministic rod-exit AOA | aborts | probability |
|---|---|---|---|
| 8.0  | 16.7° | 1/24  | 4%  |
| 8.5  | 17.7° | 1/24  | 4%  |
| 9.0  | 18.6° | 7/24  | 29% |
| 9.5  | 19.6° | 15/24 | 62% |
| 10.0 | 20.5° | 15/24 | 62% |
| 11.0 | 22.4° | 23/24 | 96% |

The probability rises as the deterministic rod-exit AOA crosses the fixed
17.5° stall angle; turbulence gusts supply the spread. With turbulence set to
0 the same rocket flies a complete 591 m flight at 9 m/s.

**Experiment 2 — abort probability vs. rod length (fixed 11 m/s wind):**

| rod | exit velocity | deterministic exit AOA | aborts | probability |
|---|---|---|---|---|
| 1.0 m | 26.7 m/s | 22.4° | 20/24 | 83% |
| 1.5 m | 31.9 m/s | 19.0° | 9/24  | 38% |
| 2.0 m | 36.3 m/s | 16.9° | 0/24  | 0%  |
| 3.0 m | 45.9 m/s | 13.5° | 0/24  | 0%  |

Holding everything else constant, the abort probability collapses to zero
exactly when the rod-exit AOA drops below the 17.5° stall constant —
confirming the mechanism through an independent variable (and providing the
practical workaround: a longer rail).

**Step data from a successful 8 m/s run** (mechanism in the open):

| t (s) | alt (m) | AOA | CP (m) | CG (m) |
|---|---|---|---|---|
| 0.064 | 0.44 | 24.5° | NaN | 0.416 |
| 0.088 | 0.96 | 17.3° | NaN | 0.416 |
| **0.092** | **1.06** | **16.5°** | **0.418** | **0.416** |
| 0.114 | 1.74 | 12.0° | 0.428 | 0.416 |
| 0.178 | 4.54 | 3.6°  | 0.452 | 0.414 |

On the rod, CP is NaN so the stall check is disarmed. At the first step off
the rod, the body-lift correction has moved the computed CP to within **2 mm
(0.08 cal)** of the CG, despite the rocket's 1.9 cal low-AOA static margin.
As the AOA decays over the next 0.1 s, CP walks back to 1.5 cal. A single
gust pushing the instantaneous AOA past 17.5° inside that window makes
`cg > cp` true for one step — which is all the abort needs.

**Failed-run symptoms:** events end `Launch rod clearance @0.092 →
Simulation abort @0.095`; `maxAltitude ≈ 1 m`; `groundHitVelocity = NaN`;
last AOA sample NaN; `getSimulatedWarnings()` empty.

## Root cause

(sources at `unstable` b02e45b5)

1. `BasicEventSimulationEngine.java:294ff` — the fin-stall check runs on
   **every time step** with no persistence requirement, firing `TUMBLE` when
   `aoa > stallAngle` and `cg > cp` hold for a single step.
2. `BarrowmanStabilityCalculator.java:39` — `STALL_ANGLE` is a fixed 17.5°.
3. `BasicEventSimulationEngine.java:658-665` — in the `TUMBLE` handler,
   `THRUST_TUMBLE_CONDITION = 0.01 N` (line 50) means any burning motor turns
   the event into `abortSimulation(TUMBLE_UNDER_THRUST)` (introduced in
   #2457, which did not consider the rod-exit transient).

**Model-validity note.** The `cg > cp` instability determination at ~18-19°
AOA is made using extensions that the OpenRocket technical documentation
(Niskanen's thesis, included in the repo as `doc/thesis.pdf`) explicitly
scopes as approximations near this regime: the Galejs body-lift correction's
CP position is described as "reminiscent of the CP of a rocket flying at an
angle of attack of 90°" (Sec. 3.2.1, Effect of body lift), and the fin force
equations are stated to be "reasonable up to angles of η ≈ 20°, above which
angle most fins stall, **which limits the validity of the equation**"
(Sec. 3.2.2). A hard, unrecoverable abort is therefore being decided by a
single time step evaluated at the boundary of the model's stated validity.

## Suggested directions (happy to work on whichever is preferred)

1. Require the stall + instability condition to persist (N consecutive steps
   or a ~0.1-0.25 s window) before emitting `TUMBLE`, so single-step gust
   transients at rod exit don't terminate the flight.
2. Alternatively, suppress the tumble check for a short interval after
   `LAUNCHROD`, where the high-AOA transient is geometrically guaranteed.
3. Independently of 1/2: surface `SIM_ABORT` causes through the warning set
   so GUI and API consumers can distinguish an aborted flight from a real
   ~1 m flight.

If one of these directions sounds right to the maintainers, I would like to
work on the fix.

### To Reproduce

1. Open the attached `tumble_abort_repro.ork`.
2. Set average wind 10-11 m/s, turbulence 10% (default), 1.0 m launch rod.
3. Run the simulation repeatedly → most runs abort at ~1 m apogee; occasional
   runs produce the full ~590 m flight. At 9 m/s the outcome is near a coin
   flip. With a 2.0 m rod (all else equal) the abort disappears entirely.

Environment: OpenRocket 24.12 jar and `unstable` sources (b02e45b5), Linux,
Temurin/OpenJDK 17. Also reproduced headless through the core API.

#### .ork file (zipped, GitHub does not accept .ork directly)
[tumble_abort_repro.zip](https://github.com/user-attachments/files/30091389/tumble_abort_repro.zip)

### OpenRocket version

24.12 (official jar); also reproduced against unstable sources (b02e45b5)

Linux

### Additional context

_No response_

sebastianvelace:
Fixes #3183.

## What this does

`recordWarnings()` already suppresses the sibling `LargeAOA` warning during the
0.25 s window after launch rod clearance; this change gates the `TUMBLE` event
emission on the same check. A one-step, gust-driven high-AOA transient at rod
exit — geometrically unavoidable in any crosswind launch — no longer emits a
spurious `TUMBLE`, and therefore no longer produces an unrecoverable
`TUMBLE_UNDER_THRUST` abort while the motor is burning.

This follows the direction suggested in the issue discussion (reuse the existing
post-rod window rather than adding a persistence timer), keeping the handling of
aerodynamic events consistent with the existing warning logic.

## Validation

Repro rocket from the issue (3FNC cardboard/balsa, 1.9 cal margin, bundled
AeroTech E15, 1.0 m rod), 24 random seeds per wind speed, 10% turbulence:

| wind (m/s) | spurious aborts **before** | spurious aborts **after** | apogee of successful flights |
|---|---|---|---|
| 8  | 8%  | **0%** | ~591 m (unchanged) |
| 9  | 38% | **0%** | ~590 m (unchanged) |
| 10 | 75% | **0%** | ~590 m (unchanged) |
| 11 | 88% | **0%** | ~589 m (unchanged) |

Regression check — a genuinely unstable (finless) rocket must still be detected
as tumbling:

| wind (m/s) | tumble/abort detected, before | after |
|---|---|---|
| 2 | 100% | **100%** |
| 5 | 100% | **100%** |

So the spurious aborts are eliminated, the flights that already succeeded are
numerically unchanged, and genuine tumble detection is preserved. The existing
`core` test suite (63 tests) passes.

## A note on scope — a cleaner "surgical" variant worth considering

`recordWarnings()` returns `false` under three conditions: (a) still on the rod,
(b) within 0.25 s of rod clearance, and (c) during descent
(`Z velocity < 20% of max`). This fix relies on **(b)** to suppress the rod-exit
transient — and because the spurious abort always occurs *under thrust*, i.e.
during high-velocity ascent, condition (c) never affects the bug itself.

However, reusing the full `recordWarnings()` also brings condition **(c)** into
the tumble path. For rockets with a recovery device this is a non-issue — the
stall check is already skipped once recovery deploys. But for **no-parachute
tumble-recovery rockets**, condition (c) becomes active from apogee onward
(where `Z velocity ≈ 0 < 20% of max`), which would suppress the `TUMBLE`
transition exactly when such a rocket is meant to detect it on descent.

If preserving descent tumble detection for that class of rocket matters, a more
surgical gate that keys only on the post-rod time window (conditions a + b,
without c) would fix #3183 without touching descent behavior — at the cost of a
small dedicated helper rather than reusing `recordWarnings()` directly. I kept
this PR to the minimal, consistent change per the issue discussion, but I'm
happy to switch to the surgical variant if you'd prefer to protect that case.

### DIFF
--- a/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
+++ b/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
@@ -308,8 +308,14 @@ public class BasicEventSimulationEngine implements SimulationEngine {
 						// If we're stable, put a warning about large AOA
 						// note -- if cp is NaN (which it is while on the rod) cg > cp is false
 						if (cg > cp) {
-							// Not stable, so transition to tumbling
-							currentStatus.addEvent(new FlightEvent(FlightEvent.Type.TUMBLE, currentStatus.getSimulationTime()));
+							// Not stable, so transition to tumbling.
+							// Gated on recordWarnings() so the brief, geometrically unavoidable
+							// high-AOA transient just after launch rod clearance (while weathercocking
+							// into the wind) does not trigger a spurious tumble -- and, under thrust,
+							// an unrecoverable TUMBLE_UNDER_THRUST abort. See issue #3183.
+							if (currentStatus.recordWarnings()) {
+								currentStatus.addEvent(new FlightEvent(FlightEvent.Type.TUMBLE, currentStatus.getSimulationTime()));
+							}
 						} else {
 							// Stable, so warning about AOA
 							if (currentStatus.recordWarnings()) {


JoePfeiffer:
We do want to record tumbling near apogee for those tumble-recovery rockets.

The more I think about it, the more I wonder if using the time-based gate on the tumble event (and possible abort) is even the right approach. Would it be better to note high angle of attack as we do now (since when angle of attack is high the aerodynamic models are invalid), but compare rocket orientation to trajectory and set tumbling if that becomes too high? This would have the effect of avoiding setting tumbling based on the rail departure transient, but would dynamically wait to set tumbling until the rocket was actually tumbling.  Thoughts?

sebastianvelace:
Thanks for taking a look at this, Joe.

To be upfront about where my PR came from: reusing `recordWarnings()` was the quick fix.
It removed the spurious aborts and I could validate it fast, but I was never sure it was
the *right* fix rather than the *smallest* one — which is why I flagged the descent problem
myself instead of quietly shipping it. So I went and tried your suggestion, and two things
came out of it that I think change the picture.

**First: this isn't really about gusts.** I built a deterministic test — wind supplied by a
listener instead of the pink noise model, so there is nothing random anywhere in the
scenario — and a *steady* 6 m/s crosswind, with no gust and no turbulence at all, aborts
every single run on current `unstable`. So the transient isn't the cause; it's just what
made the failure intermittent enough to look like one. The cause is the `cg > cp` half of
the test: it decides the rocket is unstable using a CP that the aerodynamic model produced
at an angle where we have already declared that model invalid. A sustained crosswind pushes
the angle of attack past the stall angle and that comparison does the rest.

**Second: comparing orientation to the ground track doesn't work, and I'd rather show you
why than just assert it.** A stable rocket points itself into the *air*, not along its path
over the *ground*. In a crosswind those are two different directions, so a perfectly healthy
rocket is already "crooked" relative to its ground track — no misbehaviour needed, just
geometry, and it's worst at low speed right after the rocket leaves the guide. Measuring
both angles step by step on an Alpha III at 6 m/s, I found moments where the rocket was
flying fine (angle of attack below the stall angle, model still valid) while the angle to
its ground track read almost 90°, and moments the other way round: 84° angle of attack,
deep in stall, while the ground-track angle read under 30° — i.e. "looks perfectly aligned".
False alarms and missed ones, both. The check that convinced me it wasn't my own bug: with
the wind switched off, the two angles agree exactly. All of the disagreement is wind.

**What does work is your actual idea** — wait and see whether the rocket recovers rather
than judging it on a single instant — just applied to the angle of attack we already
compute. I smooth it over the rocket's own pitch period, which the sim already calculates,
so the waiting time scales with the airframe instead of being a constant tuned to one
rocket. A stable rocket recovers within about a pitch period; a tumbling one never does.
Near-zero airflow is handled by a dynamic-pressure gate that *holds* the filtered value
rather than resetting it, so passing through apogee doesn't erase what was learned on the
way up — which is what keeps the tumble-recovery case you raised working on the way down.

Same deterministic scenario, sweeping a 0.1 s gust across the whole guide-departure window,
13 runs per cell, before → after:

| steady wind | gust | aborts before | aborts after | apogee after |
|---|---|---|---|---|
| 0 m/s | none | 0/13 | 0/13 | 133.1 m |
| 0 m/s | +6 m/s | 2/13 | 0/13 | 132.5 m |
| 2 m/s | +6 m/s | 3/13 | 0/13 | 128.5 m |
| 4 m/s | +6 m/s | 3/13 | 3/13 | 129.2 m |
| 6 m/s | none | **13/13** | **0/13** | 132.0 m |
| 6 m/s | +6 m/s | 13/13 | 4/13 | 130.9 m |

Still-air apogee is 133.1 m, so the recovered flights are normal flights rather than aborts
traded for nonsense trajectories. The rows where aborts remain are ones I checked
individually — the rocket does flip over there, so those should stay.

**A side note that may be worth its own issue.** `SimulationOptions.randomSeed` is
initialised to a random value and the pink noise wind model is seeded from it in the
constructor; `setRandomSeed` updates the field but never re-seeds the model. So turbulent
runs aren't reproducible from the seed, and I only noticed because the same sweep gave me
different numbers on consecutive runs. Anyone validating wind-sensitive behaviour will hit
this. Happy to open a separate issue if it's news to you.

**One thing I'd like your call on before I finish.** Waiting before declaring a tumble opens
a short gap the current instantaneous test doesn't have: the angle of attack is already
high, but tumbling isn't declared yet, and a `LargeAOA` warning can slip out in that gap.
That changes the warning set on two bundled examples and the event ordering in two
`FlightEventsTest` cases. Suppressing the warning during a departure fixes it, but that's a
judgement about what the warning is *for* — "the model is out of envelope on a rocket that
is otherwise flying normally", or a literal record of every excursion? The first means
suppressing it and updating those expectations; the second means emitting it and updating
the example baselines. I'd rather have your preference than pick for you. Everything else
is green, and I can finish it off either way — as a fresh PR or a second commit here,
whichever you'd rather review.

JoePfeiffer:
I didn't mean to suggest comparing against the ground track; you're right, that won't work at all.

My actual suggestion was comparing orientation against the direction of the velocity vector.  But I think your suggestion is actually better -- let's try it!

I would like to see the large AOA warning appear as an INFO level warning (I forget if it already is INFO). I want to know that there is a high enough crosswind to be a bit worrisome, and I've seen rockets on blustery days that would have a successful flight once and then go unstable on their next flight.

Yes, the bug in seeding simulations is news to me! Please open an issue, and then I expect you'll have it fixed immediately!

sebastianvelace:
Superseded by #3190, which takes the approach we settled on: waiting for sustained kinematic evidence rather than gating on the post-rod time window.

Leaving this open or closing it is your call — the diff and description here describe the design you asked me to move away from, so I would close it, but I did not want to do that unilaterally while it still has your review on it.

PR #3187, opened 2026-07-27
[#3148] Ensure sim reset/set to default apply to preferences

### LINKED ISSUE #3148: [Bug] Default launch conditions are not copied to new simulations, even when pressing the "Reset to default" button
### Describe the bug

Under Edit - Preferences, there's a tab called "Launch" where you can specify launch conditions: Wind speed and direction, latitude, longitude and altitude, and launch rod length and angle. In the top, there's a red text reading
> NOTE: the settings below apply only to newly created simulations.
> To change the settings for an existing simulation, select that simulation and edit it.
However, when new simulations are created, these settings are _not_ copied to the "Launch conditions" tab in "Edit simulation" - and pressing the "Reset to default" button doesn't work either.

Version 24.12 on Ubuntu 24.04.4 LTS.

### To Reproduce

1. Make a new rocket
2. Open Edit - Preferences, Launch tab and change default values
3. Build a rocket, define a motor, and go to the Simulations tab.
4. Click the simulation
5. The defined defaults are not copied to the simulation
6. Press the "Reset to default" button.
7. Nothing happens.

#### Screenshot(s):
Image
Image

#### .ork file:
[test.zip](https://github.com/user-attachments/files/27275613/test.zip)

### OpenRocket version

24.12

Linux

### Additional context

_No response_

SiboVG:
Fixes #3148.

### DIFF
--- a/core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
+++ b/core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
@@ -27,6 +27,7 @@ import info.openrocket.core.rocketcomponent.MassObject;
 import info.openrocket.core.rocketcomponent.Rocket;
 import info.openrocket.core.rocketcomponent.RocketComponent;
 import info.openrocket.core.simulation.RK4SimulationStepper;
+import info.openrocket.core.simulation.SimulationOptions;
 import info.openrocket.core.simulation.SimulationOptionsInterface;
 import info.openrocket.core.startup.Application;
 import info.openrocket.core.unit.DegreeUnit;
@@ -454,7 +455,10 @@ public abstract class ApplicationPreferences implements ChangeSource, ORPreferen
 	}
 	
 	public void setLaunchRodAngle(double launchRodAngle) {
-		launchRodAngle = MathUtil.clamp(launchRodAngle, -Math.PI / 6.0, Math.PI / 6.0);
+		// Use the same limit as SimulationOptions, since the launch preferences and a
+		// simulation's launch conditions are edited with the same panel.
+		launchRodAngle = MathUtil.clamp(launchRodAngle, -SimulationOptions.MAX_LAUNCH_ROD_ANGLE,
+				SimulationOptions.MAX_LAUNCH_ROD_ANGLE);
 		if (MathUtil.equals(this.getDouble(LAUNCH_ROD_ANGLE, 0), launchRodAngle))
 			return;
 		this.putDouble(LAUNCH_ROD_ANGLE, launchRodAngle);
@@ -642,7 +646,7 @@ public abstract class ApplicationPreferences implements ChangeSource, ORPreferen
 		if (isa) {
 			setLaunchTemperature(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getTemperature());
 			setLaunchPressure(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getPressure());
-			setLaunchRelativeHumidity(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchRelativeHumidity()).getRelativeHumidity());
+			setLaunchRelativeHumidity(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getRelativeHumidity());
 		}
 
 		fireChangeEvent();
--- a/core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
+++ b/core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
@@ -385,7 +385,7 @@ public abstract class ApplicationPreferences implements ChangeSource, ORPreferen
 	}
 	
 	public final boolean getLaunchIntoWind() {
-		return this.getBoolean(LAUNCH_INTO_WIND, false);
+		return this.getBoolean(LAUNCH_INTO_WIND, true);
 	}
 	
 	public final void setLaunchIntoWind(boolean check) {
@@ -463,7 +463,7 @@ public abstract class ApplicationPreferences implements ChangeSource, ORPreferen
 	
 	
 	public double getLaunchRodDirection() {
-		if (this.getBoolean(LAUNCH_INTO_WIND, false)) {
+		if (getLaunchIntoWind()) {
 			// When launching into wind, sync the launch rod direction with wind direction
 			double windDirection = this.getDouble(WIND_DIRECTION, Math.PI / 2);
 			this.setLaunchRodDirection(windDirection);
--- a/core/src/main/java/info/openrocket/core/simulation/DefaultSimulationOptionFactory.java
+++ b/core/src/main/java/info/openrocket/core/simulation/DefaultSimulationOptionFactory.java
@@ -5,57 +5,34 @@ import info.openrocket.core.preferences.ApplicationPreferences;
 
 import com.google.inject.Inject;
 
+/**
+ * Provides the launch conditions that new simulations start from, and stores the
+ * conditions of an existing simulation as the new defaults.
+ * <p>
+ * Both directions are the same copy between two {@link SimulationOptionsInterface}
+ * instances: {@link #getDefault()} copies the launch preferences into a fresh
+ * {@link SimulationOptions}, {@link #saveDefault(SimulationOptions)} copies the
+ * conditions of a simulation back into the launch preferences. Keeping a single
+ * copy implementation guarantees that both directions cover the same properties.
+ */
 public class DefaultSimulationOptionFactory {
 
-	@Inject
 	private final ApplicationPreferences prefs;
 
-	public static final String SIMCONDITION_WIND_SPEED = "SimConditionWindSpeed";
-	public static final String SIMCONDITION_WIND_STDDEV = "SimConditionWindStdDev";
-	public static final String SIMCONDITION_WIND_TURB = "SimConditionWindTurb";
-	public static final String SIMCONDITION_SITE_LAT = "SimConditionSiteLat";
-	public static final String SIMCONDITION_SITE_LON = "SimConditionSiteLon";
-	public static final String SIMCONDITION_SITE_ALT = "SimConditionSiteAlt";
-	public static final String SIMCONDITION_ATMOS_STD = "SimConditionsAtmosStd";
-	public static final String SIMCONDITION_ATMOS_TEMP = "SimConditionsAtmosTemp";
-	public static final String SIMCONDITION_ATMOS_PRESSURE = "SimConditionsAtmosPres";
-	public static final String SIMCONDITION_ATMOS_RELATIVE_HUMIDITY = "SimConditionsAtmoHumi";
-	public static final String SIMCONDITION_ROD_INTO_WIND = "SimConditionsRodIntoWind";
-	public static final String SIMCONDITION_ROD_LENGTH = "SimConditionsRodLength";
-	public static final String SIMCONDITION_ROD_ANGLE = "SimConditionsRodAngle";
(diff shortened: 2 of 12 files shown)


core/src/test/java/info/openrocket/core/simulation/DefaultSimulationOptionFactoryTest.java
senft-research, line 1: Why are the values to be tested magic numbers? Surely it would be less prone to error if we set the values as constants?

core/src/main/java/info/openrocket/core/simulation/DefaultSimulationOptionFactory.java
senft-research: Why are these static Strings still here if the code that utilizes them has been removed?

core/src/main/java/info/openrocket/core/simulation/DefaultSimulationOptionFactory.java
```diff
@@ -35,46 +36,57 @@ public DefaultSimulationOptionFactory() {
...
-			defaults.setLaunchTemperature(prefs.getDouble(SIMCONDITION_ATMOS_TEMP, defaults.getLaunchTemperature()));
-			defaults.setLaunchPressure(prefs.getDouble(SIMCONDITION_ATMOS_PRESSURE, defaults.getLaunchPressure()));
-			defaults.setLaunchRelativeHumidity(prefs.getDouble(SIMCONDITION_ATMOS_RELATIVE_HUMIDITY, defaults.getLaunchRelativeHumidity()));
-
-			defaults.setLaunchIntoWind(prefs.getBoolean(SIMCONDITION_ROD_INTO_WIND, defaults.getLaunchIntoWind()));
-			defaults.setLaunchRodLength(prefs.getDouble(SIMCONDITION_ROD_LENGTH, defaults.getLaunchRodLength()));
-			defaults.setLaunchRodAngle(prefs.getDouble(SIMCONDITION_ROD_ANGLE, defaults.getLaunchRodAngle()));
-			defaults.setLaunchRodDirection(
-					prefs.getDouble(SIMCONDITION_ROD_DIRECTION, defaults.getLaunchRodDirection()));
+			PinkNoiseWindModel preferredWind = prefs.getAverageWindModel();
```
senft-research: This seems like a bit of a bandaid fix. Whilst we are still in here, would these defaults not benefit from a builder of some sort? We are grabbing a field from inside default, to give it values from the pref's instance only then to set other fields of the defaults via the values of the prefs. Seems very round-about / messy. Surely we could use this as an opportunity to make this cleaner, to reduce the chance of similar bugs to issue 3148 happening again?

core/src/test/java/info/openrocket/core/simulation/DefaultSimulationOptionFactoryTest.java
```diff
@@ -0,0 +1,148 @@
...
+		target.copyConditionsFrom(source);
+
+		assertEquals(1, averageWindEvents.get());
+		assertEquals(1, multiLevelWindEvents.get());
+	}
+
+	/**
+	 * Asserts all launch-condition values managed by the default factory.
+	 */
+	private static void assertConditions(SimulationOptions options, double latitude, double longitude,
```
senft-research: That is a lot of assertions for one test. Considering we would want to quickly know what has failed during a build, would it not be better to break down the tests per Launch Condition?

senft-research (changes requested):
Decent changes, but a few things I would personally take a second pass at. In my opinion this bug initially happened due to the nature of how we actually set the defaults in the factory, and hence a clean up (maybe with a builder?) would be potentially ideal as a refactor. 

Furthermore, the unit tests seem to be capturing a lot of assertions in one go. Would perhaps be a good idea to break down the assertions per property, to make it easier to figure out exactly what went wrong when a build fails.

SiboVG:
Thanks — good catches, I've addressed all of them.

senft-research:
> Thanks — good catches, I've addressed all of them.

Just about to head to bed, will look at em in the morning pal. Have a nice night!

SiboVG:
> > Thanks — good catches, I've addressed all of them.
> 
> Just about to head to bed, will look at em in the morning pal. Have a nice night!

No worries, good night!

core/src/main/java/info/openrocket/core/simulation/DefaultSimulationOptionFactory.java
```diff
@@ -67,26 +44,43 @@ public SimulationOptions getDefault() {
...
+	/**
+	 * Copies all launch conditions shown on the launch conditions panel from one set of
+	 * simulation options to another. Both the launch preferences and a simulation's
+	 * options implement {@link SimulationOptionsInterface}, so this serves loading and
+	 * saving the defaults alike.
+	 *
+	 * @param source      the options to read the launch conditions from
+	 * @param destination the options to write the launch conditions to
+	 */
+	private static void copyLaunchConditions(SimulationOptionsInterface source,
```
senft-research, line 59: Much better, but once again, surely setting this up as a builder would be a lot cleaner? Not gonna die on this hill, but something certainly worth discussing in my opinion.

core/src/main/java/info/openrocket/core/preferences/ApplicationPreferences.java
```diff
@@ -642,7 +646,7 @@ public void setISAAtmosphere(boolean isa) {
 		if (isa) {
 			setLaunchTemperature(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getTemperature());
 			setLaunchPressure(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getPressure());
-			setLaunchRelativeHumidity(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchRelativeHumidity()).getRelativeHumidity());
+			setLaunchRelativeHumidity(ISA_ATMOSPHERIC_MODEL.getConditions(getLaunchAltitude()).getRelativeHumidity());
```
senft-research, line 649: Good spot, but should have probably been a separate commit, as its outside the scope of the commit message. Small thing, but worth capturing in my opinion.

core/src/test/java/info/openrocket/core/simulation/DefaultSimulationOptionFactoryTest.java
```diff
@@ -26,6 +29,41 @@
 class DefaultSimulationOptionFactoryTest {
 	private static final double EPSILON = MathUtil.EPSILON;
 
+	/** Launch conditions stored in the preferences, i.e. the input of {@link DefaultSimulationOptionFactory#getDefault()}. */
```
senft-research, line 32: Would it not have been appropriate to add the units to the field names? Not entirely in love with the idea, but it is a pain to keep having to refer back to the comments at the top for units.

core/src/test/java/info/openrocket/core/simulation/DefaultSimulationOptionFactoryTest.java
```diff
@@ -45,68 +83,249 @@ protected void configure() {
...
+		options.setLaunchRodAngle(SAVED_ROD_ANGLE);
+		options.setLaunchRodDirection(SAVED_ROD_DIRECTION);
+		options.getAverageWindModel().setAverage(SAVED_WIND_AVERAGE);
+		options.getAverageWindModel().setStandardDeviation(SAVED_WIND_STANDARD_DEVIATION);
+		options.getAverageWindModel().setDirection(SAVED_WIND_DIRECTION);
+		return options;
+	}
+
+	@Nested
+	@DisplayName("getDefault() loads the launch preferences")
```
senft-research, line 129: Much better testing layout!

core/src/test/java/info/openrocket/core/simulation/DefaultSimulationOptionFactoryTest.java
```diff
@@ -45,68 +83,249 @@ protected void configure() {
...
+		}
+
+		@Test
+		void loadsISAAtmosphere() {
+			assertEquals(PREFERRED_ISA_ATMOSPHERE, defaults.isISAAtmosphere());
+		}
+
+		@Test
+		void loadsLatitude() {
+			assertEquals(PREFERRED_LATITUDE, defaults.getLaunchLatitude(), EPSILON);
```
senft-research, line 146: I did not mention it last time, but whilst I get EPSILON being useful, is it actually required if the values are not being manipulated during the test? 

Either way no big issue with them staying.

swing/src/test/java/info/openrocket/swing/gui/dialogs/preferences/LaunchPreferencesPanelTest.java
```diff
@@ -51,79 +71,104 @@ public void setUpTest() {
...
+
+    @Nested
+    @DisplayName("New simulations start from the launch preferences")
+    class DefaultsFromLaunchPreferences {
+        private SimulationOptions defaults;
+        // The standard deviation is derived from the turbulence intensity and the average
+        private double windStandardDeviation;
+
+        @BeforeEach
+        public void editLaunchPreferences() {
```
senft-research, line 92: One thing to note,  would it not be cleaner to reset the prefs instance before each test, to avoid conditions from the previous run following over to the next test? I can't imagine it is an issue at the moment, but it is potential for false positives if we add more functionality / tests in the future.

senft-research (commented):
Overall solid changes, with just a few notes here and there to query, not much to action if not desired.

swing/src/test/java/info/openrocket/swing/ServicesForTesting.java
```diff
@@ -163,21 +165,72 @@ public java.util.prefs.Preferences getPreferences() {
...
+
+	/**
+	 * A {@link java.util.prefs.Preferences} implementation that keeps its values in memory.
+	 * <p>
+	 * The tests deliberately do not use {@link java.util.prefs.Preferences#userRoot()}: that
+	 * store is shared with the core test preferences, and the core and swing test JVMs run
+	 * concurrently, so writes of one project can overwrite those of the other. Keeping the
+	 * testing preferences in memory isolates them per JVM and makes clearing them cheap.
+	 */
+	private static class InMemoryPreferences extends AbstractPreferences {
```
senft-research: I will be blunt... what the heck am I looking at here Sibo? I was suggesting clearing the preferences per test. What in the name of coding am I looking at here? 

Why would simply adding an `@AfterEach` method to clear the prefs after each test not have been enough?

SiboVG: Calm down. I'll fix it. Learn to be more objective in your communication style.

senft-research: > Calm down. I'll fix it. Learn to be more objective in your communication style.

Let me be clear, the question was not rhetorical, and I am sorry if it came across that way. I less wanted a promise to "fix it" and more a genuine explanation for what I was looking at, as this addition seemed very out of left field mate. 

Sorry if that came across in the comment, but I very much got a bit of mental whiplash, and was genuinely confused at the time of writing that comment.

To be honest, I still am confused, as you have not explained why this was done, and such an explanation would be very much appreciated.

SiboVG:
After further review, I kept the last 2 commits and added a small follow-up cleanup.

The issue is not limited to preference state leaking between test methods. The core and Swing tests run in parallel JVMs, and both were writing to the OS-backed java.util.prefs store. I reproduced a core test failure while the Swing launch-preferences tests were running concurrently: a saved latitude disappeared and fell back to the default value of 28.61.

With the Swing test preferences stored in memory, the same concurrent run passes all 34 core tests and all 14 Swing tests. An @AfterEach clear would still access the shared OS-backed store and therefore would not prevent this cross-JVM interference. Clearing in @BeforeEach additionally ensures that each test starts with clean launch preferences.

Sorry I did not explain the concurrency context clearly earlier. Thanks for the code review.

PR #3189, opened 2026-07-29
[#3188] Seed the wind models from the simulation's random seed

### LINKED ISSUE #3188: [Bug] Simulation random seed never reaches the wind models, so turbulent flights are not reproducible
## Describe the bug

Two independently constructed but identically configured simulations, given the same
`randomSeed`, do not produce the same flight when wind turbulence is non-zero.

To be precise about the contract, since it is narrower than it first appears: re-running
a single `SimulationOptions` instance *is* already reproducible, because its wind model
keeps whatever seed it was built with. What fails is building the configuration again --
in a test, a script, or a sweep -- setting the same seed, and expecting the same flight.
Measured on `cc9b3643`: repeated runs of one instance agree exactly, while independently
built ones with the same seed differ by up to 1.36 m of apogee.

`SimulationOptions.randomSeed` exists so a simulation can be repeated exactly. It
reaches the integrator — `RK4SimulationStepper:87` and `RK6SimulationStepper:87` both
build their `Random` from `SimulationConditions.getRandomSeed()` — but it never reaches
the wind models, which are the other consumer of randomness in a simulation. The seed
therefore governs half the simulation and not the other half.

The two wind models are broken in different ways:

**`PinkNoiseWindModel`** (the *average* model) receives the seed exactly once, when the
enclosing `SimulationOptions` is constructed:

```java
// SimulationOptions.java:115, in the constructor
averageWindModel = new PinkNoiseWindModel(randomSeed);
```

and its seed field is `final` (`PinkNoiseWindModel.java:52`), so `setRandomSeed` cannot
change it afterwards:

```java
// SimulationOptions.java:552
public void setRandomSeed(int randomSeed) {
    this.randomSeed = randomSeed;   // updates the field, and nothing else
}
```

The model keeps whatever seed it was given at construction, which is the random value
`randomSeed` was initialised to (`SimulationOptions.java:94`). Any seed set later is
silently ignored.

**`MultiLevelPinkNoiseWindModel`** is never connected to the seed at all. Each level is
built with the no-argument constructor:

```java
// MultiLevelPinkNoiseWindModel.java:48
PinkNoiseWindModel pinkNoiseModel = new PinkNoiseWindModel();
```

which delegates to `this(new Random().nextInt())`. Every level draws a fresh random seed
that has no relationship to `SimulationOptions.randomSeed`, in the constructor or
anywhere else.

## To Reproduce

1. Build a simulation with non-zero wind turbulence (e.g. an Estes Alpha III, 1 m rod,
   5 m/s average wind, 20 % turbulence intensity).
2. Call `options.setRandomSeed(12345)`.
3. Run the simulation and record the apogee.
4. Repeat steps 1–3 with the same seed.
5. The apogees differ.

As a JUnit test:

```java
private double apogee(int seed) throws SimulationException {
    Simulation sim = new Simulation(TestRockets.makeEstesAlphaIII());
    sim.setFlightConfigurationId(TestRockets.TEST_FCID_0);
    SimulationOptions options = sim.getOptions();
    options.setISAAtmosphere(true);
    options.setTimeStep(0.05);
    options.setLaunchRodLength(1.0);
    options.setWindSpeedAverage(5.0);
    options.setWindTurbulenceIntensity(0.2);
    options.setRandomSeed(seed);
    sim.simulate();
    return sim.getSimulatedData().getBranch(0).getMaximum(FlightDataType.TYPE_ALTITUDE);
}

// fails on unstable
assertEquals(apogee(12345), apogee(12345), 0.0);
```

Observed on `unstable` at `cc9b3643`, two consecutive runs with seed `12345`:

| wind model | run 1 | run 2 |
|---|---|---|
| average | 1.0520754979 m | 1.0601759644 m |
| multi-level | **129.84 m** | **1.05 m** |

The multi-level row is the clearest symptom: with an identical seed, one run flies
normally and the next ends about a metre off the pad.

Sweeping further on the average model — 24 runs at seed `12345`, and the whole sweep
repeated four times — apogee ranged from **1.05 m to 131.3 m**, and the number of runs
finishing below 100 m came out as **18, 14, 17 and 17 of 24**.

That last figure is the point in miniature: not only is a flight not reproducible from
its seed, the measured failure rate is not reproducible either. Any number quoted from a
single turbulent run, or from a single sweep, is unreliable.

## Screenshots / .ork file

Not applicable — reproducible from the code above, with the bundled `TestRockets`
airframe and no external file.

## OpenRocket version

`unstable` (`cc9b3643`). The `final` seed field and the unseeded multi-level constructor
are long-standing, so earlier releases are expected to be affected as well.

## What platform are you running on?

Linux (platform-independent — this is arithmetic, not I/O).

## Additional context

**Why this survived unnoticed.** `randomSeed` is not written to the `.ork` file — no
saver or loader references it — and the UI does not expose it. The code says as much:

```java
// SimulationOptions.java, in setRandomSeed
/*
 * This does not fire an event since we don't want to invalidate simulation results
 * due to changing the seed value. This needs to be revisited if the user is ever
 * allowed to select the seed value.
 */
```

So today the seed is only reachable from code — tests, scripts, and plugins — which is
exactly the audience the bug affects. A user working through the GUI never encounters it.

**Scope.** This report is about the seed being ignored. Making the seed *persist* across
save/load, and exposing it in the UI, are separate features and are deliberately not
proposed here: even with this bug fixed, reopening a saved `.ork` will still produce a
different flight, because a new seed is generated on load. That is worth its own issue if
wanted, but it needs UI decisions this one does not.

**Impact.** Beyond reproducibility as a feature, this makes wind-sensitive behaviour
untestable and wind-sensitive bug reports unverifiable: any measurement taken from a
single turbulent run is unreliable, and a sweep repeated on the same seed will disagree
with itself. It surfaced while validating #3183, where identical sweeps returned different
abort rates on consecutive runs.

**Proposed fix.** Small and contained — I have it implemented and can open a PR:

- add `void setSeed(int)` to the `WindModel` interface (there are exactly two
  implementers);
- make `PinkNoiseWindModel.seed` non-`final` and implement `setSeed` as assigning the
  seed and calling the existing `reset()`, which discards the lazily built random source
  so it is rebuilt from the new seed;
- implement `setSeed` on `MultiLevelPinkNoiseWindModel` by giving each level a distinct
  derived seed — sharing one seed across levels would make the turbulence identical and
  perfectly correlated at every altitude;
- seed the throwaway clone in `SimulationOptions.toSimulationConditions()`, which already
  clones the wind model and is called once per run.

Seeding the per-run clone rather than the configured models matters: the configured
models' seeds stay untouched, so `equals()`/`hashCode()` and the `OUTDATED` logic in
`Simulation.getStatus()` do not need to change and no stored simulation result is
invalidated.

Total diff is +37/−1 across four files. With it applied, the test above passes exactly
(δ = 0.0) for both wind models, different seeds still give different flights, and the
`core` suite is green (1224 tests, 1221 pre-existing plus 3 new).

sebastianvelace:
Fixes #3188.

## What this does

`SimulationOptions.randomSeed` reached the integrator — `RK4SimulationStepper` and
`RK6SimulationStepper` both build their `Random` from
`SimulationConditions.getRandomSeed()` — but never reached the wind models, so any
flight with turbulence was irreproducible. The two models failed differently:
`PinkNoiseWindModel` was seeded once in the `SimulationOptions` constructor and its
`seed` field was `final`, so `setRandomSeed` could not change it; every level of
`MultiLevelPinkNoiseWindModel` was built with the no-argument constructor and so was
never connected to the seed at all.

The change adds `setSeed` to the `WindModel` interface and applies it in
`toSimulationConditions()`, to the throwaway clone that is already made there and is
built once per run.

**Seeding the clone rather than the configured models is the point of the design.** The
configured models' seeds never change, so `equals()`/`hashCode()` and the `OUTDATED`
logic in `Simulation.getStatus()` need no adjustment and no stored simulation result is
invalidated — which matters, because the existing comment in `setRandomSeed` says
explicitly that changing the seed must not invalidate results.

Levels of the multi-level model are given distinct derived seeds rather than sharing
one: seeded alike they would generate identical turbulence, making the gusts at every
altitude perfectly correlated.

## Validation

Estes Alpha III, 3 m guide, 5 m/s wind, 20 % turbulence, seed `12345`, on `cc9b3643`:

| wind model | run 1 vs run 2, before | after |
|---|---|---|
| average | 131.07 m vs 129.85 m | agree to ~6e-8 m |
| multi-level | 131.27 m vs 130.74 m | agree to ~1e-9 m |

The test asserts to a tolerance of 1e-4 m rather than exactly, because the simulator is
not bit-reproducible even with no randomness in play: on unmodified `unstable`, repeated
runs of this same flight with turbulence set to **zero** still differ by around 5e-9 m.
That floor is pre-existing and unrelated to the seed. The tolerance therefore sits about
four orders of magnitude below the effect being tested and three above the noise it has
to ignore.

The test also asserts that different seeds still give different flights, so that a change
which made the wind deterministic *regardless* of the seed could not pass; and it guards
its own fixture by requiring a real flight, because an aborted flight is reproducible as
readily as a real one and would otherwise let the test keep passing while measuring
nothing.

`core` suite green: 1224 tests, 0 failures. `:swing:compileJava` green. `WindModel` has
exactly two implementers in the repository, both updated here.

## Out of scope

`randomSeed` is not written to the `.ork` file and is not exposed in the UI, so reopening
a saved file will still produce a different flight. Making the seed persistent, and
user-selectable, is a separate feature needing UI decisions this PR does not take — worth
its own issue if wanted.

### DIFF
--- a/core/src/main/java/info/openrocket/core/models/wind/MultiLevelPinkNoiseWindModel.java
+++ b/core/src/main/java/info/openrocket/core/models/wind/MultiLevelPinkNoiseWindModel.java
@@ -174,6 +174,21 @@ public class MultiLevelPinkNoiseWindModel implements WindModel {
 		fireChangeEvent();
 	}
 
+	/**
+	 * Seed each level from the given seed.
+	 * <p>
+	 * The levels are given distinct seeds derived from the one supplied, rather than
+	 * sharing it: levels seeded alike would generate identical turbulence, making the
+	 * gusts at every altitude perfectly correlated. Levels are held sorted by
+	 * altitude, so the derivation is a function of the configuration alone.
+	 */
+	@Override
+	public void setSeed(int seed) {
+		for (int i = 0; i < levels.size(); i++) {
+			levels.get(i).model.setSeed(seed * 31 + i);
+		}
+	}
+
 	@Override
 	public ModID getModID() {
 		return ModID.ZERO; // You might want to create a specific ModID for this model
--- a/core/src/main/java/info/openrocket/core/models/wind/PinkNoiseWindModel.java
+++ b/core/src/main/java/info/openrocket/core/models/wind/PinkNoiseWindModel.java
@@ -49,7 +49,7 @@ public class PinkNoiseWindModel implements WindModel {
 	private double direction = Math.PI / 2; // this is an East wind
 	private double standardDeviation = 0;
 
-	private final int seed;
+	private int seed;
 
 	private PinkNoise randomSource = null;
 	private double time1;
@@ -68,6 +68,12 @@ public class PinkNoiseWindModel implements WindModel {
 
 	public PinkNoiseWindModel() {
 		this(new Random().nextInt());
+	}
+
+	@Override
+	public void setSeed(int seed) {
+		this.seed = seed ^ SEED_RANDOMIZATION;
+		reset();
 	}
 
 	/**
--- a/core/src/main/java/info/openrocket/core/models/wind/WindModel.java
+++ b/core/src/main/java/info/openrocket/core/models/wind/WindModel.java
@@ -32,5 +32,17 @@ public interface WindModel extends Monitorable, Cloneable, ChangeSource {
 	 */
 	CoordinateIF getWindVelocity(double time, double altitude);
 
+	/**
+	 * Set the seed of the model's random source, so that a simulation run with a
+	 * given seed reproduces the same wind.
+	 * <p>
+	 * This discards any random state already generated, so it must only be called
+	 * while configuring a simulation and never from within a running one: reseeding
+	 * mid-flight would change the wind partway through.
+	 *
+	 * @param seed the seed value.
+	 */
+	void setSeed(int seed);
+
 	WindModel clone();
 }
--- a/core/src/main/java/info/openrocket/core/simulation/SimulationOptions.java
+++ b/core/src/main/java/info/openrocket/core/simulation/SimulationOptions.java
@@ -788,7 +788,10 @@ public class SimulationOptions implements ChangeSource, Cloneable, SimulationOpt
 		conditions.setGeodeticComputation(getGeodeticComputation());
 		conditions.setRandomSeed(randomSeed);
 
+		// Seed the throwaway clone rather than the configured model, so that the seed
+		// governs the run without becoming part of the configuration's identity.
 		WindModel windModel = getWindModel().clone();
+		windModel.setSeed(randomSeed);
 		conditions.setWindModel(windModel);
 		conditions.setAtmosphericModel(getAtmosphericModel());
 
--- a/core/src/test/java/info/openrocket/core/simulation/WindModelSeedReproducibilityTest.java
+++ b/core/src/test/java/info/openrocket/core/simulation/WindModelSeedReproducibilityTest.java
@@ -1,0 +1,130 @@
+package info.openrocket.core.simulation;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import info.openrocket.core.document.Simulation;
+import info.openrocket.core.models.wind.MultiLevelPinkNoiseWindModel;
+import info.openrocket.core.models.wind.WindModelType;
+import info.openrocket.core.rocketcomponent.Rocket;
+import info.openrocket.core.simulation.exception.SimulationException;
+import info.openrocket.core.util.BaseTestCase;
+import info.openrocket.core.util.TestRockets;
+import org.junit.jupiter.api.Test;
+
+/**
+ * A simulation run twice with the same random seed must produce the same flight.
+ * <p>
+ * The seed reaches the integrator -- {@code RK4SimulationStepper} builds its
+ * {@code Random} from {@code SimulationConditions.getRandomSeed()} -- but it did
+ * not reach the wind models, so any flight with wind turbulence was irreproducible.
+ * {@code PinkNoiseWindModel} was seeded once when the enclosing
+ * {@code SimulationOptions} was constructed and its seed field was final, so
+ * {@code setRandomSeed} could not change it; every level of
+ * {@code MultiLevelPinkNoiseWindModel} was built with the no-argument constructor
+ * and so was never connected to the seed at all.
+ */
+public class WindModelSeedReproducibilityTest extends BaseTestCase {
+
+	private static final int SEED = 12345;
+	private static final int RUNS = 5;
+
+	/** Turbulence has to be non-zero, or there is no randomness to reproduce. */
+	private static final double TURBULENCE = 0.2;
+
+	private static final double WIND_SPEED = 5.0;
+
+	/**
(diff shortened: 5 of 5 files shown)


sebastianvelace:
Two corrections after further review, both now pushed.

**The test did not test what it claimed.** All the reproducibility assertions fix the seed, so a model that ignored the value it was given and seeded from a constant would satisfy every one of them. The obvious complement — different seeds give different flights — does not close the gap either, because the integrator draws on the same seed: two flights differ under two seeds even when the wind is identical. I confirmed this by mutation, replacing the multi-level derivation with a constant, and the test still passed.

Added `testSeedControlsTheWindItself`, which samples the models directly. That removes the integrator from the measurement and asserts only the property in question — the same seed reproduces the same wind exactly, a different seed does not — and the same mutation now fails it.

**I have also corrected the wording of #3188**, which overstated the fault. Re-running a single `SimulationOptions` instance *is* already reproducible, since its wind model keeps the seed it was constructed with. What fails is building an equivalent configuration again, setting the same seed, and expecting the same flight — which is what a test, a script or a parameter sweep does. Measured on `cc9b3643`: repeated runs of one instance agree exactly, independently built ones with the same seed differ by up to 1.36 m of apogee.

One question I would rather ask than assume: adding `setSeed` to `WindModel` makes it a required method. There are only two implementers in the repository and both are updated here, but if `WindModel` is considered part of the plugin API I can make it a `default` instead, or avoid touching the interface.

JoePfeiffer:
Thank you for your contribution! I wonder if this would let us tighten the bounds on the assertions for other simulation tests?

Also, exposing the random number seed in the simulation options with a checkbox indicating the seed is fixed and an associated numeric text entry field does seem worth doing (but I agree with your assessment that it should be its own feature and PR, not bloat on this one).

sebastianvelace:
I wasn't sure either, so I went and measured it before answering.

Short version: yes, but not on its own. Merging this doesn't tighten anything by itself, since a test only gets a reproducible wind if it asks for one by calling `setRandomSeed`, and the ones with loose bounds don't.

Where it does help is the multi stage scenario in `FlightEventsTest`. On the centre booster's branch the tumble time wanders by about 0.15 s from run to run, against the 0.25 s the test allows, so roughly two thirds of the allowed slack was just noise. Pin the seed and it goes to exactly zero across 12 runs.

The single stage test barely budges, about 2 ms on ground hit, because the default wind is gentle and the flight is short. It drops to zero too, but there wasn't much there to reclaim. So the win sits in the longer, multi branch flights rather than spread evenly across the suite.

Fun thing I ran into while looking: `NaturalFrequencyTest`, `DampingRatioTest`, `CorrectiveMomentCoefficientTest` and `DampingMomentCoefficientTest` already pin `setRandomSeed(0xC0FFEE)`. They were written expecting this behaviour and only actually get it now.

Happy to take it on as its own PR: pin the seed across the simulation tests and tighten the bounds to whatever the pinned spread supports, with before and after numbers for each one. I'd keep it clear of #3190 so the tumble change stays easy to review on its own.

And thanks for merging this, and for opening #3191. Glad it was useful.

PR #3190, opened 2026-07-29
[#3183] Detect tumbling from a sustained angle of attack, not an instant one

### LINKED ISSUE #3183: [Bug] Stable rocket randomly aborts (TUMBLE_UNDER_THRUST) at launch rod clearance in wind — silent ~1 m apogee, non-reproducible results

sebastianvelace:
Fixes #3183. Replaces #3186, which took the approach you and I agreed to abandon.

## What this does

The previous criterion combined an instantaneous angle of attack above the stall angle
with a `cg > cp` stability comparison. Following your suggestion to wait for real
kinematic evidence, tumbling is now decided from how long a high angle of attack has
persisted, measured against the rocket's own pitch period.

Both halves of the old test were unsound, and the second is the one that mattered. The
angle of attack is measured against the apparent airflow, so a single integration step
can carry it across the threshold without the rocket's state having changed. And the CP
the stability comparison consumes is produced by an aerodynamic model that is outside its
validity envelope above the stall angle — that is, in exactly the regime where the test
is evaluated. It drew its conclusion from a quantity it had itself declared unreliable.

`TumbleDetector` low-pass filters the angle of attack with a time constant taken from
`TYPE_NATURAL_FREQUENCY`, so the wait scales with the airframe rather than being a
constant tuned to one rocket. No aerodynamic coefficient enters the decision, and no
flight-phase gating is needed in either direction: the guide departure transient does not
survive the filter, and descent tumbling does — which is the tumble-recovery case you
raised against #3186.

Since you asked, this measures the angle against the **air-relative** velocity, which is
what `TYPE_AOA` already is. So it is your original suggestion — orientation against the
velocity vector — with the dwell added.

## The warning

`Warning.LargeAOA` is now `MessagePriority.LOW`, which `ExampleFilesTest` already counts
as "informative". There is no `INFO` level in `MessagePriority`; if you would rather have
one added than reuse `LOW`, say so and I will.

It is now raised during a departure as well, since that is the first sign of what
follows — but **not for a separated stage**. A spent booster reaches a large angle of
attack every time, by design, so reporting it says only that the stage did what stages
do. It would appear on every flight of every multi-stage rocket, and a warning that
always appears is one nobody reads. That judgement is easy to reverse if you disagree:
it is one condition in `BasicEventSimulationEngine`.

The happy consequence is that no example baseline changes. `ExampleFilesTest` is
untouched by this PR.

## Validation

Estes Alpha III, 1 m guide, wind supplied by a listener rather than the pink noise model
and the integrator's seed pinned, so there is nothing random anywhere in the scenario and
every row below is reproducible exactly.

**Steady crosswind, no gust:**

| wind | before | after |
|---|---|---|
| 0 m/s | 133.1 m | 133.1 m |
| 2 m/s | 132.2 m | 132.2 m |
| 4 m/s | 130.7 m | 130.7 m |
| 6 m/s | **abort, 1.1 m** | **132.0 m** |
| 8 m/s | abort, 1.1 m | abort, 5.5 m |

The first three rows are numerically identical before and after: flights that already
worked are untouched. The 6 m/s row is issue #3183 — a steady crosswind, no gust and no
turbulence, aborting every time on `unstable`.

**A 0.1 s, +6 m/s gust swept across the whole guide-departure window, 12 positions:**

| steady wind | aborts before | aborts after |
|---|---|---|
| 0 m/s | 2/12 | **0/12** |
| 2 m/s | 3/12 | **0/12** |
| 4 m/s | 3/12 | 3/12 |
| 6 m/s | **12/12** | 4/12 |

**The rows where aborts remain are ones I checked individually, and they should stay.** A
1 m guide is short: this airframe leaves it at about 10 m/s, so a 6 m/s crosswind puts
the angle of attack at roughly `atan(6/10) ≈ 32°` at the moment of departure, past the
stall angle, with the fins unable to generate a restoring moment. Tracing one of them,
the angle climbs monotonically from 27° to 94° without recovering while `vz` falls under
thrust, and `TYPE_NATURAL_FREQUENCY` is `NaN` throughout — the simulator's own statement
that the rocket has no static stability. Those rockets really do flip over.

One honest note on the "after" column: at 2 m/s steady the worst surviving flight reaches
99.7 m rather than the usual ~130 m. It recovers and flies, but it loses altitude doing
it.

Regression check, from `TumbleDetectorTest`: removing the fins makes the airframe
genuinely unstable, and that is still detected.

## Tests

`core` suite green, 1231 tests. `:swing:compileJava` green.

`FlightEventsTest` now pins the simulation seed in the two multi-stage cases. Tumbling is
declared after a dwell rather than at a single step, which makes its time sensitive to
the integrator's random perturbation: the centre booster's tumble time ranged over 2.25 s
to 2.77 s across 100 runs, which is wider than the tolerance the test allows. Pinning the
seed collapses that to a single value, so the asserted times are a property of the rocket
rather than of the run. I preferred that to widening the tolerance until it passed.

Every assertion added here was checked by mutation — the separated-stage condition, when
removed, is caught by three tests.

## Note

This branch is based on `unstable`, not on #3189, so the two are independent. #3189 would
make the wind in these tests reproducible as well; here I sidestepped it with a listener.

### DIFF
--- a/core/src/main/java/info/openrocket/core/logging/Warning.java
+++ b/core/src/main/java/info/openrocket/core/logging/Warning.java
@@ -50,7 +50,9 @@ public abstract class Warning extends Message {
 		 */
 		public LargeAOA(double aoa) {
 			this.aoa = aoa;
-			setPriority(MessagePriority.NORMAL);
+			// Informative: it reports that the aerodynamic model is out of its envelope,
+			// which is worth knowing on a windy day, not that the flight has gone wrong.
+			setPriority(MessagePriority.LOW);
 		}
 
 		public double getAOA() {
--- a/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
+++ b/core/src/main/java/info/openrocket/core/simulation/BasicEventSimulationEngine.java
@@ -291,31 +291,53 @@ public class BasicEventSimulationEngine implements SimulationEngine {
 //					}
 //				}
 				
-				// Check for fin stall and either set tumbling or LargeAOA warning depending on
-				// rocket stability margin
+				// Check for tumbling, and otherwise for fin stall.
+				//
+				// Tumbling is decided from how long a high angle of attack has persisted,
+				// measured against the rocket's own pitch period, rather than from its value at
+				// a single step.  A gust cannot sustain it and a tumble cannot avoid it, so no
+				// flight-phase gating is needed in either direction: the launch guide departure
+				// transient does not survive the filter, and descent tumbling does.  Notably the
+				// CP/CG comparison is gone: the CP it consumed was produced by an aerodynamic
+				// model already outside its validity envelope at the angles where the test fired.
+				//
+				// A large instantaneous angle of attack still means the aerodynamic model is out
+				// of its envelope.  It is reported at low priority, since it says the crosswind
+				// is worth knowing about rather than that the flight has failed -- including
+				// while the rocket is departing, where it is the first sign of what follows.
+				//
+				// Not for a separated stage, though.  A spent booster left to fall reaches a
+				// large angle of attack every time, by design, so reporting it says only that
+				// the stage did what stages do.  A warning raised on every flight of every
+				// multi-stage rocket is one nobody reads.
+				//
 				// Inhibited if already tumbling, parachutes deployed, or on the ground
 				if (!currentStatus.isTumbling() &&
 					(currentStatus.getDeployedRecoveryDevices().size() == 0) &&
 					!currentStatus.isLanded()) {
-					final double cp = currentStatus.getFlightDataBranch().getLast(FlightDataType.TYPE_CP_LOCATION);
-					final double cg = currentStatus.getFlightDataBranch().getLast(FlightDataType.TYPE_CG_LOCATION);
-					final double aoa = currentStatus.getFlightDataBranch().getLast(FlightDataType.TYPE_AOA);
-					final double margin =
-						currentStatus.getSimulationConditions().getAerodynamicCalculator().getStallAngle() - aoa;
+					final FlightDataBranch branch = currentStatus.getFlightDataBranch();
+					final double aoa = branch.getLast(FlightDataType.TYPE_AOA);
+					final double stallAngle =
+						currentStatus.getSimulationConditions().getAerodynamicCalculator().getStallAngle();
+					// FlightConditions does not reach here, but its velocity is recoverable:
+					// the recorded Mach number is the air-relative speed over the speed of sound.
+					final double airSpeed = branch.getLast(FlightDataType.TYPE_MACH_NUMBER)
+							* branch.getLast(FlightDataType.TYPE_SPEED_OF_SOUND);
 
-					// large AOA -- stalling.					
-					if (margin < 0) {
-						// If we're stable, put a warning about large AOA
-						// note -- if cp is NaN (which it is while on the rod) cg > cp is false
-						if (cg > cp) {
-							// Not stable, so transition to tumbling
-							currentStatus.addEvent(new FlightEvent(FlightEvent.Type.TUMBLE, currentStatus.getSimulationTime()));
-						} else {
-							// Stable, so warning about AOA
-							if (currentStatus.recordWarnings()) {
-								currentStatus.addWarning(new Warning.LargeAOA(aoa));
-							}
-						}
+					final boolean tumbling = currentStatus.getTumbleDetector().update(
+							currentStatus.getSimulationTime(),
+							currentStatus.isLaunchRodCleared(),
+							aoa,
+							airSpeed,
+							branch.getLast(FlightDataType.TYPE_AIR_DENSITY),
+							branch.getLast(FlightDataType.TYPE_NATURAL_FREQUENCY));
+
+					if (tumbling) {
+						currentStatus.addEvent(new FlightEvent(FlightEvent.Type.TUMBLE, currentStatus.getSimulationTime()));
+					} else if (aoa > stallAngle
+						&& !currentStatus.isSeparatedStage()
+						&& currentStatus.recordWarnings()) {
+						currentStatus.addWarning(new Warning.LargeAOA(aoa));
 					}
 				}
 
@@ -535,6 +557,7 @@ public class BasicEventSimulationEngine implements SimulationEngine {
 
 					// Create a new simulation branch for the booster
 					SimulationStatus boosterStatus = new SimulationStatus(currentStatus);
+					boosterStatus.setSeparatedStage(true);
 
 					// Prepare the new simulation branch
 					boosterStatus.setFlightDataBranch(new FlightDataBranch(boosterStage.getName(), boosterStage, currentStatus.getFlightDataBranch()));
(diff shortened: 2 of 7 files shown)


JoePfeiffer:
Very nice!

My suggestion of setting LargeAOA's level to INFO was based on remembering the filters for displaying call it "informational" and me not checking the code. Changing all low priority warnings to informational is probably a good idea for consistency, but isn't the subject of this PR.

Something else to think about in a different PR is LargeAOA warnings on boosters -- you're right that all low power rocket boosters are supposed to tumble, but most high power are stable.  This is similar to the "open airframe" warning that is present on high power boosters but not low power; the heuristic that's used to decide whether to flag that warning is whether the booster has a recovery device. That's probably also a good idea for the LargeAOA.

PR #3204, opened 2026-08-02
[#2092] Fix recovery device deployment with motors in pods

### LINKED ISSUE #2092: [Bug] Wacky simulation results with motors in pods
Running an unstable build of 22.02.

I took the "Simple Model Rocket" example and added some pods to it, with motors (was preparing a tutorial for TRF).  The sim results are crazy:
image

The Apogee results show the effect of the very early deployment on the -0 and -2 second delays of the pod motors.  But the deployment velocity is exactly the same for both, as is max velocity.  This makes absolutely no sense.

Here is a sim plot of the first one:
image

OK, this is weird... it shows a deployment at first motor burnout (about 1 second), which is the burnout of the A10... but then it shows *another* deployment at 5 seconds, which is the ejection charge of the B4-4.  But there is only one parachute in the rocket.  

And this is why it's showing a 10 mph deployment velocity... it's only showing the *second* deployment (which is not really a deployment at all), and at that point the rocket is descending peacefully on the parachute that was deployed at 1s.  You can see the second deployment isn't real because the descent velocity doesn't change.

So basically there is all sorts of confusion in here in how the pod motors are being treated.

Here's the ORK file.
[pod_motors_broken.ork.zip](https://github.com/openrocket/openrocket/files/10867109/pod_motors_broken.ork.zip)

SiboVG:
This PR fixes #2092. There turned out to be two separate bugs behind the strange deployment behavior.

First, `DeploymentConfiguration.DeployEvent.EJECTION` matched ejection charges to recovery devices using stage numbers alone. Because pods belong to their parent stage, a pod motor’s ejection charge could incorrectly deploy the sustainer’s parachute. The check now compares the motor mount and recovery device by their innermost `ComponentAssembly`, reflecting that an ejection charge only pressurizes its own airframe assembly. If the firing motor is unavailable, the previous stage-based check remains as a fallback. Stage-separation behavior is unchanged.

Second, recovery devices could deploy multiple times. Each later ejection charge queued another `RECOVERY_DEVICE_DEPLOYMENT` event, and `FlightData.calculateInterestingValues` used the last event to calculate deployment velocity, producing the incorrect value reported in #2092. Deployment is now ignored when the device is already deployed, both when queueing and handling the event to cover simultaneous ejection charges. This also matches the UI description: “First ejection charge of this stage.”

The new assembly check exposed a dormant bug in `RocketComponent.getAssembly`, which never advanced to the parent and could loop indefinitely. The first commit fixes that and adds a `MotorMount.getAssembly` pass-through matching `getStage`.

---

One behavior change is worth noting: a recovery device inside a pod is now triggered only by motors in that same pod, including inline coaxial pods. Designs that place a parachute in a coaxial pod while keeping the motor in the main airframe will need another trigger.

### DIFF
--- a/core/src/main/java/info/openrocket/core/rocketcomponent/DeploymentConfiguration.java
+++ b/core/src/main/java/info/openrocket/core/rocketcomponent/DeploymentConfiguration.java
@@ -2,6 +2,7 @@ package info.openrocket.core.rocketcomponent;
 
 import info.openrocket.core.l10n.Translator;
 import info.openrocket.core.simulation.FlightEvent;
+import info.openrocket.core.simulation.MotorClusterState;
 import info.openrocket.core.startup.Application;
 import info.openrocket.core.unit.UnitGroup;
 import info.openrocket.core.util.ArrayList;
@@ -25,6 +26,14 @@ public class DeploymentConfiguration implements FlightConfigurableParameter<Depl
 			public boolean isActivationEvent(DeploymentConfiguration config, FlightEvent e, RocketComponent source) {
 				if (e.getType() != FlightEvent.Type.EJECTION_CHARGE)
 					return false;
+
+				// An ejection charge only pressurizes the airframe assembly its motor sits in, so a motor
+				// in a pod cannot deploy a recovery device in the parent airframe, and vice versa.
+				if (e.getData() instanceof MotorClusterState motorState) {
+					return motorState.getMount().getAssembly().equals(source.getAssembly());
+				}
+
+				// Fall back on the stage of the ejection charge if we don't know which motor fired it
 				RocketComponent charge = e.getSource();
 				return charge.getStageNumber() == source.getStageNumber();
 			}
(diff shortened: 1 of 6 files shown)


neilweinstock:
> First, DeploymentConfiguration.DeployEvent.EJECTION matched ejection charges to recovery devices using stage numbers alone. Because pods belong to their parent stage, a pod motor’s ejection charge could incorrectly deploy the sustainer’s parachute. The check now compares the motor mount and recovery device by their innermost ComponentAssembly, reflecting that an ejection charge only pressurizes its own airframe assembly. If the firing motor is unavailable, the previous stage-based check remains as a fallback. Stage-separation behavior is unchanged.

This seems like a very good change, eliminating the need to use boosters in this situation.

JoePfeiffer:
If someone is curious to play with before&after, here's a little .ork that triggers both bugs (before) and works properly (after)
[3204.zip](https://github.com/user-attachments/files/30991128/3204.zip)

PR #3213, opened 2026-08-08
[#3209] Refresh UI elements after theme changes

### LINKED ISSUE #3209: When changing UI theme, some UI elements do not update correctly until app is restarted
On Windows 11: I changed from dark mode to light mode, and then my window looked like this:

Image

Both the title bar and the coloring of the caliper controls are still showing as if in dark mode. Also the roll-over highlight of the caliper handles.  After closing and re-opening the app everything displayed correctly.  Similar problem going in the other direction.

SiboVG:
Fixes #3209 by also updating root panes (untested, I didn't have my Windows machine on hand) and the calipers after a UI theme change. Calipers now also remain in view when changing the theme.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/figure3d/RocketFigure3d.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/figure3d/RocketFigure3d.java
@@ -106,6 +106,8 @@ public class RocketFigure3d extends JPanel implements GLEventListener {
 
 	private static Color backgroundColor;
 	private Color customBackgroundColor = null;
+	/** Strong reference required because UI theme listeners are stored weakly. */
+	private final Runnable themeChangeListener;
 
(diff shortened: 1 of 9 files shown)


neilweinstock:
@jimmiedave discovered another one earlier this evening: when changing theme, the colors of the rocket figure display are initially displayed in their default colors, rather than the view settings.  Dunno if you covered that in your fixes above, I can test tomorrow.

jimmiedave:
> the colors of the rocket figure display are initially displayed in their default colors, rather than the view settings.

Repro: 3D figure only. Light mode, set custom 3D bg color. Switch to dark mode, bg is default. Open settings for figure view: take no action and set bg color reappears. 

2D seems to work as expected.

SiboVG:
Right, should be fixed now.

PR #3215, opened 2026-08-09
[#2318] Put appearance 2D and 3D in dedicated sections

### LINKED ISSUE #2318: Change layout of Appearance tab
I propose the following updated layout for the Appearance tab:

image

There is no new functionality here, only changes in layout and labeling to enhance clarity and usability.  Rationale:
1) Figure Style appearance is _rarely_ used.  Hence put it on the bottom.
2) Group the two separate appearances using field sets (or whatever they're called in Swing) to make it clear they're separate, and to provide labels for each separate group of settings. 
3) Label the second one "2D Figure Style", because I don't know if most folks even realize what "Figure Style" means.  By putting "2D" in there, it gives a good enough hint, especially as opposed to 3D appearance label up top.
4) Since the field sets are already appropriately labeled, don't need a redundant label on the first checkbox to enable default appearance mode.

SiboVG:
This PR fixes #2318 and puts the 2D figure style widgets and 3D appearance widgets in dedicated panels with a title border.

image

### DIFF
--- a/core/src/main/resources/l10n/messages.properties
+++ b/core/src/main/resources/l10n/messages.properties
@@ -1414,8 +1414,7 @@ MotorDbUpdate.Failed.title = Motor database update
 MotorDbUpdate.Failed.message = Unable to download and install the motor database update.
 
 ! AppearanceConfig
-AppearanceCfg.lbl.Appearance = Appearance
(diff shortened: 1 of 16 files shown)


neilweinstock:
Thank you for this one, good improvement.

Issue #2489, opened 2024-05-31
[Maybe Bug] OR shows significantly less weathercocking than RockSim or RASAero


JoePfeiffer:
### Describe the bug

A user on TRF compared weathercocking between OR, RockSim, and RASAero.  We show significantly less; this should be investigated to determine whether we have an error.

### To Reproduce

See https://www.rocketryforum.com/threads/weathercock-simulations-all-over-the-place.186584/

#### Screenshot(s):
*(drag-and-drop the screenshot(s) here)*

#### .ork file:
*(drag-and-drop the file here as a .zip file)*

### OpenRocket version

Tests were with 23.09

### What platform are you running on?

Linux

senft-research:
I know it has been age since this ticket, but something really doesn't sit right with me.

I am not a big believer in coincidences but, from the tabular data from the user on the forum post? The 2 negative up wind values add up to exactly -2048... 

Either its a huge coincidence that the only 2 negative upwind values came to a power of 2, or there might very well be a bug somewhere here.

PR #3220, opened 2026-08-09
Account for body-in-fin interference in CNa

SiboVG:
## Problem

As reported in #2489, OpenRocket predicts noticeably less weathercocking than RASAero II and RockSim for the tested rocket.

One likely contributor is that OpenRocket currently underestimates the fin-set normal-force slope, CNa. It accounts for the increased force on the fins caused by the body, but omits the corresponding force induced on the body by the fins. This missing contribution is most noticeable for relatively short fins mounted on a large-diameter body.

A lower CNa produces less aerodynamic normal force for a given angle of attack. It also reduces the aerodynamic weighting of the aft fin section, moving the calculated total CP forward. Together, these effects produce a weaker restoring moment in a crosswind, so the simulated rocket turns into the wind less strongly.

## What changed

OpenRocket previously used the classical Barrowman fin-in-body correction

`K_F(B) = 1 + τ`

where

`τ = r_t / (s + r_t)`,

`r_t` is the body radius and `s` is the exposed fin span.

This PR adds the omitted body contribution so that the combined subsonic correction becomes

`K_total = (1 + τ)²`.

The extra contribution used by the implementation is therefore

`K_body,effective = (1 + τ)² - (1 + τ) = τ(1 + τ)`.

The combined factor is applied through Mach 0.9. Between Mach 0.9 and 1.5, the additional body contribution is linearly reduced to zero. Above Mach 1.5, OpenRocket retains the existing `1 + τ` correction because calculating the body contribution there requires the fin-tip Mach-cone geometry from the full NACA method.

The additional force is currently assigned to the fin-set center of pressure. It is not included in cant-driven roll forcing because it represents force on the body rather than on the canted fin surface.

## Formula derivation and references

The existing `1 + τ` correction is the simplified fin-body interference factor from [Barrowman’s original NARAM-8 report](https://www.nakka-rocketry.net/articles/Barrowman.NARAM-8.pdf).

The combined square follows from the slender-body factors in [NACA Report 1307, equation 14](https://ntrs.nasa.gov/api/citations/19930091008/downloads/19930091008.pdf#page=8) and [equation 21](https://ntrs.nasa.gov/api/citations/19930091008/downloads/19930091008.pdf#page=10). If their long shared numerator is written as `Q(x)`, the equations reduce to

`K_F(B) = Q(x) / (1 - x)²`
Screenshot 2026-08-10 at 01 36 22

with `x = r/8`. I won't try to type out `Q(x)`... but it's the entire numerator, including 2/PI

and

`K_B(F) = [(1 - x²)² - Q(x)] / (1 - x)²`.

Screenshot 2026-08-10 at 01 36 38

Their sum is therefore

`K_F(B) + K_B(F) = (1 - x²)² / (1 - x)² = (1 + x)²`.

NACA uses `x = r/s`, with `s` measured from the body axis to the fin tip. In OpenRocket notation this is `x = r_t / (s + r_t) = τ`.

The same decomposition, along with the supersonic fin-tip Mach-cone limitation, appears in [Barrowman’s 1967 thesis, sections 3.31–3.32 and equations 3-96–3-99](https://ntrs.nasa.gov/api/citations/20010047838/downloads/20010047838.pdf#page=41).

This PR only uses the combined slender-body identity. It does not yet implement NACA’s separate geometry-dependent fin and body factors or their separate centers of pressure. A roadmap for that fuller implementation has been added to the technical documentation.

## CNa comparison to RASAero II

Old OpenRocket CNa plot of 'Two Stage High Power' example rocket for the two fin sets (blue = in Sustainer, red = booster):
Old

New CNa plot:
New

Rocket design exported to RASAero II and CNa plot of the sustainer (equivalent of the blue line in the OR plots):
Screenshot 2026-08-09 at 17 14 53

OR still underestimates CNa, but it's already closer to what RASAero calculates.

**EDIT**: I forgot that the nose cone still adds '2' to CNa of the sustainer. So, you need to shift the blue curves above upwards by '2'.

## Weathercocking comparison to RASAero II and RockSim

Unfortunately, I'm not knowledgeable enough to make a reliable comparison of weathercocking between OR, RASAero, and RockSim. It would be great if someone could chip in on this!

I also don't know how to reliably plot CNa in terms of Mach in RockSim.

---

Related to #2489, but I don't want to close it yet. Let's first implement the complete CNa implementation.

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/barrowman/FinSetCalc.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/barrowman/FinSetCalc.java
@@ -147,13 +147,13 @@ public class FinSetCalc extends RocketComponentCalc {
 			break;
 		}
 				
-		// Body-fin interference effect
+		// Combined body-fin interference effect on the normal force
 		double r = bodyRadius;
 		double tau = r / (span + r);
-		if (Double.isNaN(tau) || Double.isInfinite(tau))
+		if (Double.isNaN(tau) || Double.isInfinite(tau)) {
 			tau = 0;
-		cna *= 1 + tau; // Classical Barrowman
-		//		cna *= pow2(1 + tau);	// Barrowman thesis (too optimistic??)
+		}
+		cna *= calculateBodyFinInterferenceFactor(tau, conditions.getMach());
 		//		logger.debug("Component cna = {}", cna);
 		
 		// TODO: LOW: check for fin tip mach cone interference
@@ -169,7 +169,8 @@ public class FinSetCalc extends RocketComponentCalc {
 		// Without body-fin interference effect:
 		//		forces.CrollForce = fins * (macSpan+r) * cna1 * component.getCantAngle() / 
 		//			conditions.getRefLength();
-		// With body-fin interference effect:
+		// The body-in-fin lift does not act through the canted fin surface, so
+		// roll forcing retains only the classical fin-in-body correction.
 		forces.setCrollForce((macSpan + r) * cna1 * (1 + tau) * cantAngle / conditions.getRefLength());
 		
 		if (conditions.getAOA() > STALL_ANGLE) {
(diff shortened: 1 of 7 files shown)


JoePfeiffer:
It's a bit startling to see that giant Q(x) term cancel out like that! Still looking at it...

SiboVG:
> It's a bit startling to see that giant Q(x) term cancel out like that! Still looking at it...

I've edited the original post with the full formulas, that should make it more clear.

JoePfeiffer:
> > It's a bit startling to see that giant Q(x) term cancel out like that! Still looking at it...
> 
> I've edited the original post with the full formulas, that should make it more clear.

Oh, I didn't mean to imply it wasn't clear. I checked against the NACA report; it's still startling!

SiboVG:
> > > It's a bit startling to see that giant Q(x) term cancel out like that! Still looking at it...
> > 
> > 
> > I've edited the original post with the full formulas, that should make it more clear.
> 
> Oh, I didn't mean to imply it wasn't clear. I checked against the NACA report; it's still startling!

Oh yes. The best part about math is making things disappear.

neilweinstock:
Just tested it out and CP definitely moves back, which is a good thing.  Probably still somewhat more conservative than RockSim but any move in the right direction is very welcome.

PR #3231, opened 2026-08-10
Fix broken star history chart in README

Mubelotix:
The star history chart currently displayed in the README is broken. GitHub now restricts the stargazer API to owner-only access, which impacted the upstream star-history service. This PR points the chart to a small mirror of the project that fetches stars via an alternative data source, so no API token (and no token write access) is required. The existing hashtag link as well as the dark/light variants and legend/theme options are preserved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,10 +151,10 @@ OpenRocket is proudly open-source under the [GNU GPL](https://www.gnu.org/licens
 ⭐ Please give us a star if you find OpenRocket useful, and spread the word! ⭐
 
 
-<a href="https://www.star-history.com/?type=date&repos=openrocket%2Fopenrocket">
+<a href="https://star-history.dera.page/#openrocket/openrocket&type=date&legend=top-left">
  <picture>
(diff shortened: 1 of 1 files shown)


SiboVG:
Are you still seeing this issue? I already modified the star chart last week. It looks correct on my machine, even in incognito mode:

image

Mubelotix:
That's right my branch wasn't up-to-date when I forked so I didn't see it. You might however prefer this solution as it doesn't require giving away a token with write permissions to a third-party

SiboVG:
Thanks for the PR, and for flagging the broken chart.

Small correction though: the sealed-token flow doesn't need write access. The stargazers endpoint only requires a fine-grained token with Metadata: Read-only, and it's encrypted before it goes in the README. Write permissions apply to the Action-based alternatives, which push a generated SVG into the repo.

The custody concern underneath is fair, but this swaps one third party for a newer single-maintainer one with a dataset we can't verify. If it becomes an issue, I'd rather render the SVG in CI and drop the third party entirely.

Closing this one — nothing against the fork, and good luck keeping the tool alive!

Mubelotix:
> Only requires a fine-grained token with Metadata: Read-only

It was the case on day 1 but Github stepped in to increase the necessary permissions. You probably forgot but it wouldn't work if you hadn't added this scope

Also keep in mind that "encryption" is misleading, the third-party has to decrypt it in order to use it, so it lies somewhere in plain text in their systems for at the very least a few seconds every day

I totally understand your concerns though, no solution is perfect

PR #3236, opened 2026-08-11
Fix transonic fin CNa endpoint derivative

SiboVG:
The transonic interpolation used the queried Mach number when calculating the derivative constraint at Mach 0.9. This produced a different interpolation polynomial for each query and overestimated CNa inside the transonic region.

The interpolator interpolates like this:

```java
cnaInterpolator.interpolate(mach, subV, superV, subD, superD, 0);
```

So, from the last subsonic value (`subV`, `subD`) to the first supersonic value (`superV`, `superD`). `superV` and `superD` already correctly used a fixed mach, `CNA_SUPERSONIC`. `sq` in `subV` and `subD` also used `CNA_SUBSONIC` correctly, but `subD` still used `mach` instead of `CNA_SUBSONIC`.


SiboVG:
Old CNa of fin in 'A simple model rocket':
image


New CNa:
image

To be fair, the difference is hardly noticeable (look after Mach 0.9), but oh well 🙂

JoePfeiffer:
Dang that code could use some sensible variable names.

PR #3237, opened 2026-08-11
Fix body friction fineness ratio

SiboVG:
OR's tech docs Equation (3.85) applies the cylindrical-body skin-friction correction

$$
1+\frac{1}{2f_B},
$$

where the body fineness ratio is defined as

$$
f_B=\frac{L}{D}.
$$

However, in out code, we calculated `fB` using the maximum radius:

```java
double fB = (maxX - minX + 0.0001) / maxR;
double correction = (1 + 1.0 / (2 * fB));
```

which doubled the fineness ratio and halved the correction above 1. For example, a body with a true fineness ratio of 10 received a correction of 1.025 instead of 1.05.

This PR replaced R by D...

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanDragCalculator.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanDragCalculator.java
@@ -163,8 +163,8 @@ public class BarrowmanDragCalculator implements DragCalculator {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


SiboVG:
The effect is that $f_B$ now halved, which in turn increases the $(1+\frac{1}{2f_B})$ term, causing a slightly higher $(C_D)_{friction}$:

image

I'll post a few component analysis plots soon to show the difference.

PR #3244, opened 2026-08-13
Implement powered base drag correction

SiboVG:
OpenRocket previously calculated body base drag using the full coasting base area, even while motors were firing.

This conflicted with the existing [aerodynamic technical documentation](doc/techdoc/chapter-aerodynamic-properties.tex#L2088-L2155), which describes reducing the effective base area by the area occupied by active motor exits. Consequently, OpenRocket overestimated body base drag during powered flight relative to its documented model.

Rocket exhaust is known to change base pressure and drag; this was measured in free-flight tests documented in [NACA RM L50I18](https://ntrs.nasa.gov/citations/19930086428).

## Solution

The simulation now determines the projected area of motors currently in the `THRUSTING` state and subtracts it from the wake area. These areas are grouped by their associated core, booster, or pod wake.

image

For each independent wake \(b\), the effective powered base area is:

$$A_{\mathrm{base,powered},b}=\max\left(0,A_{\mathrm{base},b}-\sum_{i \in \mathcal{T}_b}N_i\pi\left(\frac{d_i}{2}\right)^2\right)
$$

Here, $\(\mathcal{T}_b\)$ contains the thrusting motor configurations that exhaust into wake $\(b\), \(N_i\)$ is the number of physical motors represented by the configuration, and $\(d_i\)$ is the motor diameter.

In simplified pseudocode:

```text
motorAreaByWake = {}

for each active motor cluster:
    if motor is THRUSTING:
        wake = assembly containing the motor mount
        motorAreaByWake[wake] +=
            motorCount * circularArea(motorDiameter)

for each exposed terminal body base:
    wake = assembly containing the body base
    totalBaseArea = instanceCount * projectedBaseArea

    poweredBaseArea = max(
        0,
        totalBaseArea - motorAreaByWake[wake]
    )

    add baseCoefficient * poweredBaseArea / referenceArea
```

This association is important for rockets with parallel boosters, powered pods, different motor sizes, or clustered motors. A motor can reduce only the base area of the wake into which it exhausts. If its area exceeds that base area, the result is clamped locally instead of reducing the base drag of another body.

The existing Mach-dependent base-drag coefficient is unchanged; only the effective area used to scale it is reduced.

Because OpenRocket motor definitions do not contain nozzle-exit diameter, the motor casing diameter is used as the available approximation. The base area refers to the full projected low-pressure wake bounded by the body’s external radius, not only the material area of a thin tube wall.

The correction applies only to exposed terminal body bases. Internal diameter steps and fin trailing-edge drag remain unchanged. When no motor is thrusting in a wake, its original coasting result is recovered exactly.

The approximation follows the method cited by the OpenRocket documentation from Eugene Fleeman’s [*Tactical Missile Design*, 2nd edition](https://openlibrary.org/books/OL9760680M/Tactical_Missile_Design_%28Aiaa_Education_Series%29), p. 23.

## Effect on Simulations

During motor thrust, the effective base area is reduced, resulting in less base drag and slightly higher acceleration and velocity. This can increase the simulated apogee.

For rockets with boosters, pods, or multiple motor groups, each motor now affects only its associated body wake. Coasting flight is unaffected because the correction applies only while a motor is actively thrusting.

'A simple model rocket' example design before:
image

after:
image

Comparison to RockSim:
image

As you can see, OR appears to underestimate apogee compared to RockSim. With this PR, we're growing a bit closer.


JoePfeiffer:
I'd been pondering trying to do this for a while, and the nozzle exit diameter has been the stumbling block. It is always smaller, and frequently much smaller, than the motor casing diameter. I've got an M1350 handy; its case diameter is 75mm and its exit diameter is 47mm.  I don't want to open the packaging to check, but I expect many of the motors in my stock have an exit diameter of half or less of the case diameter.

The estimate should be no greater than half the casing diameter, and there needs to be a way for a user who knows the actual exit diameter to enter it and to save it with the .ork file.

SiboVG:
> I'd been pondering trying to do this for a while, and the nozzle exit diameter has been the stumbling block. It is always smaller, and frequently much smaller, than the motor casing diameter. I've got an M1350 handy; its case diameter is 75mm and its exit diameter is 47mm. I don't want to open the packaging to check, but I expect many of the motors in my stock have an exit diameter of half or less of the case diameter.
> 
> The estimate should be no greater than half the casing diameter, and there needs to be a way for a user who knows the actual exit diameter to enter it and to save it with the .ork file.

Is it not more about the spread of the flame? I don't have much experience with this, but after exiting the nozzle, does the flame not expand to roughly the motor diameter? In that case, it still disturbs the wake.

neilweinstock:
> Is it not more about the spread of the flame? I don't have much experience with this, but after exiting the nozzle, does the flame not expand to roughly the motor diameter? In that case, it still disturbs the wake.

That's what I thought, but I'm not sure.  It should be easy enough to verify this.

JoePfeiffer:
My understanding is that the downstream plume is not very important: there is no mechanism by which it can "feed back" to affect the rocket.

What does matter is that, assuming perfect expansion (which is itself problematic), the rocket nozzle is at the same pressure as ambient. So we have the lower pressure at the aft end of the rocket causing base drag, but the area the pressure is acting on is reduced by the exit area of the nozzle. I thought I'd seen something about it in the RASAero documentation, but I'm not finding it right now.

Hmm, it looks like I ought to be able to get some papers related to it from the NMSU library...

SiboVG:
I think both effects need to be distinguished here. The plume can affect the near-base recirculation region and base pressure, so it is not quite correct that nothing downstream can feed back onto the rocket; [NACA RM L50I18](https://ntrs.nasa.gov/citations/19930086428) specifically measured changes in base pressure caused by motor operation.

However, that does not mean plume diameter or motor casing diameter belongs in Fleeman’s simplified formula. Fleeman uses the nozzle exit area $A_e$, and RASAero also asks explicitly for nozzle exit diameter when calculating power-on drag.

For the 75 mm casing / 47 mm exit example, using casing diameter makes the excluded area about $(75/47)^2=2.54$ times too large.

I would also avoid a universal “half the casing diameter” estimate. The given M1350 already has an exit diameter of 0.63 times its casing diameter, and the ratio will depend considerably on the motor design.

Maybe the most defensible implementation would be:

1. Add an optional nozzle-exit diameter associated with the selected motor configuration and save it in the .ork file.
2. Use the actual nozzle-exit area when provided.
3. When unknown, either apply no correction, as RASAero does, or use a clearly documented estimate—not silently treat the casing diameter as the nozzle diameter.

It's annoying that the motor data does not include information about the nozzle. Manufacturers also don't typically specify it...

neilweinstock:
The clearly-documented estimate in option #3 could just be something very conservative, like 25% or 30%.  Every motor has to have *some* kind of exit.  Of course, it won't make much difference to the simulation.

SiboVG:
> The clearly-documented estimate in option #3 could just be something very conservative, like 25% or 30%. Every motor has to have _some_ kind of exit. Of course, it won't make much difference to the simulation.

Should we add a checkbox `Using experimental motor` which has a small chance of a 300% plume expansion, followed by a tumbling flight event?

JoePfeiffer:
> I think both effects need to be distinguished here. The plume can affect the near-base recirculation region and base pressure, so it is not quite correct that nothing downstream can feed back onto the rocket; [NACA RM L50I18](https://ntrs.nasa.gov/citations/19930086428) specifically measured changes in base pressure caused by motor operation.

Yes, but they weren't attributing it to the downstream plume, they were looking at the exit pressure in the immediate vicinity of the nozzle, as shown in their Figure 9. And, just to make things more fun, this could either increase or decrease the base pressure depending on nozzle exit pressure.

> However, that does not mean plume diameter or motor casing diameter belongs in Fleeman’s simplified formula. Fleeman uses the nozzle exit area A e , and RASAero also asks explicitly for nozzle exit diameter when calculating power-on drag.
> 
> For the 75 mm casing / 47 mm exit example, using casing diameter makes the excluded area about ( 75 / 47 ) 2 = 2.54 times too large.
> 
> I would also avoid a universal “half the casing diameter” estimate. The given M1350 already has an exit diameter of 0.63 times its casing diameter, and the ratio will depend considerably on the motor design.
> 
> Maybe the most defensible implementation would be:
> 
>     1. Add an optional nozzle-exit diameter associated with the selected motor configuration and save it in the .ork file.
> 
>     2. Use the actual nozzle-exit area when provided.
> 
>     3. When unknown, either apply no correction, as RASAero does, or use a clearly documented estimate—not silently treat the casing diameter as the nozzle diameter.

My choice would be to take the same approach as RASAreo: add a "nozzle diameter" somewhere in the motor selection UI, with an initial value of 0.  This ends up just not correcting the base drag for power-on and works like it does now. When the user enters a nozzle diameter it is used to adjust base drag; it is also stored in the .ork file for next time. This is sort of a mix of all three options above.

A separate PR would adjust base pressure and thrust for altitude, assuming complete expansion at sea level and adjusting for air pressure. This gets even more tricky, since most (but not all!) rockets actually have the nozzle a little ways aft of airframe...

JoePfeiffer:
This one seems to have fallen off the radar -- I'd really like to see it get finished and merged.

SiboVG:
> This one seems to have fallen off the radar -- I'd really like to see it get finished and merged.

I'll try to make some time for it in the coming days.

SiboVG:
Done. The motor chooser dialog now has an entry field for the nozzle exit diameter. If this equals 0 (default value), powered base drag correction is not applied. The value is also saved in the .ork.

image

PR #3254, opened 2026-08-20
Add SignPath Windows signing workflow

SiboVG:
## Summary

We decided that we would not buy a code signing certificate anymore, partially due to the high cost, and to instead let SignPath code sign the Windows installers for us. This is a free service for open source projects. They keep the private key and only sign approved builds from our GitHub workflow, so it avoids the “one developer owns the key” problem.

They do control the signing service, so if they disappeared we’d need another provider for future releases. But already signed and timestamped releases would remain signed; they don’t need SignPath to keep operating. They also don’t control the repo, source code or distribution.

This PR adds a GitHub Actions workflow that builds the x86-64 and Arm64 Windows installers and submits them to SignPath for approval and Authenticode signing.

After signing, the workflow verifies both signatures and publishes the installers as a GitHub Actions artifact. Release maintainers download this artifact, test the installers, and upload them to the GitHub release manually.

The `Sign Windows installers` needs to be **manually** triggered, so no need to worry about code signing something on every new commit.

## Other platforms

- **macOS:** Continues to be built locally with install4j, including Apple Developer ID signing, notarization, and QuickLook extension integration.
- **Linux:** Continues to be built locally with install4j and does not require code signing.
- GitHub releases and their assets are still published manually.

## Additional changes

- Disables the previous Windows PFX signing configuration.
- Pins the bundled Windows JRE.
- Adds the SignPath code-signing policy and setup documentation.

## Testing

Once this PR is merged, we should run a test run of the workflow. We have configured two different SignPath policies: `release-signing` and `test-signing`. The former is for releases to be published, the latter is for testing the workflow. Currently, we have a repo secret (`Settings > Secrets and variables > Actions > Repository secrets`) `SIGNPATH_SIGNING_POLICY_SLUG ` set to `test-signing`. For official releases, we must modify this to `release-signing`.

### DIFF
--- a/.github/workflows/sign-windows.yml
+++ b/.github/workflows/sign-windows.yml
@@ -1,0 +1,198 @@
+name: Sign Windows installers
+
+on:
+  workflow_dispatch:
+
+# SignPath reads the workflow artifact and build metadata through GitHub.
+permissions:
+  actions: read
+  contents: read
+
+concurrency:
+  group: sign-windows-installers
+  cancel-in-progress: false
+
+jobs:
+  build-and-sign:
+    name: Build and sign Windows installers
+    runs-on: windows-latest
+    timeout-minutes: 60
+
+    env:
+      INSTALL4J_VERSION: '12.0.3'
+      INSTALL4J_DOWNLOAD_SHA256: '957fe72f9207fafae0b31f612cdd69a42f5b491108096eb16dfb79eea7c21e54'
+      SIGNPATH_ORGANIZATION_ID: ${{ vars.SIGNPATH_ORGANIZATION_ID }}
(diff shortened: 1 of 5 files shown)


SiboVG:
To clarify the process:

Push the final Release commit to `unstable` or `master` (the SignPath workflow is configured to work on either branch)
      ↓
Manually trigger the `Sign Windows installers` GH action.
      ↓
GitHub Actions builds Windows installers
      ↓
Unsigned installers sent to SignPath
      ↓
Maintainer approves in SignPath
      ↓
Signed installers returned to GitHub Actions
      ↓
We download, test, and upload them to the GitHub release

PR #3258, opened 2026-08-20
[#3251] Keep horizontal caliper label on-screen

### LINKED ISSUE #3251: Distance value is slightly off-screen when dragging horizontal calipers
When you drag a horizontal caliper, the real-time distance value displayed underneath it is slightly off-screen to the left.
Image

(I also think the text should be smaller.)

SiboVG:
Fixes #3251 by keeping the horizontal caliper’s live position label within the visible viewport. The label font is also reduced to match the vertical caliper display, with regression tests covering shifted viewport positioning.


neilweinstock:
Looks great.

PR #3260, opened 2026-08-21
Add Monte Carlo landing-dispersion analysis

### LINKED ISSUE #1050: Landing spread?
Are there future plans to have a visual display of potential landing locations, the same way Rocksim Pro does? Just curious

SiboVG:
This PR adds a Monte Carlo landing-dispersion analysis to the Simulations panel. Fixes #1050.

The analysis runs the selected simulation repeatedly while varying weather, launcher, vehicle, propulsion, and recovery inputs. It then plots the resulting landing cloud and calculates containment statistics for every recovered body or independently landing separated stage.

The original simulation and rocket are not modified by the analysis.

## Demo

https://github.com/user-attachments/assets/7d4dc230-3fcb-4388-b0d3-4c284e1cd7c0

## Screenshots

image

image

image

## How the analysis works

Each analysis consists of one nominal trajectory followed by the requested number of dispersed trajectories.

Before starting any trajectories, OpenRocket generates every sample from a signed 32-bit master seed. Each parameter has its own random stream, and simulation seeds use another independent stream. This has two useful properties:

- repeating the same simulation, settings, and master seed reproduces the same samples;
- adding or changing one parameter does not reshuffle the samples generated for the other parameters.

The sampled runs are executed concurrently using the configured maximum thread count. Each run receives an independent copy of the simulation, rocket, options, wind models, extensions, and aerodynamic state. Samples are generated before parallel execution begins, and results are returned in run-number order, so thread scheduling does not affect the analysis definition.

Variations are applied through simulation listeners rather than by editing rocket components:

- launch and wind inputs are applied to the copied simulation options;
- atmospheric density, mass properties, aerodynamic forces, and thrust are adjusted during calculation;
- ignition and recovery events are rescheduled through the event listener;
- recovery drag is adjusted after deployment.

The analysis records the east/north position when each eligible flight-data branch reaches the ground. A body is included when it descends under a deployed recovery device or represents an independently simulated separated stage. Stable component IDs are used to correlate the same body between runs, rather than relying on branch creation order.

A ballistic primary body is not treated as a valid landing body. If the nominal simulation contains neither a recovered body nor an independently landing separated stage, the analysis stops with an explanation.

Failures are retained per run and per body. A failed booster branch, for example, does not discard a successful sustainer landing from the same trajectory.

## Available uncertainty inputs

The setup dialog currently exposes the following inputs:

| Group | Parameter | How it is varied | Default spread |
|---|---|---|---:|
| Weather | Wind speed | Adds an offset to the average speed at every selected wind level | 1 m/s |
| Weather | Wind direction | Rotates the direction at every selected wind level | 10° |
| Weather | Air density | Scales atmospheric density through pressure at the model temperature | 2% |
| Launcher | Guide tilt | Adds a physical launch-guide angle error | 1° |
| Launcher | Guide azimuth | Adds an azimuth alignment error after resolving launch-into-wind | 5° |
| Vehicle | Total mass | Scales mass and inertia while preserving the separately sampled CG offset | 2% |
| Vehicle | Axial CG | Moves the center of gravity forward or aft | 5 mm |
| Vehicle | Axial drag | Scales the aerodynamic axial-force coefficient | 10% |
| Vehicle | Normal force and moments | Scales normal force, pitching moment, and damping moment together | 10% |
| Propulsion | Motor thrust | Scales the thrust from every active motor | 5% |
| Propulsion | Ignition timing | Shifts air-start and upper-stage ignition events | 0.1 s |
| Recovery | Recovery drag area | Scales axial drag while recovery devices are deployed | 15% |
| Recovery | Deployment timing | Shifts recovery-device deployment events | 0.5 s |

These defaults are intended as practical starting values, not as validated uncertainty data. They can be restored with **Reset to default**, while **All fixed** sets every explicit spread to zero.

The initial pad ignition is deliberately not shifted: changing it only moves the entire flight in time and does not disperse the landing point.

A single recovery-drag multiplier currently covers all deployed devices, so a drogue and main are correlated rather than sampled independently.

## Distributions and spread

Three distributions are available:

- **Normal**: the spread is one standard deviation around the nominal value.
- **Uniform**: the spread is the maximum positive or negative deviation from nominal.
- **Log-normal**: the spread is one standard deviation of the natural logarithm of a multiplier.

Normal and Uniform can be used for additive offsets and relative multipliers. Log-normal is only available for relative parameters such as mass, density, drag, and thrust. It keeps the multiplier positive and has a median equal to the nominal value; as expected for a log-normal distribution, its mean is slightly higher than nominal.

A spread of zero keeps that input fixed.

Relative Normal and Uniform samples are limited to a minimum multiplier of `0.01`, preventing mass, density, drag, or thrust from becoming zero or negative. Wind speed is limited to zero, launch-guide angle follows OpenRocket’s existing physical limit, and an event shifted before its trigger is placed 1 ms after that trigger. Very wide distributions can therefore have clipped tails.

Wind-speed variation changes the mean wind speed without changing the configured turbulence standard deviation. This preserves the stochastic wind model instead of accidentally scaling its turbulence together with the mean.

Inputs are currently sampled independently; this version does not yet provide a correlation matrix or empirical distributions.

## Results

The results dialog provides:

- an east/north scatter plot of dispersed landings;
- nominal and sample-mean landing markers;
- one-, two-, and three-sigma covariance ellipses;
- empirical R50, R90, and R95 radii around the sample mean;
- selection between recovered bodies and independently landing stages;
- a table containing every run’s status, landing position, maximum altitude, and flight time;
- PNG export of the displayed plot;
- CSV export of the complete analysis.

The plot keeps equal scale on both axes and supports panning, mouse-wheel zoom, zoom buttons, and fitting all landing points. Its palette is theme-aware and uses both color and line style to distinguish the covariance ellipses.

The exported CSV includes the master seed, analysis settings, per-run simulation seed, sampled input deviations in explicit SI units, stable body ID, branch index, landing output, altitude, flight time, and failure information. This should make an analysis reproducible and easier to inspect outside OpenRocket.

For a bivariate normal cloud, the 1σ, 2σ, and 3σ covariance ellipses contain approximately 39.3%, 86.5%, and 98.9% of the population. They are descriptive for skewed, clipped, or multimodal clouds; the empirical R50/R90/R95 values are the direct containment measures.

## Caching

The most recent completed result is cached in memory per simulation.

When the setup dialog is reopened with matching settings, **Plot cached analysis** opens that result without rerunning the trajectories. The action disappears as soon as the run count, master seed, distribution, or spread changes.

The cache also records the simulation state used to generate the result. It is invalidated when any relevant input changes, including:

- rocket or active flight configuration;
- simulation options and wind configuration;
- drag or stability lookup data;
- simulation extensions or their configuration.

The cache uses weak references to simulations, so cached results do not keep closed documents alive. Only completed analyses are cached; cancelled or failed analyses do not replace the previous result.

Results themselves are intentionally not written to the `.ork` file.

## Saving settings in `.ork` files

The run count, master seed, selected distributions, and non-zero spreads are stored on the individual `Simulation`.

Valid changes are saved when an analysis starts or when the setup dialog closes. Reopening the dialog restores those values. Simulations that have never configured landing dispersion do not receive an unnecessary settings block.

The `.ork` format now supports an optional element similar to:

```xml
<landingdispersion runs="1000" seed="123456">
  <uncertainty parameter="wind-speed"
               distribution="normal"
               spread="1.0"/>
  <uncertainty parameter="total-mass"
               distribution="log-normal"
               spread="0.02"/>
</landingdispersion>
```

Spreads are stored in OpenRocket’s internal SI representation. The loader validates the run count, seed, parameters, distributions, and spreads. Invalid or unknown entries produce warnings and are ignored rather than preventing the document from opening.

The analysis result is not persisted, so opening an `.ork` restores the setup but still requires a new run.

## Supporting simulation changes

A few lower-level changes were needed to make repeated parallel simulations safe and accurate:

- Added `Simulation.duplicateForIndependentSimulation()` to create a deep, document-independent simulation copy suitable for concurrent execution.
- Wind-model clones now own their listener lists and correctly relay changes through cloned `SimulationOptions`.
- Pink-noise wind models can change average speed while preserving their configured standard deviation.
- Cloned wind models reset their stochastic state, allowing the per-run simulation seed to control turbulence reproducibly.
- `FlightDataBranch` now retains the stable ID of the component represented by the branch, allowing separated stages to be matched between runs.
- Euler integration now fires the post-aerodynamic-calculation listener for recovery and tumble aerodynamics, matching the RK steppers.
- `RigidBody.equals()` now includes the center of mass, and its hash implementation is consistent with its tolerance-based equality.
- Simulation extensions must explicitly declare that they are safe for repeated concurrent Monte Carlo execution. Extensions with file, console, network, scripting, or UI side effects remain disabled by default.
- Built-in side-effect-free extensions opt in; arbitrary Java and scripting extensions remain blocked.

## Limitations

This first version does not yet include:

- correlations between uncertainty inputs;
- empirical or historical weather distributions;
- independent per-stage mass and propulsion errors;
- thrust-vector or fin-alignment errors — well, we currently don't support off-axis thrust in general;
- wind-shear variation;
- radial CG offsets;
- sensitivity attribution or convergence confidence intervals;
- latitude/longitude, terrain, or range-boundary overlays.

The analysis should therefore be treated as a tool for exploring a characterized uncertainty model, not as a substitute for validating that model or documenting its assumptions.

## Testing

Coverage was added for:

- Normal, Uniform, and Log-normal sampling;
- deterministic sampling and independent parameter streams;
- application of each uncertainty type;
- concurrent trajectory execution and result ordering;
- recovered bodies, separated stages, and partial branch failures;
- landing statistics and covariance ellipses;
- wind-model cloning and seed reproducibility;
- cache reuse and invalidation;
- `.ork` save/load round trips and invalid input handling;
- CSV export;
- setup-dialog persistence;
- results rendering helpers.


neilweinstock:
Dumb question: is landing dispersion the main thing that people generally want from Monte Carlo analysis?

SiboVG:
> Dumb question: is landing dispersion the main thing that people generally want from Monte Carlo analysis?

That's what I assumed, but I guess you can also use it to get apogee variation and other metrics.

neilweinstock:
> > Dumb question: is landing dispersion the main thing that people generally want from Monte Carlo analysis?
> 
> That's what I assumed, but I guess you can also use it to get apogee variation and other metrics.

Just want to make sure that the output of this covers the obvious use cases.  Like, do we also want dispersion plots of those other metrics?

Looks fantastic BTW.

JoePfeiffer:
For some reason when I run build 3517 I'm not getting the Monte Carlo button.

But two things: 

We definitely want to also show dispersion of other simulation results (there was a suggestion of error bars on plots, which would be a good way to display some of it).

We need to make sure that if there is no recovery device deployment then we won't do the landing dispersion. We want to make sure we can't be used for targeting.

SiboVG:
> We need to make sure that if there is no recovery device deployment then we won't do the landing dispersion. We want to make sure we can't be used for targeting.

Already implemented. Boosters (without a recovery device) are currently allowed to be plotted, but normal stages will give a warning message.

SiboVG:
Ok, I updated it to be a more generic 'Monte Carlo' simulation run. You can now choose between 'Landing spread' and 'Flight metrics'. In the flight metrics tab, you can choose between plotting the data in a histogram, or a box plot.

https://github.com/user-attachments/assets/f839a3bf-8d20-4520-859b-db4a7829580f

neilweinstock:
Very nice!!!  On extremely minor nit: Should be just "Apogee", not "Apogee Altitude".

SiboVG:
> Very nice!!! On extremely minor nit: Should be just "Apogee", not "Apogee Altitude".

Right, of course. It now uses existing translation keys from the sim table.

A few more improvements

image

image


In light mode (*yuck*):

image

image

SiboVG:
Decided to add one more thing: I thought it might be useful for range safety officers to retrieve a full report of landing spread + other Monte Carlo results as a PDF. In the 'Pint design info', you can now also print a Monte Carlo analysis report.

It's unchecked by default, as it's more of an advanced feature that doesn't need to be printed by default:
image

Demo:

https://github.com/user-attachments/assets/e2eda1b2-d879-48f1-a1c0-b2d278b48cfc

You can check which simulations to add. If the sim has not been configured/run before, a new Monte Carlo popup will open.

Example report:
[test.pdf](https://github.com/user-attachments/files/31346824/test.pdf)

kjhambrick:
This looks wonderful !

One Q:  Would PR 3260 eventually eliminate the need for [Standard Deviation] and [Turbulance Intensity] from the Wind Model interface ?

Thanks for all that all'y'all do !

-- kjh

PR #3261, opened 2026-08-22
Harden motor database update installation

SiboVG:
Improves the safety of motor database updates:

- Adds download timeouts and cancellation
- Prevents concurrent and stale installs
- Validates SQLite integrity before installation
- Uses unique temporary files
- Rolls back if the database and metadata cannot both be installed

### DIFF
--- a/core/src/main/java/info/openrocket/core/database/MotorDatabaseRemoteUpdater.java
+++ b/core/src/main/java/info/openrocket/core/database/MotorDatabaseRemoteUpdater.java
@@ -276,9 +276,11 @@ public class MotorDatabaseRemoteUpdater {
 							", sha256_db=" + actualDbSha256 + ")");
 				}
 
+				// Validate the full database before replacing the local one.
 				validateDownloadedDatabase(tmpDbFile);
 			}
 
+			// Cancellation and the total deadline also apply to validation time, not just network reads.
 			deadline.throwIfExpired();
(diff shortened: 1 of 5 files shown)


SiboVG:
Also see: https://github.com/openrocket/motor-database/pull/5

senft-research:
> Please make use of atomic commits. This means: don't fix 10 different issues and cram them in one commit. Split up commits into smaller commits that fix only one issue/feature.

Should this not be broken down into smaller commits, as it is addressing 5 different areas for improvement for the db? I know 5 is not 10 (I checked... twice!) but I am pretty sure the principle stands in this case haha

EDIT: After looking at it, I think this applies to [openrocket/motor-database#5](https://github.com/openrocket/motor-database/pull/5) aswell

SiboVG:
> > Please make use of atomic commits. This means: don't fix 10 different issues and cram them in one commit. Split up commits into smaller commits that fix only one issue/feature.
> 
> Should this not be broken down into smaller commits, as it is addressing 5 different areas for improvement for the db? I know 5 is not 10 (I checked... twice!) but I am pretty sure the principle stands in this case haha
> 
> EDIT: After looking at it, I think this applies to [openrocket/motor-database#5](https://github.com/openrocket/motor-database/pull/5) aswell

Sure, I'll split it up. I'll try to get to it tonight.

Issue #3196, opened 2026-07-31
[Bug] Low-aspect-ratio fin CP jumps across a pole at supersonic Mach numbers


SiboVG:
### Describe the bug

The supersonic fin center-of-pressure calculation can become singular for low-aspect-ratio fins, causing very large and discontinuous changes in both the fin-set CP and the rocket’s total CP as Mach changes.

The formula used for the CP position along the mean aerodynamic chord is:

```
f(x) = (x - 0.67) / (2x - 1)
```

where: `x = AR * beta` and `beta = sqrt(Mach² - 1)`

This formula has a pole at `AR * beta = 0.5`.

For a sufficiently low aspect ratio, this pole occurs at an ordinary supersonic Mach number.

### To Reproduce

1. Open `A simple model rocket`.
2. Modify the trapezoidal fin set to:
    - Root chord: 6 cm
    - Tip chord: 4 cm
    - Height: 0.5 cm
3. Open Component Analysis.
4. Step through Mach, especially near 2.69 (fin set AR = 0.2, so the `beta = 0.5 / 0.2 = 2.5`, thus Mach pole at approx. `Mach = sqrt(beta² + 1) ≈ 2.69`
5. Fin set CP goes loco

https://github.com/user-attachments/assets/c7c49a94-23a8-41c2-a4dc-cd5fbde4f5be

### OpenRocket version

unstable

### What platform are you running on?

macOS

### Additional context

`FinSetCalc` and `TubeFinSetCalc` contain equivalent versions of this CP logic and should use the same constants and limiting behavior.

SiboVG:
The component analysis parameter sweep actually shows the effect much better

<img width="963" height="612" alt="Image" src="https://github.com/user-attachments/assets/18a6e65c-bc5e-4737-930e-161a369d6b51" />

deepspace28:
I'd like to work on this one.

While reproducing it I hit a second singularity in the same CP path that isn't in the description. `calculatePoly()` divides its coefficients by `(1 - 3.4641*ar)^2`, and since `3.4641` is `2*beta` at Mach 2, that denominator vanishes at `ar = 0.288675`. So the transonic interpolation blows up for a different set of fins than the supersonic pole does — at that aspect ratio I measure a fin-set CP of about `-7.8e5` m at Mach 0.5025, well below the Mach range in the report.

[...]

PR #3263, opened 2026-08-24
Complete the NACA 1307 fin-body interference model

SiboVG:
This completes the fin-body interference work from #3220 using the method published in [NACA Report 1307](https://ntrs.nasa.gov/citations/19930091008).

The previous implementation treated fin-body interference as a single multiplier applied at the fin CP. That gives the correct slender-body total normal force, but loses the separate load carried by the body. This PR keeps the fin and body contributions separate, calculates the body-load CP, and recombines their moments. As a result, the rocket CP can now move with Mach and fin geometry instead of remaining tied to the isolated-fin CP.

## What changed

- Implement equation 14 for the fin load in the presence of the body.
- Recover the body carryover load from equation 21.
- Use equation 19 for cant-driven roll forcing.
- Calculate the subsonic body-load CP from chart 16.
- Calculate the supersonic body load and moment using the report’s planar-pressure equations.
- Use chart 15 for low-aspect-ratio supersonic configurations.
- Include the available cylindrical afterbody, including contiguous equal-radius body tubes.
- Interpolate force and moment through Mach 0.9–1.5 to avoid transonic CP discontinuities.
- Fair the small-angle model back to the existing fallback between 10° and 20° angle of attack.
- Retain the previous approximation for geometries outside the published chart range.
- Record the chart digitization, interpolation rules, panel mapping, and tolerances in doc/techdoc/naca1307-chart-digitization.md.

## Practical effect

At subsonic speeds, total fin-set CNa remains unchanged because the two NACA terms still sum exactly to (1 + τ)². The difference is that the fin and body loads now act at their own centers of pressure, so the combined CP can move.

Through the transonic and supersonic ranges, both CNa and CP may change for supported trapezoidal fin sets. The direction and size of the CP shift depend on taper, sweep, aspect ratio, body radius, afterbody length, and Mach; it is not a uniform forward or aft correction.

Cant-driven roll also changes. The old calculation used the angle-of-attack interference factor for roll, while NACA 1307 treats fin incidence separately. The equation-19 factor is generally smaller, so simulations with canted fins may predict less roll authority. This correction is applied consistently to trapezoidal, elliptical, and free-form fins because it depends only on the body-radius/semispan ratio.

### Comparison between master, unstable, and this branch

Below you can find the CP results from this branch, unstable, and master (= OpenRocket 24.12). The first plot shows the shift relative to master in millimetres; the second shows the absolute CP position measured from the nose.

example-cp-mach-relative

example-cp-mach-absolute

You will notice that much of the subsonic CP change was already present on unstable. This comes mainly from #3220, implemented in c9ed669d7, which added the body-in-fin normal-force contribution. Because that additional force was applied at the existing fin CP, it generally moved the whole-rocket CP aft. The contribution is removed through the transonic range and is absent above Mach 1.5, which is why unstable and master often converge again at Mach 2 and 3.

This branch replaces that approximation with the NACA Report 1307 model introduced in #3220. It separates the force on the fins from the force transferred to the body and applies each at its own CP.

At subsonic speeds, the total fin-set CNa remains unchanged because the two NACA terms still sum to (1 + τ)²; only their moment distribution changes. The resulting CP difference from unstable is therefore usually limited to a few millimetres at low Mach numbers.

The larger differences appear for supported trapezoidal fins at supersonic speeds, where the body load and its CP are calculated from the NACA pressure model. Depending on the geometry, this can move the whole-rocket CP aft by several centimetres. Compared with unstable, the shift is about 51 mm for the ARC payload rocket at Mach 2, 46 mm for the dual-parachute example, and approximately 100 mm for the simulation-extension examples at Mach 3. The direction and size of the shift depend on fin sweep, taper, aspect ratio, body radius, and available afterbody length, so it is not a uniform correction.

Three examples need some additional context:

- Airstart timing: The large subsonic shift comes from the earlier #3220, not from the complete NACA model. Its fin geometry uses the fallback path, which is why the current and unstable results are identical.

You can see the same change in RASAero too. Here is the normal Barrowman result:
image

And here is Roger's modified Barrowman:
image

- Pods—powered with recovery deployment: The example was replaced in 2b10806ac because the old version was an overly complex and unstable two-stage design with unmodelled off-axis booster thrust. Master and unstable therefore contain substantially different rockets, so their CP values are not directly comparable.

- Tube fin rocket: This change is caused by #3196.

## Applicability

The complete CP and normal-force calculation is limited to the range covered by the NACA charts:

- Trapezoidal fins
- Constant-radius cylindrical parent body
- r/s ≤ 0.6
- Taper ratio between 0 and 1
- No forward-swept leading edge
- No aft-swept trailing edge
- Forward flow and pre-stall angles of attack

Other geometries retain the existing scalar correction. The NACA result is used fully through 10° angle of attack and smoothly removed by the existing 20° stall angle.

## Validation

The tests cover:

- The exact equation-14/equation-21 slender-body identity
- The equation-19 wing-incidence factor
- The Mach 2 calculation from table I of NACA Report 1307
- The table II wind-tunnel comparison for triangular wing-body combination 2a
- Interior and endpoint values from charts 4, 15, and 16
- Both supersonic-leading-edge branches and their common sonic limit
- Continuity at the chart-15/planar-pressure boundary
- Continuity at Mach 0.9 and Mach 1.5
- Fin planform and radius-family interpolation
- Finite and extended cylindrical afterbodies
- Fallback behavior for unsupported geometries
- Roll forcing for supported and fallback fin shapes
- CP and CD regression values for every bundled example design across six Mach numbers
- Reproducible example simulations with fixed wind randomness

The digitized chart tests use tolerances based on the source scan resolution and plotted line thickness rather than treating the plotted values as exact data.


wcravens:
The justification / support documentation in the techdocs is really good (at least at first glance - haven't looked at the rendered version yet).  But can these ideas/changes be supported by references, preferably to published docs/papers?

neilweinstock:
> The justification / support documentation in the techdocs is really good (at least at first glance - haven't looked at the rendered version yet). But can these ideas/changes be supported by references, preferably to published docs/papers?

There's a link to the NASA report in the first post in this thread.

wcravens:
@neilweinstock Roger that.  I didn't really make my point clear.  I mean, can the reference be added to the techdoc.  In this thread is one thing, but in the techdoc would be good.

I'd do it myself, but figured I leave it the active branch, and those closest to the changes.

SiboVG:
> The justification / support documentation in the techdocs is really good (at least at first glance - haven't looked at the rendered version yet). But can these ideas/changes be supported by references, preferably to published docs/papers?

Done.

senft-research:
Due to my Aerospace Background, I would also like to review this PR if that is okay with you @JoePfeiffer @SiboVG ?

JoePfeiffer:
> Due to my Aerospace Background, I would also like to review this PR if that is okay with you @JoePfeiffer @SiboVG ?

In spite of your expertise, it would be best if I look at it.

senft-research:
> > Due to my Aerospace Background, I would also like to review this PR if that is okay with you @JoePfeiffer @SiboVG ?
> 
> In spite of your expertise, it would be best if I look at it.

Understood. Looking forward to the review!

JoePfeiffer:
Looking at calculateNacaApplicabilityWeight and I'm a little confused by the angle of attack
 check -- can't AoA legitimately be negative?

SiboVG:
> Looking at calculateNacaApplicabilityWeight and I'm a little confused by the angle of attack check -- can't AoA legitimately be negative?

AoA can be signed in other conventions, but OpenRocket represents it as an unsigned angle between 0 and π, with lateral airflow direction stored separately in theta. Both `FlightConditions.setAOA()` overloads clamp it to that range, and the simulation calculates it using `acos()`. So the negative check here is just defensive. Reverse flow corresponds to AoA above 90°, which the existing stall cutoff already excludes.

JoePfeiffer:
> > Looking at calculateNacaApplicabilityWeight and I'm a little confused by the angle of attack check -- can't AoA legitimately be negative?
> 
> AoA can be signed in other conventions, but OpenRocket represents it as an unsigned angle between 0 and π, with lateral airflow direction stored separately in theta. Both `FlightConditions.setAOA()` overloads clamp it to that range, and the simulation calculates it using `acos()`. So the negative check here is just defensive. Reverse flow corresponds to AoA above 90°, which the existing stall cutoff already excludes.

D'oh! Thanks.

PR #3273, opened 2026-08-25
Remove AWT/Swing (java.desktop) from OpenRocket Core

thzero:
### Summary
Decouples the headless `core` module from the `java.desktop` module. Previously the physics/engine core pulled in AWT and Swing — partly for pure math that happens to live in `java.desktop`, and partly for genuine GUI code (dialogs, file choosers) that never belonged in a headless engine. 

After this change **`core/src/main` has zero references to `java.awt` / `javax.swing` / `javax.imageio`.**

### Motivation
Motivation came from examination of other people who have been using the OpenRocket `core` module in different ways including use of alternative compilers.   These approached have yielded some code that had to be worked around in order to be compiled into alternative languages, and the AWT/Swing being one of them.  The AWT and Swing components were obvious candidates to propose to get removed from `core`.

### What changed
- **Geometry** — Replaced `java.awt.geom.{Line2D,Point2D,Rectangle2D}` usage (fin self-intersection, distances, bounding boxes) with a small native `Geo2D` helper. The segment-intersection routine reproduces `Line2D` semantics (incl. collinear overlaps).
- **Database-loading dialogs** — Motor/preset loaders no longer construct Swing dialogs. They report failures through a `DatabaseLoadingErrorHandler` callback; `swing` registers the handler that shows the dialog. Headless callers just log and continue. Also stripped HTML markup out of `RASPMotorLoader` exception messages.
- **Colors** — Migrated `java.awt.Color` in SVG export and the color preferences to the native `ORColor`, and moved `ORColor`'s AWT bridge into `swing`'s `ColorConversion`.
- **Preset images** — `BaseComponentDTO` drops its `BufferedImage`/`ImageIO` accessors and works with the raw PNG `byte[]` it already stored; encode/decode moves to the GUI.
- **File filters** — `SimpleFileFilter` is now a plain `java.io.FileFilter` in core; a new `SwingFileFilter` adapter in `swing` provides the `JFileChooser` type.
- **Example extension** — `CSVSave` throws a `SimulationException` (surfaced by the sim framework) instead of popping a `JOptionPane`.

### Approach
`core` exposes toolkit-neutral types (`Geo2D`, `ORColor`, `byte[]`, `java.io.FileFilter`, callbacks/exceptions); `swing` keeps all AWT/Swing and converts at the boundary via thin adapters (`ColorConversion`, `SwingFileFilter`, the loading-error handler). Public API signatures that used AWT types moved to the neutral equivalents, with GUI callers updated.

### Testing
`core` and `swing` compile clean; existing geometry, SVG-export, preferences, and fin-validation tests pass. AWT-bridge assertions moved from `core` tests to `swing` (`ColorConversionTest`); the preset-image round-trip test now asserts on `byte[]`.


jimmiedave:
Generally supportive of more/better modularity, better potential for re-/different-use.

Agnostic on merging for 2026 beta.

Pessimistic on potential for some unintended consequence someplace (maybe a small one).

Seems like this represents a really foundational change (prob. a good one) to the architecture. Robust testing seems a must here.

thzero:
Yeah, don't disagree.   Just one of those annoying things where core had gui stuff in it.

neilweinstock:
I'm labeling this one "next release" for now.  It's a good change but not critical to go in before the beta, and we need to focus on other things at the moment.

thzero:
> I'm labeling this one "next release" for now. It's a good change but not critical to go in before the beta, and we need to focus on other things at the moment.

I'd agree.   Most of my recent PRs probably fall in same boat.  These all coming out of watching other auxiliary projects and seeing where they patched OR code and why... And was it potentially worth a change/port back to OR.

And some like this were on my list back when I started the gradle work and really wanted core to be the engine and swing the ui, with a nice separation.

PR #3276, opened 2026-08-25
Keep calipers behind rocket information

### LINKED ISSUE #3271: [Bug] Calliper is in front of rocket info in 2d view
### Describe the bug

The caliper tool in 2D view obscures information about the rocket. In my opinion, the expected behavior would be for secondary tools to be behind the key information, not in front, to prevent this sort of obscuring. 

### To Reproduce

1. Enable the calipers tool
2. Move one of the calipers to a position where it overlaps with on-screen stats.

#### Screenshot(s):

Image

#### .ork file:

### OpenRocket version

Unstable

Linux

### Additional context

_No response_

SiboVG:
Fixes #3271. The calipers are now drawn behind the flight info. To be fair, it's still a bit difficult to read, but I don't see another easy way out other than dimming the caliper right behind the flight info boundary boxes, or to add a semi-transparent background box behind the flight info when the caliper is behind it. In my opinion, this solution is ok for now.

image

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/scalefigure/RocketFigure.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/scalefigure/RocketFigure.java
@@ -106,7 +106,7 @@ public class RocketFigure extends AbstractScaleFigure {
 	
 	private final ArrayList<FigureElement> relativeExtra = new ArrayList<>();
 	private final ArrayList<FigureElement> absoluteExtra = new ArrayList<>();
-	/** Elements painted last, on top of everything (including absoluteExtra), using the rocket transform. */
+	/** Elements painted over the rocket but below screen-space HUD elements, using the rocket transform. */
 	private final ArrayList<FigureElement> relativeTopExtra = new ArrayList<>();
 
 	private static Color motorFillColor;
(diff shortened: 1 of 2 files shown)


neilweinstock:
I'm not entirely sure I agree with this one.  When I click on an element to drag it I expect it to be on top.  You can always move the calipers or hide them entirely to read the other stuff in there.  

But I guess it's OK.  I'm far more bothered by the red in the toolbar.

jimmiedave:
Couple ideas here:

1. XOR drawing mode - probably looks a little rough during overlap, a little retro, but can ensure contrast.
Screenshot 2026-09-07 at 10 31 12 PM



<br /><br />
2. 75% opacity white (or black for dark mode, I guess) rectangle behind textual elements: can read the text, see the shaded caliper at the same time.

Screenshot 2026-09-07 at 10 35 49 PM

<br /><br />
3. Like 2 above, but the caliper labeling appears below the textual areas. Note also (near nosecone) that the caliper has a 2px white outline to improve contrast. This is a visual effect which doesn't affect measurements.
Screenshot 2026-09-07 at 10 58 28 PM

neilweinstock:
Honestly, I don't think anything fancy is required here, and in fact I prefer it the way it started, with the caliper rendered on top of the other stuff.

Reasons:
1) The other info is relatively static.  You don't need to be reading it all the time.  If you really really need to read it while using the calipers, just temporarily hide the calipers and then reenable them.

2) The calipers are generally to be used temporarily: enable them, do some stuff, then disable them.  While using them, **they are the most important thing**, and their readability should be prioritized.

You might argue that I am not foreseeing other use cases, that someone might want the calipers on screen all the time, or whatever.  If so, we can address that later.  Existing caliper implementation is already *far* beyond MVP, I don't think we need to keep polishing it beyond perfection at this point.

All of which is a long way of saying: I would defer this PR until after the Beta, and then let's see.  Our time is better spent elsewhere right now.

JoePfeiffer:
I'm with @neilweinstock on this. One's intuition would be that putting the calipers behind the rocket information would be better, but looking at the result of doing it I prefer what we've got now. @jimmiedave 's ideas for putting more work into keeping the calipers from obscuring might be worth doing, but we've got bigger fish to fry.

PR #3282, opened 2026-08-26
Depth-test origin axes against each other

### LINKED ISSUE #3272: [Bug] Blue Origin Axis of "Show origin axes" feature moves when moving rocket side to side?
### Describe the bug

Not sure how to describe this one (see video below) but when moving the rocket side-to-side, one of the axis (blue) moves along with it? 

I do not know how better to describe it than that's not how axis work? Unless we are using a system I am not familiar with here. Just because the rocket 

### To Reproduce

1. Move the rocket side-to-side with the "Show origin axes" feature on.
2. See the blue axis line... moves? 
3. 
...

#### Screenshot(s):

https://github.com/user-attachments/assets/c932d5c3-40a3-49c5-b603-5f3b1b78bc74

### OpenRocket version

Unstable

Linux

### Additional context

_No response_

SiboVG:
Fixes #3272. The 'issue' in #3272 was caused by depth perspective in the axes, but, the rendering of the axes was actually not done well. They had a weird render order, causing certain axes to always be drawn on top.

The axes now correctly intersect and render:

https://github.com/user-attachments/assets/8dfe3975-6247-4591-9aec-b0cfcf70bf09


senft-research:
What was the rationale behind the original rendering order?

PR #3283, opened 2026-08-26
Respect concave fin outlines in 3D

### LINKED ISSUE #3281: Fin shape is wrong in 3D view on this file
Image shows 2d view (correct) and 3d view (incorrect).  File is attached.

This is snapshot build e1a3a8c.

Image

[booster_shot.zip](https://github.com/user-attachments/files/31472819/booster_shot.zip)

SiboVG:
Fixes #3281. The 3D mesh generator used unconstrained triangulation, which could fill across indentations in freeform fin outlines. This uses constrained triangulation and preserves correct face winding.

image

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/figure3d/geometry/components/FinSetGenerator.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/figure3d/geometry/components/FinSetGenerator.java
@@ -14,10 +14,10 @@ import info.openrocket.swing.gui.figure3d.scene.properties.GraphicsQualitySettin
 import info.openrocket.swing.gui.figure3d.scene.properties.RenderingConfiguration;
 import org.joml.Vector2f;
 import org.joml.Vector3f;
-import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.Polygon;
-import org.locationtech.jts.triangulate.DelaunayTriangulationBuilder;
+import org.locationtech.jts.triangulate.polygon.ConstrainedDelaunayTriangulator;
+import org.locationtech.jts.triangulate.tri.Tri;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


senft-research:
What was the rationale of using unconstrained Delaunay initially?

senft-research:
Also looked at the `jts` library and the javadocs for the new triangulator are quite vague.

> The Constrained Delaunay Triangulation of a polygon is a set of triangles covering the polygon, with the maximum total interior angle over all possible triangulations.

That "maximum total interior angle over all possible triangulations" is certainly a bizarre / borderline nonsensical way of describing a constrained Delaunay Triangulation haha. I have made an issue on the jts repo for clarification. But it might be a shout to look at the code for the constrained triangulator to confirm what it is actually doing? 

Seems like it would be sensible to investigate further before we merge this. Sure it has fixed this problem, but knowing exactly how this new algo actually works would allow us to get ahead of potential complications going forward.

senft-research:
Were the concerns I put above discussed in private? Or was it deemed not important to clarify what the actual underlying code for the main way to triangulate the mesh is actually doing?

jimmiedave:
The upstream jts project's decisions on what to implement, why, and how are their own. It's not something the OpenRocket team needs to spend its limited time on. 

This PR was merged because it solves the demonstrated OpenRocket issue.

senft-research:
> The upstream jts project's decisions on what to implement, why, and how are their own. It's not something the OpenRocket team needs to spend its limited time on. 
> 
> This PR was merged because it solves the demonstrated OpenRocket issue.

I appreciate this, but when it comes to the brittle nature of LWJGL in terms of how it renders things, I do think this view is not considering how easy it is for secondary effects to pile up in our new rendering framework mate. 

There is a very real chance this fix could cause issues down the line, with us unable to find the cause as we have no idea the side effects this method of triangulation is causing. 

But ultimately I respect the final say of the maintainers.

Issue #3285, opened 2026-08-27
[Bug] Airframe continuity check is unit-dependent — checkGeometry compares diameters/positions as rounded display strings


thzero:
## Summary

`BarrowmanStabilityCalculator.checkGeometry(...)` decides whether adjacent airframe components are continuous — whether the body diameter "steps," and whether two components leave an axial gap/overlap — by formatting both values to their **display string** and comparing the strings, instead of comparing the numbers with a defined tolerance. Because the formatting rounds the value *in the user's configured display unit*, the geometry decision — and therefore which warnings a simulation emits — depends on the user's unit preference. The same design can produce a `DIAMETER_DISCONTINUITY` warning for an imperial user and no warning for a metric user.

## Location (`core`)

`core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanStabilityCalculator.java`, method `checkGeometry(...)`:

- **~line 164** — diameter discontinuity:
  ```java
  if (!UnitGroup.UNITS_LENGTH.getDefaultUnit().toStringUnit(2.0 * sym.getForeRadius())
          .equals(UnitGroup.UNITS_LENGTH.getDefaultUnit()
                  .toStringUnit(2.0 * prevComp.getAftRadius()))) {
      actualWarnings.add(Warning.DIAMETER_DISCONTINUITY, prevComp, sym);
  }
  ```
- **~line 181** — axial gap/overlap, identical technique on absolute X-positions (`symXfore` vs `prevXaft`).

The comparison routes through `Unit.toString(double)` (`core/src/main/java/info/openrocket/core/unit/Unit.java:100`), which does `toUnit(value)` (SI → display unit) → `roundForDecimalFormat(...)` (`:170`, ~3 significant figures) → `DecimalFormat`. `getDefaultUnit()` is the user-selected unit (`cm`/`in`).

## Why it's a correctness bug

1. **Unit-dependent output.** The effective tolerance is "equal to 3 significant figures in the user's display unit." Switching the default length unit changes the rounding grid, so it changes which continuity warnings are produced for an unchanged design. A physics/validation decision should not depend on a display setting.

[...]

PR #3287, opened 2026-08-28
Fix unit-dependent airframe continuity check by comparing diameters/positions numerically instead of as formatted display strings (#3285).

thzero:
### Problem

BarrowmanStabilityCalculator.checkGeometry(...)` decided whether adjacent airframe
components were continuous — whether the body diameter "steps," and whether two
components leave an axial gap/overlap — by formatting both values to their display
string via `Unit.toStringUnit` and comparing the strings. Because that rounds the
value *in the user's configured display unit*, the geometry decision (and therefore
which warnings a simulation emits) depended on whether the user had cm or inches
selected. The same design could produce a `DIAMETER_DISCONTINUITY` warning for an
imperial user and none for a metric user.

### Fix

Replaced both string comparisons with direct numeric tolerance checks via
`MathUtil.equals(a, b, epsilon)` — the same approach the rest of the geometry code
already uses (e.g. `SymmetricComponentCalc`, `SymmetricComponent.getNextSymmetricComponent`).
The result is now independent of display units. Removed the now-unused `UnitGroup` import.

### ⚠️ Reviewer note — tolerance value

The tolerance is a hardcoded constant:

```java
private static final double CONTINUITY_EPSILON = 0.0001; // 0.1 mm
```

- **Left as a `static final` constant, not a preference**, on purpose — this is an
  internal validation threshold, not a user-tuning knob. 
- **`0.1 mm` was chosen** as coarse enough to ignore sub-manufacturing modelling noise
  but fine enough to flag a real step. Deliberately *not* `MathUtil.EPSILON` (1e-8 m),
  which would over-warn on floating-point-scale differences.
- **Behavior change to be aware of:** 0.1 mm absolute is *stricter* than the old metric
  rounding for large-diameter airframes (old cm display precision hid steps up to ~1 mm
  on a ~20 cm tube). A few very-near-continuous large designs may now surface a
  `DIAMETER_DISCONTINUITY` that they didn't before. This is arguably more correct, but
  flagging it in case a different threshold (or a relative tolerance for diameters vs.
  absolute for positions) is preferred. 

- **The old behavior wasn't an absolute tolerance, it was "equal to 3 significant figures in the display unit," which is effectively a relative tolerance. Concretely, what the old code actually did in the default metric unit (cm):

diameters under 10 cm → resolution ~0.1 mm (my constant already matches this)
diameters over 10 cm → resolution jumps to ~1 mm

```java
// diameter: relative, mirrors the "3 significant figures" behavior across all sizes
double d1 = 2.0 * sym.getForeRadius();
double d2 = 2.0 * prevComp.getAftRadius();
if (Math.abs(d1 - d2) > CONTINUITY_REL_TOL * Math.max(d1, d2))   // e.g. 0.005 (0.5%)
    actualWarnings.add(Warning.DIAMETER_DISCONTINUITY, prevComp, sym);
````

~0.5% (0.005) is a reasonable central match to the old 3-sig-fig behavior. At 100 mm that's a 0.5 mm threshold; at 20 cm, ~1 mm — close to what old metric did on large tubes, so it removes the "new warnings on large airframes" concern.

### Testing

Added `BarrowmanCalculatorTest.testContinuityIndependentOfDisplayUnit()`: builds an
airframe with a 0.4 mm diameter step (a case that produced no warning in cm but a
discontinuity in inches under the old code), runs `checkGeometry` under both `cm` and
`in` and asserts the emitted warning set is identical (and non-empty). All existing
`BarrowmanCalculatorTest` tests continue to pass.

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanStabilityCalculator.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/BarrowmanStabilityCalculator.java
@@ -25,7 +25,6 @@ import info.openrocket.core.rocketcomponent.PodSet;
 import info.openrocket.core.rocketcomponent.Rocket;
 import info.openrocket.core.rocketcomponent.RocketComponent;
 import info.openrocket.core.rocketcomponent.SymmetricComponent;
-import info.openrocket.core.unit.UnitGroup;
 import info.openrocket.core.util.Coordinate;
 import info.openrocket.core.util.CoordinateIF;
 import info.openrocket.core.util.MathUtil;
@@ -39,6 +38,14 @@ public class BarrowmanStabilityCalculator implements StabilityCalculator {
 	private static final double STALL_ANGLE = 17.5 * Math.PI / 180;
 	private static final String BARROWMAN_PACKAGE = "info.openrocket.core.aerodynamics.barrowman";
 	private static final String BARROWMAN_SUFFIX = "Calc";
+
+	/**
+	 * Tolerance (in SI metres) below which two airframe diameters or axial positions are
+	 * considered continuous in {@link #checkGeometry}. Coarse enough to ignore sub-manufacturing
+	 * modelling noise, fine enough to flag a real step. Deliberately unit-independent: the
+	 * emitted continuity warnings must not depend on the user's display-unit preference.
+	 */
+	private static final double CONTINUITY_EPSILON = 0.0001; // 0.1 mm
 
 	private final WarningSet ignoreWarningSet = new WarningSet();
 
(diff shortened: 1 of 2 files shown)


jimmiedave:
Numeric comparison: yes, please. Good thinking; could make some of the nonsense noise warnings go away.

Relative tolerance sounds like a better choice to me: .1mm can be felt with fingers on a BT-20 model, probably wouldn't be noticed on a 4-in one.

**further thinking on this**
- Itching my brane is the thought that perhaps the textual comparison is based upon the numbers user can actually see on the screen (which should inspire user confidence in the result, _although it causes way too many warnings..._), while an exact numeric comparison instead could differ based on truncation/rounding/numeric conversion noise in the least significant digits.
- I'm thinking that the tolerance factor/would could eliminate that effect of "noise". Is that the case?

jimmiedave:
Testing:
Quick test case. Add a nose cone - default one. Add a shoulder, 48 OD shoulder, 50 shoulder length. Add a body tube - default one. Set body tube to 50.01 diameter (turning off "Automatic").

Observe that the text displays for both say "50", but there's a warning onscreen about a discontinuity in the body tube.
&nbsp;
Screenshot 2026-09-12 at 6 20 37 PM

neilweinstock:
I'm pushing this to next release, which for now simply means "after the upcoming beta".  Good feature, might need some fine tuning.

PR #3292, opened 2026-08-29
Fix Design Info Panel Save-as and Cancel bug

### LINKED ISSUE #3288: [Bug] "Save as..." brings up "Save Design Info" panel, with empty "Design name:", kills rocket name.
### Describe the bug

Double-clicking on a rocket in the component tree brings up the "Rocket configuration" dialog, which shows the "Design name" if one has been set (the examples here assume the design has a set name), and 5 other text values.

Choosing menu option "Save as..." brings up the similar "Save Design Info" (note incorrect OR capitalization style: OR uses Sentence case, not Title Case) dialog. However, this dialog comes up with an empty "Design name:" field. If one decides to save the file at this point, the "Design name:" field remains empty, but the design's name changes to "Rocket".

If instead one decides to Cancel the "Save as..." operation, two additional problems occur:

1. Immediately upon pressing "Cancel" in the "Save Design Info" dialog, the rocket's Design name changes to "Rocket" (i.e. this is not what a cancellation of the pending action should do)
2. The "Save as..." file dialog opens (which it should not, because we've already Cancelled.)

So dismissing the Save Design Info dialog through either OK or Cancel (or the <close window> gadget - a synonym for "Cancel") wipes out the "Design name:" with which the file was loaded.

### To Reproduce

1. Open a design with a "Design name:" (create one and save it first if you don't have one)
2. Double-click the topmost item in the component tree - it's the one labeled with the Design name.
3. The "Rocket configuration" dialog opens. Note the "Design name:" and the other 5 fields as filled out or not.
4. Click "Cancel", making no changes to the dialog.
5. Select `File>Save as...` menu item.
6. The "Save Design Info" dialog opens. (Note that the dialog title is in Title Case - incorrect style in OR.)
7. Note that the "Design name:" field is blank. (the other 5 fields don't appear to be affected at all by this issue).
8. Note that the topmost item in the component tree still has the "Design name" filled in.
9. Click "Cancel"
10. Note that the topmost item in the component tree is now called "Rocket", and the actual "Design name:" has disappeared.
11. Note that the "Discard your changes?" dialog appears, though no changes have been made by user.
12. Click "Yes" to discard.
13. Note that the "Save as..." file dialog appears, though the user has just decided to cancel (the cancelled operation shouldn't proceed past the cancellation).
14. Cancel the file dialog, close the file without saving and reopen it.
15. Perform steps 1-8 again
16. Click "OK" to approve the "Save Design Info" dialog (without adding anything to the blank "Design name:" field.
17. Note that the topmost item in the component tree is now called "Rocket", and the actual "Design name:" has disappeared.
18. Continuing to name and save the file at this point saves with an empty "Design name:". If you'd instead typed the correct name into the "Design name:" field, it would've been saved with the design, but wouldn't change the behavior of the "Save Design Info" dialog next time the file is opened.

### Screenshots / .ork file

Image

Image

Image

### OpenRocket version

24.12 released

macOS

### Additional context

Macbook Pro, external monitor and laptop screen open
M3 MAX Apple Silicon CPU
macOS Tahoe 26.6.2

wcravens:
Closes #3288

### DIFF
--- a/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
+++ b/core/src/main/java/info/openrocket/core/rocketcomponent/Rocket.java
@@ -541,6 +541,10 @@ public class Rocket extends ComponentAssembly {
 		this.functionalModID = source.functionalModID;
 		this.refType = source.refType;
 		this.customReferenceLength = source.customReferenceLength;
+		this.designer = source.designer;
+		this.revision = source.revision;
+		this.designType = source.designType;
+		this.kitName = source.kitName;
 		this.stageMap = new ConcurrentHashMap<>(source.stageMap);
 
 		// these flight configurations need to reference the _this_ Rocket:
--- a/core/src/test/java/info/openrocket/core/document/OpenRocketDocumentUndoRedoTest.java
+++ b/core/src/test/java/info/openrocket/core/document/OpenRocketDocumentUndoRedoTest.java
@@ -1,0 +1,52 @@
+package info.openrocket.core.document;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import info.openrocket.core.rocketcomponent.DesignType;
+import info.openrocket.core.rocketcomponent.Rocket;
+import info.openrocket.core.util.BaseTestCase;
+import info.openrocket.core.util.TestRockets;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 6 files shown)


senft-research:
It is good practice to identify what you have actually done to resolve the issue, to make it easier to review. 

Thanks for the contribution!

senft-research:
Is there a reason this was closed?

PR #3295, opened 2026-08-29
Add a 3D flight path export (KML / CSV / GPX) with custom templates

thzero:
## Summary

OpenRocket can export a simulation's numeric data as CSV, but there is no way to get the flight's ground track and altitude out in a form that mapping tools understand. This adds a **3D Path** tab to the simulation edit dialog that exports the flight path to KML (for Google Earth), a waypoint CSV (for Google My Maps and similar tools), or GPX.

OpenRocket already computes a latitude and longitude for every simulation step: it starts from the launch site coordinates and applies the flight's horizontal motion, using the same geodetic calculation already used for Coriolis and wind drift. The exporter just reads those values, so the exported track reflects wind, the
chosen Earth model, and stage drift. The only requirement is that the launch latitude and longitude are set in the Launch Conditions.

## What the tab does

The new **3D Path** tab sits next to **Export data** and contains:

- A format dropdown: KML (Google Earth), Waypoint CSV, GPX track, and any custom templates the user has installed.
- Unit selectors for altitude and distance.
- Checkboxes for which points of interest to mark: pad, liftoff, burnout, apogee, recovery deployment, landing, max velocity, max acceleration.
- Options for the path itself: include the airborne line, include the ground track, and a   "keep every Nth point" setting to thin out long flights.

The write runs on a background thread with the same progress dialog the CSV export uses, so a large file does not freeze the UI. If the launch latitude and longitude are both zero (the default), the tab warns before writing, because the track would otherwise land off the coast of Africa. The chosen format and options are remembered between sessions.

## Templates

The output formats are not hard-coded. Each one is a Mustache template that is filled from a data model built from the flight. The three built-in templates (KML, CSV, GPX) ship with the application. Users can drop their own `.mustache` files into an `ExportTemplates` folder in the OpenRocket user directory and they appear in the format dropdown automatically, with no rebuild. This is how someone reproduces a particular club's waypoint format or a custom KML style.

The user guide documents the folder location, the naming convention, and the full list of tokens available to a template.

## Implementation notes

- New core package `info.openrocket.core.file.flightpath`: the data model   (`FlightPathModel`), a builder that turns a simulation into that model (`FlightPathModelBuilder`), a template repository (`FlightPathTemplateRepository`), and the exporter (`FlightPathExporter`). This code is Swing-free and unit tested.
- Rendering uses jmustache (`com.samskivert:jmustache`), a small logic-less Mustache engine. The output escaping is chosen by file extension: XML escaping for `kml`, `gpx` and `xml`, quote doubling for `csv`, none otherwise.
- The Swing side is the new `SimulationFlightPathExportPanel` plus the tab wiring in `SimulationConfigDialog`. The existing CSV **Export data** tab is unchanged.

## Testing

`FlightPathExportTest` covers the model building (distance and bearing from the pad, unit conversion, the summary maxima) and the rendering of each built-in template, including XML escaping and the flight-path/ground-track toggles. Existing tests are unaffected.


kjhambrick:
I love this idea !

PR 3295 will be a **HUGE** step-saver for me.

My pre-launch workflow with OR 2412 is this:

1. run an OR MultiLevel Wind sim using an OpenMeteo forecast for 'launch day'
2. export the OR **sim.csv,**
3. parse **sim.csv** as **lat-lon-alt.csv** and **pins.csv**
4. import **lat-lon-alt.csv** and **pins.csv** into [GPS Visualizer](https://www.gpsvisualizer.com/map_input?form=googleearth) to produce a **gpsviz.kml** file for google earth,
5. load the **gpsviz.kml** into google earth,
6. repeat until I find 'the best' motor candidates and launch time(s) for the launch day.

All I will need to do is steps 1 and 6 with PR 3295 !!

Very nice, thanks @thzero.

-- kjh

jimmiedave:
Generally, I like the idea.

Thoughts: 

- Launch lat/lon: since this is required, rather than warn on (0,0), can user be prompted to enter one (and offered the widgets to do so, with the input stored in the "regular" place for it) if not set?
- Thinning out points on long flights: rather than keep every Nth point, could this be optimized based on keeping a vector over several points? i.e. 6 points in a straight line are represented as well by the line's endpoints, while 6 in a curve represent a bunch more information. Could the points be thinned based on closeness to a straight line (i.e. dropped if nearly irrelevant) automatically? This could cut down a bunch of points during ascent and during linear descents in calm winds, without sacrificing a bunch of useful detail. 
- Thinking further on what I've just suggested, probably good to understand that "every Nth point" keeps uniform timing between samples (think: the frame rate of a film) which might be important if user is intent on creating an animation. Worth thinking through what's desired at end-product.
- Wondering if, now that there are tabs for what would've been the simulation edit, sim options, plot, export, and now 3D path, we need separate "Edit simulation" and "Plot / Export" **buttons** for this tabbed UI. There could be a single button that opens this panel.

jimmiedave:
Testing: Opening a saved design, exporting a KML sets the "dirty" flag for the file and asks for a "Save" action upon closing. It's not apparent to me, acting as User, what might have changed in the file requiring a save: I was assiduous to change nothing.

**RESOLVED**:
This file had not had any simulation run, and obvs requires a simulation to export KML. Running the simulation dirties the file. The simulation ran on its own when I clicked the "Plot / Export" button (but interestingly, not when I clicked the "Edit simulation" button - though it ran on its own when I closed the dialog after coming in from "Edit simulation").

jimmiedave:
Testing: using the Two-Stage example rocket, it's not apparent that both stages are captured correctly in the KML file. Zooming out, "Burnout" appears, on the ground.

Also, the flight appears to be happening in a neighborhood in the SW USA. (In the past, we've used KSC in FL, USA as a default launch location). If this flight is supposed to have happened in the SW USA, we'll need to equip the example rockets with much more precise lat/lon positions to maintain realism (and possibly be concerned that we might reveal contributors' street addresses, etc).

Entering data with significant precision I was able to accurately target our club's marquee launch site in Georgia (after first finding my launch occurring in China: flipped sign on Longitude).
&nbsp;

Screenshot 2026-09-12 at 4 06 52 PM

thzero:
Odd, all my flights without a lat/lng used ksc.   I'll check that out.

Thanks for feedback .  I'll see about checking on the two others.

Anything odd about the 2 stager?

jimmiedave:
The two-stager is "Two stage high power rocket" from `File>Open example`

The Lat/Lon in the file does appear to be in the Southwest (maybe that's @JoePfeiffer 's place).

jimmiedave:
Testing: Same imported KML appears quite differently in map view vs. satellite view. 

Certainly possible that I've turned on "labels" or whatever only for map view (I don't use Earth often), but way different, and I don't know why only the 3 labels that appear above are in the Satellite view (why not all or none?).
&nbsp;
Screenshot 2026-09-12 at 6 27 32 PM
&nbsp;
Screenshot 2026-09-12 at 6 27 47 PM
&nbsp;

kjhambrick:
@thzero --

Should I be able to build my own copy of **PR 3295** for testing or is there a .JAR file that I missed ?

Without having built a copy of this PR 3295, is the **\<altitudeMode\>** user-selectable so that the KML file is displayed as desired in Google Earth ?

I am asking because all I see are **\<altitudeMode\>relativeToGround\<\/altitudeMode\>** pics.

While the "**relativeToGround**" and "**absloute**" \<altitudeModes\> are dramatic to view in Google Earth ( not to mention the beautiful \<extrude\>1\<\/extrude\> attribute ), the most useful mode for me personally when DriftCasting is "**clampToGround**" mode so that I can most easily see the launch site hazards which may come into play for my intended flight(s) for the launch window.

And I believe the **\<altitudeMode\>** ( and also \<**extrude**\> ) properties do need to be run-time settings when the KML file is generated so that the file is interpreted correctly when imported into Google Earth ( I've never explored changing the \<altitudeMode\> from within GE ).

Here is a q&d google search:  [what is the domain of AltitudeMode in a KML file ?](https://www.google.com/search?q=what+is+the+domain+of+AltitudeMode+in+a+KML+file+%3F&sca_esv=444dee17e46a57e2&sxsrf=APpeQnthLFR2mPFjFySefnY6gArZAXD38w%3A1789296007476&ei=weeiauvnJpWtqtsP0t77qAE&biw=1868&bih=863&uact=5&sclient=gws-wiz-serp&fbs=ABfTbFVyMZGZf1hfvX9uKjN_-G8c4u0nXx4bEIpwm1lnNH832UV2GkpKEK36HumGR10Evz0DE9hkPgm5pwJD6Y4g4gdUnrwTgnTiuY9xU8D56OFy27CWno0tmsB3TNHX8y-8cBcj9NEJapOKomqRyLG0NeYB48TKlPl-3BTtFoVamK7F0S_gV7G_TI02xC0sBDhBchs8Px36ZdR6TMrSvwuxatlKqnSsFg&aep=10&ntc=1&mstk=AUtExfDpRsjMzX6l5blLJIbscVlNmb0Lgr8V-dX5wjcSiT3Vrwdqyk5riPtRSmAVAtkJf2lIHVjQCXM2cOPyBuhl6u1jl2DxoCNxfTURmEa0xNCN2ikOPb_N5VbAxBcyfR7VxE6pJFE2n1YKgzNGDaY0DRmyjr2tiYY5KzL2lqYHJsvFCYtcMoAAAfTw9it1INq3QBHurxORaTUoBu6kXb6AAUGPgGxabwXNDcfE2jNMBipJdE6yT5vkPEnLOU21sBwyxHrUbf8qcAtjcqcf24mYgZy0RCRx2Dd66IRYY9nZz4GHFnspf93ed2lwupbxqyau-TM77AJsPoXnHg&aioh=3&csuir=1&cs=0&udm=50&atvm=2&mtid=432matPEBa3fwbkPq_yBSA)

Concrete examples from other KML generators that I use ...  

This is a screen shot of the GPS Visualizer web app along with the relevant help file.

My screenshot tool is not able to captide the on-screen [**Altitude Mode**] dropdown ( left-center screen ) so I added the help file to the top left:

Screenshot_20260913_052355


Similarly, GPS DC calculates KML files directly from OpenMeteo hourly data but then it sets the GE \<altitudeMode\> via one of three Export Buttons below the on-screen GE map inage:

Screenshot_20260913_052604

In any case, the content of the KML file need not change, only the \<altitudeMode\> and the optional \<extrude\> variables.

HTH and THANK YOU !

-- kjh

p.s. so is there a .JAR file I can download or should I be able to build my on copy of OR 26XX Snapshot with PR 3211 + PR 3295 ?

kjhambrick:
Oops ... I meant to include this Google Earth screenshot.

This was a candidate flight ( LPV2 on an AT E30T ) for the Tripoli Houston Launch yesterday at Hearne Muni Airport created by Google Earth.

 There are two DriftCast sources::  

1.  OpenRocket PR 3211 Export / KML created by GPS Visualizer for a12:00 pm fkight ( bold red tracks with yellow annotated pins )
2. GPS DC for the same altitude and weathercocking distance at 10:00, 11:00, 12:00, 13:00,14:00 CDT.

Screenshot_20260913_063456

Thanks again !

-- kjh

kjhambrick:
One last thing ... I can't see Open Rocket DriftCasts ever replacing @dbellhorn's GPSDC and especially not his RSO edition -- they are two completely different animals.

thzero:
> This file had not had any simulation run, and obvs requires a simulation to export KML. Running the simulation dirties the file. The simulation ran on its own when I clicked the "Plot / Export" button (but interestingly, not when I clicked the "Edit simulation" button - though it ran on its own when I closed the dialog after coming in from "Edit simulation").

> SimulationConfigDialog's OK handler calls getOutdatedSimulations() → SimulationRunDialog on every OK press, regardless of tab (SimulationConfigDialog.java:441-446). That's why the sim ran when you hit Plot/Export, and why it also ran when you closed the dialog with OK after "Edit simulation" — the Edit button itself only opens the dialog, so nothing runs until OK. 

So yeah thats why you saw the results you did as far as the dirty and saving goes.

thzero:
> The two-stager is "Two stage high power rocket" from `File>Open example`
> 
> The Lat/Lon in the file does appear to be in the Southwest (maybe that's @JoePfeiffer 's place).

Yeah, its definitely the sample file that has some lat/lng in it.  Seems like something we should clean up, and like have all the sample files fly out of KSC or something.

thzero:
> One last thing ... I can't see Open Rocket DriftCasts ever replacing @DBellhorn's GPSDC and especially not his RSO edition -- they are two completely different animals.

why not? its very simple code, very very simple rocket code too :(.   the only reason I can see is its very simple, so can be easily run without any real setup prior, and OR would require of course an Ork file.

thzero:
> @thzero --
> 
> Should I be able to build my own copy of **PR 3295** for testing or is there a .JAR file that I missed ?
> 
> p.s. so is there a .JAR file I can download or should I be able to build my on copy of OR 26XX Snapshot with PR 3211 + PR 3295 ?

yeah, you'd have to build your own copy of a .jar file. or send me a DM


> Without having built a copy of this PR 3295, is the **<altitudeMode>** user-selectable so that the KML file is displayed as desired in Google Earth ?

No, but probably could be.

thzero:
> Testing: Same imported KML appears quite differently in map view vs. satellite view.
> 
> Certainly possible that I've turned on "labels" or whatever only for map view (I don't use Earth often), but way different, and I don't know why only the 3 labels that appear above are in the Satellite view (why not all or none?).  

Dunno, I'll admit I only use map view, I have not tried to do satellite view or determine the difference.

thzero:
> Testing: using the Two-Stage example rocket, it's not apparent that both stages are captured correctly in the KML file. Zooming out, "Burnout" appears, on the ground.
> 
> Also, the flight appears to be happening in a neighborhood in the SW USA. (In the past, we've used KSC in FL, USA as a default launch location). If this flight is supposed to have happened in the SW USA, we'll need to equip the example rockets with much more precise lat/lon positions to maintain realism (and possibly be concerned that we might reveal contributors' street addresses, etc).
> 
> Entering data with significant precision I was able to accurately target our club's marquee launch site in Georgia (after first finding my launch occurring in China: flipped sign on Longitude).  
> 

Scratching head on this one.   This is a shot of one of my 2 stages doing like 3400' feet.   The built in one does 4400' or so, and yet....

image

kjhambrick:
> **why not? its very simple code,** very very simple rocket code too :(. the only reason I can see is its very simple, so can be easily run without any real setup prior, and OR would require of course an Ork file.

It would be a lot more work to select the best motor and launch window for a recoverable flight using OR alone.

Mostly because IMO, OR is not the right tool for that job because one exports a single simulation, and not a set of simulations all at one time.

The main limitation IMO, are the underlying data structures in the .ORK:   Not counting the physical properties of the rocket, each \<simulation\> block is more-or-less a standalone data structure ( see below (*)).

I suppose someone could clone a motor N-times and then import a single hourly weather forecast, into a separate hourly \<simulation\> structure.

But I can achieve the same result in GPSDC for an altitude and launch window as shown above in just a jiffy wothout polluting my .ORK any more that it alreadt is with all the motors I sim :)

But it gets even more complicated if you have no idea what will be safe to fly before doing any OR sims,

This is where the RSO Edition of GPSDC really shines -- it queries OpenMeteo for a user-selectable launch window, to retrieve an entire set of hourly forecasts all at once.

The user enters a range of apogees and whether the rocket will be flying single or dual deployment.

If flying dual deployment then you will also enter the altitude AGL where the main chute is deployed.

For example, this is GPSDC RSO edition for a series of  arbitrary DD flights with apogees from 2,000 ft to 10,000 ft at 2,000 ft intervals.

The run was set up so the main would be deployed at 500 ft AGL and the launch window was from 08:00 to 17:00 for yesterday's Tripoli Houston launch at Hearne:

Screenshot_20260913_085845

GPSDC RSO Edition produced this KML is just a jiffy and then I opened it in Google Earth and eyeballed it.

Before running GPSDC RSO edition, I really had no idea what would be safe to fly at Hearne ( it is notorious for the rivers and trees and ponds and powerlines outsite the fences about 900 ft on either side of the pad and all around the Airport ).

One quick glance at GPSDC RSO Edition told me to fly something with an apogee between 2000 and 3000 ft.

That info in turn gave me a set of candidate OR motors to sim via PR 3211 and then to plot in Google Earth the particular rocket that I wanted to fly ...

GPSDC RSO Edition is a wonderful tool for the task at hand.

This would be untenable if all I had to work with was OR.

That's why I said that OR will never replace GPSDC ...

-- kjh

(*) - @SiboVG has apparently worked around the single flight per simulation .ork file issue in [Add Monte Carlo landing-dispersion analysis- #3260](https://github.com/openrocket/openrocket/pull/3260) by storing the results of many (many, many) sims in memory ,,,

EDIT:  fix link to PR 3260

kjhambrick:
> Should I be able to build my own copy of **PR 3295** for testing or is there a .JAR file that I missed ?

Never mind on the JAR file access ...

Another **BIG THANKS** to @neilweinstock who showed me how to find the JAR files for the PR build logs !

I'll test PR 3295 a little as I find some free time away from work.

Thanks again, Neil and thanks @thzero for the new export features.

-- kjh

neilweinstock:
Pushing this to next release, which simply means "beyond the upcoming beta".  Same story: great feature, needs a bit more time.

JoePfeiffer:
> The two-stager is "Two stage high power rocket" from `File>Open example`
> 
> The Lat/Lon in the file does appear to be in the Southwest (maybe that's @JoePfeiffer 's place).

I had never noticed that there was a particular location we'd used for our example sims in the past! Yes, the "pad" in the KML is my house (it's easy enough to find my street address on line, and I'm old enough to remember back when the phone company put out a book every year doxing everyone in town, so I'm not really worried about it. But yes, just as a matter of principle not using a contributor's address as a launch location makes sense).

thzero:
> > The two-stager is "Two stage high power rocket" from `File>Open example`
> > The Lat/Lon in the file does appear to be in the Southwest (maybe that's @JoePfeiffer 's place).
> 
> I had never noticed that there was a particular location we'd used for our example sims in the past! Yes, the "pad" in the KML is my house (it's easy enough to find my street address on line, and I'm old enough to remember back when the phone company put out a book every year doxing everyone in town, so I'm not really worried about it. But yes, just as a matter of principle not using a contributor's address as a launch location makes sense).

Yeah.. I remember phone books also. :)

And yeah probably should go through the sample rockets and make sure!

thzero:
> Pushing this to next release, which simply means "beyond the upcoming beta". Same story: great feature, needs a bit more time.

all good :)

thzero:
@kjhambrick Ok, check out the presets...  i haven't gotten through them all yet with samples, but wanted to get it out since you had asked about those two options.

jimmiedave:
> > The two-stager is "Two stage high power rocket" from `File>Open example`
> > The Lat/Lon in the file does appear to be in the Southwest (maybe that's @JoePfeiffer 's place).
> 
> I had never noticed that there was a particular location we'd used for our example sims in the past! Yes, the "pad" in the KML is my house

hence #3323

kjhambrick:
@thzero --

PR3295  is VERY nice !

I love the KML Export Dialog with the altitudeMode settings !

I did a q&d test where I manually imported a saved OpenMeteo csv file for Sep 5 at 12:00 and did an OR Multi-Simulation Edit to set the Multilevel Wind for all of my configs and sims.

Then I manually set (lat,lon), site altitude and the temperature, pressure and humidity the same settings that PR 3211 returned for the 2026-09-05T12:00 flight

Then I ran your new [3D Path] Export:

pr3295-kml-dialog-01-Screenshot_20260915_081359

BEAUTIFUL !

I selected [DriftCast] and checked only the **Pad, Apogee, Ejection and Landing** Waypoints and let-er-rip.

Then I imported the resulting KML into GoogleEarth (GE) without any problems at all:

Screenshot_20260915_083005

Nice !

Then I loaded the KML produced by Exporting the same sim data via PR 3211 and processing it in GPS Visualizer:

Screenshot_20260915_083105

looks like the same Driftcast at a glance, and it **IS** the same when I selected both KML files at one time in GE:

Screenshot_20260915_082754

**Suggestion[ 1 ]**:  Blue-on-Blue is a little washed out on the Hutto plowed field ...

So one suggestion would be to allow the end user to select the color of the Tracks and Waypoints in the PR 3295 Export Dialog ?

Speaking of colors:  Wow !  Color me IMPRESSED !!

Much more testing and feedback to come and I need to learn how to wrap my head around the 3D plots :)

I am analysing a recent 45kft two stage flight where I've got Blue Raven Data along with an OR sim that matches the actual flight pretty well.

I will ask for permission to share the .ORK here along with athe Blue Raven data.

But it might take a while -- the owner of the data and the ORK is heading to XPRS and then BALLS for another EPIC flight and I don't want to distract him before the events ...

Thanks for letting me play along !

-- kjh

p.s.  one other suggestion might be similar to what I do with my OR-export.csv parser ...

Maybe allow the end uset to annotate the WayPoints with a user-definable 'Mission String' ...

This way one could load multiple KML files into a single GE session and the GroundTracks and Waypoints would be unique per OR simulation ...

Anyhow just a thought ...

One can also edit the WayPoints and the Track in GE after all :)

thzero:
Wish I was going to XPRS and BALLS

thzero:
> p.s. one other suggestion might be similar to what I do with my OR-export.csv parser ...
> 
> Maybe allow the end uset to annotate the WayPoints with a user-definable 'Mission String' ...
> 
> This way one could load multiple KML files into a single GE session and the GroundTracks and Waypoints would be unique per OR simulation ...
> 
> Anyhow just a thought ...
> 
> One can also edit the WayPoints and the Track in GE after all :)

Something like this...

image
image

kjhambrick:
Woo Hoo Yes !

Just like that !!

Thanks.

-- kjh

thzero:
> ## Per-stage track colors and a user-definable mission name
>
> ### Mission name
>
> A new Mission box with a Name field. Whatever you type is put in front of the document name, the stage folders and the track names — so a mission named Sod Blaster exports Sod Blaster Simulation 1 containing Sod Blaster Sustainer and Sod Blaster Booster. Load half a dozen of these at once and each flight is identifiable in the tree and switchable on its own.
>
> Waypoint markers keep their stage-qualified names by default (Sustainer Apogee), because a near-vertical flight already packs its markers into a small patch of screen and longer names overlap more. Prefix waypoints extends the naming to them for when two flights' markers genuinely overlap on the map.
>
> A name that already begins with the mission name isn't prefixed twice, so naming a mission after the rocket doesn't produce Sod Blaster Sod Blaster Sustainer. An empty mission name — the default — names everything exactly as before this change.
>
> ### Stage colors
>
> Each stage is drawn from a fixed ten-color palette, shared with the plot window so a stage keeps its color between the graph and the map. One color does the whole stage: the flight path line at full strength, the ground track in a darkened shade of it, and the waypoint pins tinted with it. Stage colors... opens a dialog for overriding them, one swatch per stage, with Reset to defaults.
>
> It's a dialog rather than inline controls because the stage count varies with the design and the tab has no room for a variable-length list.
>
> ### Not persisted, deliberately
>
> Neither the mission name nor the stage colors are remembered between exports. Both describe one particular export rather than how you like the exporter set up, and a remembered one would quietly mislabel or miscolor the next file. The Prefix waypoints checkbox is remembered, since that's a working preference.
>


ThreeDPathExport

DBellhorn:
> I can't see Open Rocket DriftCasts ever replacing @DBellhorn's GPSDC and especially not his RSO edition.

Integrating KML generation of a rocket's flight path makes sense to me.  I am more ambivalent toward the idea of duplicating GPSDC's multiple simulations at different times of the day.  Introducing that type of functionality feels counter to the OR focus on performing a single launch simulation as realistically as possible.

That being said, I'm not particularly opposed to the idea.  Whatever the core OR devs decide on this path forward is fine with me.

I do feel strongly that GPSDC "RSO Edition" functionality should not be incorporated into OR.  It calculates approximate landing zones based on drift paths from a range of altitudes.  Somehow requiring multiple rocket designs simulating launches to each of those altitudes sounds very counterproductive.

thzero:
Add flight summary balloons to the 3D path KML export
The KML export wrote geometry and labels only, with every number behind the picture left in OpenRocket. Some of the ideas here came from Rocket Trajectory Analyzer at https://www.pazziperirazzi.it/RTA, which reports peak altitude, duration, maximum distance and landing position next to its trajectory. A KML description is the right place to put those.
- The document balloon carries the flight summary, each stage folder its own range and landing, and each waypoint its time, altitude, distance and bearing from the pad plus the deploying device.
- Custom templates gain maxRange, timeToApogee, flightTime and the velocity and acceleration unit labels at flight level, and maxRange, hasLanding, landingDistance, landingBearing and landingTime per stage.
- Maximum range is the farthest point from the pad, not the landing point, since a rocket can drift downrange and partway back.
- A flight cut short by its time limit records no ground hit, so the landing lines sit behind a flag and disappear rather than inventing a touchdown.
- The balloon markup is written pre-escaped rather than in CDATA: values are escaped by the template engine, and inside CDATA those escapes are not decoded, so a rocket named "Bill & Ted" would reach the balloon as literal entity text.
- The Mustache compiler now treats an empty string and a zero as false so optional lines can vanish entirely. The change is additive and existing templates are unaffected.

kjhambrick:
OMG @thzero !

There are only a few things that might suck me back to my desk after dinner but your latest rev of PR3295 is definitely one thing that did it !

I've got a copy of the JAR file and I'll re-plot that 2-stage flight at FAR with the latest.

I'll let you know tomorrow morning.

THANK YOU !

-- kjh

kjhambrick:
Well ... I couldn't stand not looking :)

Installed pr3295 build 3638 and re-exported the KML.

This is the lost booster:

Screenshot_20260921_173738

It is WONDERFUL !

I especially love the position and bearing from the pad.

My only suggestion would be to add the (lat,lon) to the summary balloon.

Either as an ordered pair, (lat,lon),  or as separate labeled lines ... you've got a good eye so I'll leave aesthetics to you :) :)

Thanks $1M  -- this is a keeper !

-- kjh

kjhambrick:
Thanks for build 3649, @thzero !

I like having (lat,lon) coordinates along with distance and bearing in the summary bubbles !

Screenshot_20260923_041050

And the parachute info on the pin is a plus !

Screenshot_20260923_041122

-- kjh

PR #3296, opened 2026-08-30
3288 bug save as brings up save design info panel with empty design name kills rocket name

### LINKED ISSUE #3288: [Bug] "Save as..." brings up "Save Design Info" panel, with empty "Design name:", kills rocket name.

wcravens:
Closes #3288

# Overview

Two commits, two independent bugs in the "Save As..." → *Save Design Info* dialog.

## Commit 1 — empty design name wipes the rocket name

**Cause.** `SaveDesignInfoPanel` overrode `updateFields()` as a no-op. That override
existed only to dodge an NPE: the base `RocketComponentConfig.updateFields()`
dereferences `infoLabel` and `multiCompEditLabel`, which are created in `addButtons()` —
a method `SaveDesignInfoPanel` overrides without creating them.

The side effect was the real bug. `componentNameField` is populated *only* by
`updateFields()`, so "Design name" came up empty. On focus loss the text listener saw
`""` ≠ the rocket's name and called `setName("")`, which `RocketComponent` maps to the
default component name — hence the rocket being renamed to "Rocket".

**Fix.**
- `RocketComponentConfig.updateFields()` — null-guard `infoLabel` and
  `multiCompEditLabel`; subclasses may override `addButtons()` without creating them.
  (Most of the diff is re-indenting the existing info-label logic inside the new `if`.)
- `SaveDesignInfoPanel` — delete the no-op override so the inherited one runs.

## Commit 2 — Cancel does not discard its edits

**Cause 1 — no undo position.** `BasicFrame.showSaveRocketInfoDialog()` builds the panel
in a plain modal `JDialog`, bypassing `ComponentConfigDialog.showDialog()` — the thing
that pushes `document.addUndoPosition(...)`. With no snapshot of the dialog's opening
state, commit `56cdced69` had to strip the inherited `document.undo()` from the cancel
handler, because cancelling was rolling back whatever the user did *before* opening the
dialog (#2680). That removed the symptom and left the cause.

**Cause 2 — undo could not restore the design info anyway.** `Rocket.loadFrom()` never
copied `designer`, `revision`, `designType` or `kitName` back from the snapshot. These
live on `Rocket` rather than `RocketComponent`, so `copyFrom()` does not cover them
either, and undo restored only the design name. This affected Edit ▸ Undo everywhere,
not just this dialog.

**Fix.**
- `Rocket.loadFrom()` — copy the four design-info fields alongside `refType` /
  `customReferenceLength`.
- `BasicFrame.showSaveRocketInfoDialog()` — `document.addUndoPosition(...)` before
  `setVisible(true)`. It must be before: the dialog is modal, so `setVisible` blocks.
- `SaveDesignInfoPanel` — snapshot `rocket.getModID()` at construction; on cancel call
  `undoChangesSince(document, modIDAtOpen)`, which undoes **only if the rocket actually
  changed while the dialog was open**. That guard is what keeps #2680 from returning:
  with no edits the document is in a clean state, and `undo()` would step back past the
  undo position into the user's earlier work.
- `SaveDesignInfoPanel` — also drop `ComponentConfigDialog.clearConfigListeners = false`
  from the same cancel handler. Survivor of the same copy-paste `56cdced69` trimmed. The
  flag is a one-shot "skip `component.clearConfigListeners()` on the next
  `ComponentConfigDialog.windowClosed`", and this panel's parent is a plain `JDialog`, so
  `disposeDialog()` takes the `parent.dispose()` branch and that handler never runs —
  nothing on this path reads or resets the flag. A dead write, not a behaviour change;
  removing it keeps the handler honest about what it touches. Detail in
  `issue-draft-clearconfiglisteners-leak.md`.

**Tests.**
- `OpenRocketDocumentUndoRedoTest` (core) — undo/redo round-trip over name, designer,
  revision, design type, kit name.
- `SaveDesignInfoPanelTest` (swing) — cancel reverts the dialog's edits; cancel without
  edits preserves unsaved prior work (#2680 regression); cancel reverts *only* the
  dialog's edits when earlier changes are pending.

All four failed against the old code and pass now; full `:core:test`, `:swing:test` and
checkstyle are green.

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
@@ -7,6 +7,7 @@ import info.openrocket.core.document.OpenRocketDocument;
 import info.openrocket.core.l10n.Translator;
 import info.openrocket.core.rocketcomponent.RocketComponent;
 import info.openrocket.core.startup.Application;
+import info.openrocket.core.util.ModID;
 
 import info.openrocket.swing.gui.components.StyledLabel;
 
@@ -26,8 +27,13 @@ public class SaveDesignInfoPanel extends RocketConfig {
     private static final Translator trans = Application.getTranslator();
     private static final ApplicationPreferences preferences = Application.getPreferences();
 
+    /** State of the rocket when this panel was created, used to detect whether Cancel has anything to undo. */
+    private final ModID modIDAtOpen;
+
     public SaveDesignInfoPanel(OpenRocketDocument d, RocketComponent c, JDialog parent) {
         super(d, c, parent);
+
+        this.modIDAtOpen = d.getRocket().getModID();
 
         // (Optional) Fill in the design information for this file
         StyledLabel label = new StyledLabel(trans.get("SaveDesignInfoPanel.lbl.FillInInfo"), StyledLabel.Style.BOLD);
@@ -63,8 +69,8 @@ public class SaveDesignInfoPanel extends RocketConfig {
                 int resultYesNo = JOptionPane.showConfirmDialog(SaveDesignInfoPanel.this, msg,
                         trans.get("RocketCompCfg.CancelOperation.title"), JOptionPane.YES_NO_OPTION, JOptionPane.WARNING_MESSAGE);
                 if (resultYesNo == JOptionPane.YES_OPTION) {
-                    ComponentConfigDialog.clearConfigListeners = false;		// Undo action => config listeners of new component will be cleared
                     disposeDialog();
+                    undoChangesSince(document, modIDAtOpen);
                 }
             }
         });
@@ -82,5 +88,23 @@ public class SaveDesignInfoPanel extends RocketConfig {
         buttonPanel.add(okButton);
 
         this.add(buttonPanel, "newline, spanx, growx");
+    }
+
+    /**
+     * Discard the design info edits made in this dialog, restoring the state the rocket was in when the dialog
+     * was opened. The caller is expected to have added an undo position just before opening the dialog (see
+     * {@code BasicFrame.showSaveRocketInfoDialog()}), which is what that undo restores.
+     * <p>
+     * Nothing is undone when the rocket was not modified while the dialog was open: the document would then be in
+     * a clean state, and {@link OpenRocketDocument#undo()} would step back past the undo position and roll back
+     * whatever the user did <em>before</em> opening the dialog (see issue #2680).
+     *
+     * @param document      the document being edited
+     * @param modIDAtOpen   the rocket's modification ID at the time the dialog was opened
+     */
+    static void undoChangesSince(OpenRocketDocument document, ModID modIDAtOpen) {
+        if (document.getRocket().getModID() != modIDAtOpen) {
+            document.undo();
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
```diff
@@ -7,6 +7,7 @@
 import info.openrocket.core.l10n.Translator;
 import info.openrocket.core.rocketcomponent.RocketComponent;
 import info.openrocket.core.startup.Application;
+import info.openrocket.core.util.ModID;
```
SiboVG, line 10: I haven't checked the undo/redo code in a while, but the changes in this file feel a bit like hack. I can't look into it in detail now, but want to mention this before we merge this.

SiboVG (changes requested):

swing/src/main/java/info/openrocket/swing/gui/configdialog/SaveDesignInfoPanel.java
wcravens: Yep, because the whole design is a bit of a hack.  The fix is just a minimal amount of change to fix the issue without completely redesigning how change control is managed in these dialogs.

Issue #3247, opened 2026-08-18
[Feature Request] Refactoring of Simulation / Util Code (I.e. Tidy Up Initiative)


senft-research:
**Is your feature request related to a problem? Please describe.**
The Simulation side of the core module has a lot of code smells / gremlins that we haven't gone back to tidy up. This has led to difficulties refactoring some of the simulation code in the past. Furthermore, it makes the on boarding of new contributors more difficult (for example, key methods, directly linked to equations within Sampo's original paper, are not signposted to, leaving it as an exercise for the reader to deduce)

**Describe the solution you'd like**
Simply? A Tidy up from the bottom, working our way upwards. I have already started to do some refactoring of the `Quaternion.java` class in my own private repo, and this is what ultimately inspired this issue. 

**Describe alternatives you've considered**
The alternative would be to simply leave the code as is, and continue to accept that some aspects of codebase will be difficult for new contributors to begin parsing through / understanding. 

**Additional context**
This issue would not be resolved by a singular PR, but rather be an ongoing effort of many small PRs that would link directly to this issue. 

I also believe anyone contributing to this issue should get the benefit of learning more about OR's core codebase from their efforts. 

As such:

- I believe this issue can be marked as "good first issue", with the idea that they are spending the time looking at core classes (such as those in util / simulation), gaining a comprehensive understanding of the class, and using that understanding to propose refactors. 

- I believe this issue could very well be a signposted "good first issue" i.e. a primary suggested issue to consider doing PRs for. A lot of contributors bounce off of OR due to the messiness of some of the codebase (god knows I was one of them for a time haha), so leaning into that by having an issue that is designed to address that problem could see us retain some more potential contributors. 

- I believe this issue's related PRs should not include LLM work, as the main benefit of this issue is not the cleanup, but the understanding that would be gained by those contributing to it. 

- Any work on classes that utilize calculations / concepts from Sampo's original paper (e.g: Such as Quaternions) should reference the equation number / page number corresponding to the code within its javadocs). 

- When refactoring a class, its corresponding Unit Tests should be analyzed and refactored too if there is missing test-cases, lacking code coverage etc. Essentially, no class' Unit Tests should be assumed to be ideal when they are the focus of a PR relating to this issue.

senft-research:
An example of a refactor:

In `Quaternion.java` there are several fields for the scalar / vector components with 0 context. For a new contributor, they would have to go out of their way to look up what they mean and why, with 0 context provided. 

```java
[...]

PR #3302, opened 2026-09-02
Move common code from RK4SimulationStepper and RK6SimulationStepper into new AbstractRKSimulationStepper

JoePfeiffer:
There is a tremendous amount of duplicated code between the two Runge-Kutte simulation steppers. This PR creates a new abstract class AbstractRKSimulationStepper containing that shared code so we don't have to try to keep it in sync.

It should be noted that this is strictly a code cleanup which makes no changes to to computations, so the existing unit tests remain valid.


senft-research:
Small suggestion, but as per #3247 , could this be an opportunity to signpost the RK4 / RK6 sources (i.e. the equation numbers in Sampo's paper and the paper you sent me a week or 2 back) within the java docs?

JoePfeiffer:
> Small suggestion, but as per #3247 , could this be an opportunity to signpost the RK4 / RK6 sources (i.e. the equation numbers in Sampo's paper and the paper you sent me a week or 2 back) within the java docs?

About all I can think that would be marginally useful would be to say "this is an implementation of section 4.2.4 of the technical documentation". And as I look at it, the tech doc has stuff that came from the thesis, and really doesn't need careful explanation like quaternions and RK4.

senft-research:
> About all I can think that would be marginally useful would be to say "this is an implementation of section 4.2.4 of the technical documentation".

I would recommend literally just that mate. Not a bad shout to make it easier to signpost.

PR #3317, opened 2026-09-06
Allow selecting macOS app bundles as graphics editors

### LINKED ISSUE #3303: Selecting graphic editor doesn't work on Mac
This is a regression the crept in at some point, as far as I know applies only to the Mac.

If I try to select a graphic editor in the Preferences -> Graphics tab, I can't select a .app bundle because the file browser will always cd into it.  See screen recording below.  A .app bundle should be treated as an application, not a directory.

https://github.com/user-attachments/assets/ce4efeff-bbd2-405f-8a04-e867c81e9bb2

SiboVG:
Fixes #3303. You can now select `.app` folders on macOS as the graphics editor.
image

### DIFF
--- a/swing/src/main/java/info/openrocket/swing/gui/util/GraphicsEditorChooser.java
+++ b/swing/src/main/java/info/openrocket/swing/gui/util/GraphicsEditorChooser.java
@@ -2,11 +2,14 @@ package info.openrocket.swing.gui.util;
 
 import java.awt.Component;
 import java.io.File;
+import java.util.Locale;
 import java.util.Optional;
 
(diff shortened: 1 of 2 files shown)


neilweinstock:
It now works if you double-click an app, but does not work if you single click and the select "Open".  It just does nothing.  Wouldn't normal file selector behavior allow you to select an app that way?

https://github.com/user-attachments/assets/c29d405c-1a11-4ef3-a12a-990d349105c8

neilweinstock:
I am receptive to merging this now and fixing the last remaining issue later.  @SiboVG how would you like to proceed?

PR #3318, opened 2026-09-06
Add a subsonic pressure-drag floor for stubby stored-table nose cones

thzero:
# Summary

Short, blunt nose cones (fineness ratio below ~1.8) built from the stored experimental drag tables were being assigned essentially zero subsonic pressure drag, which is non-physical — a stubby ellipsoid or Von Kármán has real, measurable subsonic pressure drag. This PR adds a fineness-dependent floor to the subsonic region for those shapes coming from the mmrocket-sim project.

# Problem

SymmetricComponentCalc.calculateNoseInterpolator() builds the pressure-drag curve two ways:

CONICAL / OGIVE — computed analytically, with a proper Cd(M=0) anchor. These are fine.
ELLIPSOID / POWER / PARABOLIC / HAACK — taken from stored fineness-ratio-3 tables (NASA TR-R-100) and extrapolated to the actual fineness ratio.

For the stored-table shapes whose table is zero at drag divergence (POWER > 0.5, PARABOLIC > 0.5, HAACK), the subsonic power-law fit is skipped (the old minValue < 0.001 early return), and the fineness extrapolation is multiplicative, so 0 → 0 at every fineness ratio. The result is a stubby nose that carries ~0 subsonic pressure drag and then jumps at its drag-divergence Mach. The remaining table shapes (ELLIPSOID and the analytically-blended POWER/PARABOLIC cases) do get the fit, but it's anchored at a tiny cdMach0 for a blunt body, so their low-Mach end is unrealistically small too.

# Approach

Refactored the three subsonic-fit early returns into a single guard that is the exact De Morgan negation of the originals, so every nose that took the fit before takes it identically now — but control flow no longer skips past the new floor.
Added applyStubbyNoseFloor(...), applied only to stored-table nose shapes with fineness ratio > 0 and < 1.8. It raises the whole subsonic range (including the leading point) to at least the floor via max(existing, floor), so the curve doesn't sit low and then jump at drag divergence.
The floor is a fraction of this model's own cone value at the same fineness ratio:
cone = 0.8/(1+4f²) — the class's own conical Cd(M=0).
roundness = 1/3 — a rounded stubby nose as a fraction of that cone, bracketed from Mercer's Centuri TIR-100 measured deltas and an independent CFD estimate (~0.12 on frontal area at fineness ratio 0.5).
taper = 1 − (f/1.8)² — fades the floor to 0 at the 1.8 cutoff (TIR-100 §8 finds no significant nose-shape variation above ~1.8), so there is no discontinuity there.

Erring high charges more drag and predicts less altitude — the conservative direction. This is a bracket on the endpoint and shape of the law, not a calibrated curve; the region between fineness ratio 0.5 and the 1.8 cutoff is interpolation.

# Scope / non-goals

CONICAL and OGIVE are deliberately excluded — their analytic Cd(M=0) already exceeds this floor, so max() would ignore it anyway.
Only true noses (fore radius ≈ 0, expanding aft) are affected; mid-body transitions and boattails are out of scope.
Noses at fineness ratio ≥ 1.8 are untouched — behavior is identical to before.

# Testing

testStubbyNoseSubsonicFloor — Von Kármán (HAACK, param 0; the pure zero-at-divergence case) at fineness ratio 0.5 asserts subsonic Cd equals the floor (0.12304527) at M = 0.0, 0.3, 0.6. Locks the constants and the taper law.
testTallNoseNoSubsonicFloor — same nose at fineness ratio 2.0 asserts subsonic Cd stays ~0, locking the 1.8 cutoff.
Existing testEllipseNoseconeDrag / conical / ogive tests still pass, confirming the fit refactor is behavior-preserving.

### DIFF
--- a/core/src/main/java/info/openrocket/core/aerodynamics/barrowman/SymmetricComponentCalc.java
+++ b/core/src/main/java/info/openrocket/core/aerodynamics/barrowman/SymmetricComponentCalc.java
@@ -51,6 +51,23 @@ public class SymmetricComponentCalc extends RocketComponentCalc {
 	private final double wetArea;
 	private final double sinphi;
 
+	/**
+	 * Fineness ratio at/above which nose shape stops affecting subsonic pressure
+	 * drag. Centuri TIR-100 section 8 measures no significant variation across the
+	 * catalogue nose shapes from fineness ratio 4.0 down to 1.8.
+	 */
+	private static final double STUBBY_NOSE_FINENESS_LIMIT = 1.8;
+
+	/**
+	 * A stubby rounded nose's subsonic pressure drag as a fraction of this class's
+	 * own cone value at the same fineness ratio. Bracketed from measured deltas
+	 * (Mercer's TIR-100 series and a CFD estimate for a stubby nose), which put a
+	 * rounded stubby nose near 1/3 of this model's cone, ~0.12 on frontal area at
+	 * fineness ratio 0.5. Erring high charges more drag (conservative). This is a
+	 * bracket on the endpoint, not a calibrated curve.
+	 */
+	private static final double STUBBY_NOSE_ROUNDNESS = 1.0 / 3.0;
+
 	public SymmetricComponentCalc(RocketComponent c) {
 		super(c);
 		if (!(c instanceof SymmetricComponent)) {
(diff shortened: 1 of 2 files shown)


jimmiedave:
Responding "have had eyes on it".

Question: beyond "the software tests out as working", how might the results be validated? Is it gonna take a college student with wind tunnel access, or a bunch of flights matching sims better, or can a curve just be plotted and compared to existing experimental results?

BTW: this PR is written up well. I could learn a thing or two.

neilweinstock:
Assigning it to "Next Release", hope to get it in sometime after the upcoming beta.

I have same question as jddj regarding validation.

PR #3319, opened 2026-09-06
mass inertia correction

thzero:
# Problem

When a component's mass is overridden with "override mass of all subcomponents" enabled, MassCalculation.calculateStructure() only zeroed the children's center-of-mass weight. Their RigidBody entries stayed in the bodies list at their geometric masses, so the moment-of-inertia sum reflected the geometric mass while the reported total mass was the overridden mass.

So the rocket flew with mass X but the rotational/longitudinal inertia of a different mass — roll and pitch inertia off by exactly the override factor. This threw off weathercocking and drift; apogee was largely unaffected.

Found via a cross-check against mmrocket-sim, which surfaced the mass/inertia discrepancy on overridden subcomponents.

$ Fix

When the override covers subcomponents, keep the geometric mass distribution but rescale it to the overridden total:

scale = overrideMass / geometricSubtreeMass

Only the inertia bodies are rescaled — mass and CG bookkeeping are untouched. The no-override and self-only-override paths reduce to the previous values exactly, so they're bit-identical.

Rescaling proportionally to the existing geometry is the physically-consistent default (for fixed geometry, MOI scales linearly with mass), but it's a deliberate behavior choice worth a reviewer's eye.

# Testing

Full core test suite passes.
Added testSubcomponentMassOverrideScalesInertia: pins a stage over its subcomponents at factors 1.0 / 1.8 / 3.6 / 9.1 and asserts mass and both inertia axes scale linearly. Before the fix, the inertia asserts failed for every factor ≠ 1.
Updated testFalcon9HeavyBoosterStageMassOverride: its recorded MOI values were the bug — a 0.5 kg override carrying the full geometric boosters' inertia — so they're updated to the mass-consistent values.

### DIFF
--- a/core/src/main/java/info/openrocket/core/masscalc/MassCalculation.java
+++ b/core/src/main/java/info/openrocket/core/masscalc/MassCalculation.java
@@ -70,7 +70,20 @@ public class MassCalculation {
 	public void addInertia( final RigidBody data ) {
 		this.bodies.add( data );
 	}
-	
+
+	/**
+	 * Scale the mass and moment of inertia of every accumulated body by the given
+	 * non-negative factor.  Used when a mass override covers all subcomponents: the
+	 * geometric mass distribution is preserved but rescaled to the overridden total,
+	 * keeping the moment of inertia consistent with the mass.
+	 *
+	 * @param factor  non-negative scale factor
+	 */
+	void scaleInertia( final double factor ) {
+		for (int i = 0; i < this.bodies.size(); i++)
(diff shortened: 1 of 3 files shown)


jimmiedave:
To my depth of understanding (I understand the problem as described, don't have the depth on the current code), this seems like a reasonable fix. Responding here to indicate I've read, and assent to merge if there aren't other objections.

neilweinstock:
This one seems good to me, need a code review before merge.

PR #3320, opened 2026-09-08
Expands the Design Report (print/export) with more static design data and adds a CSV export.

thzero:
This was done because I wanted more information on PDF print outs but also need a CSV output of some of the main specifications of the rocket.

### Printed report
- New **Rocket Static Stats** section: length, max diameter, mass (empty/loaded), CG (empty/loaded), CP, fineness (L/D), stability (cal and %), drag coefficient at the default Mach, normal-force slope (CNα), and pitch/roll inertia — shown for the whole rocket and, for multi-stage designs, per stage.
- Layout cleanup: bold, larger rocket name; two-column metadata (Stages | Design type, Designer | Kit); more consistent spacing.
- **Design Notes** (revision + comment) on their own page.
- Motors moved to their own page under a **Motors** header (like Parts Detail).
- New **Include motors** checkbox in the print/export dialog to omit the motor/flight tables.
- 
### CSV export
- New **Save as CSV** button. Exports the textual design info only (no diagram): metadata + statistics as `Scope,Field,Value,Unit`, with value and unit in separate columns.
- Whole-rocket and per-stage statistics come from the same `DesignReport.computeStaticStats(...)` as the PDF, so the two never diverge.
- Adds **nose-to-fin-root** top/bottom distances per stage and per fin set (duplicate fin-set names disambiguated with #1/#2).
- UTF-8 BOM so spreadsheets render unit symbols (e.g. `oz·in²`) correctly.


neilweinstock:
Looks pretty good.  One small issue: the "Include motors" should be a checkbox in the main content selection list up top (probably named "Motor summary" or something like that) rather than on its own in the bottom.

thzero:
> Looks pretty good. One small issue: the "Include motors" should be a checkbox in the main content selection list up top (probably named "Motor summary" or something like that) rather than on its own in the bottom.

Done...

Issue #3325, opened 2026-09-14
[Feature Request] Add a Blender-specific OBJ export preset


shawn89890916:
### Description

OpenRocket already exports high-quality Wavefront OBJ geometry and appearance data, and the export documentation explicitly calls out Blender as a use case. Importing a design into Blender still requires users to know which generic rendering options map to Blender's unit and color expectations.

On the current `unstable` branch, `OBJOptionChooser.optimizeSettingsForRendering()` enables motors and appearances, disables triangulation, selects normal LOD, and applies a scale of `20` with the comment `Idk, pretty arbitrary`. It does not enable the existing sRGB option, even though the OBJ export documentation and the original OBJ PR describe sRGB as important for Blender color consistency.

### Requested behavior

Provide a discoverable Blender-oriented export profile that configures the existing OBJ exporter for a predictable Blender import. This should reuse the current exporter and should not add a Blender runtime dependency or launch an external process.

A possible profile would:

- export component appearances and motors
- keep non-triangulated editable geometry
- enable sRGB colors
- use a documented, deterministic scale appropriate for Blender units (the exact value should be agreed with maintainers)
- preserve the user's component, child, instance, and origin selections unless the profile needs a clearly documented value

The UI could expose this as a separate **Blender** preset or refine the current **Rendering** preset if that is preferable.

### Documentation

A short OpenRocket-to-Blender workflow could document:

1. selecting the complete rocket/component set
2. applying the Blender profile
3. importing the generated OBJ/MTL and texture assets into Blender
4. verifying dimensions and component groups

### Acceptance criteria

- an options-level test covers the preset values
- a standard example rocket imports into a supported Blender version at the documented dimensions
- exported colors use the intended color space
- existing 3D Printing and custom export workflows remain unchanged
- OpenRocket does not depend on an installed copy of Blender

### Context

I maintain a small Blender-native, dimension-driven rocket builder at https://github.com/shawn89890916/openrocket-to-blender. While reviewing whether that functionality belonged in OpenRocket, I found that OpenRocket's existing OBJ exporter already provides the geometry layer. A focused Blender profile and workflow appears to fill the remaining integration gap without duplicating OpenRocket's component exporters.

If the maintainers agree with this direction, I would be happy to implement the preset, tests, translations, and documentation against `unstable`.

PR #3326, opened 2026-09-14
[#3325] Add Blender OBJ export preset

shawn89890916:
## Summary

This draft addresses #3325 by adding a **Blender** preset to the existing OBJ export chooser.

The current generic Rendering preset enables appearances and motors but uses an arbitrary scale of 20 and leaves the color-space choice to the user. OpenRocket already has all geometry and appearance export capabilities needed by Blender, so this change only provides a predictable configuration of those existing options.

## Blender preset

- exports motors
- exports component appearances
- exports colors in sRGB
- keeps editable, non-triangulated geometry
- uses normal level of detail
- uses scale `1`, matching OpenRocket's meter coordinates to Blender's default 1 BU = 1 m
- preserves component/children/instances/origin choices from the current dialog

The existing 3D Printing and Rendering presets remain unchanged. Changing an advanced option, including sRGB, updates the highlighted preset.

## Other changes

- adds the English localization keys for the button and tooltip; other locales can follow the existing Crowdin workflow
- updates the feature list to mention Blender rendering
- adds focused tests for the applied options and preset matching

## Verification

- `./gradlew :swing:test --tests info.openrocket.swing.gui.choosers.OBJOptionChooserTest`
- `./gradlew :swing:checkstyleMain :swing:checkstyleTest`
- both commands pass with Microsoft OpenJDK 17

## Review question

Would maintainers prefer a dedicated Blender preset as implemented here, or should these values replace/refine the generic Rendering preset? The draft status leaves that product decision open before finalizing documentation or screenshots.


neilweinstock:
Setting this to "Next Release", no time to evaluate before the upcoming beta.  Hopefully we can get it in by second beta or final release.

swing/src/main/java/info/openrocket/swing/gui/choosers/OBJOptionChooser.java
```diff
@@ -516,6 +535,30 @@ private boolean isOptimizedForRendering(OBJExportOptions options) {
                 options.getLOD() == ObjUtils.LevelOfDetail.NORMAL_QUALITY && options.getScaling() == 20;
     }
 
+    private void optimizeSettingsForBlender() {
+        OBJExportOptions options = new OBJExportOptions(rocket);
```
SiboVG: Many settings are already set by the rendering optimization. In this function, first call `optimizeSettingsForRendering` and then set specific settings for Blender, such as sRGB

SiboVG (changes requested):

swing/src/main/java/info/openrocket/swing/gui/choosers/OBJOptionChooser.java
shawn89890916: Implemented in ea767d8. `optimizeSettingsForBlender()` now calls `optimizeSettingsForRendering()` first and only applies the Blender-specific sRGB and 1:1 scale overrides. I also updated the tests to verify that the shared rendering options are preserved. The targeted tests and Checkstyle both pass. Thanks for the suggestion!

shawn89890916:
Rebased this PR onto the latest `unstable` (`6deae5079`). The requested rendering-preset reuse remains in `a52c998`, and the targeted Blender preset tests plus Swing main/test Checkstyle all pass.

@SiboVG, when you have time, could you please take another look? Thanks!

PR #3327, opened 2026-09-16
Add Thrust Compensation for Air Pressure

JoePfeiffer:
# Description

Adjust motor thrust due to pressure at altitude compared to sea level. Published motor thrust curves provide thrust at sea level. At higher altitudes the air pressure is lower, so the thrust is increased by A * (P<sub>sea</sub> - P<sub>alt</sub>) where A is the motor's nozzle exit area, P<sub>sea</sub> is pressure at sea level, and P<sub>alt</sub> is pressure at the current altitude.

It uses the nozzle diameter added in #3244 to calculate area. As with that PR, if the nozzle exit area is 0 there is no correction.

The thrust correction is saved as a flight data variable, in case users are interested in seeing that.

# Testing

A unit test has been added as MotorPressureCorrectionTest.java. It independently calculates the expected thrust at altitude, and compares with the thrust calculated by OR.

It has also been compared against results from RASAero. A two stage rocket powered by an N5800 in each stage was created in OR and ported to RA. Unfortunately, as is far too common, OR's altitude estimate was somewhat more optimistic than RA's; the surface roughness of the two models were adjusted until their altitude estimates were similar at 79,204 ft for OR and 79, 031 ft for RA (Imperial units are used as RA does not support metric). These results were with a nozzle diameter of 0.

One gratifying result, not directly related to this PR, is that the two rocket's centers of pressure were very similar: the full stack had its CP at 118 in with OR and 115.67 with RA, and the sustainer CP was 81.434 in with OR and 82.8032 in with RA. 

Assuming a nozzle diameter of 2.5 in (the actual nozzle exit diameter of an N5800), the apogees become 85,910 ft with OR and 85,336 ft with RA. The difference provided by the changed nozzle diameter is comparable, though greater with OR (note that this includes the differences in both base drag and thrust).

The OR and RA models, with nozzle diameter 0, are attached.

# AI assistance

No AI was used anywhere in developing this PR.

[n-to-n.ork.zip](https://github.com/user-attachments/files/32265713/n-to-n.ork.zip)
[n-to-n.CDX1.zip](https://github.com/user-attachments/files/32265726/n-to-n.CDX1.zip)


SiboVG:
@JoePfeiffer I'm not seeing the `MotorPressureCorrectionTest`. Did you include it in your VCS?

One concern: `if (thrust > 0)` checks the thrust of all motors together. Consequently, for mixed clusters, a zero-thrust motor receives a pressure correction whenever another motor produces thrust

We should also reset `store.thrustCorrection` before the pre-listener.

For completeness, we might also want to reference the standard rocket thrust equation, or link to https://www1.grc.nasa.gov/beginners-guide-to-aeronautics/rocket-thrust/.

JoePfeiffer:
> @JoePfeiffer I'm not seeing the `MotorPressureCorrectionTest`. Did you include it in your VCS?

Could have sworn I did that... added now.

> One concern: `if (thrust > 0)` checks the thrust of all motors together. Consequently, for mixed clusters, a zero-thrust motor receives a pressure correction whenever another motor produces thrust

I don't think so:  doesn't getThrustingNozzleExitAreas() return the sum of the areas of the currently thrusting motors (and doesn't contain the area of anything not currently thrusting)? Under that assumption this code should be correct.

> We should also reset `store.thrustCorrection` before the pre-listener.

No, the thrust passed on to a post-listener ought to be the thrust as calculated by us, which includes the correction. Also, I want that correction to be available to users in case they have some use for it (if we just reset it, there's no reason to have it in the store in the first place).
 
But this raises a question I hadn't noticed before -- are the contents of the store available to listeners? They ought to be....

> For completeness, we might also want to reference the standard rocket thrust equation, or link to https://www1.grc.nasa.gov/beginners-guide-to-aeronautics/rocket-thrust/.

Good idea. Added in the code; where is a good place to add it in the docs? Is the intent of https://openrocket.readthedocs.io/en/latest/user_guide/thrust_curves.html to document our handling of thrust curves in general? All that's in there at the moment is a mention of custom thrust curves with a link to the thrust curve folder documentation.

SiboVG:
Thanks, I checked the latest changes.

The mixed-cluster concern is resolved: `getThrustingNozzleExitArea()` only includes motors for which `isThrusting()` is true, so that part looks fine.

I do still think `store.thrustCorrection = 0` should be moved before the pre-thrust listener. My point wasn't to clear the correction before the post-listener. Currently, if a pre-listener overrides thrust, `calculateThrust()` returns early and `store.thrustCorrection` keeps its previous value (or NaN initially). So I suggest:

```java
store.thrustCorrection = 0;

// Pre-listeners
thrust = SimulationListenerHelper.firePreThrustCalculation(status);
if (!Double.isNaN(thrust)) {
    return thrust;
}
```

I would also set `store.thrustCorrection = 0.0` in `landedValues()` alongside `thrustForce` and `dragForce`.

One physics/documentation point: in the standard thrust equation

`F = mdot*Ve + (Pe - Pa)*Ae`

`Pe` is the nozzle exit pressure, not standard atmospheric pressure. The implemented correction is still reasonable, but it is better derived by treating the thrust curve as measured at a reference ambient pressure:

`F = F_ref + (P_ref - P_ambient) * Ae`

where we currently assume `P_ref = 101325 Pa`.

So I would avoid saying that the thrust-curve values are simply the `mdot*Ve` term, and instead document that we assume published thrust curves represent total measured thrust at approximately standard sea-level ambient pressure.

One minor test issue as well: `MotorPressureCorrectionTest` sets the pressure and then calls `calculateFlightConditions()`, which replaces `store.flightConditions` and recalculates the atmosphere. The test passes, but I think it would be clearer to call `calculateFlightConditions()` first and then explicitly set the pressure before `calculateThrust()`.

Other than that, the implementation looks good to me.

JoePfeiffer:
> I do still think `store.thrustCorrection = 0` should be moved before the pre-thrust listener. My point wasn't to clear the correction before the post-listener. Currently, if a pre-listener overrides thrust, `calculateThrust()` returns early and `store.thrustCorrection` keeps its previous value (or NaN initially). So I suggest:
> 
> ```java
> store.thrustCorrection = 0;
> 
> // Pre-listeners
> thrust = SimulationListenerHelper.firePreThrustCalculation(status);
> if (!Double.isNaN(thrust)) {
>     return thrust;
> }
> ```

I misread; you are correct.

> 
> I would also set `store.thrustCorrection = 0.0` in `landedValues()` alongside `thrustForce` and `dragForce`.
> 
Got it.

> One physics/documentation point: in the standard thrust equation
> 
> `F = mdot*Ve + (Pe - Pa)*Ae`
> 
> `Pe` is the nozzle exit pressure, not standard atmospheric pressure. The implemented correction is still reasonable, but it is better derived by treating the thrust curve as measured at a reference ambient pressure:
> 
> `F = F_ref + (P_ref - P_ambient) * Ae`
> 
> where we currently assume `P_ref = 101325 Pa`.
> 
> So I would avoid saying that the thrust-curve values are simply the `mdot*Ve` term, and instead document that we assume published thrust curves represent total measured thrust at approximately standard sea-level ambient pressure.

Yeah, I was trying to avoid having to define the terms since ; having cited the thrust equation ignoring it in favor of just using F_ref  also didn't seem like a good idea. So... tried to clarify a bit on this round.

> 
> One minor test issue as well: `MotorPressureCorrectionTest` sets the pressure and then calls `calculateFlightConditions()`, which replaces `store.flightConditions` and recalculates the atmosphere. The test passes, but I think it would be clearer to call `calculateFlightConditions()` first and then explicitly set the pressure before `calculateThrust()`.

Note quite sure why one is clearer than the other, but I found some other opportunities to clarify as well.
> 
> Other than that, the implementation looks good to me.

